PR #134, opened 2019-03-12
get out enhancements for IE

tujoworker:


tujoworker:
Also, new icons where fetched and automatically added

PR #264, opened 2019-09-05
ftr: Icon compatibility

arnemolland:
## Summary
I had a problem with broken files in Figma, Illustrator and icon fonts. All icons should now be more compatible, e.g. Illustrator, Figma and icon fonts, as strokes are converted to paths and unneeded/incompatible attributes are removed. The resulting icons are identical to the old ones 🧹

I'm suspecting this may be related to the use of [svgo](https://github.com/svg/svgo) (this is by the way the recommended tool in the portal), as it broke the icon compatibility when I used it for optimizing. Instead, I've used [svgcleaner](https://github.com/RazrFalcon/svgcleaner) which have resulted in no broken files for me so far.

## Test plan
A comparison between the old and new icons in Figma. The UI in the portal etc. is unaffected.
Screenshot 2019-09-05 at 23 08 02


tujoworker:
Thak you for making the first pull request on Eufemia! 🥳

I/we have to investigate more on this, as the toolling is set up to be 100% automated on a CI from the Figma master file *Eufemia - Web* to the `dnb-ui-lib` ready for release. 

I agree that we can future improove the tooling. So I will make a quick POC if it is as simple as swapping out **svgo** with **svgcleaner**.

I may also have a chat with you on Slack.

tujoworker:
I close this PR since it is not the best sollution to convert all icons to paths. The overall space comsuption will almost get twice as large, as paths often use more data points to get the same result.

But for sure, will request of having path only SVG icons **more often in future**, we could automaticly create two versions of the icons as SVG files.

PR #566, opened 2020-07-01
feat: add new #accordion component

tujoworker:


tujoworker:
The accordion will be included in #rc/v7.2-beta

PR #642, opened 2020-09-15
repo: use build dir and Node.js v14 as the target version

tujoworker:
This PR adds local library development support for:
- Newer Node.js version - as for now v14
- Use /build forlder only dyring a library build

- [ ] We still need to thest an actual release


tujoworker:
v7.2.0-beta.18 got released sussfessfully with the new setup. #653

.circleci/config.yml
```diff
@@ -6,7 +6,7 @@ version: 2
 jobs:
   test:
     docker:
-      - image: circleci/node:10.6
+      - image: circleci/node:14
```
ivoiv, line 9: Will this one use the same version of 14 every time, or will it go for the latest 14.x.x ?

If it goes for the latest 14 we should maybe also set Engine and Volta env versions to >= 14.0.0 instead of just 14.10.1 as well?

PR #667, opened 2020-10-07
feat: ready to use #payment-card component

tujoworker:
# Principles
The intention is to make it easy to integrate a payment card design and to have a single source of truth.

## Where to place it

While Eufemia components are daily building block components, we have to find another place to have it. 
Since the "patterns" section of the lib has very much undefined content & parts, we may have it placed there.
Also, we  may considder to actually rename patterns to extentions. Would that be a more fitting name for most of the content there?  

## To do

- [x] First integration, into patterns
- [x] Find alternative to daggy
- [x] Create more tests
- [x] Write info texts
- [x] Write properties referance
- [x] Hide Events tab in docs
- [ ] Test with screen reader
- [x] Write release notes


SindreMTeigland (approved):

MrHaze:
@tujoworker Maybe we can put it under `examples` or `addOns`?

tujoworker:
Looks like there has to be made more enhancements for screen readers, so I opened a [new PR](https://github.com/dnbexperience/eufemia/pull/691) with a first commit of fixing the SVGs logos with a title.

PR #686, opened 2020-11-05
chore: test setting variant and size of button

langz:
Testing the following logic of a Button, https://github.com/dnbexperience/eufemia/blob/a72bb99dde7b2f446173cb04357a66137a71640d/packages/dnb-ui-lib/src/components/button/Button.js#L240-L255


tujoworker:
Awesome 💙🙏

PR #691, opened 2020-11-09
fix: enhance PaymentCard for screen readers

tujoworker:
# Principles
The intention is to make it easy to integrate a payment card design and to have a single source of truth.

## Where to place it

While Eufemia components are daily building block components, we have to find another place to have it. 
Since the "patterns" section of the lib has very much undefined content & parts, we may have it placed there.
Also, we  may considder to actually rename patterns to ~extentions~addons. Would that be a more fitting name for most of the content there?  

## To do

- [x] First integration, into patterns
- [x] Find alternative to daggy (state handling)
- [x] Create visual tests
- [x] Create ARIA test
- [x] Create integration tests
- [x] Make it work for SSR
- [x] Write info texts
- [x] Write properties referance
- [x] Hide Events tab in docs
- [x] Test with screen reader
- [x] Enhance screen reader supprt so the card **type** can be read
- [x] Write release notes
- [x] Make #skeleton animate
- [x] Create logic tests


SindreMTeigland:
Maybe also move the figcaption above the graphical card. That way the screenreader will read the cards productname first, and then the content of the card

SindreMTeigland:
There is an issue with some gradient prop when translating it to JS on MCBlack.

image

The logo should look like the one on the right

SindreMTeigland:
https://github.com/dnbexperience/eufemia/blob/aa2dc01ba86d6eb6800bece5eeb60957ad8694c3/packages/dnb-ui-lib/src/patterns/payment-card/style/_payment-card.scss#L259

This can be removed for now, but If we want to support the more compact design adding this class will create that style.
image

Not sure how the smaller design should display statuses like blocked and expired if this is implemented

tujoworker:
Do you know more about the transparent SVG issue?

Is the compact design a feature that is requested by someone? Did your initial component include that already (in addition to the styles)? We can then for sure enable/add that feature in this PR 😉

SindreMTeigland:
> Do you know more about the transparent SVG issue?
> 
> Is the compact design a feature that is requested by someone? Did your initial component include that already (in addition to the styles)? We can then for sure enable/add that feature in this PR 😉

I did not look into the SVG issue at the time and just imported the svg file instead. That worked fine.

The compact design is something I added based on som [figma illustrations](https://www.figma.com/file/Y2Q7ms668IlhiAkRVe87XCY4/Mobilbank---iOS?node-id=20419%3A600) for mobile bank, but I did not finalize it due to it not fitting with the application I was working on. 
image
You can see here that expired/blocked cards here is whited out and the state of the card is displayed below (this goes for other states as well)
I talked to someone in UX and they said that the compact cards are used in the new dnb.no designs

PR #761, opened 2021-02-15
break: rename Number component and import to NumberFormat resp. number-format

tujoworker:
This is a breaking change PR.

## Also

If you where to import utils functions, like `format`, then you have to change the import path to be:

- `import { format }  from 'dnb-ui-lib/components/number-format/NumberUtils'` 

This is due to better TypeScript support.


packages/dnb-ui-lib/src/components/number-format/NumberUtils.d.ts
```diff
@@ -0,0 +1,64 @@
+export type formatTypes = 'phone' | 'org' | 'ban' | 'nin' | 'currency';
```
dagerikhl, line 1: This is suberb! Great that you export these utils. Been wishing for this functionality exported recently, actually :D Now I don't have to create my own implementation that could easily break.

dagerikhl (approved):
Looks good to me! Feels a little better to not overwrite JS globals.

It's such a vast repo, so I don't feel I can confidently say that you've replaced every instance of `Number` with `NumberFormat`, but I trust that you have control over that 👍

tujoworker:
Closing this PR as it is merged into the alpha release branch. A first release is out https://www.npmjs.com/package/@dnb/eufemia

PR #767, opened 2021-03-01
feat: New name and design for silver card. Added missing product

SindreMTeigland:
## Summary

Three of the card products have been rebranded from Silver to Pluss card with new name and new design.
Both name changes and new design have been added.

Missing  product *VO1* have been added aswell
## Test plan

All test where green with command *yarn test -u*
Changes has been also been verified by viewing new design and changes in the documentation.
image


tujoworker:
Very very good job 🎉

I updated also the visual snapshots. They did not fail on CI, because the pixel change was 5.42% – that is under the ratio set during a CI run, because of font rendering differance (mac vs linux).

PR #778, opened 2021-03-25
fix: add to prop to Button for use when using routing libraries

asmadsen:
## Summary

Add `to` as a prop type to the Button, this is because when in use with routing libraries they often use `to` as opposed to `href`.

## Test plan

Tests are not really relevant since this will only result in changed types. However since Button snapshot tests use prop types to generate mock props the snapshop tests are updated.


tujoworker:
Thank You so much 👌🙏🏻

PR #803, opened 2021-04-10
feat: add progress indicator of type linear

langz:
## Summary

Adds a progress indicator of type linear.
The component is planned to be used in the new loans hub, and it's already created a "local component" for this in corporate.
https://bitbucket.tech.dnb.no/projects/CPP/repos/ccp-web-commons/browse/src/components/progress-bar

## Test plan

https://user-images.githubusercontent.com/1359205/114255210-35c1fa00-99b4-11eb-9598-6ec107e76808.mov

https://user-images.githubusercontent.com/1359205/114252607-5edb8e00-99a6-11eb-83ea-9c9e5030a1b9.mov

https://user-images.githubusercontent.com/1359205/115674538-0a44f500-a34e-11eb-9045-c756d17ace46.mov


packages/dnb-eufemia/src/components/progress-indicator/style/themes/dnb-progress-indicator-theme-ui.scss
```diff
@@ -44,4 +44,28 @@
       stroke: var(--color-emerald-green);
     }
   }
+  // circular variant
+  &__linear {
```
langz, line 48: The motivation for having different sizes(I personally only need default as of now), is to align with corporate, who seems to use different sizes already.

https://bitbucket.tech.dnb.no/projects/CPP/repos/ccp-web-commons/browse/src/components/progress-bar/ProgressBar.module.scss#23-41

packages/dnb-eufemia/src/components/progress-indicator/style/themes/dnb-progress-indicator-theme-ui.scss
```diff
@@ -44,4 +44,28 @@
...
+      height: 1.5rem;
+      border-radius: 0.75rem;
+    }
+    &--huge {
+      height: 2rem;
+      border-radius: 1rem;
+    }
+
+    &__bar {
+      background-color: var(--color-sea-green);
```
langz, line 68: as of now, I only need the color `color-sea-green`, but I could see us needing to support `color-fire-red` and `color-success-green`. See https://bitbucket.tech.dnb.no/projects/CPP/repos/ccp-web-commons/browse/src/components/progress-bar/ProgressBar.module.scss#49,53

What's the best way to to facilitate for this option? 🤔

packages/dnb-eufemia/src/components/progress-indicator/style/_progress-indicator.scss
```diff
@@ -105,6 +104,42 @@
...
     }
   }
+
+  // linear variant
+  &__linear {
+    position: relative;
+    overflow: hidden;
+    width: 100%;
+
+    // default/basis size
```
langz, line 114: The following default size (combination of `height` and `border-radius`) are currently not a subset/option of the available sizes(small, medium, large, huge) in dnb-progress-indicator-theme-ui.scss. 
I am not sure what's best?

The reason it was done this way is because i followed the same "pattern" as the circular type.

langz:
I see there have been some changes in the files **Pagination.test.js.snap**, and **Autocomplete.test.js.snap**.
Not sure if this is scary or not.

langz (commented):
I've looked through it, and it looks fine from my perspective :)

packages/dnb-eufemia/src/components/progress-indicator/ProgressIndicatorLinear.js
```diff
@@ -0,0 +1,155 @@
...
+    size: null,
+    visible: true,
+    complete: false,
+    progress: null,
+    onComplete: null,
+    callOnCompleteHandler: null
+  }
+  static getDerivedStateFromProps(props, state) {
+    state.progress = parseFloat(props.progress)
+    state.visible = props.visible
```
langz, line 31: Don't think I need this

packages/dnb-eufemia/src/components/progress-indicator/ProgressIndicatorLinear.js
```diff
@@ -0,0 +1,155 @@
...
+    state.complete = props.complete
```
langz, line 32: Don't think I need this

packages/dnb-eufemia/src/components/progress-indicator/style/themes/dnb-progress-indicator-theme-ui.scss
tujoworker: One way is to offer a prop like `color`, but use React inline `style` to set this color. But add example of using CSS vars:

```jsx
<ProgressIndicator
color="success-green"
/>
```

Which would translated to:

```jsx
<span
style={{color: `var(--color-${color})`}}
/>
```

I think this would be ok, even if IE is then not supported. Another option is to actually give variants. Like if a color is set, then have this in the component CSS. Which would be better to have just "some" options.

But should we add this later on, once it is actually needed?

tujoworker: Another option is for sure to translate it to:

```jsx
<span
style={{'--step-indicator-color': `var(--color-${color})`}}
/>
```

tujoworker:
Very very good work @langz!

I rebased from main, and made some small changes i an additional commit. Removed some demo examples. Huge is not actually a documented size, it's only used for development 😁

packages/dnb-eufemia/src/components/progress-indicator/style/themes/dnb-progress-indicator-theme-ui.scss
langz: Thanks for the input.
Yes, I am fine with waiting until it's actually needed, that sounds smart 👍

tujoworker (approved):
Thank you so much for the wonderful work you have done here 🙌💙

PR #805, opened 2021-04-10
chore: tertiary button requires an icon

langz:
## Summary

Tertiary button in global-status/Examples.js requires an icon.
Removed the following info message that occurs when visiting the portal for GlobalStatus: 
image


langz:
Not sure why continuous-integration/drone/push build is failing.
Also, when pressing the link(Details), I get a 404 - https://cloud.drone.io/dnbexperience/eufemia/3507

image

PR #821, opened 2021-05-04
fix: extend icon property description of the Button component

rodynenko:
## Summary

Extend Button's `icon` property description. Add details what icons can be used as a string.


tujoworker (approved):
Nice! Thank You 👍

PR #833, opened 2021-05-09
fix: change url to WCAG 2.1

langz:
Link to WCAG 2.1 did not work.
When clicking, it sent me to https://eufemia.dnb.no/design-system/www.w3.org/TR/WCAG21/

https://eufemia.dnb.no/design-system/about#standards


tujoworker:
AWESOME! 🙌

PR #834, opened 2021-05-09
fix: change contact email for Tobias

langz:
I think Tobias's email is incorrect, `_tobias.hoegh@dnb.no_.`

https://eufemia.dnb.no/design-system/contact


tujoworker:
AWESOMEx2! 🙌 (looks like I tried to make my self less work. Now this era has an end 😂)

PR #851, opened 2021-05-19
feat: add MediaQuery and useMediaQuery to @dnb/eufemia/shared

tujoworker:
## What
This PR will add two new shared components/hooks:

- `MediaQuery`
- `useMediaQuery`

## Why

Eufemia needs this helpers internal to make the new StepIndicator work properly as designed.

But also projects will get benefits by these helpers as I have seen several times the usage of `window.addEventListener('resize', ...)`.

## How

By using the [matchMedia](https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia) API we do not need to listen to e.g. `window.addEventListener('resize', ...)` which is a performance waste, even with a debounce helper.

By using `matchMedia` we only render when the requested media query actually changes.

The docs takes in place a new page in the layout section: `/uilib/usage/layout/media-queries`

## Usage

Example usage:

```jsx
import { MediaQuery } from '@dnb/eufemia/shared'

<MediaQuery ssr when={{ min: 'medium' }}>
  matches all above medium screens
</MediaQuery>
```

```jsx
import { useMediaQuery } from '@dnb/eufemia/shared'

function Component() {
  const match = useMediaQuery({
    ssr: true,
    when: { min: 'medium' },
  })

  return match ? 'true' : 'false'
}
```


tujoworker:
The big question is how to handle SSR:

While there are hacky methods to simulate the environment on the server and based on guesses, there is not good way to deliver a good solution to this problem.

Thats why I for now have just added the `ssr` property. If this one is set, it will match to true, if the `window` object is not available.

carlodiaz (commented):
Looks good, the only thing I noticed is that the breakpoints seem to be different to what is used on the open pages, but I suppose the ones here are the correct values...

tujoworker:
I added support for providing custom properties types:

```jsx
import { Provider } from '@dnb/eufemia/shared'
...
<Provider
  value={{
    breakpoints: {
      xsmall: '20em',
      medium: '30em',
      large: '60em',
    },
  }}
>
  ...
  <MediaQuery when={{ min: 'xsmall' }}>
    matches all above xsmall screens
  </MediaQuery>
  ...
</Provider>
```

carlodiaz (approved):

PR #853, opened 2021-05-20
fix: ensure that a button still has type button

tujoworker:
This PR fixes a wrong defined check for adding type

- Added a test to ensure we have type=button
- Because a button will else be able to submit a <form>
- An Anchor does not need that


rodynenko (approved):
Looks good

PR #856, opened 2021-05-21
fix: add role attribute to form-status

asmadsen:
## Summary

Add role attribute to FormStatus component, so when an "error" is shown it's picked up by a screen reader.

The FormStatus component will now add `role="alert"` unless the `disableRole` prop on FormStatus is set to `true`.

## Test plan

I have created two tests, testing that the `role="alert"` is only set if the `disableRole` prop is false.

I also added a new section in the storybook, that toggles the FormStatus when the submit button is clicked. On mac with VoiceOver this triggers the screen reader to read the contents of the status.


tujoworker:
Thank You very much! Will look into this one asap!

tujoworker (approved):
Looks good 🙌

PR #859, opened 2021-05-24
[WIP] feat: use ES with ES6 as the default module format

tujoworker:
This PR will make the Eufemia ES version as the default –  with ES6 as the basis. By now we used ESM, which compiled down to ES5.

In teori, this should be fine, as everyone I'm aware of uses a compiler / bundler.

This also fixes a codesandbox.io issue when importing modules from the package root as named imports.

```jsx
import { Button, DatePicker } from '@dnb/eufemia'
```


tujoworker:
Closing this for now. We may going for it on a later stage. For now we can't afford too many larger changes.

PR #863, opened 2021-05-28
feat: add #number-format `currency_breakpoint` to hide a currency sig…

tujoworker:
This PR enhances the NumberFormat component with an additional property  so the sign of a currency will be hidden at a given media query breakpoint (screen width). 

- It's an opt-in feature. It has to be set carefully by developers on places where it is very clear WHAT currency the number represents.
- It is made responsive, so it will react on screen size changes / device rotation.
- The breakpoint can be custom defined. Theoretical it could respond to more complex media queries like `prefers-color-scheme` or `orientation`. - But we keep the documentation to a basic usage and refer to the [Eufemia breakpoints](https://eufemia.dnb.no/uilib/usage/layout/media-queries#media-queries-properties-table).
- It uses `small` as the default, if true is given.
- It does not interfere with the "copy to clipboard" feature as.
- It will NOT remove the currency name/sign for screen readers.

The property name needs to be tied to "currency" as it is a feature only for currency signs.
And somehow it needs to tell that you can customize the breakpoints.
But better and more clear property names are welcome.

As of now, `currency_breakpoint` is used.

Not sure, maybe we should make it more clear in the docs somewhere, that; if you use this feature, it has to be done carefully, as the currency should be still reflected to the user, even if the user enters the page with a small screen.

Any good idea on how to write such a sentence? Or what a dev has to care about – e.g. if no UX designs are given?


tujoworker:
Heres a quick playground with that version: https://codesandbox.io/s/eufemia-responsive-currency-sign-d0qv2?file=/src/App.tsx

tujoworker:
Conditional hiding of information should be avoided from an accessibility standpoint of view. Use rather:

```jsx
<NumberFormat currency currency_display="" />
```

to hide the currency sign. But still give screen readers the full currency userd.

PR #901, opened 2021-06-18
fix: fix #number-format when percent should be displayed and the valu…

tujoworker:
# The problem
This PR fixes an issue when `percent` is used, but the given value has decimals. 

## Solution 
Now we check how many decimals the number has, and return the same amount, as long as not the amount of `decimals` are defined.


tujoworker:
Heres a quick CSB https://codesandbox.io/s/eufemia-starter-forked-6y9zt?file=/src/App.tsx

tujoworker:
Thank you Marte for the PR review!

PR #910, opened 2021-06-25
fix: fix #dropdown #autocomplete value on useEffect

tujoworker:
This fixes an issue where a Dropdown changes its initial state right after mount.

- Reproduction (Thank you Nils): https://codesandbox.io/embed/cocky-haibt-j2d0u
- Added a test to verify that.
- Will add a working code sandbox example as soon as a package is created by CSB.


OsloNils (approved):
Changes verified on CodeSandbox.

PR #929, opened 2021-07-23
fix: add variant to proptypes of dropdown

langz:
## Summary

The motivation for making this change is that we migrate our apps to TypeScript, and it seems like the variant prop of the Dropdown component is not included in the Dropdown's declaration file, as I believe this is generated from the component's proptypes where variant is missing. 
The property also seems like it's missing from the documentation/list of properties - https://eufemia.dnb.no/uilib/components/dropdown/properties, even though it exists demos of a Dropdown using the variant prop - https://eufemia.dnb.no/uilib/components/dropdown/demos#dropdown-as-tertiary-variant.

When using the variant prop in our TypeScript project, I get the following errors/warnings:
Screenshot 2021-07-23 at 10 10 44
Screenshot 2021-07-23 at 10 11 33


tujoworker:
Looks good! I committed some small additions in order to actually support the variants visually and to satisfy the snapshot tests. Also, merging it today, so you have tomorrow a version ready to use. Hope that this was ok?

Heres a quick CSB about the visual changes: https://codesandbox.io/s/eufemia-pr-929-dropdown-variant-4xdjx?file=/src/App.tsx:173-509

langz:
Sounds great, thanks for the assistance @tujoworker 👏

PR #932, opened 2021-08-02
fix: currency param should accept string and boolean values

langz:
The motivation for making this change is that we migrate our apps to TypeScript, and it seems like the currency param of the number format function only accepts a boolean value(according to the declaration file), but I believe/think the actual implementation accepts a string(Currency code (ISO 4217)) as well.

https://github.com/dnbexperience/eufemia/blob/e412541451c1bc1ab44aea645ae79e954bcde70f/packages/dnb-eufemia/src/components/number-format/NumberUtils.js#L45

https://github.com/dnbexperience/eufemia/blob/e412541451c1bc1ab44aea645ae79e954bcde70f/packages/dnb-eufemia/src/components/number-format/NumberUtils.js#L117

We use the number format function quite a few places in our apps, and especially in our tests 🙏

### DIFF
--- a/packages/dnb-eufemia/src/components/number-format/NumberUtils.d.ts
+++ b/packages/dnb-eufemia/src/components/number-format/NumberUtils.d.ts
@@ -50,8 +50,8 @@ export interface formatOptionParams {
   nin?: boolean;
   /** percent type */
   percent?: boolean;
-  /** currency type */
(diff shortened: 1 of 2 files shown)


tujoworker:
Look good to me 👍 nice work 🙌

PR #960, opened 2021-08-30
Enhance #auto-complete for whitespace and numeric usage

tujoworker:
This PR will enhance Autocomplete to act smarter:

- Introduce `search_numbers` so numeric entries will match if the users types numbers without being too precise and if it includes additional numeric chars like `.,-`
- Prioritize the result order of first written word chunks 
- Change the styling of results to have black-55 on second lines and always `basis` font-weight on first line
- ~~Typing a several words without whitespace will still match~~ we may use that in future, but for now, we release only the `search_numbers` feature.


tujoworker:
Thanks to Sivert Randen for the PR review and pair programming session

PR #968, opened 2021-09-06
fix: remove main role from modal and enhance usage of aria linking

tujoworker:
This PR contains changes to the Modal/Drawer component.

The reason: axe-core reports failure about the usage of several role="main"

- Remove `role="main"` on Mac/iOS/Safari
- All other enviroments will use `role="dialog"`
- If no modal title is given, a `aria-label="Dialog"` will be present, defiend in the locale/translation files


tujoworker:
Thanks to Pernilla Svedberg for repoting this issue! 🙏

PR #979, opened 2021-09-09
fix: add type definitions to stylis plugin

tujoworker:
Also remove unneeded tsconfig settings, like:

- Creating declaration files
- Creating source maps

Just because we only use tsc with --noEmit

### DIFF
--- a/packages/dnb-eufemia/tsconfig.json
+++ b/packages/dnb-eufemia/tsconfig.json
@@ -6,12 +6,10 @@
     "esModuleInterop": true,
     "allowSyntheticDefaultImports": true,
     "isolatedModules": true,
-    "declaration": true,
     "types": ["node", "jest"],
-    "sourceMap": true,
     "outDir": "./out",
     "rootDir": ".",
     "baseUrl": "."
   },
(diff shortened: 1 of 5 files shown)


packages/dnb-eufemia/tsconfig.json
```diff
@@ -6,12 +6,10 @@
...
     "isolatedModules": true,
-    "declaration": true,
     "types": ["node", "jest"],
-    "sourceMap": true,
     "outDir": "./out",
     "rootDir": ".",
     "baseUrl": "."
   },
-  "exclude": ["node_modules", "build"]
+  "exclude": ["./node_modules", "build"]
```
dinarosv, line 14: Isn't it necessary to add `./` in front of build as well here?

dinarosv (approved):
Yes looks good to me 👍

packages/dnb-eufemia/tsconfig.json
tujoworker: The [docs](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html) does not have it 🤷‍♂️

PR #1000, opened 2021-09-27
fix: typescript types for step-indicator

langz:
Seems like it was a ```s``` too much in the variable name of step indicator's TypeScript types.
```stepIndicatorPropsTypes``` vs. ```stepIndicatorPropTypes```.

I think this is what causes the definition file to not include the correct types.

_NB: Make sure you include *PropType in the variable name. This also effects references inside a single file._ - https://eufemia.dnb.no/uilib/development/types#sharing-proptypes-between-components


tujoworker:
Nice catch! Thank You 🙏

PR #1003, opened 2021-10-01
fix: fix doc for default value of GlobalStatus's show property

langz:
Just came over this while browsing the property documentation of GlobalStatus - https://eufemia.dnb.no/uilib/components/global-status/properties


tujoworker:
Thank You 🙏💙

PR #1029, opened 2021-10-11
fix: use id when setting data-dnb-modal-active

tujoworker:
**use id when setting data-dnb-modal-active**

Also, we now set it on the HTML element, instead of the body.

This makes it possible to access this state within the head tag, if needed (e.g. style tag with styles).


dinarosv (approved):
Was nice with the explanation of why this is changed 👍

PR #1030, opened 2021-10-11
fix: set data-dnb-drawer-list-active with id on HTML

tujoworker:
This PR make it possible to let CSS styles know if another drawer-list is "open" and what ID it has. 

We may considder to add docs for this as well.


dinarosv (approved):
👏

PR #1033, opened 2021-10-11
chore: lower visual tests threshold on CI

tujoworker:


dinarosv (approved):
Sounds like a good idea. Will it have any side effects?

PR #1035, opened 2021-10-11
fix(build): fix css minify and file naming of sass to css

tujoworker:
We now also test the whole filanmes including the extentions and the minification, just to make sure.


dinarosv (approved):
Good work, you figured that out quickly 🤩

PR #1037, opened 2021-10-11
chore: add yarn version to volta project

langz:
Not sure if we "need to" specify the yarn version(not sure if this is the correct version) for volta, but lately, whenever I run any yarn command, I always get the following message error.
```
Volta error: No Yarn version found in this project.
Use `volta pin yarn` to select a version (see `volta help pin` for more info).
```

Could just be that the problem exists between chair and keyboard, if then, feel free to delete this 🙏

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,8 @@
     "yarn": ">=1.17.0"
   },
   "volta": {
-    "node": "14.17.0"
+    "node": "14.17.0",
+    "yarn": "1.22.10"
   },
   "packageManager": "yarn@3.1.0-rc.8"
 }


tujoworker:
What Volta version do you have? run `volta -v`

I have 1.0.5 I did not seen the error. You can update Volta with that command: `curl https://get.volta.sh | bash`

The thing is, now that yarn is not on NPM, Volta will try to install it from NPM, as it does not have any custom instructions, like Node.js has. I bet there will be support for that in future.

So, Yarn v3 is not on NPM, we rather download it directly from source when running e.g. `yarn set version 3`. And that source, Volta could use to do the "same".

PR #1041, opened 2021-10-13
chore: make title of tabbar optional

langz:
Usually when running eufemia locally, `yarn start`, I get the following errors in the console:
image
This PR should fix these errors.


tujoworker:
I just had the same issue here as well. I had never really had time to find out whats happening. But now I had a little dive into that – before I saw your PR 🙏 And I found out that the graphql data layer did contain the title - but not the build, so I hope PR #1042 will probably solve this, even it could be a different plase as well. But an initial test looks promising.

Even the initial PR #1042 fixes someting different, I hope this small change from:

`global.nodesCache = global.nodesCache || {}` 

to

`global.nodesCache = {}`

will help.

PR #1094, opened 2021-11-03
chore: skip fragile visual tests

tujoworker:
I have seen this visual test fail too often, so we may skip it on CI. Not sure if then just the next one will fail. Probably. But from my experience, the href test has been a fragile thing before.


dinarosv (commented):
I probably just didn't understand the purpose of this... 
You said "skip it on CI", but here it only runs on the CI? or?

tujoworker:
You catched it 👍😍

dinarosv (approved):

PR #1095, opened 2021-11-03
chore: only run CI on push, and not on PR + push

tujoworker:
In order to "overwrite" existing branches, named `--visual-reports`, we can force push with lease as a solution.


dinarosv (approved):
Because CI runs already on push?

tujoworker:
yes, it does. We run it twice. Its very common to do this as so. But we do not need for now.

PR #1102, opened 2021-11-04
chore: refactor Portal data structure and sourcing

tujoworker:
This PR changes how the Portal source data. The old method was made before Gatsby had `createSchemaCustomization` and `createResolvers`, which should ensure we do not miss the "title" anymore, as this should be much more stable.


dinarosv (approved):
Nice! looking forward to this :D

PR #1107, opened 2021-11-09
Enhance Portal Setup

tujoworker:
This PR is made with Yarn PnP in mind – and it works fine, except that both VSCode integrations for ESLint and Prettier do not work properly with PnP. While there are other extensions which do, I think we go for the time being with node_modules.

All other enhancements are made with separate commits.


.vscode/settings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "search.exclude": {
+    "**/public": true,
+    "**/.cache": true,
+    "**/.yarn": true,
+    "**/.pnp.*": true
+  }
+}
```
dinarosv, line 8: These files are commited on purpose, right?

packages/dnb-design-system-portal/.ncurc.json
```diff
@@ -1,3 +1,12 @@
...
+    "eslint",
+    "ora",
+    "react-router-dom",
+    "stylelint",
+    "stylelint-config-prettier",
+    "stylelint-config-recommended",
+    "stylelint-config-standard",
+    "stylelint-scss"
+  ]
 }
```
dinarosv, line 3: What does this file do?

packages/dnb-design-system-portal/src/docs/design-system/changelog.md
```diff
@@ -5,8 +5,8 @@ draft: true
 order: 1
 ---
 
-import ChangeLog from 'Root/../../CHANGELOG.md'
-import ReleasesInfo from 'Pages/design-system/changelog/info-about-releases'
```
dinarosv, line 9: Great!

dinarosv (approved):
Great work!👏 And great amount of work 😄

.vscode/settings.json
tujoworker: the .yarn folder we have, the .pnp file we dont. But its just to let vscode know, what to ignore by default when you as a dev search the repo.

packages/dnb-design-system-portal/.ncurc.json
tujoworker: I may mention these in the docs! Even its not the purpose that other devs do update the deps. But they are a run command config for [ncu](https://www.npmjs.com/package/npm-check-updates).

dinarosv: Aaah, yes might be good to mention in the docs 🙌

PR #1109, opened 2021-11-10
chore: rename the Eufemia change to EUFEMIA_CHANGELOG.md

tujoworker:


dinarosv (commented):
Why are we changing this again? 🤔

tujoworker:
Just to avoid confusion about release change logs and Eufemia change log as a design system "change log". What do you think?

dinarosv (approved):

PR #1111, opened 2021-11-10
chore: refactor usage of alias'es inside the Portal

tujoworker:
This PR simplifies the usage of Webpack aliases inside and between the docs.


dinarosv (approved):
👍

PR #1112, opened 2021-11-10
chore: replace jsconfig with jsconfig

tujoworker:
VSCode uses this to let its internal TS to work, but it gets too little config, so we rather sue tscofnig and give it more details there.


dinarosv (approved):
Great! 🙌🌟

PR #1114, opened 2021-11-11
chore: split up GitHub Actions

tujoworker:
In order to lower the CI duration, this PR splits up the actions into several workflows. Now only some actions do use macos when running the visual tests and during icon fetch.
We also cache all `node_modules` on ubuntu directly and safe with that some seconds. A quick test shows that the same does take over 8 min on macos, so not an option.

The visual tests do only run when the commit does contain files inside `./packages/dnb-eufemia/src/**/*`. So they will not run when only docs are updated. Is this good?


.github/workflows/release.yml
```diff
@@ -0,0 +1,102 @@
...
+      - name: Release
+        if: (github.ref == 'refs/heads/release' ||
+          github.ref == 'refs/heads/beta' ||
+          github.ref == 'refs/heads/alpha')
+        run: yarn workspace @dnb/eufemia publish:ci
+
+      - name: Slack
+        uses: 8398a7/action-slack@v3
+        with:
+          channel: '#eufemia-dev'
```
dinarosv, line 95: Denne eufemia-dev slack kanalen er en lukket gruppe?

tujoworker: ja. Jeg la deg til 👍

dinarosv (approved):
Very good work 👏  

How much time did we save on this?

tujoworker:
Depends. When the visual tests run, then we are stil on 15 min – else we are down on 5 min

PR #1115, opened 2021-11-11
fix: fix Breadcrumb Storybook Docgen issue

tujoworker:
This PR makes Storybook running again. After we added tsconfig it had some problems due to the underlaing docgen usage and our ts setup.

Also, remove flag in gatsby-config for much faster dev startup time.


dinarosv (approved):
🙌

PR #1119, opened 2021-11-15
chore(Build): ensure we compile tsx to js but keep the original tsx files

tujoworker:
During a peer dev session I found an issue. `tsx` and `ts` files are transpoiled to js, but without chaing the file extention. 

What we expect is to have both. 

Create React Script v4 did work fine, while v3 did not. 

This PR fixes that.


dinarosv (approved):
👍  
Are we keeping the original tsx in the build?

tujoworker:
Yes, we are keeping the original tsx and ts files in the build

PR #1120, opened 2021-11-15
Fix Autocomplete data value handling

tujoworker:
This PR fixes two bugs – similar code wise, but with a different outcome. 
In both cases, we get a new `data` array, but used internally the old one, while using the new `value`.


dinarosv (approved):

OddKMS (approved):
Good stuff!

PR #1121, opened 2021-11-15
Use TypeScript ESLint parser and plugin

tujoworker:
This makes TS coding more plesent, as we now actually run ESLint with better/correct ESLint setup/config.


dinarosv (approved):
Smart!

PR #1125, opened 2021-11-15
chore(Portal): make React import optional

tujoworker:
This is a React v17 feature. With this PR we can use it today already. Not sure if I personally will use it that often. But its a feature of Gatsby 4.1, so why not enable it right.


dinarosv (approved):
Oh cool! Did not know this was a new thing from React 🌼

PR #1126, opened 2021-11-15
feat(InputMasked): ATTENTION! make allowLeadingZeroes default to true

tujoworker:
Eufemia did inherit this setting for text mask, the dependency.

## About the change

This looks like very simple change. But it can affect apps which do not want to allow negative zeros. Therefore the question. Shall we release this as a feature or as a breaking change? The latter requires more work. 

## The motivation? 

A project/app did release their forms without actively enable leading zeroes. But they actually needed it. Users could not enter low currency amounts. 

Having leading zeroes enabled, is probably a more common thing that forbid it? If you where using a vanilla input, it would be allowed, right? now, when it should really be forbidden from the front-end, then the devs should do both checks and tests for verify this behaviour.


dinarosv:
What is actually allow leading zeros? When typing `00034` or that the value becomes `00034` after typing as well?

dinarosv (approved):

tujoworker:
> What is actually allow leading zeros? When typing `00034` or that the value becomes `00034` after typing as well?

Yes ✨

PS: Hope you are doing well, Dina – and thanks again for all the work 🙏

packages/dnb-design-system-portal/src/docs/uilib/about-the-lib/releases/eufemia/v10-info.md
```diff
@@ -126,6 +126,10 @@ When you convert from `<Modal />` to `<Dialog />` – follow these steps:
 
 1. New **Definition List** layout direction: `direction="horizontal"` including `Dl.Item` [demo](https://eufemia.dnb.no/uilib/elements/lists/#definition-list-in-horizontal-direction).
 
+### [InputMasked](/uilib/components/input-masked)
+
+1. In v10, InputMasked **allows leading zeros**. To prevent that behaviour, the property `disallowLeadingZeroes` has changed to `disdisallowLeadingZeroes`.
```
langz, line 131:
```suggestion
1. In v10, InputMasked **allows leading zeros**. To prevent that behaviour, the property `allowLeadingZeroes` has changed to `disallowLeadingZeroes`.
```

packages/dnb-design-system-portal/src/docs/uilib/components/input-masked/properties.md
```diff
@@ -52,7 +52,7 @@ Defaults to Norwegian number format.
 | `integerLimit`              | (number): Limit the length of the integer number. Defaults to `null` for unlimited.                                                    |
 | `requireDecimal`            | (boolean): Whether or not to always include a decimal point and placeholder for decimal digits after the integer. Defaults to `false`. |
 | `allowNegative`             | (boolean): Whether or not to allow negative numbers. Defaults to `true`.                                                               |
-| `allowLeadingZeroes`        | (boolean): Whether or not to allow leading zeroes. Defaults to `false`.                                                                |
+| `disallowLeadingZeroes`     | (boolean): Whether or not to allow leading zeroes. Defaults to `false`.                                                                |
```
langz, line 55:
```suggestion
| `disallowLeadingZeroes`     | (boolean): Whether or not to disallow leading zeroes. Defaults to `false`.                                                                |
```

Not sure if we need to change this though 🤔

langz (approved):

PR #1127, opened 2021-11-16
fix(InputMasked): fix manual range selection

tujoworker:
When selecting partial or all input masked content, the selection got lost. This PR fixes this behaviour by checking if the user has selected more than one char.


dinarosv (approved):
Looks good 🌟

PR #1134, opened 2021-11-17
fix(Section): Text color on --fire-red-8

dinarosv:
Needed to change this for the new GlobalStatus. Considered it would be best as an independent PR.

Fire-red-8 is a very bright version of fire red, so changed default color to black-80 instead of white (for better contrast).


tujoworker (approved):
👌

PR #1136, opened 2021-11-17
fix(InputMasked): inputmode usage on Android phones

tujoworker:
Among issues like: a dot appears on the Android soft keyboard, but this lets the user not enter a coma.

Android sends just 229 on most of the keycodes when using the keyDown event.

It always returns 229 because it is a W3C standard to return 229 when an "Input Method Editor" is processing and the event is keydown.  It appeared in a 2010 W3C spec linked below.

A [codepen](https://codepen.io/ashconnolly/pen/WyWgPG?__cf_chl_captcha_tk__=30wzbiTaXd90vfTkcOT6ijFBIqzQToyJk6R65lT3k9U-1637164880-0-gaNycGzNB2U) on that topic.

Because the InputMasked is using the keycode in several ways – we have to do someting.
As of now, we simply avoid using `inputmode`.

Reprod from Marius K. [CSB](https://codesandbox.io/s/optimistic-herschel-uu43h?file=/src/App.jsx)

### DIFF
--- a/packages/dnb-eufemia/package.json
+++ b/packages/dnb-eufemia/package.json
@@ -95,7 +95,7 @@
     "classnames": "2.3.1",
     "core-js": "3.19.1",
     "date-fns": "2.25.0",
-    "keycode": "2.2.0",
+    "keycode": "2.2.1",
     "prop-types": "15.7.2",
     "what-input": "5.2.10"
   },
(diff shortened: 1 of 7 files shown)


dinarosv (approved):
Nice.
Have you tried it on an android phone too?

PR #1137, opened 2021-11-17
feat(GlobalStatus): New design

dinarosv:
Change the design for component GlobalStatus.

Check out the related [GlobalStatus - Figma](https://www.figma.com/file/PomHecnAnpJJQb9HPowWSa/?node-id=5556%3A16178) and the related [EDS-188 Jira Issue](https://jira.tech.dnb.no/browse/EDS-188).


packages/dnb-eufemia/src/components/global-status/style/_global-status.scss
```diff
@@ -46,18 +46,19 @@
...
 
       padding-top: 1rem;
       padding-bottom: 1rem;
-      padding-left: 3rem;
       padding-right: 5rem;
+      padding-left: 2.5rem;
+      margin-bottom: 0.5rem;
 
-      font-size: var(--font-size-medium);
+      // font-size: var(--font-size-medium);
```
tujoworker, line 61: This is not used anymore? Maybe remove it then ...

packages/dnb-eufemia/src/components/global-status/GlobalStatus.js
```diff
@@ -843,49 +854,45 @@ const isElementVisible = (elem, callback, delayFallback = 1e3) => {
...
+
 ErrorIcon.propTypes = {
   title: PropTypes.string,
 }
+
 ErrorIcon.defaultProps = {
   title: 'error',
 }
 
 const InfoIcon = (props) => (
```
tujoworker, line 891: Looks like the GlobalStatus has the same icons as the FromStatus now? 

https://github.com/dnbexperience/eufemia/blob/fb1aa76f4863899d18e84b7a0f7c6ba8d713dd38/packages/dnb-eufemia/src/components/form-status/FormStatus.js#L511

Because we already have docs on how to use the exported FormStatus icons, we probalby can use them from there, to avoid duplication. Also, these are size optimized.

tujoworker (commented):
Looks very good so far. Just a few things we have to considder/change.

packages/dnb-eufemia/src/components/global-status/style/_global-status.scss
dinarosv: Thanks for the review! Will fix this 🌟

packages/dnb-eufemia/src/components/global-status/GlobalStatus.js
dinarosv: Ahh, yeah that might be. I will fix this too 👍

tujoworker (approved):
Good work 🙌 You don't need it – but I suggest to make a rebase -i here and merge the two commits into one, because they belong togetter, I would argue ;)

PR #1138, opened 2021-11-18
fix(Breadcrumb): Remove padding on left and right side

dinarosv:
By mentioned in an [issue on Slack](https://dnb-it.slack.com/archives/CMXABCHEY/p1637238042059400?thread_ts=1637048941.041200&cid=CMXABCHEY), it is not desired to have padding on the sides of the Breadcrumb component. 

(Figma is also without this vertical padding, had a chat with Mats).

Also updated the related screenshots and snapshots.


tujoworker (approved):
Well done! 🤩

PR #1141, opened 2021-11-19
Fix(NumberFormat): types for NumberUtils

sivran:
export format and cleanNumber as functions and not types

## Summary

format and cleanNumber were exported as `type` instead of as a function.
This makes typescript complain about it being used as a type and not a value


tujoworker (approved):
Thanks for the PR 👌 I will add you as a dev in the repo 👍

Issue #876, opened 2021-06-04
Improve warning for tertiary button without icon property


langz:
## 🐛 Bug Report

This is just nitpicking, but it's actually a scenario I've experienced.

The following code `<Button text="Select" variant="tertiary"/>` results in Eufemia giving the following warning `"Eufemia: A Tertiary Button requires an icon. Please declare an icon to: null"`.

https://github.com/dnbexperience/eufemia/blob/cc23d02b71a659ed8beaeb9d15919bc89c5d436e/packages/dnb-eufemia/src/components/button/Button.js#L243

I interpreted this as "set the property icon to: null", therefore writing the following code `<Button text="Select" variant="tertiary" icon={null} />`, which will result in the same error again 😂 

## To Reproduce

https://codesandbox.io/s/new-silence-4qu50?file=/src/App.js
Then open the console.

## Expected behavior

Not exactly sure, but a clearer description of what to do in this scenario.

tujoworker:
Thanks for the very good report 😍
Yes, true. The reason is, because a tertiary button needs an icon to be WCAG valid (it has no underline) – but the message should be improved.

PR #1145, opened 2021-11-23
chore: Improve warning for tertiary button without icon (#876)

dinarosv:
Closing issue #876 

Updated warning message and also added more information in the docs.

### DIFF
--- a/packages/dnb-eufemia/src/components/button/Button.js
+++ b/packages/dnb-eufemia/src/components/button/Button.js
@@ -135,7 +135,8 @@ export default class Button extends React.PureComponent {
       icon !== false
     ) {
       warn(
-        `A Tertiary Button requires an icon. Please declare an icon to: ${content}`
+        `Icon required: A Tertiary Button requires an icon to be WCAG compliant in most cases, because variant tertiary has no underline.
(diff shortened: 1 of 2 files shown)


packages/dnb-eufemia/src/components/button/Button.js
```diff
@@ -135,7 +135,8 @@ export default class Button extends React.PureComponent {
       icon !== false
     ) {
       warn(
-        `A Tertiary Button requires an icon. Please declare an icon to: ${content}`
+        `Icon required: A Tertiary Button requires an icon to be WCAG compliant in most cases, because variant tertiary has no underline.
+        (Override this warning using icon={false}, or consider using one of the other variants)`
```
tujoworker, line 139: The indent can be remove, it can start all the way to the left. Else it will have spaces, instead of only a newline.

tujoworker (approved):
Looks good – had just a litle comment on the indent.

packages/dnb-eufemia/src/components/button/Button.js
dinarosv: Updated with this fix 🌟

PR #1148, opened 2021-11-23
Enhance type generation

tujoworker:
With this PR, its now possible to referance prop types as a static function or class property.


dinarosv:
So this was not possible before this commit? Has it something to do with the build/publishing stage?

tujoworker:
Yes, the generation of TypeScript definition files during build – it was already possible to link to other files with propTypes, but only with a named export/import – but now it's possible in several ways – and most importantly – `Button.propTypes` support from a default export/import. The docs shows some examples and updated info.

dinarosv (approved):
Nice!

PR #1149, opened 2021-11-23
feat(Inputs): use sea-green as the caret color

tujoworker:
– disscuss with UX first


tujoworker:
Closing this in favor of leaving it as is for now. This is based on the Eufemia meeting held 30.11.2021

PR #1150, opened 2021-11-24
fix(InputMasked): only use inputmode=numeric/decimal when safe to use

tujoworker:


dinarosv (approved):
Great work! You found a solution to the Android-bug too 😁

PR #1152, opened 2021-11-25
Fix/input focus color

tujoworker:
This PR fixes an issue where the focus state on inputs had a wrong color (sea-green -> emerald-green). Now, with that change, I run the visual tests with `colorThreshold` of 0 – and found that there are a good amount of snaptshos we should have updated before, but they where so suptile, that they dident got picked up. But its added on an extra commit.


dinarosv (approved):
Wow, even I could not see the difference 😄

tujoworker:
Yes, no body has repored it, I just saw it when I changed the caret color on the textarea, that the textarea had an darker outline during focus state.

tujoworker:
Here is the diff:

Screenshot 2021-11-25 at 11 20 19

PR #1154, opened 2021-11-26
fix(Breadcrumb): Spacing prop + contrast colors

dinarosv:
Changes:
- Spacing disabled by default, have to send in prop `spacing` to include padding in both breadcrumb and the following collapse 
- Added some contrast colors, because the breadcrumb disappeared when using `styleType="sea-green`. ~~It is still not optimal with contrast colors on hover and focus, but that is a bigger change I believe...~~

(Also changed spacing to be identical top and bottom, have to wait for *go* on this from the designers)


packages/dnb-eufemia/src/components/section/style/themes/dnb-section-theme-ui.scss
```diff
@@ -95,14 +95,16 @@
...
+  &--fire-red .dnb-button--tertiary,
+  &--emerald-green .dnb-button--tertiary,
+  &--sea-green .dnb-button--tertiary {
     color: var(--color-white);
     &:hover {
-      color: var(--color-sea-green);
+      color: var(--color-white);
     }
     &:focus {
       color: var(--color-white);
```
tujoworker, line 107: We need to have the fakeFocus set to white, else it has too low contrast on dark backgorunds, and the focus ring is even not visible on emerald-green.

Screenshot 2021-11-30 at 12 28 36

tujoworker (approved):
Except for the focus ring contrast – it looks good to me. Good work Dina 🌼

packages/dnb-eufemia/src/components/section/style/themes/dnb-section-theme-ui.scss
dinarosv: Better now? 😊

PR #1158, opened 2021-12-01
Enhance Modal / Drawer Performance

tujoworker:
This PR aims to enhacne performance issues when opening the Modal / Drawer while a large amount of table rows are already in the DOM.

The first approach was to lower the amount of DOM maniulations during the invalidatione phase. But;

- Tests confirm that we can't remove `tabindex=-1` 
- And the same tests do also confirm, that making the attribute changes do take actualy the lowest amount of all CPU usage. On a table with 10 000 rows, it takes under 0.5s

So what takes the most CPU? Its the browser re-paint phase. Why does the browser need to make a repoiant when the Modal opens and closes? Because it changes every visible pixle:

- we need to change some styles on the body and html element to lock scrolling
- we show an overlay with opacity

Now, this PR still includes a couple of enhancements:

- We delay the `tabindex=-1` until the modal is opened (sadly, with 10 000 table rows, we still have almost the same delay)
- Use opacity for the overlay (its hardware accelerated) – instead of background-color with a alpha channel (opacity is hardware accelerated by defualt, and do not need the will-change property)

🍀 Now, what can project do to enhance their amount of DOM content? There are a couple of options. One is to use a virtualized scroller for that particular table – which only puts the visible parts in to the DOM.

### DIFF
--- a/packages/dnb-eufemia/src/components/modal/ModalContent.js
+++ b/packages/dnb-eufemia/src/components/modal/ModalContent.js
@@ -121,24 +121,65 @@ export default class ModalContent extends React.PureComponent {
   constructor(props) {
     super(props)
     this._contentRef = React.createRef()
+
+    // NB: The ""._id" is used in the __modalStack as "last._id"
     this._id = props.id
-    this._ii = new InteractionInvalidation()
-    this._ii.setBypassSelector([
-      '.dnb-modal__content',
-      `#dnb-modal-${props.root_id || 'root'}`,
-    ])
   }
 
   componentDidMount() {
+    // Add it to the index at first
+    // we use it later with getListOfModalRoots
     this.addToIndex()
 
+    // Because of nested modals/drawers, we run this regardless
+    // has to be run at first – so the scrollbar gets removed
+    this.removeScrollPossibility() // forces browser to re-paint
+
+    this.setFocus()
+    this.setAndroidFocusHelper()
+
+    const id = this.props.id
+    dispatchCustomElementEvent(this, 'on_open', {
+      id,
+    })
+
+    if (
+      isTrue(this.props.no_animation) ||
+      process.env.NODE_ENV === 'test'
+    ) {
+      this.lockBody() // forces browser to re-paint
+    } else {
+      this._lockTimeout = setTimeout(() => {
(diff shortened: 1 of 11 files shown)


packages/dnb-eufemia/src/components/modal/ModalContent.js
```diff
@@ -121,24 +121,65 @@ export default class ModalContent extends React.PureComponent {
...
   }
 
   componentDidMount() {
+    // Add it to the index at first
+    // we use it later with getListOfModalRoots
     this.addToIndex()
 
+    // remove it every time, because of nested modals/drawers
+    // has to be run at first – so the scrollbar gets removed
+    this.removeScrollPossibility() // force browser to re-paint
```
dinarosv, line 136: Is it this part that is the time consuming part? Because of the comment

packages/dnb-eufemia/src/shared/helpers.js
```diff
@@ -421,3 +418,14 @@ export async function copyToClipboard(string) {
...
 }
+
+export const warn = (...e) => {
+  if (
+    typeof process !== 'undefined' &&
+    process.env.NODE_ENV !== 'production' &&
+    typeof console !== 'undefined' &&
+    typeof console.log === 'function'
+  ) {
+    console.log('Eufemia:', ...e)
```
dinarosv, line 429: Have you considered using console.warn(...)?

tujoworker: Yes – but some browsers will show the trace stack as well, which takes a huge amount of place in the console.

dinarosv (approved):
Good work finding the source of the problem 🌟 
Had some comments, but that's more for my own understanding 😅

packages/dnb-eufemia/src/components/modal/ModalContent.js
tujoworker: I will re-write the comments 🤦‍♂️ but yes.

packages/dnb-eufemia/src/shared/helpers.js
dinarosv: Aaah, I understand 👍

mariuskalvo:
LGTM 🚀

PR #1163, opened 2021-12-03
[WIP] feat: autocomplete limit

### LINKED ISSUE #1100: Autocomplete limit rendered items
In short: filtering a relatively long array is fast / cheap - but rendering a lot of elements is expensive.

Im running into performance problems with the autocomplete component where i have a long list of options.

Is there any option to limit the automcomplete component to only render e.g. max 25 items at a time when suggesting?

(I know i could implement a on_type ahndler and pre-filter the array and send to updateData - but it feels like functionality that might allready exist but i cant find, or one that more users might find usefull)

tujoworker:
Closes #1100


vron (approved):
lgtm - but not very knowledgeable about the general codebase..

tujoworker:
Right now I play arround of not using a limit – but rather a driver for a thid party virtual scroller – because it works ok when filtering, because of the re-ordering – but what of you show everthing at first? A list with 10 000+ will then be show to scroll through.

Now, with a virtualized scroller, only the needed items are rendered – its just, have a almost Plug & Play method would be best – any thoughts on that?

PR #1165, opened 2021-12-06
chore(CI): Check if PR is fork during Slack notify

tujoworker:


dinarosv (approved):
Nice solution 👍  So it will not post a slack message if it comes from a fork now? Or will it post from the main branch later anyways?

tujoworker:
It will not notify. From what I have read, it fails because of the fact that the GH_TOKEN is not accessible on forks (internal token). Now, we could get around this by providing a custom one I think. But it’s both with the work for now I argue.

PR #1167, opened 2021-12-06
feat(Modal): Update Drawer styles

dinarosv:
Updated appearance of Drawer (Modal) to match the new Figma design.

Still need to:
- [ ] Update Modal docs with Breadcrumb navigation


dinarosv:
Styles will be updated after the new logic of Modal/Drawer is implemented, because the change from Figma was larger than expected and regards logic in the components too

PR #1176, opened 2021-12-14
chore(Portal): include publish version in portal footer

### LINKED ISSUE #1060: Display version of Eufemia in the Portal
## 🚀 Feature Proposal

Every day, many developers use the portal to figure out how to use components of Eufemia.
When doing so, it would be really awesome to see which version of Eufemia that's used in the portal they are viewing.

A common scenario seen is that developers are not sure if the version of Eufemia used in the portal is the same as used in their project or if it's the latest. This sometimes cause confusion regarding versions, new/old props, changed props, styling, etc.

## A suggestion for displaying version of Eufemia in the portal

Screenshot 2021-10-22 at 10 20 53
Screenshot 2021-10-22 at 10 21 03

If any have some thoughts regarding making this possible, please add it as a comment, and I or someone else could perhaps look into this 🙏

tujoworker:
closes #1060

This PR will include the release version in the footer of the portal. It will only display a version if the current branch is one of the release branches.
Else it will display `Not released` (I'm glad for better ideas there 😀)

It will look like this:

Screenshot 2022-02-02 at 08 52 32

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -64,7 +64,6 @@ steps:
       branch:
         include:
           - release
-          - release-*
           - portal
           - beta
           - alpha
@@ -99,7 +98,6 @@ steps:
       branch:
         include:
           - release
-          - release-*
           - portal
           - beta
           - alpha
@@ -166,7 +164,6 @@ steps:
         include:
           - main
           - release
-          - release-*
           - beta
           - alpha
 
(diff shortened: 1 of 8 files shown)


dinarosv (approved):
Good work! 🌟  I think this was a great idea!

PR #1177, opened 2021-12-14
Update Modal to Typescript

dinarosv:


packages/dnb-eufemia/src/components/Section.d.ts
```diff
@@ -0,0 +1,14 @@
...
+ */
+
+/**
+ * Library Index section to autogenerate all the components and extensions
+ * Used by "prepareSections"
+ */
+
+import Section from './section/Section'
+export * from './section/Section'
+export default Section
```
tujoworker, line 14: I guess we do not need this file for now, as it is only referancing to files? Could you check that out?

packages/dnb-eufemia/src/components/button/Button.d.ts
```diff
@@ -0,0 +1,229 @@
+import * as React from 'react';
```
tujoworker, line 1: We should really create a routine, where we update all the docs, now that we add type definition files. We will else go qickly out of sync.

packages/dnb-eufemia/src/components/modal/__tests__/__snapshots__/Modal.test.tsx.snap
```diff
@@ -5,69 +5,7 @@ exports[`Modal component have to match snapshot 1`] = `
...
-  1={
-    Object {
-      "displayName": "ModalRoot",
-      "props": Object {
-        "children": "children",
-        "direct_dom_return": "direct_dom_return",
-        "id": "id",
-        "root_id": "root_id",
-      },
+      "props": Object {},
```
tujoworker, line 8: The lib tells it should support TS: https://github.com/typicode/react-fake-props

But maybe the types are too advanced?

packages/dnb-eufemia/src/components/step-indicator/__tests__/__snapshots__/StepIndicator.test.js.snap
```diff
@@ -2622,7 +2622,6 @@ Array [
             overlay_class={null}
             prevent_close={false}
             prevent_core_style={false}
-            root_id="root"
```
tujoworker, line 2625: Do you know why the `root_id="root"` is missing? Maybe we do not fall back to a default=root somewhere?

tujoworker (approved):
Very good work – Dina! ✨

My comments are not holding back this PR – given that everything works. Lets hope the tests do cover all use cases 🤞

packages/dnb-eufemia/src/components/button/Button.d.ts
dinarosv: So the `d.ts` files will not be updated on build if we commit them?

I agree with the components in typescript, an idea is to extract the commented descriptions to a markdown table in docs.. But that is of course more work

packages/dnb-eufemia/src/components/modal/__tests__/__snapshots__/Modal.test.tsx.snap
dinarosv: Yes that might be.. Do you think it can affect the tests in a bad way?

packages/dnb-eufemia/src/components/step-indicator/__tests__/__snapshots__/StepIndicator.test.js.snap
dinarosv: Fixed it 👍

tujoworker: nice ✨ 🎉

dinarosv:
I wrote down the mock prop generation and docs routine for a later discussion 👍

PR #1179, opened 2021-12-15
fix: StepIndicator clickable NVDA issue

tujoworker:
NVDA reports that a non clickable StepIndicator step is clickable. This is only due the fact that NVDA checks is there is an click event listener attached to the DOM element. This PR ensures we do not add a listener when not needed.


dinarosv (approved):
Nice work 🙏

PR #1180, opened 2021-12-15
fix(InputMasked): allow leading zero when custom mask is given

tujoworker:
This PR fixes an issue where the user could not enter a leading zero when a custom mask was used.


dinarosv (approved):
SÅ bra du kunne fikse det! 🌟🌟👏

PR #1183, opened 2021-12-16
feat(Tag): Add onClick prop

thaytharma:
Adds `onClick`-prop to Tag.

It uses the `Button`-component (unstyled) as base.


langz:
Seems like the visual tests are failing, do we need to generate new snapshots?
https://github.com/dnbexperience/eufemia/runs/4545510995?check_suite_focus=true

`yarn test:screenshots:update` - https://github.com/dnbexperience/eufemia/blob/main/package.json#L28

packages/dnb-eufemia/src/components/tag/style/_tag.scss
```diff
@@ -16,11 +16,40 @@
...
   }
+
   &__text {
-    color: var(--color-black-80);
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+  }
+
+  &--clickable {
```
langz, line 34: For all of this styling, is there a way to reuse what's been done in other "clickable" elements?
Like reusing mixins like `@include hover()`, `@include focus()`, `@include active()`, etc?

langz, line 34: Something about this styling that makes it look strange in my version of Safari:
image

langz (changes requested):
Awesome work @thaytharma 🚀

packages/dnb-eufemia/src/components/tag/style/_tag.scss
thaytharma: Thanks for the heads-up! This has been updated 😊

langz: https://eufemia.dnb.no/uilib/helpers/classes#sass-and-mixins

langz: Looks great now :)

langz: Not sure if we should reuse these mixins or do as we did here, adding our own selectors, like `&:hover`

langz (approved):

packages/dnb-eufemia/src/components/tag/style/_tag.scss
tujoworker: If we can wait until starting of jan 22 – I would like to have a peer dev session. Basically two topics:

- Overall accessibility with tags and what semantic to use
- Try "one more time" to use the button component – it gives so much for free (like href support with correct sematnics)

thaytharma: Sure! 😃  Let's do that @tujoworker 👍

thaytharma: @tujoworker 
This PR be re-done to using `<Button variant="unstyled" ../>`. 
Can you do a re-review when you get a chance? 😊

dinarosv (approved):
Like this component a lot! 🤩  
I'm thinking we might need more information on when to use and when to not use the clickable tag, so no one puts them where a Button would be more appropriate 💭

tujoworker:
Indeed! Very good work! I will have a last check with Theyan on Friday – just to verify that my commits are ok 👍

tujoworker (approved):
Indeed! Very good work! I will have a last check with Theyan on Friday – just to verify that my commits are ok 👍

PR #1184, opened 2021-12-16
release of v9.18.0

tujoworker:


dinarosv (approved):
Are we releasing now because of the new component? :)

PR #1185, opened 2021-12-16
fix(InputMasked): inherit of provider/context props

tujoworker:
The InputMasked did not properly handle provider/context props as well as accepting local props when a custom mask only is used. This PR fixes both.


dinarosv (approved):
I like the new change! Looks clean 🌟👍

PR #1188, opened 2022-01-03
fix(Breadcrumb): forward props in BreadcrumbItem to span as well

dinarosv:
Fixing a reported issue when sending down props to `<Breadcrumb.Item data-testid="breadcrumb-item-card" onClick={null}/>` without interaction. Check out the [slack thread](https://dnb-it.slack.com/archives/CMXABCHEY/p1640792268004600).

Successfully ran tests and screenshot tests, as well as manual testing 👍


langz (approved):
👍 ✅

PR #1189, opened 2022-01-04
fix(Breadcrumb): improve accessibility

thaytharma:
Improved accessibility of the current breadcrumb-item by using [aria-current="page"](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current) instead of aria-current="true".

From MDN: _"...In a breadcrumb list, when a link within a set of pagination links is styled to indicate the user is currently on that page, aria-current="page" should be set on that link"_


tujoworker:
Awesome – Thanks 🙌

tujoworker (approved):

PR #1190, opened 2022-01-05
[WIP] feat: Separate Drawer and Dialog from Modal

dinarosv:
The first purpose of this work was to separate Drawer into a new component folder, to get two smaller components instead of Modal handling everything (including tests). 

The most valuable purpose for this PR is to make it easier to change Dialog and Drawer in the future, both in styling and logic. There are incoming style changes for both these components, with information/confirmation variants for Dialog and styling update for Drawer.

I also appreciate feedback on the logic solution (the separation), what could be done better?

## Dialog and Drawer are Modal variations
Browsing and trying out different solutions which had to provide backward compatibility, resulted in this solution. The Modal is now the logic behind Drawer and Dialog components (much like the [Material UI's Modal](https://mui.com/components/modal/)). For backward compatibility, meaning the usage of `<Modal mode="drawer" />`, the components come with a `<DialogContent>`/`<DrawerContent>` which is imported into Modal. The Dialog is simply a `<Modal>` wrapping a `<DialogContent>`. 

But we still have to include the original tests and documentation for Modal with the two modes, so that V1 still can be used. This results in some redundant code before we release a breaking change.

## Props
The new components Drawer and Dialog now use camelCase props, while Modal still use snake_case. Because we want new components to use camelCase, but I did not want to change the Modal component too much in this PR.

## Parts
Drawer and Dialog comes with different parts, similar to `Modal.Content`. I changed some of the parts names, to more accurately fit the purposes they have. Drawer now come with `Drawer.Navigation`(i.e. HeaderBar), `Drawer.Header`, `Drawer.Body`(i.e. Content), and Dialog has the same.


tujoworker (changes requested):
Very nice work! Just have some comments ...

tujoworker (approved):
This PR looks good 🙌 Impressive work, Dina ✨

PR #1191, opened 2022-01-07
feat: add Avatar component

langz:
## Summary

Link to Figma: https://www.figma.com/file/cdtwQD8IJ7pTeE45U148r1/Eufemia---Web?node-id=17869%3A0

TODO in the future:
- Incorporate "badges" (see Figma sketches):
image


langz (commented):
To me it looks okay, just have to verify the sizes with what's used in pm-netbank today.

packages/dnb-design-system-portal/src/docs/uilib/components/avatar/properties.md
```diff
@@ -0,0 +1,19 @@
...
+showTabs: true
+---
+
+## Properties
+
+### `Avatar` properties
+
+| Properties                                  | Description                                                                                                                       |
+| ------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------- |
+| `size`                                      | _(optional)_ Size of the avatar. Options: `small` \| `medium` \| `large` \| `x-large`.                                            |
```
thaytharma, line 11: Suggestion: 

It seems like the `default`-size is the same size as the `medium`. Should we consider removing `default` and use `medium` as default value instead? Also, I think it's beneficial to update the description here with the default size.

packages/dnb-design-system-portal/src/docs/uilib/components/avatar/properties.md
```diff
@@ -0,0 +1,19 @@
...
+| `children`                                  | _(optional)_ Content of the component.                                                                                            |
+| `alt`                                       | _(optional)_ Used in combination with `src` to provide an alt attribute for the `img` element.                                    |
+| `src`                                       | _(optional)_ Specifies the path to the image                                                                                      |
+| `skeleton`                                  | _(optional)_ Applies loading skeleton.                                                                                            |
+| `imgProps`                                  | _(optional)_ [Image properties](/uilib/elements/image) applied to the `img` element if the component is used to display an image. |
+| `variant`                                   | _(optional)_ Override the variant of the component. Options: `primary` \| `secondary` \| `tertiary`.                              |
```
thaytharma, line 17:
```suggestion
| `variant`                                   | _(optional)_ Override the variant of the component. Options: `primary` \| `secondary` \| `tertiary`. Default to `primary`.                             |
```

thaytharma:
Looks great! 🤩

packages/dnb-design-system-portal/src/docs/uilib/components/avatar/properties.md
langz: Correct, the default/fallback size when not providing any size is `medium`.
As far as I see it, the value  `default` is not supported for the size property. So "removing `default`" shouldn't be needed/possible.

I'll add a description/text explaining that the default value of size is `medium` ✅

packages/dnb-design-system-portal/src/docs/uilib/components/avatar/properties.md
langz: Nice find, I'll add that 🙇

packages/dnb-design-system-portal/src/docs/uilib/components/avatar/properties.md
thaytharma: Perect!

tujoworker (approved):
Yes, indeed! Awesome work 🤩

tujoworker:
The only thing we need to improve is the requirement for additional, screen reader information. E.g. to require with a prop more info?

langz:
> The only thing we need to improve is the requirement for additional, screen reader information. E.g. to require with a prop more info?

@tujoworker, I am not sure if I fully understood this.
Could you try to elaborate in a bit more details? Which attributes/properties are we thinking of, etc?
Then I could try to implement this improvement.

PR #1193, opened 2022-01-12
fix: fixed global context bug that removed component props

dinarosv:
Global context did not work on Typescript components because of object ordering

### DIFF
--- a/packages/dnb-eufemia/src/shared/Context.js
+++ b/packages/dnb-eufemia/src/shared/Context.js
@@ -46,12 +46,13 @@ export const prepareContext = (props = {}) => {
       return context.translation
     },
     locales,
-    ...props,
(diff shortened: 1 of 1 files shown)


packages/dnb-eufemia/src/shared/Context.js
```diff
@@ -46,12 +46,13 @@ export const prepareContext = (props = {}) => {
...
     },
     locales,
-    ...props,
     translation, // make sure we set this after props, since we update this one!
     // All eufemia components because of Typescript:
     Breadcrumb: {},
     BreadcrumbItem: {},
     Tag: {},
+
+    ...props,
```
tujoworker, line 55: Translation should still come after ...props

tujoworker (approved):
Nice 👍

PR #1196, opened 2022-01-13
fix: fix reset_button_text property of DatePicker

langz:
Doesn't seem like property `reset_button_text` of DatePicker was used at all 🤔 

The reported bug can be read here: https://dnb-it.slack.com/archives/CMXABCHEY/p1642090852041200
CodeSandbox: https://codesandbox.io/s/eufemia-starter-forked-isxjv?file=/src/App.tsx

### DIFF
--- a/packages/dnb-eufemia/src/components/date-picker/DatePickerFooter.js
+++ b/packages/dnb-eufemia/src/components/date-picker/DatePickerFooter.js
@@ -19,12 +19,15 @@ export default class DatePickerFooter extends React.PureComponent {
     onSubmit: PropTypes.func,
     onCancel: PropTypes.func,
     onReset: PropTypes.func,
+
+    resetButtonText: PropTypes.string,
   }
 
   static defaultProps = {
     onSubmit: null,
     onCancel: null,
     onReset: null,
+    resetButtonText: null,
   }
 
   onSubmitHandler = (args) => {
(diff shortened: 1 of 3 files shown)


dinarosv (approved):
Nice work 💯 🤩  That was quick!

packages/dnb-eufemia/src/components/date-picker/DatePickerFooter.js
```diff
@@ -100,8 +103,11 @@ export default class DatePickerFooter extends React.PureComponent {
       return <></>
     }
 
-    const { submit_button_text, cancel_button_text, reset_button_text } =
-      this.context.translation.DatePicker
+    const {
+      submit_button_text,
+      cancel_button_text,
+      reset_button_text: reset_button_text_translation,
```
ahmetacer5, line 109: I think we don't need renaming here.

langz: True.
The reason for renaming was to try to differentiate between the translation, and the prop.
By keeping it as `reset_button_text` we would have both `reset_button_text` and `resetButtonText` which to me became a bit confusing, as in my head it's the exact same names just in two different formats.

Now with the renaming, we have `reset_button_text_translation`, and `resetButtonText`.

That was my thoughts at least.

ahmetacer5: yes that's right I didn't think that way 👍

PR #1197, opened 2022-01-14
fix: fix rendering of DatePickerFooter when show_reset_button is prov…

langz:
DatePickerFooter wasn't rendered when providing `show_reset_button={true}`.
This PR renders DatePickerFooter when `show_reset_button={true}` is provided to DatePicker.


dinarosv (approved):
🌟 🌟

PR #1199, opened 2022-01-14
chore: add missing components to components docs

langz:
Seems like we've forgotten about updating this page when adding new components, https://eufemia.dnb.no/uilib/components

The description texts used for the components(see [here](https://eufemia-choreaddmissingcomponentstocom.gtsb.io/uilib/components)), is just copy+pasted from the description in their respective .md file. These descriptions could be improved, suggestions are welcomed 🙏


thaytharma (approved):
Well done! 🚀

PR #1200, opened 2022-01-17
fix: added export of Bar in Pagination

dinarosv:
Pagination (mode pagination) rendered loading spinner as content when no children was given. 

Added a test for the reported issue as well

### DIFF
--- a/packages/dnb-eufemia/src/components/pagination/Pagination.js
+++ b/packages/dnb-eufemia/src/components/pagination/Pagination.js
@@ -320,6 +320,10 @@ Pagination.Content = PaginationContent
 const PaginationWrapper = Pagination
 const InfinityMarkerWrapper = InfinityMarker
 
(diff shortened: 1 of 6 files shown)


tujoworker (approved):
Good work 👍

packages/dnb-eufemia/src/components/pagination/Pagination.js
```diff
@@ -320,6 +320,10 @@ Pagination.Content = PaginationContent
 const PaginationWrapper = Pagination
 const InfinityMarkerWrapper = InfinityMarker
 
+export const Bar = (props) => (
```
tujoworker, line 323: Having a comment about, why we export it, or at least, that it is not documented, and that a dev rather should get in touch with us, than just using it, would be beneficial 👍

tujoworker (approved):

PR #1202, opened 2022-01-17
feat: ensue ESM imports are fully specified

tujoworker:
Right now, there is an issue sing Eufemia with Create React App v5 because it now enforces all node modules to be fully specified – if they have `"type": "module"` inside their `package.json`.

We have three options in order to make it work:
1. we remove `"type": "module"`
2. during build we transform all of our files to include an extension by using [this plugin](https://github.com/tujoworker/fully-specified/tree/main/babel-plugin-fully-specified) (the current state of that PR)
3. Use/create an ESLint plugin that ensures we actually add the extension in the code.

All these options are based on that we continue to run the dev code in CJS as of today. A fourth option could have been to define `"type": "module"` inside the repo. But that requires the most work.

### Option 1.

What Webpack/Node.js on the other hand needs to do, is to lookup "what" extension was meant to use, because it's not specified. 

### Option 2.

It can potentially still fail. Even Webpack actually loads all files, and shows us all errors, not just the "next" one, which again gives me more confidence, that it works fine.

### Option 3.

This needs some habituation from every dev. Also, currently, TS warns us about not adding an extension, so we need to change that somehow.

It has to fail the build, when the lint rule errors. Also, if it includes a "fix" feature, then we can run this on all files and make the change for all existing files in one commit.

Also here we gain performance during app development and their builds.

Screenshot 2022-01-19 at 20 16 22

Screenshot 2022-01-19 at 20 15 48

Any thoughts?


langz:
Great work looking into this 🎊 
I like option 1 seen in the context of a potential release v9.19.0 today(26.01.2022), but I think I would prefer option 3.
A reason for me liking option 3 the most, is that it seems to be the most "direct approach", not hiding it behind builds that magically transform files, but rather putting the responsibility in the hands of all devs. These are just my thoughts 🤔

langz (approved):
Great work looking into this 🎊 
I like option 1 seen in the context of a potential release v9.19.0 today(26.01.2022), but I think I would prefer option 3.
A reason for me liking option 3 the most, is that it seems to be the most "direct approach", not hiding it behind builds that magically transform files, but rather putting the responsibility in the hands of all devs. These are just my thoughts 🤔

tujoworker:
@langz Yes - I agree and share your thoughts! Sadly, it will require a lot of work, especially the parts, where we, as of today, would have to import things from `.../shared/index.js` - instead of `.../shared`.

And more issues:

- I checked it out, and could not find any solution to the ESLint/TypeScript issue, where TS complains when we add an extension to TS files.
- And second, we still could have to find a solution, to rename then `*.tsx` to `*.js` during build time. And there we need a babel plugin anyway.

packages/dnb-eufemia/src/components/avatar/Avatar.tsx
```diff
@@ -15,13 +15,10 @@ import {
   extendPropsWithContext,
   warn,
 } from '../../shared/component-helper'
-import { Img } from '../..'
 
 // Internal
 import AvatarGroup, { AvatarGroupContext } from './AvatarGroup'
 
-export * from './AvatarGroup'
```
dinarosv, line 23: Why are there changes in Avatar (and in Tag) in this PR? 💭

packages/dnb-eufemia/package.json
```diff
@@ -145,6 +144,7 @@
     "@wojtekmaj/enzyme-adapter-react-17": "0.6.5",
     "audit-ci": "5.1.2",
     "babel-jest": "27.3.1",
+    "babel-plugin-fully-specified": "1.3.0",
```
dinarosv, line 147: Do we need this package in option 1(this PR)?

dinarosv (commented):
Super that you found out of this ⭐ I think we can live with this solution for a while...

I just have two comments on some of the code 👍

packages/dnb-eufemia/src/components/avatar/Avatar.tsx
tujoworker: They generate a circular dependency issue when using Rollup.js – and the build now fails. I found that the shell scripts did not fail, even not failed tests. But now they will.

packages/dnb-eufemia/package.json
tujoworker: I intended for option 2 – most arguments go this way in my eyes. 

With option 1 – we make actually a large change in the eye of Node.js/Webpack. We move from ESM back to CJS, but with ESM imports. Even Webpack did not care about our faulty imports before (with no extensions).

So with option 2 we fix that issue.

PR #1205, opened 2022-01-19
Fix dialog being omitted in accessability tree

thebreiflabb:
This omits the modal root from getting
aria-hidden="true" added to it, which
removes the modal from the accessibility tree

## Summary

The modal root component got `aria-hidden="true"` added to it when opening the dialog, rendering the content inside not accessible in the accessibility tree.

This adds additional ignore of the modal root component in the `aria-hidden` handler logic.

## Test plan

I added a new test, for checking that a button outside still gets `aria-hidden="true"`, and the modal root no longer get it. However I could not get the test to work, I tried both using `document.querySelector` and `Comp.find` as seen in the tried test. As well as with and without the `{attachTo: attachToBody()}` option.

I have verified the described effect/change in Storybook, as well as when using it as a component in our application.

### DIFF
--- a/packages/dnb-eufemia/src/components/modal/ModalContent.tsx
+++ b/packages/dnb-eufemia/src/components/modal/ModalContent.tsx
@@ -124,6 +124,7 @@ export default class ModalContent extends React.PureComponent<
         [
           // Bypass modal content
           '.dnb-modal__content *',
+          `#dnb-modal-${this.props.root_id || 'root'}`,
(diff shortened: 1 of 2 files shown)


tujoworker:
Thank You Sondre for the good PR ❤️ so far. I will check with @dinarosv if we should wait before this large [PR](https://github.com/dnbexperience/eufemia/pull/1190) or not.

tujoworker (approved):
Thanks again for the PR 🙏

I fixed the tests and we will take this inn now 👍

PR #1207, opened 2022-01-21
chore: remove unused style for payment card

langz:
Doesn't seem like the following variable is used, ```--dnb-payment-shadow-figma```


dinarosv (approved):
Yup, you are correct! Nice catch 🌟

PR #1210, opened 2022-01-23
docs: update "new" status

thaytharma:
Removed "new" status from docs for components that are older than 1-2 years.


tujoworker (approved):
Looks good - super duper  🤩

PR #1211, opened 2022-01-23
docs: remove Upload

thaytharma:
Deleted Upload-component from docs for now, as it has been marked as "wip" for some time.
I did not see any tasks for this for Jira. Feel free to decline this PR if this still should be a "wip".


tujoworker (approved):
It has been in the drawing board for years - by no code has been made so far.

It’s an important component we should get it

The motivation for marking it as wip is that devs get in touch and ask, so we get a meaning on prioritization. 

But I think we can remove it for now, as you suggest with this pr.

PR #1215, opened 2022-01-25
fix(DrawerList): ensure DrawerList is unsetting its global attributes on unmount

tujoworker:
This PR ensures the DrawerList will "reset" global attributes is has set during its open event.


langz (approved):
Looks great 👍

dinarosv (approved):
🙌

PR #1216, opened 2022-01-25
fix syntax highlighting

tujoworker:
This PR fixes both some outdated docs about the font usage and ensures there will be a scroll-bar when not enough space is left to show the content inside the `<code>` element.


dinarosv (approved):
Nice fix 🌟 🌟

langz (approved):

PR #1217, opened 2022-01-25
feat(Tag): add onDelete-prop

thaytharma:
Added `onDelete`-prop to `Tag`-component.


langz (approved):
I think it looks very nice 💯 

I have one comment that's not directly related to this:
I see that in a lot of our demos/examples we use `space={{right: 'x-small'}}`, etc. 
When not doing it, it seems to be a bit "crammed":
image

Should perhaps the tag group add some spacing as standard, and removing the spacing on the last item?
Might be that it's intentional to have this little space between the tags though, if so, then ignore my thoughts 👯

andrewjrhill:
🤩   Looks great.

thaytharma:
> I think it looks very nice 💯
> 
> I have one comment that's not directly related to this: I see that in a lot of our demos/examples we use `space={{right: 'x-small'}}`, etc. When not doing it, it seems to be a bit "crammed": image
> 
> Should perhaps the tag group add some spacing as standard, and removing the spacing on the last item? Might be that it's intentional to have this little space between the tags though, if so, then ignore my thoughts 👯

Great suggestion! Will have a discussion with the designers about the default group spacing etc.

PR #1218, opened 2022-01-25
fix: use discriminated union for props alt and src

langz:
Not sure if we want this or not.
This is just a suggestion to make a property required when an other property is available.
In these examples, either both `alt` and `src` is required, otherwise both are optional.
So if only providing `alt` or `src`, typescript will tell us that the missing property is required.

### DIFF
--- a/packages/dnb-eufemia/src/components/avatar/Avatar.tsx
+++ b/packages/dnb-eufemia/src/components/avatar/Avatar.tsx
@@ -22,16 +22,17 @@ import AvatarGroup, { AvatarGroupContext } from './AvatarGroup'
 
 export * from './AvatarGroup'
 
+type AltSrcProps =
+  | {
+      alt?: string
+      src?: string
+    }
+  | { alt: string; src: string }
+
 export type AvatarSizes = 'small' | 'medium' | 'large' | 'x-large'
 export type AvatarVariants = 'primary' | 'secondary' | 'tertiary'
 
 export interface AvatarProps {
-  /**
-   * Used in combination with `src` to provide an alt attribute for the `img` element.
-   * Default: null
-   */
-  alt?: string
-
   /**
    * Custom className on the component root
    * Default: null
@@ -57,12 +58,6 @@ export interface AvatarProps {
   size?: AvatarSizes
 
   /**
-   * Specifies the path to the image
-   * Default: null
-   */
-  src?: string
-
-  /**
    * Props applied to the `img` element if the component is used to display an image.
    * Default: null
    */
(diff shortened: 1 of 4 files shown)


packages/dnb-eufemia/src/components/avatar/Avatar.tsx
```diff
@@ -22,16 +22,17 @@ import AvatarGroup, { AvatarGroupContext } from './AvatarGroup'
...
+
 export type AvatarSizes = 'small' | 'medium' | 'large' | 'x-large'
 export type AvatarVariants = 'primary' | 'secondary' | 'tertiary'
 
 export interface AvatarProps {
-  /**
-   * Used in combination with `src` to provide an alt attribute for the `img` element.
-   * Default: null
-   */
-  alt?: string
```
tujoworker, line 33: Could this work instead?

1. define it
```tsx
type AltSrcType =
  | {
      alt?: string
      src?: string
    }
  | { alt: string; src: string }
```

2. use it, instead of `string`
```tsx
alt?: AltSrcType
```

This way we would still have the overall property description as before.

langz: Hmm, wouldn't this mean that the alt property would expect to be like this?
`<Avatar alt={{ alt: 'custom_alt_label', src: 's' }} />`? 🤔

tujoworker: You are right. I just thought it was so much extra code for this to make it happen.
But actually – I looked into it, and I could not this to get work.
I would say, it would be within the term of [conditional types](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html).

Even if I provide all props as an "either / or". It just does not give the result we want.

tujoworker: I found a solution!

tujoworker: We can use `never` on alt and src and extend the either/or with Omit:

Screenshot 2022-02-03 at 14 22 02

tujoworker: We can use never on alt and src and extend the either/or with Omit:

Screenshot 2022-02-03 at 14 23 38

This works 👍

tujoworker: I think, we should close this PR. I mean, it works. But its not nice to maintain, and also, the `never` type is kind of hard to grasp, just because we want this dependency between several props.
I hope TypeScript will come with a better solution in the future. Then we may rewrite these types.

langz: Sounds good to me 💯

PR #1221, opened 2022-01-25
fix(StepIndicator): fix height animation issue when CSS is loaded too slow

tujoworker:
This PR fixes an issue as described like so:

> The findings where that Chromium loads and runs the the JS earlier than the styles. So the button has initially 33px (no styling) and changes almost frame by frame down to 27px during the rendering phase, before the styles pops in place – but that was too late, much later than the didMount event, so the component did use and set the lower height.

The fix is basically, that we ensure, the element gets an of `height: auto` when the from/to are the same values. This way, we let chrome do its thing, but only care about animation, when the values are not equal.


dinarosv (approved):
Looks very good 😃

PR #1225, opened 2022-01-26
fix(Tag): remove data-prop from TagGroup

thaytharma:
Removed data-prop from TagGroup after a discussion with @tujoworker 😊
EDIT: This change is done to be able to avoid circular dependencies.


tujoworker (approved):
Nice Work ✨

for those who have not been in the meeting: the reason is to be able to avoid circular dependencies 👍

PR #1226, opened 2022-01-26
chore: make Avatar inline

tujoworker:
This PR makes the Avatar component usable in body text.

Here an example of a whole Avatar "group":

Screenshot 2022-01-26 at 13 47 16


langz (approved):
Thanks, looks nice 🙇

PR #1227, opened 2022-01-26
chore: ensure correct element order and use z-index instead

tujoworker:
This is because of the order a screen reader consumes the content.


langz (approved):
🙇

PR #1229, opened 2022-01-26
fix(TagGroup): make Tag component inline, so it can be used in body text

tujoworker:
Making the TagGroup component inline, should not effect layout issues, when used with flex or grid.

This is how it is with this commit:

Screenshot 2022-01-26 at 22 33 09

This is how it was before:
Screenshot 2022-01-26 at 22 13 50

Also, because this commit moves some "it" inside "description" we have to delete/rewrite the snapshots


langz (approved):
Nice nice 👍 
For context, the same was done for Avatar in https://github.com/dnbexperience/eufemia/pull/1226

thaytharma (approved):
Looks great 😇

PR #1230, opened 2022-01-27
fix(CodeHighlighting): ensure a scrollbar is shown when not enough space is available (pre tag)

tujoworker:
I completely missed the correct tag. The scrollbars / overflow should have been added to the pre tag and not the code tag 🤦‍♂️


dinarosv (approved):
Oh! I also didn't notice this in the last PR 😄

PR #1232, opened 2022-01-27
portal: update portal Algolia deps and visual test (CI setup)

tujoworker:
This PR updates

- Algolia deps and enhances the indexName handling
- Ensure we always run visual test


dinarosv (approved):
Sounds okay, did you check that nothing broke after the update? 👍

tujoworker:
> Sounds okay, did you check that nothing broke after the update? 👍

Looks like we did not run the visual tests, so we also ensure now they run when `yarn.lock` changes.

PR #1233, opened 2022-01-27
chore(repo-utils): add repo-utils to check CI instead of external dependency

tujoworker:
This PR fixes a GitHub Actions change, where they lately changes the CI env variable from being a boolean to a string. This is not based on research, but from own tests.

We change with this PR to our own check.

### DIFF
--- a/packages/dnb-design-system-portal/src/uilib/search/searchHelpers.js
+++ b/packages/dnb-design-system-portal/src/uilib/search/searchHelpers.js
@@ -1,4 +1,4 @@
-const { isCI } = require('ci-info')
+const { isCI } = require('repo-utils')
 
 // Finds current index name for the Algolia search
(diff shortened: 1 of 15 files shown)


packages/dnb-design-system-portal/src/uilib/search/searchHelpers.js
```diff
@@ -1,4 +1,5 @@
-const { isCI } = require('ci-info')
+const { isCI } = require('repo-utils')
+console.log('isCI', isCI)
```
dinarosv, line 2: Should the `console.log` stay here?

dinarosv (commented):
Nice ⭐ 

Do you have a good reason for adding another package just for the isCI check? It could not be a local file?

dinarosv (approved):

PR #1234, opened 2022-01-28
feat: add usePropsWithContext hook for hook based components

tujoworker:
This PR is all about better TS support.

The new hook `usePropsWithContext` replaces `extendPropsWithContext` and returns the type given as the first property. This ensures the new object returned functions with the wanted type. 

As an example, the `Tag.tsx ` did use a not typed variant of unstyled – it was valid because we got back `any`.

EDIT: This PR relies on PR #1235


dinarosv (approved):
Good work 👏 😄

PR #1235, opened 2022-01-30
chore: fix CI setup

tujoworker:
This commit fixes a couple of issues related to the last CI setup changes and isCI check

It also moves the isCI test to the build check, instead of the verify stage. I think it rather belongs there.


dinarosv (approved):
Looks good to me! Nice catch 👏

PR #1236, opened 2022-01-31
chore: es6 functions in new components

thaytharma:
This PR changes our new components to use ES6-functions. We tend to add new components based on these files, so we might apply them now if we want to keep it this way.

This might just be a personal preference, 
so please feel free to decline this PR.

### DIFF
--- a/packages/dnb-eufemia/src/components/avatar/Avatar.tsx
+++ b/packages/dnb-eufemia/src/components/avatar/Avatar.tsx
@@ -83,7 +83,7 @@ export const defaultProps = {
   children: null,
 }
 
-function Avatar(localProps: AvatarProps & ISpacingProps) {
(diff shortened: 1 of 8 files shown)


tujoworker (changes requested):
I'm fine with it. I think its good that you guys form the decision around such styles.

Could you squash the last two commits into the first one? Or you will squash it during the PR merge?

tujoworker (approved):

thaytharma:
> I'm fine with it. I think its good that you guys form the decision around such styles.
> 
> Could you squash the last two commits into the first one? Or you will squash it during the PR merge?

Thanks! Squashed and merged 🙌🏼

dinarosv:
I have no particular meaning about functions vs. const for components, so we can go with this 👍

PR #1237, opened 2022-01-31
chore: remove yarn-plugin-ignore-install-options

tujoworker:
Gatsby Cloud now supports Yarn v2 (and with that, it should support v3 as well)

Based on this feedback: https://gatsby.canny.io/gatsby-cloud/p/support-yarn-2-and-pnp-api

Background:

Because Gatsby Cloud did run its own install command, with with flags, which got removed in yarn v2 and v3 - so the install command could not get executed.

To overcome this, I created a plugin, that removed all install flags. But with this PR we remove this plugin.


dinarosv (approved):
Looks good to me 🌟

PR #1238, opened 2022-01-31
feat(TypeScript): enhance overall typing and add verify check

tujoworker:
This PR adds a new check in the CI pipeline, where all types get verified. But it also includes a couple of type fixes.

Run `yarn test:types` during development or even `yarn test:types:watch`.

This PR should be merged with all the commits separated, as several include a fix decoration in order to be a part of the change log.


packages/dnb-eufemia/src/components/timeline/Timeline.tsx
```diff
@@ -67,12 +68,8 @@ const Timeline = (localProps: TimelineProps & ISpacingProps) => {
...
       {...props}
     >
-      {data?.map((timelineItem: TimelineItemProps) => (
-        <TimelineItem
-          key={timelineItem.name}
-          skeleton={skeleton}
-          {...timelineItem}
-        />
+      {data?.map((timelineItem, i) => (
+        <TimelineItem key={i} skeleton={skeleton} {...timelineItem} />
```
dinarosv, line 72: You should have gotten a warning here with index is not permitted as a key?

dinarosv (approved):
Nice TS-work! 🤩 

Had one small comment for index in mapping, but it's only nitpicking 👍

packages/dnb-eufemia/src/components/timeline/Timeline.tsx
tujoworker: We used `Item.name` there before – but I think it is defined as a ReactNode, but that gave us an error. Using i here is ok as well.

dinarosv: Thats true, the order of the data doesn't probably change here. 
I just checked out [Keys | React docs](https://reactjs.org/docs/lists-and-keys.html#keys), where they say you _can_ use index as key (actually it's default if you don't provide a key) - but it is still not recommended if the order of the list changes. No worries here though 😊

PR #1240, opened 2022-01-31
docs: remove signal-orange from Section

thaytharma:
Looks like `signal-orange` is a deprecated `style_type` for `Section`.

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/components/section/properties.md
+++ b/packages/dnb-design-system-portal/src/docs/uilib/components/section/properties.md
@@ -25,7 +25,6 @@ You can easily [customize the color](/uilib/components/section#customize-color).
 | `pistachio`     | uses `--color-pistachio`.                                                                           |
 | `black-3`       | uses `--color-black-3`.                                                                             |
 | `emerald-green` | uses `--color-emerald-green`.                                                                       |
-| `signal-orange` | uses `--color-signal-orange`.                                                                       |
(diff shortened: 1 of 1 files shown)


langz:
I am not sure about this, I see that if I tried setting style_type to signal-orange, it worked:
image

I also see we have some perhaps related code in https://github.com/dnbexperience/eufemia/blob/main/packages/dnb-eufemia/src/components/section/style/themes/dnb-section-theme-ui.scss#L123-L131

But it's probably correct/safe to make the change you've done 🤔 

@tujoworker, can you take a look at this?

thaytharma:
> I am not sure about this, I see that if I tried setting style_type to signal-orange, it worked: image
> 
> I also see we have some perhaps related code in https://github.com/dnbexperience/eufemia/blob/main/packages/dnb-eufemia/src/components/section/style/themes/dnb-section-theme-ui.scss#L123-L131
> 
> But it's probably correct/safe to make the change you've done 🤔
> 
> @tujoworker, can you take a look at this?

Yes, it works, but it looks like the styling isn't working as expected. The text color has some contrast issues, especially with the link being sea-green. I was expecting white. 🤔

@tujoworker What do you think?

tujoworker (approved):
Yes! Thanks! Nice you found this one ❤️

tujoworker:
Screenshot 2022-02-02 at 11 54 40

Its deprecated – so we may remove the css later 👍

PR #1243, opened 2022-01-31
chore: retry up to 5 times when visual tests fails

tujoworker:
This commit will ensue:

If a visual test fails, jest will retry it on the failed one, up to five times.

How did I test it?

1. I could not find any indication that jest-screenshot not uses the default test runner
2. changed the retryTimes to 300
3. changed button examples to something invalid
4. run the visual button tests

The result:

It kept waiting, until I reverted the change – because the tests where run a the dev portal, it changed the example in runtime and the test did succeed.


dinarosv (approved):
Nice 🤩

tujoworker:
Nice - it looks like it works fine 👍

Screenshot 2022-02-01 at 15 56 26

PR #1244, opened 2022-02-01
fix: label of avatar group to be react node

langz:


thaytharma (approved):
Great great! 
Should we update the docs (`properties.md`) as well?

langz:
> Great great! Should we update the docs (`properties.md`) as well?

Yeah, was thinking the same, but I feel that the description _The label description of the group of avatars_ still fits well.

thaytharma:
> > Great great! Should we update the docs (`properties.md`) as well?
> 
> Yeah, was thinking the same, but I feel that the description _The label description of the group of avatars_ still fits perfectly well.
> 
> Not sure if we want or need more specifics regarding the type of the property description?

We can leave it like this. I just see other components add the "valid" types to the description.

PR #1245, opened 2022-02-01
fix: label of tag group to be react node

langz:


thaytharma (approved):
Thanks for fixing this!
Adding the same suggestion as in `AvatarGroup` https://github.com/dnbexperience/eufemia/pull/1244#pullrequestreview-868899371

PR #1246, opened 2022-02-01
feat(VisuallyHidden): Add VisuallyHidden component

thaytharma:
Added `VisuallyHidden`-component

I started with using `dnb-sr-only`-class, but ended up using the the`srOnly`-mixin instead as I felt it was a "cleaner".
I also added a `focusable`-option which I saw being used in other libraries like in [Bootstrap](https://getbootstrap.com/docs/5.0/helpers/visually-hidden/)

I'm open to suggestions as this is an initial version of the component 😊


packages/dnb-eufemia/src/components/visually-hidden/VisuallyHidden.tsx
```diff
@@ -0,0 +1,74 @@
...
+   * Hide an element by default, but to display it when it’s focused (e.g. by a keyboard-only user)
+   * Default: false
+   */
+  focusable?: boolean
+
+  /**
+   * Root element of the component
+   * Default: div
+   */
+  element?: string
```
tujoworker, line 31: Maybe use ReactNode here? This way its even possible to send in an existing component.


```jsx
const Component = () => <div>...</div>
render(<VisuallyHidden element={Component}>hidden content</VisuallyHidden>)
```

tujoworker (approved):
This is awesome! Fantastic work ✨

packages/dnb-eufemia/src/components/visually-hidden/VisuallyHidden.tsx
thaytharma: Great suggestion!
This has been updated 🚀

PR #1251, opened 2022-02-02
feat: update icons: credit_note, deleted_invoice, speedometer, travel

tujoworker:
Update latest icons.

I notices that the speedometer svg was huge. So we need to change that in the Figma file first.

### DIFF
--- a/packages/dnb-eufemia/src/components/input-masked/InputMaskedHooks.js
+++ b/packages/dnb-eufemia/src/components/input-masked/InputMaskedHooks.js
@@ -13,6 +13,7 @@ import {
   isTrue,
   dispatchCustomElementEvent,
 } from '../../shared/component-helper'
+import { safeSetSelection } from './text-mask/createTextMaskInputElement'
 
 import TextMask from './TextMask'
 import createNumberMask from './addons/createNumberMask'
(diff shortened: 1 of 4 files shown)


packages/dnb-eufemia/src/icons/icons-meta.json
```diff
@@ -1403,6 +1403,36 @@
...
+    "variant": "secondary"
+  },
+  "calculator": {
+    "tags": [],
+    "created": 1643787002153,
+    "name": "objects/calculator_16",
+    "variant": "secondary"
+  },
+  "travel": {
+    "tags": ["insurance", "plane"],
```
thaytharma, line 1419:
```suggestion
    "tags": ["insurance", "plane", "flight"],
```

Small suggestion!

tujoworker: Thank you @thaytharma – I will forward this suggestion to Mats – our system grabs the SVGs directly from Figma via an API and transforms them.

We have just to figure out why the speedometer size gets so huge.

thaytharma: Oh, I see! That sounds good 🚀

PR #1253, opened 2022-02-02
fix(InfoCard): change text props to React.ReactNode

langz:


AadneRo (approved):

thaytharma (approved):
Great 🙌🏼

PR #1254, opened 2022-02-02
fix(AvatarGroup): remove left border from first avatar

langz:
This PR removes the left border from the first avatar in an avatar group.
Reason for wanting to remove the left border is to not always have 2px border(looks like a margin) on the left side of the first avatar in an avatar group. We could see that this added some "unwanted spacing" on the left side of the avatar when used in pm-netbank.


thaytharma (approved):
Looks good! 🙌🏼

PR #1256, opened 2022-02-02
docs: fix menu order in side-menu

tujoworker:
This commit fixes an issue where automatic ordering gets messed up. By using the same logic as before, it works fine after increasing the increment starting number.

Also, I think fragments can still be at the bottom. Any thoughts on that?


dinarosv (approved):
Very nice 🙌

tujoworker:
The second commit will ensure we still get the correct color / aria-current on the actual item – if a sub page is shown. This not only applies to the new nested components, but was an issue before anyway.

Screenshot 2022-02-02 at 13 19 24

PR #1260, opened 2022-02-02
chore(Release): switch order of portal deploy with release action

tujoworker:
This commit ensures we release first, and then deploy the portal. In case the release is not successful – we abort and can retry. Its easier to manually deploy the portal at its own via the portal branch.


dinarosv (approved):
Probably smart to do it this way, yes 👌

Issue #1118, opened 2021-11-12
Autocomplete selecting wrong item


vron:
Hi,

I am still struggling with details of the autocomplete component (in async mode searching towards a server that can return different results at different times).

1. When typing to search (first time after loading a page - does not seem to reoccure once i have allready selected) i get some options. If i click the first item the call "await pricingStore.setClient(e.data.selected_key" from below is called with the correct selected_key (as is setVal call one the line above). *However* the autocomplete will show the selected_key from the second item in the list, despite the "console.log("rendering", val);" call showing that the last render was with the val from the first entry.

(I am also 90% sure that it sometimes renders with a completely different value - but can't seem to reliably re-create it now)

2. calling emptyData in the onTypeHandler will completely remove the input the user has typed - I though it should only clear the suggestions data

I am rendering it thus:

    export const ClientFind = observer(() => {
        const [val, setVal] = useState('');
        console.log("rendering", val);
        return (
            <Autocomplete
                value={val}
                mode="async"
                on_type={onTypeHandler}
                on_change={action(async (e) => {
                    if (e.data === undefined) {
                        setVal('');
                        return await pricingStore.setClient(undefined);
                    }
                    setVal(e.data.selected_key);
                    await pricingStore.setClient(e.data.selected_key);
                })}
                no_scroll_animation="true"
            />
        )
    });

using a on_type handler thus:

        const onTypeHandler = ({
            value,
            showIndicator,
            hideIndicator,
            updateData,
            debounce,
            emptyData
        }: any) => {
            showIndicator();
            debounce(
                async ({ value }: any) => {
                    let res = await clientsStore.findClients(value);
                    let data = [...res.map(c => ({
                        selected_key: c.name,
                        selected_value: c.name,
                        content: [
                            c.name,
                            <NumberFormat org>{c.orgno}</NumberFormat>,
                        ],
                    }))];
                    //emptyData();
                    updateData(data);
        
                    hideIndicator();
                }
                , value }, 250
            )
        }

vron:
I should add to make it clear - it "mostly works" since the callback on_change is called with the correct item - so i get the correct key to use further in my logic.

It seems to be after the on_change callback is finished that the autocomplete itself changes the displayed value (the correct one flashes past)

tujoworker:
Can You check if this version works?

Just replace the **verison number** of Eufemia in `package.json` with this https://pkg.csb.dev/dnbexperience/eufemia/commit/12cce683/@dnb/eufemia

[...]

PR #1265, opened 2022-02-03
chore: add test regarding issue #1118

tujoworker:
Just to have a dedicated test for this case as well – ##1118


dinarosv (approved):
👍  smart!

PR #1268, opened 2022-02-04
fix: fix InputMasked delete key usage

### LINKED ISSUE #1113: InputMasked and del key
Hi,

I'm sorry - I keep reporting stuff I finde - hopefully it's usefull, if not just tell me to shut up :-)

So the input masked is working much better - but one small thing we notice:

If you have foprmated number in with with space sepperation, e.g 5 000 000:

If you now place the cursor just after the 5 (before the space) and press the [del] key on the keyboard nothing happens.

I would expect it to delete one 0 and re-format the number with spaces in the correct places.

(typicall occureswhen the user want to delete a lot of tailing numbers, will then "meet" a space sooner or later and not be able to delete further)

tujoworker:
closes #1113


dinarosv (approved):
Looks good to me 😄

PR #1269, opened 2022-02-04
feat(Version): run Eufemia.version in your browser console

tujoworker:
I often wanted to have an easy way to check the Eufemia version while debugging. Now, that we check what version we release in beforehand, we can also make it available by typing in the browser console:

`Eufemia.version`


dinarosv (approved):
This is a really nice feature 👏

PR #1270, opened 2022-02-04
fix(NumberUtils): "replace" did not exists on IE11 

tujoworker:
More details about the issue [here](https://dnb-it.slack.com/archives/CMXABCHEY/p1643204645078400)


dinarosv (approved):
Tricky bug, nice that you fixed this 👏

PR #1271, opened 2022-02-04
docs: fix change log styling

tujoworker:
While we moved the EUFEMIA_CHANGELOG.md from the docs into the repo root, we missed our styles, because the mdx transformer does not work when the file is located outside. I tried to make some changes to the configs, but with no success. Anyway, there are good arguments to keep it in the repo root, then we fill find a solution, else we move it back into the /docs folder with this PR.

Screenshot 2022-02-04 at 15 24 38


dinarosv (approved):
This looks good! 😃 
Think we have forgotten about the changelog for a while, the new components should've been added?

PR #1277, opened 2022-02-07
Fix Avatar circle and logo 

tujoworker:
This PR removes the usage of border to ensure we get a perfect circle. Instead we use box-shadow (FakeBorder).

Before
Screenshot 2022-02-07 at 18 21 17

After
Screenshot 2022-02-07 at 18 44 33

It also centers the logo component

Before
Screenshot 2022-02-07 at 18 52 28

After
Screenshot 2022-02-07 at 18 52 52


langz (approved):
Thanks, looks nice :)
I am just a bit unsure about the visual diff here, https://github.com/dnbexperience/eufemia/pull/1277/files#diff-cdc3575d637f3e18889f79b3e7e309b3601dc79cfc751ae2d9e87719e5a1368c, where it looks like the border/overlay for the 1st avatar is different.

image

tujoworker:
@langz ohh, nice catch! Its probably the last one which will need `box-shadow: none` ?

langz:
> @langz ohh, nice catch! Its probably the last one which will need `box-shadow: none` ? Or just leave it there?

I am not exactly sure what will fix this 🤔 
If the designers think it looks great, then I am also fine with leaving it as it is now.

dinarosv (approved):
🌟 🌟 
So that's the reason behind using box-shadow instead of border, nice! 🤯

dinarosv:
After watching your comments, I discovered it is still doing this for one of the avatar groups:
Skjermbilde 2022-02-08 kl  08 53 09

PR #1279, opened 2022-02-08
feat(Tabs): add no_border prop which removes tablist border

thaytharma:
Added `no_border`-prop to remove the default bottom border under the tablist.


tujoworker (approved):
Fantastic! ✨🤩

PR #1282, opened 2022-02-08
fix(Modal): ensure IE11 is placing the close button to the right side

tujoworker:
[CSB](https://codesandbox.io/s/eufemia-modal-ie11-bug-r8upy?file=/src/App.tsx) to test the PR version on.

Before
Screenshot 2022-02-08 at 20 46 02

After
Screenshot 2022-02-09 at 12 17 44


dinarosv (approved):
👍

PR #1283, opened 2022-02-09
fix: css change to fix aliasing artifacts for ProgressIndicator

JulianNymark:
## Summary

There is an aliasing artifact (checked on both FF and chrome) when 2 same sized svg shapes are drawn on top of one another, it's not really visible when drawing 'dark on light', but it's more so when 'light on dark'. This is somewhat of a quickfix (slightly hacky?). So there's probably a more "proper" solution by not overlapping shapes/paths of similar size, animating both the increasing light shape but also the decreasing dark shape?

I think loading indicators are some of the first things users will notice on any element / page that requires loading, and they are usually the "only thing on the page" surrounded by whitespace, and if it looks "cheap" like bucket fill in ms-paint with 'edge aliasing artifacts', then that probably leaves a negative impression :)

image below should show the 'artifacts'.

image

## Test plan

I tested it by looking at the component before and after in the storybook, running it by going into the sandbox directory and running `yarn run start`. I tried to follow what I could find at the contribution guidelines, but hopefully this is a decent starting point :)

### DIFF
--- a/packages/dnb-eufemia/src/components/progress-indicator/__tests__/__snapshots__/ProgressIndicator.test.js.snap
+++ b/packages/dnb-eufemia/src/components/progress-indicator/__tests__/__snapshots__/ProgressIndicator.test.js.snap
@@ -223,6 +223,21 @@ exports[`ProgressIndicator scss have to match default theme snapshot 1`] = `
 .dnb-progress-indicator__circular__line.dark .dnb-progress-indicator__circular__circle {
   stroke: var(--color-emerald-green); }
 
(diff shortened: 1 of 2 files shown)


tujoworker (changes requested):
Very nice you did fine a sleek "hack" for hiding the artifacts 👍

Now that you have a Node.js version with icu-full – you can simply run the test with update again and amend these files.

`yarn test:update`

tujoworker:
Hers a [Codesandbox](https://codesandbox.io/s/eufemia-progress-indicator-articfacts-ylx5q) to checkout new builds

PR #1284, opened 2022-02-09
chore: update classWithCamelCaseProps with a prop change check

dinarosv:
Moved this into a separate PR to keep it simple. I also added a new test to verify 😃


tujoworker (approved):
Awesome 🙌

PR #1285, opened 2022-02-09
chore(CI): remove .drone.yml and outdated deploy rutines

tujoworker:
As we use GitHub Actions, we do not need all the droneCI setup anymore. Also, auto-committing pages we do not use, as we create our pages manually.


langz (approved):
Nice job cleaning up 🧹

PR #1286, opened 2022-02-09
fix: set Eufemia.version during release

tujoworker:
A tiny little bug made that we did not write the version in the file. Now, we the tests in place, we have our contract that it works 👍


dinarosv (approved):
Good catch 👍

PR #1292, opened 2022-02-16
feat(TypeScript): remove original ts/tsx files and only leave type definition files

tujoworker:
There are setups and cases where it is not recommended to have `.tsx` files inside `node_modules`. Until now we have just published the raw ts source files in order to let projects use the types of our ts components, while we had definition files for the rest.

With this PR – we change this, so we only end up with definitions files.

I will squash all commits on pr merge. But the intention is to get out a **beta** version first.

Here is the [beta release](https://github.com/dnbexperience/eufemia/releases/tag/v9.23.0-beta.1) and the [ci logs](https://github.com/dnbexperience/eufemia/runs/5213124846)


dinarosv (approved):
So great that you fixed this! 👏

tujoworker:
A second beta was released in corporation with the issue reporter.

PR #1293, opened 2022-02-16
Fix disabled tertiary dropdown

AadneRo:
Fix bug where the disabled dropdown of type tertiary don't have tertiary styling

Now the tertiary dropdown will look like the tertiary button when disabled


tujoworker (approved):
Awesome 🙌 Thank you so much! 

Just one thing from my side; If you either reword the git commit message by starting with `fix: ...` or make that change during the pr merge by choosing squash and give it that message prefix.

More on that topic here: https://eufemia.dnb.no/contribute/style-guides/git#commit-messages

packages/dnb-eufemia/src/components/dropdown/Dropdown.js
```diff
@@ -527,9 +527,12 @@ class DropdownInstance extends React.PureComponent {
       ),
     }
 
+    /* fixes issue where disabled tertiary dropdown have border */
+    const enableTriggerStyles = !(disabled && variant == 'tertiary')
+
     const triggerParams = {
       className: classnames(
-        'dnb-dropdown__trigger',
+        enableTriggerStyles && 'dnb-dropdown__trigger',
```
dinarosv, line 535: Instead of disabling this class, which have some dependencies in `_dropdown.scss`, would you consider changing the css in `dnb-dropdown-theme-ui.scss` instead? Found the applied weird border styling there :)

dinarosv (changes requested):
Nice that you fixed this so fast 😍  Just had one comment to maintain consistency in this component :)

langz (approved):

packages/dnb-eufemia/src/components/dropdown/Dropdown.js
tujoworker: yeah, so basically use CSS to fix it, instead JavaScript. That makes sense.

AadneRo: Removing that results in:
Screenshot 2022-02-16 at 14 50 44

dinarosv: But also just doing this for variant tertiary of course 😊  @AadneRo

tujoworker: @AadneRo We can have a quick chat tomorrow maybe?

AadneRo: That would help, thank you

AadneRo: I think I managed @tujoworker

tujoworker: Yes! Wonderful ✨

Two more, but small things 🙏 

- run `yarn test:update dropdown`
- run `git commit --amend` and write `fix` lower case? It would else not be picked up and displayed on the [releases page](https://github.com/dnbexperience/eufemia/releases)

And tell me more about what makes you happy 😆

Screenshot 2022-02-17 at 11 51 08

dinarosv (approved):
Wiiii, nice fix 🙌 ✨

packages/dnb-eufemia/src/components/dropdown/Dropdown.js
AadneRo: 😂 I tried to import an icon first, and got frustrated when it didn't work. That led to angy-face.

Updated the commit message now, and with the test:update

PR #1295, opened 2022-02-17
feat(esm): remove esm directory in package build

tujoworker:
Because we copy all content from `/esm` to one level back, it contains almost only duplications. This redundancy we remove with this merge.

Package reduction is from `9.26 MB` to `8.74 MB`

Also here a beta version was released: https://github.com/dnbexperience/eufemia/releases/tag/v9.23.0-beta.3


dinarosv (approved):
Good! 🤩

PR #1296, opened 2022-02-17
fix(Styles): ensure independent styling packages have declared their dependencies

tujoworker:


dinarosv (approved):
Good! Did you find each manually or some automatic magic?

PR #1297, opened 2022-02-17
feat(ProgressIndicator): Add aria-valuenow value

thaytharma:
Add `aria-valuenow`values to `ProgressIndicator` when progress property is defined for accessibility reasons.

From MDN Docs: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-valuenow
 - The aria-valuenow enables defining a current numeric value between the minimum and maximum values. The minimum and maximum values are defined with [aria-valuemin](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-valuemin) and [aria-valuemax](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-valuemax). 
 - If there is no known value, like when a progress bar is in an indeterminate state, don't set an aria-valuenow attribute.

### DIFF
--- a/packages/dnb-eufemia/src/components/progress-indicator/ProgressIndicatorCircular.js
+++ b/packages/dnb-eufemia/src/components/progress-indicator/ProgressIndicatorCircular.js
@@ -166,6 +166,7 @@ export default class ProgressIndicatorCircular extends React.PureComponent {
       params.role = 'progressbar'
       params['aria-label'] = title
       params['title'] = title
+      params['aria-valuenow'] = progress
     } else {
       params.role = 'alert'
       params['aria-busy'] = true
--- a/packages/dnb-eufemia/src/components/progress-indicator/ProgressIndicatorLinear.js
+++ b/packages/dnb-eufemia/src/components/progress-indicator/ProgressIndicatorLinear.js
@@ -79,6 +79,7 @@ export default class ProgressIndicatorLinear extends React.PureComponent {
       params.role = 'progressbar'
       params['aria-label'] = title
       params['title'] = title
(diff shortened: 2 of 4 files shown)


tujoworker:
Ohh, your thoughts are good!

I double checked it with VO and NVDA, so with this PR change, we get:

`role="progressbar" aria-label="87,13&nbsp;%" title="87,13&nbsp;%" aria-valuenow="87.12636482948174"`

which results in this for VO:

Screenshot 2022-02-18 at 09 01 06

and NVDA:

Screenshot 2022-02-18 at 09 13 18

I can't remember if its possible to avoid or use a different "formatter"? But having the ability to show a custom formattet title, would probably be a better choice to present numbers.

What do you think?

thaytharma:
> Ohh, your thoughts are good!
> 
> I double checked it with VO and NVDA, so with this PR change, we get:
> 
> `role="progressbar" aria-label="87,13&nbsp;%" title="87,13&nbsp;%" aria-valuenow="87.12636482948174"`
> 
> which results in this for VO:
> 
> Screenshot 2022-02-18 at 09 01 06
> 
> and NVDA:
> 
> Screenshot 2022-02-18 at 09 13 18
> 
> I can't remember if its possible to avoid or use a different "formatter"? But having the ability to show a custom formattet title, would probably be a better choice to present numbers.
> 
> What do you think?

We already have the title in place, right? 🤔
I can decline this PR until we find a better solution! 😊

tujoworker:
Yes, we have a title. But only for percentage I think – as of now.

PR #1298, opened 2022-02-18
fix(Tag): fix Tag.Group spacing

tujoworker:
Old
Screenshot 2022-02-18 at 14 24 15

New
Screenshot 2022-02-18 at 14 24 25

Flex `gap` ensures we do not get extra spacing at the bottom or on the right side. So its a good option.

But `gap` is not the best option when it comes to flexibility. But for the Tag component it may be just good enough. Because – in case where a custom spacing is used, it comes in addition. But that's probably OK.


dinarosv (approved):
This is a great solution! Nice teamwork 👏

sivran (approved):

PR #1301, opened 2022-02-21
fix(Tabs): align to correct UI

tujoworker:
Align to UX defined "Tab menu" UI.

From
Screenshot 2022-02-21 at 21 34 39

To
Screenshot 2022-02-22 at 14 35 04

- [x] This PR relies on #1299 and needs to get rebased before merge.


dinarosv (commented):
Skjermbilde 2022-02-22 kl  14 26 13

I think it did look better with the smaller focus ring, which got approved by Casper when we changed it last time (in november). But I can't remember why it is different in Figma, has it been changed since? 🤔

tujoworker:
> has it been changed since?

Not that I'm aware of. I can change that back and we then change it in Figma.

PR #1302, opened 2022-02-22
feat(Dialog): add new variant confirm

dinarosv:
Added variant `confirm` to Dialog.
See the [Figma file](https://www.figma.com/file/cdtwQD8IJ7pTeE45U148r1/Eufemia---Web?node-id=21007%3A9783) for the different variant design.


tujoworker (approved):
Looks good to me now. Just a last question, should we provide the Dialog with `description` or not?

dinarosv:
> Looks good to me now. Just a last question, should we provide the Dialog with `description` or not?

Think we should do that. There were also a few other things I will add, that we talked about, before I merge :)

packages/dnb-design-system-portal/src/docs/uilib/components/dialog/prop-table.md
```diff
@@ -3,6 +3,7 @@
 
 | Properties         | Description                                                                                                                                                                                                                                  |
 | ------------------ | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `variant`          | _(optional)_ The dialog variant. Can either be `inform` or `confirm`. Defaults to `inform`.                                                                                                                                                  |
```
tujoworker, line 6: What do you think about `variant="information"` and variant="confirmation" – just to be in line with other "noun" variants.

information = noun
inform = verb

tujoworker (approved):

packages/dnb-design-system-portal/src/docs/uilib/components/dialog/prop-table.md
tujoworker: It’s good to be in line with UX here. Devs can see the name of UX components. So in this case, the whole word is descriptive witch is actually good.

dinarosv: I'm convinced 😃 Will change this too!

packages/dnb-eufemia/src/components/dialog/parts/DialogAction.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+  /**
+   * For dialog actions, give a custom text for the confirm button.
+   */
+  confirmText?: string
+
+  /**
+   * For variant confirmation, handle the confirm action click.
+   */
+
```
tujoworker, line 25:
```suggestion
```

tujoworker (approved):

PR #1305, opened 2022-02-23
chore(CI): run Slack action on failures

tujoworker:
Ensure we run the Slack action when an action fails.

### DIFF
--- a/.github/workflows/visual-test.yml
+++ b/.github/workflows/visual-test.yml
@@ -85,4 +85,4 @@ jobs:
         env:
           SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK }}
           MATRIX_CONTEXT: ${{ toJson(matrix) }}
-        if: github.repository_owner == 'dnbexperience'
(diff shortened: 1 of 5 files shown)


.github/workflows/visual-test.yml
```diff
@@ -85,4 +85,4 @@ jobs:
         env:
           SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK }}
           MATRIX_CONTEXT: ${{ toJson(matrix) }}
-        if: github.repository_owner == 'dnbexperience'
+        if: ${{ always() && github.repository_owner == 'dnbexperience' }}
```
dinarosv, line 88: What does the `always()` do? In an if-check it seems like it should always be true?

dinarosv (approved):
⭐ ⭐

.github/workflows/visual-test.yml
tujoworker: yeah, it looks wired and illogical. But that's how it works:

> Causes the step to always execute, and returns true, even when canceled.

https://docs.github.com/en/actions/learn-github-actions/expressions#always

PR #1306, opened 2022-02-23
fix(Internal): ensure usePropsWithContext will use context regardless

tujoworker:
While default props are listed, with functional components and TS we do not need to list all anymore just to ensure that we use the context. So with this PR we use the context as long as the prop is not set.


dinarosv (approved):
Ah yes, this makes sense 👍

PR #1307, opened 2022-02-23
fix(Modal): containerPlacement bug for left side

dinarosv:
Fixes the reported [issue](https://dnb-it.slack.com/archives/CMXABCHEY/p1645613615303419) with Modal/Drawer being placed in the center of the screen when using `container_placement="left`. 

Also added a screenshot-test to make sure we don't break this in the future :)


tujoworker (approved):
Awesome and good we have these extra tests now as well 🙌

PR #1308, opened 2022-02-23
fix(docs): add more information on converting from Modal

dinarosv:
To prevent misinterpretation and errors when trying to convert from Modal to Drawer or Dialog 😊


tujoworker (approved):
Nice! Thanks 🙌

Let me know what do you think – but in order to make it visible in the change log, we can use the "fix" decoration to bold out that change. I'v the pr title already. If you agree, you simply can squash merge.

PR #1309, opened 2022-02-23
fix(docs): fix event docs of InfoCard

tujoworker:
- fix event docs of InfoCard
- run visual tests with a more narrow width a given width


dinarosv (approved):
Good! 🌟 🌟

PR #1310, opened 2022-02-23
chore: remove unused defaults props with null as value

tujoworker:
Not sure if this is something that makes things better, but at least, we do not need them anymore.


dinarosv (commented):
This is only used for when the extendPropsWithContext is used, right? I have had some bugs removing the `null` props before

dinarosv (approved):
Then it looks good! 😊

PR #1312, opened 2022-02-24
chore(Portal): clean up and fix reset command

tujoworker:
- Make the total `reset` command actually work
- Remove unused deps in portal


dinarosv (approved):
👍

PR #1313, opened 2022-02-24
chore(Storybook): upgrade storybook to use Webpack v5

tujoworker:
This pr #1304 made Storybook not working anymore, as it relied on Webpack v4.
But with this RP we use a different builder be in sync with the the repo usage of Webpack v5.


dinarosv (approved):
Good update ✅

PR #1314, opened 2022-02-24
feat(Tag): omit non-existing group warning with new prop: hasLabel

tujoworker:
For reference https://dnb-it.slack.com/archives/CMXABCHEY/p1645714272401489


dinarosv (approved):
This was a good solution 😊

PR #1315, opened 2022-02-24
fix(Dropdown): fix empty data entry handling

tujoworker:
Thank you Mateusz!

For reference: https://dnb-it.slack.com/archives/CMXABCHEY/p1645701889827709


dinarosv (approved):
Good ⭐

PR #1317, opened 2022-02-25
feat(FormStatus): add marketing status

thaytharma:
Added a new `marketing` state. 
https://www.figma.com/file/cdtwQD8IJ7pTeE45U148r1/Eufemia---Web?node-id=16838%3A6822

I'm not sure which `aria-role` to use here in this case. Inputs are appreciated!

Skjermbilde 2022-02-25 kl  12 06 15


tujoworker (approved):
Fantastic!

PR #1318, opened 2022-02-25
feat(Avatar): add hasLabel to omit non-existing avatar group warning

thaytharma:
Did the same for Avatar as we did in [Tag](https://github.com/dnbexperience/eufemia/pull/1314).


tujoworker (approved):
Ahh nice! Looks good! I guess, its good to have, even the Avatar is probably not used inside of a table or definition list. But there may come edge cases we can't think of – most probably 😉

PR #1322, opened 2022-02-28
feat(Pagination): change style of next/previous page buttons

dinarosv:
- moved the next/previous page buttons from the sides to below and changed to tertiary button (to match Figma)
- updated number of buttons shown in small/large viewport (to match Figma)
- converted `PaginationBar` to functional component in TypeScript
- changed the small/large viewport handling from css logic to js
- bugfix: updated the `calculatePagination()` so it don't jump around as much as it did


tujoworker (approved):
Awesome work! ❤️

tujoworker:
I think we still need to make some CSS enhancements. Maybe even `gap: 1rem` can be used here?

Screenshot 2022-03-01 at 10 39 01

dinarosv:
@tujoworker Ideally, it shouldn't wrap like that when the viewport gets smaller... It should get fewer buttons 🤔 But I thought I fixed that hmm..

tujoworker:
Yeah, right. We may look into that asap. Maybe we rather should use [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) – we use it already in three components.

PR #1323, opened 2022-02-28
fix(Dialog): set role alertdialog for confirmation variant

dinarosv:
To improve screenreader experience, this PR adds role `alertdialog` for the Dialog with variant `confirmation`. 

An alert dialog should have role 'alertdialog' ✅ , an 'aria-labelledby'/'aria-label' ✅ , and an 'aria-describedby' ✅  to be fully accessible (read more at [W3C Group Alert Dialogs](https://www.w3.org/TR/wai-aria-practices/#alertdialog)).


tujoworker (changes requested):
Nice enhancement! Two things:

- We need to test this with screen readers before release (to test the actual enhancement)
- Include a jest test that verifies the extra attribute

tujoworker (approved):
Both things are done 👍

PR #1328, opened 2022-03-03
fix(Button): make tertiary button icon inherit button height property

tujoworker:
This PR makes it easier to change the sizes of the button


dinarosv (approved):
Very good!

PR #1329, opened 2022-03-03
fix(Modal): containerPlacement bug for top

dinarosv:
This had the same cause as the previous reported issue with containerplacement left, where the default alignment was different for mode drawer in the original version. Hope and think this will fix the reported issue with the background too


tujoworker (approved):
Looks good 🙌

PR #1332, opened 2022-03-03
fix(Pagination): handle unknown / undefined page_count

tujoworker:
During jest test runs after merge of #1322 I saw React key warnings. The reason was that page_count was undefined in some tests. This PR will ensure we do not return `[[ 1 ], [], [ undefined]]` – but rather `[[1]]` – will this be correct @dinarosv ?

### DIFF
--- a/packages/dnb-eufemia/src/components/pagination/PaginationBar.tsx
+++ b/packages/dnb-eufemia/src/components/pagination/PaginationBar.tsx
@@ -136,8 +136,9 @@ const PaginationBar = (innerProps: PaginationBarProps) => {
       getTranslation(innerProps).Pagination
     )
 
-  const prevIsDisabled = currentPage === 1
-  const nextIsDisabled = currentPage === pageCount || pageCount === 0
+  const prevIsDisabled = currentPage > -1 ? currentPage === 1 : true
+  const nextIsDisabled =
+    currentPage > -1 ? currentPage === pageCount || pageCount === 0 : true
 
   const paginationBarRef = useRef(null)
   const currentScreenSize = useResizeObserver(paginationBarRef)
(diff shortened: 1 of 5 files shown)


packages/dnb-eufemia/src/components/pagination/PaginationBar.tsx
```diff
@@ -136,16 +136,17 @@ const PaginationBar = (innerProps: PaginationBarProps) => {
       getTranslation(innerProps).Pagination
     )
 
-  const prevIsDisabled = currentPage === 1
-  const nextIsDisabled = currentPage === pageCount || pageCount === 0
+  const prevIsDisabled = currentPage > -1 ? currentPage === 1 : true
```
dinarosv, line 139:
```suggestion
  const prevIsDisabled = currentPage <== 1
```
For faster readability... Think it would act the same (the only difference her is that it will be `true` for currentPage==0, which also should disable previous button. Do you have any inputs?

packages/dnb-eufemia/src/components/pagination/PaginationBar.tsx
```diff
@@ -136,16 +136,17 @@ const PaginationBar = (innerProps: PaginationBarProps) => {
...
+  const nextIsDisabled =
+    currentPage > -1 ? currentPage === pageCount || pageCount === 0 : true
```
dinarosv, line 141:
```suggestion
    currentPage < 0 || currentPage === pageCount || pageCount === 0
```

dinarosv (approved):
Thanks for fixing this ❤️  
I had just some small things I thought could be changed, but not really necessary :)

packages/dnb-eufemia/src/components/pagination/PaginationBar.tsx
tujoworker: I want `prevIsDisabled` to be true, if an invalid value was given. I don't think `currentPage <== 1` will give us that?

dinarosv: Ah, I understand. Then your solution is the better

PR #1333, opened 2022-03-04
feat(Portal): Suffix page titles with "| Eufemia"

thaytharma:
Suffixed page titles with "| Eufemia" in the portal.

This is just a suggested change of common practice on how page titles should be.
It makes it easier to find and search through logs etc.

Used pipe instead of dash since [DNB.no](https://www.dnb.no/forside-ung) uses this pattern as well.

Feel free to ignore this PR if this is not what we want.


dinarosv (approved):
Think this is a good improvement 👍

PR #1334, opened 2022-03-04
fix(Dropdown): ensure the ellipsis text overflow actually works

tujoworker:
This fixes an overflow issue reported by Jens, UX designer.


dinarosv (approved):
Good work 🌟

PR #1335, opened 2022-03-04
feat(Pagination): enhance accessibility and align styles to UX

tujoworker:
- use vertical separator on dots
- use aria-live for current page
- omit setting focus when prev/next buttons are used
- align styles to UX (default button sizes)
- remove outline when button goes from enabled to disabled state when activeElement


dinarosv (approved):
Looks nice!

PR #1340, opened 2022-03-07
chore(Icons): refactor Figma icons lib setup

tujoworker:
After the Figma lib where re-organized, we needed to change how we process the icons after connecting to the Figma API.


dinarosv (approved):
Good! But do we still provide two versions of every icon now? When they gathered two in one with a variant instead?

tujoworker:
> do we still provide two versions of every icon now?

Yes! I like the idea of having then in one. But it comes with some negative sides regarding additional app size.

Anyway, it would have to be done in a separate and much larger pr.

PR #1345, opened 2022-03-09
fix(FormStatus): wider the min-width to 30rem and react on window width change

tujoworker:
- This PR makes the `FormStatus` respond to the window resize event.
- It also widens the min-width from 12rem to 30rem, when it is used together with another component, that is smaller
- The MarketingIcon got only auto-formatted with prettier


dinarosv (approved):
Good work 😄 
So we went away from stretching the formstatus with the input?

tujoworker:
> So we went away from stretching the FormStatus with the input?

If the input stretches, the FormStatus should not. It was always like that. Had just to clarify it yesterday with Magnus.

PR #1347, opened 2022-03-10
fix(Avatar): Allow Avatar to get skeleton prop from context

langz:
Motivated by #1344

PR #1350, opened 2022-03-10
fix(Avatar): Allow Avatar.Group to set skeleton prop

langz:
This PR just adds the property and type for skeleton to the Avatar.Group, so that it will be easier/possible to use the component with the skeleton prop. The functionality is/was already in place, take a look at https://eufemia.dnb.no/uilib/components/avatar and add the skeleton property to one of the demos/examples.

```
<Avatar.Group skeleton label="label">
   <Avatar>A</Avatar>
</Avatar.Group>
 ```

### DIFF
--- a/packages/dnb-eufemia/src/components/avatar/AvatarGroup.tsx
+++ b/packages/dnb-eufemia/src/components/avatar/AvatarGroup.tsx
@@ -9,6 +9,7 @@ import { AvatarSizes, AvatarVariants } from './Avatar'
 import Context from '../../shared/Context'
 import { ISpacingProps } from '../../shared/interfaces'
 import { usePropsWithContext } from '../../shared/hooks'
+import { SkeletonShow } from '../skeleton/Skeleton'
 
 export interface AvatarGroupProps {
   /**
@@ -46,6 +47,12 @@ export interface AvatarGroupProps {
    * Default: primary.
    */
   variant?: AvatarVariants
+
+  /**
+   * Skeleton should be applied when loading content
+   * Default: null
+   */
+  skeleton?: SkeletonShow
 }
 
 export const defaultProps = {
@@ -55,6 +62,7 @@ export const defaultProps = {
   size: 'medium',
   children: null,
   variant: 'primary',
+  skeleton: null,
 }
 
 export const AvatarGroupContext = React.createContext(null)
--- a/packages/dnb-eufemia/src/components/avatar/AvatarGroup.tsx
+++ b/packages/dnb-eufemia/src/components/avatar/AvatarGroup.tsx
@@ -6,6 +6,7 @@ import { createSpacingClasses } from '../space/SpacingHelper'
 import { AvatarSizes, AvatarVariants } from './Avatar'
 
 // Shared
+import { validateDOMAttributes } from '../../shared/component-helper'
 import Context from '../../shared/Context'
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: packages/dnb-eufemia/src/components/avatar/AvatarGroup.tsx [hunk 1]
```diff
@@ -46,6 +47,12 @@ export interface AvatarGroupProps {
    * Default: primary.
    */
   variant?: AvatarVariants
+
+  /**
+   * Skeleton should be applied when loading content
+   * Default: null
+   */
+  skeleton?: SkeletonShow
```
langz, line 55: Not sure if/how I should use this prop in the component.
As it doesn't seem like I actually need to make more code changes to be able to do:

```
<Avatar.Group skeleton label="label">
   <Avatar>A</Avatar>
</Avatar.Group>
 ```

File: packages/dnb-eufemia/src/components/avatar/AvatarGroup.tsx [hunk 2]
```diff
@@ -55,6 +62,7 @@ export const defaultProps = {
   size: 'medium',
   children: null,
   variant: 'primary',
+  skeleton: null,
```
dinarosv, line 65:
```suggestion
  skeleton: false,
```

dinarosv (changes requested):

packages/dnb-eufemia/src/components/avatar/AvatarGroup.tsx
langz: Done ✅

dinarosv (approved):

langz:
Not really sure what we've done now, when running eufemia locally, I get the following error when visiting http://localhost:8000/uilib/components/avatar/demos

```
Warning: Received `false` for a non-boolean attribute `skeleton`.
If you want to write it to the DOM, pass a string instead: skeleton="false" or skeleton={value.toString()}.
If you used to conditionally omit it with skeleton={condition && value}, pass skeleton={condition ? value : undefined} instead.
```

I'll try looking into it tomorrow.

langz:
What I wanted to achieve with this PR was the following:
image

At the moment, it's not working ,and I am not exactly sure why, but it seems to "work again", if I remove the following line https://github.com/dnbexperience/eufemia/pull/1347/commits/d2622c5022eaccc1b7e7c188e588507df4b31c7c#r826913486

dinarosv:
> At the moment, it's not working when passing skeleton to Avatar.Group, and I am not exactly sure why, but it seems to "magically work again" if I remove the following line [#1347 (comment)](https://github.com/dnbexperience/eufemia/pull/1347#discussion_r826913486)

I think you might need to pass down `skeleton` from AvatarGroup to the clone function for the Avatar? Or another option is to include it in the AvatarGroup context?

langz:
> > At the moment, it's not working when passing skeleton to Avatar.Group, and I am not exactly sure why, but it seems to "magically work again" if I remove the following line [#1347 (comment)](https://github.com/dnbexperience/eufemia/pull/1347#discussion_r826913486)
> 
> I think you might need to pass down `skeleton` from AvatarGroup to the clone function for the Avatar? Or another option is to include it in the AvatarGroup context .... edit: what if you change the order of `{skeleton: context?.skeleton}` and `avatargroupcontext` in Avatar?

Thanks, that did it 🙇 🙏

PR #1351, opened 2022-03-10
fix(Tag): Allow Tag.Group to set skeleton prop

langz:
This PR just adds the property and type for skeleton to the Tag.Group, so that it will be easier/possible to use the component with the skeleton prop. The functionality is/was already in place, take a look at https://eufemia.dnb.no/uilib/components/tag and add the skeleton property to one of the demos/examples.

```
<Tag.Group skeleton label="tags">
   <Tag />
</Tag.Group>
 ```

### DIFF
--- a/packages/dnb-eufemia/src/components/tag/TagGroup.tsx
+++ b/packages/dnb-eufemia/src/components/tag/TagGroup.tsx
@@ -54,7 +54,12 @@ const TagGroup = (localProps: TagGroupProps & ISpacingProps) => {
     className,
     children: childrenProp,
     ...props
-  } = usePropsWithContext(localProps, defaultProps, context?.TagGroup)
+  } = usePropsWithContext(
+    localProps, 
+    defaultProps, 
+    context?.TagGroup,
+    { skeleton: context?.skeleton }
+  )
 
   let children = childrenProp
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/dnb-eufemia/src/components/tag/TagGroup.tsx [hunk 1]
```diff
@@ -29,12 +30,19 @@ export interface TagGroupProps {
    * Default: null
    */
   children?: React.ReactNode
+
+  /**
+   * Skeleton should be applied when loading content
+   * Default: null
+   */
+  skeleton?: SkeletonShow
```
langz, line 38: Not sure if/how I should use this prop in the component.
As it doesn't seem like I actually need to make more code changes to be able to do:

```
<Tag.Group skeleton label="tags">
   <Tag />
</Tag.Group>
 ```

**Review by thaytharma — APPROVED** [review]
Looks good to me!

File: packages/dnb-eufemia/src/components/tag/TagGroup.tsx [hunk 2]
```diff
@@ -29,12 +30,19 @@ export interface TagGroupProps {
...
 }
 
 export const defaultProps = {
   label: null,
   className: null,
   children: null,
+  skeleton: null,
```
dinarosv, line 45:
```suggestion
  skeleton: false,
```

langz: Done ✅

PR #1353, opened 2022-03-11
docs(Tag): Change heading levels by only one factor

langz:
Fixes the following warning from the browsers console:

```
Eufemia Heading levels can only be changed by factor one! Got: 3 and had before 1 - The new level is 2 
```


tujoworker (approved):
❤️

PR #1354, opened 2022-03-11
docs(VisuallyHidden): Change heading levels by only one factor

langz:
Fixes the following warning from the browsers console when running locally:

```
Eufemia Heading levels can only be changed by factor one! Got: 4 and had before 2 - The new level is 3 
```


tujoworker (approved):
❤️

PR #1356, opened 2022-03-14
feat(Theme): create theme handler Gatsby Plugin

tujoworker:
This PR changes how the portal imports theme styles. Everything is handled via Webpack, where we split out theme CSS files into separate packages and load them as soon as the user selects a different theme. But all logic is separated in a Gatsby Plugin.

Using a different theme is possible today already. This PR is all about letting the user or the app change a theme. 

This solution does not lower the LH score 👍

Future tasks

- This Gatsby plugin can easily be used by applications as well. We may a release of that NPM package.
- Create visual tests for other brand themes. It should be pretty straight forward, as we now can load a certain theme CSS via `https://eufemia.dnb.no/?dnb-theme=ui`

Only Eufemia Portal related; What should we call the default Eufemia DNB theme? In code its called `ui`. Right now its called DNB light. Any better ideas?

Screenshot 2022-03-15 at 11 51 03

### DIFF
--- a/packages/dnb-design-system-portal/gatsby-config.js
+++ b/packages/dnb-design-system-portal/gatsby-config.js
@@ -106,6 +106,19 @@ const plugins = [
   },
   'gatsby-plugin-sass',
   'gatsby-plugin-emotion',
+  {
+    resolve: 'gatsby-plugin-eufemia-theme-handler',
+    options: {
+      themes: {
+        ui: { name: 'DNB light' }, // universal identity
+        eiendom: { name: 'DNB Eiendom (WIP)' },
+      },
+      defaultTheme:
+        process.env.GATSBY_CLOUD && currentBranch.includes('eiendom')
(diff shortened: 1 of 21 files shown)


dinarosv (approved):
Wow! Really nice solved 👍  

I think the name "DNB light" represents the default theme well. (Does that mean we are close to a "DNB dark" 😲 !).

dinarosv:
And I was only thinking that the DNB eiendom theme looks a bit odd to have in a released version of the portal? Any way we could fix it up or maybe add another one as an example?

tujoworker:
> And I was only thinking that the DNB eiendom theme looks a bit odd to have in a released version of the portal? Any way we could fix it up or maybe add another one as an example?

I agree! Soon we will get more concrete info about it. Maybe we just add wip; `DNB Eiendom (WIP)`

PR #1357, opened 2022-03-14
chore: add skeleton definition to context

dinarosv:
The typescript files which are extracting `context.skeleton` in the new PRs needs the reassurance that skeleton exists.


langz (approved):
Thanks 👍

PR #1358, opened 2022-03-15
feat(Theme): add visual test example for Eiendom theme

tujoworker:
Just a quick POC how to possibly make visual tests for other themes. Technically, we could even make "the same" tests run on every theme automatically. 

- Or at least, it could be enabled per test basis?
- Maybe this is better in terms of scalability? 
- Or it would mean too much magic and too little control? 

Example:

Instead of having this extra file: `Button.eiendom.screenshot.test.js`

We could add an property to the original test config:

```js
setupPageScreenshot({ url: '/uilib/components/button/demos', additionalThemes: ['eiendom'] })
```
or

```js
const screenshot = await testPageScreenshot({
  selector: '[data-visual-test="button-primary"]',
  simulate: 'focus',
  additionalThemes: ['eiendom']
})
```

or other properties like: `includeAllThemes: true` could be an option too.

Any input or ideas are welcome.


dinarosv:
So then there will be a large increase in stored images, if we are going to add them to all tests 🤔 
I think it is a tricky problem.. Maybe having one additional screenshot test to just show the basics when the component has a theming file? But then I think about setting the theme for that single test:
`setupPageScreenshot({ url: '/uilib/components/button/demos', theme: 'eiendom' })`

Could potentially give some unwanted hickups though, when theme changes during the tests...

tujoworker:
> So then there will be a large increase in stored images

Yes, that's a good point to address. 

So I also think; creating additional visual tests for the parts that actually changes, will benefit the most. So having extra test files is with that possibly the best option then? Because of the extra declarativity and isolation. Which may result in duplication, but with a possibly "good" one – right?

dinarosv:
> So having extra test files is with that possibly the best option then?
Because of the extra declarativity and isolation. Which may result in duplication, but with a possibly "good" maintainability?

Yes that sounds like a good solution, for a better developer experience and maintainability 👍

thaytharma:
> > So having extra test files is with that possibly the best option then?
> > Because of the extra declarativity and isolation. Which may result in duplication, but with a possibly "good" maintainability?
> 
> Yes that sounds like a good solution, for a better developer experience and maintainability 👍

I agree!

dinarosv (approved):

PR #1361, opened 2022-03-18
feat(Drawer): less spacing in new styles

dinarosv:
Updates Drawer to match Figma, with less spacing between parts.

This PR also updates closebutton with set default size for Dialog and Drawer - again to match Figma.


tujoworker (approved):
Very good work! Thank you so much 🙏

Now, that we removed the header min-height, we can actually remove the CSS property as well:

```diff
--drawer-header-min-height: 6rem;
```

PR #1362, opened 2022-03-18
chore: give boolean values to button from Modal

dinarosv:
Fixing console error arising from Modal sending string props to button:

Skjermbilde 2022-03-18 kl  11 05 57


tujoworker (approved):
Supert!

PR #1363, opened 2022-03-18
feat(DatePicker): use Drawer when opening the DatePicker

tujoworker:
Instead of inline (as of now) or React Portal, this PR will use the Drawer when opening the DatePicker. 

This change will also result in only showing ONE month view when `range` is used.

To do:
- [ ] Update tests
- [ ] Update documentation
- [ ] Align with Figma


OddKMS:
Any news on this PR, or any way I can help?

This is one of the remaining blockers for the Transfers application before it can move on to user testing

OddKMS:
I've picked this up as I've got another issue that coincides with a lot of the changes here.
I'm also trying to port over the `setDirectionObserver` from `DrawerList` so that the calendar portion of the `DatePicker` gets moved up whenever it spawns or is scrolled down and out of the viewport. :)

PR #1364, opened 2022-03-22
docs: improve headers in list summary from pages

langz:
Fixes the following warning from the browsers console when running locally:

```
Eufemia Can not decrease to heading level 1! Had before 2 - The new level is 2 
```


langz (commented):
Looks fine to me 🙏

PR #1365, opened 2022-03-22
docs: fix <div> cannot appear as a descendant of <p>

langz:
This is an alternative fix for the following warning from the browsers console when running the list docs locally:

```
Warning: validateDOMNesting(...): <div> cannot appear as a descendant of <p>.
```

Screenshot as well:
image

I think this fix is "okay", but I don't like that the only way I found of solving this was to include prettier-ignore, as prettier would "format away" the fix if not including prettier-ignore.


tujoworker (approved):
Maybe we just use `Nested ol list by using ".dnb-ol--nested"` instead of this extra pre tag? I'm fine with both at least 👍

PR #1371, opened 2022-03-25
fix(VisuallyHidden): use span as default element so it can be used inside paragraphs

tujoworker:
I think, because `VisuallyHidden` will mostly be used for text messages primarily, we should use span as the default element. Typically when used inside a paragraph, we have to use a span, and not a div. Now, it also can be used inside a heading too.


thaytharma (approved):
Great! This change makes totally sense. 🚀

PR #1375, opened 2022-03-30
docs(HTML Elements): Removes docs for elements in HTML elements overview docs

langz:
This PR removes docs for the specific html elements from the html elements overview doc, as the elements are already linked to in the docs, and have their own doc/page.

In my opinion, this will make the html elements overview doc a bit cleaner, as it has quite a lot of text [today](https://eufemia.dnb.no/uilib/elements), scroll scroll 📜


langz (commented):
Looks fine to me 🤓

PR #1381, opened 2022-04-01
fix(Skeleton): Missing properties in PropTypes

langz:
Fixes the following warning when running `yarn build:types`:

```
✖ The property "aria_busy" is not defined in PropTypes!
Component: Skeleton
File: ./src/components/skeleton/Skeleton.js


✖ The property "aria_ready" is not defined in PropTypes!
Component: Skeleton
File: ./src/components/skeleton/Skeleton.js
```


tujoworker (approved):
Thank you 🙌

PR #1384, opened 2022-04-01
fix(Dropdown): Missing properties in PropTypes

langz:
Fixes the following warning when running `yarn build:types`:

```
✖ The property "enable_body_lock" is not defined in PropTypes!
Component: Dropdown
File: ./src/components/dropdown/Dropdown.js


✖ The property "min_height" is not defined in PropTypes!
Component: Dropdown
File: ./src/components/dropdown/Dropdown.js


✖ The property "page_offset" is not defined in PropTypes!
Component: Dropdown
File: ./src/components/dropdown/Dropdown.js


✖ The property "observer_element" is not defined in PropTypes!
Component: Dropdown
File: ./src/components/dropdown/Dropdown.js
```


langz:
Something wrong with the visual tests here. Not able to reproduce the same failing test locally.

tujoworker:
Sorry for the late reply. I think we may try to move the prop types over to the helpers instead of adding them directly to the component. 
With that, the error messages/warnings are gone as well.
I made a quick commit to see if that helps when running on CI.

tujoworker (approved):
Thank You for taking care of these warnings ❤️

langz:
Thanks for the great fix 🙏

PR #1390, opened 2022-04-04
chore: upgrade to yarn v3.2 for faster installs

tujoworker:


langz (approved):
I checked out this branch, and ran `yarn install`, and `yarn serve`.
These scripts seem to work as expected at least 👏

PR #1391, opened 2022-04-05
feat(Modal): add "bypass_invalidation_selectors" property 

tujoworker:
This property can be used to omit element invalidation outside of a modal.

From the docs:

> Define an array with HTML class selectors (`['.element-selector']`) which should not get invalidated when the modal opens/closes. Use this in order to let some parts of your site still be accessible by screen readers.


packages/dnb-eufemia/src/components/modal/types.ts
```diff
@@ -307,6 +307,11 @@ export interface ModalContentProps {
    */
   overlay_class?: string
 
+  /**
+   * A list of HTML class selectors that not should get invalidated during modal opening.
```
langz, line 311:
```suggestion
   * A list of HTML class selectors that should not get invalidated during modal opening.
```

packages/dnb-eufemia/src/components/modal/types.ts
```diff
@@ -307,6 +307,11 @@ export interface ModalContentProps {
...
+   */
+  bypass_invalidation_selectors?: Array<string>
```
langz, line 313: Should we include a description for this prop in some .md doc?
Would help for my understanding of what this is used for 🙏

tujoworker: Yes. Will add it. Thanks for the heads up 🙏

PR #1395, opened 2022-04-22
chore: unexposes enzyme's shallow

langz:
Just a suggestion, as we are not supposed to "shallow test".

_Do not shallow test, but test components like a user would interact (use mount or render to also test their children)._ - https://eufemia.dnb.no/uilib/usage/best-practices/for-testing#integration-tests


langz (commented):
✅

tujoworker (approved):
Good point 👌

PR #1397, opened 2022-04-26
feat: make Accordion support a separate icon for the expanded state

tujoworker:
By providing an array with icons, the second one will be used/shown as the expanded icon.

New in this PR:

Screenshot 2022-04-26 at 09 46 56

How it looks&feels:

https://user-images.githubusercontent.com/1501870/165249901-ac438c99-3395-465a-ae1d-de17aff8e4c0.mov

Should we rather use an object? 

```jsx
      icon={{
        closed: addIcon, expanded: subtractIcon
      }}
```

Screenshot 2022-04-26 at 09 46 05


svedbergp:
@tujoworker it looks good (y) Maybe the the alternative with object would be more clear on what is the closed and expanded state.

langz:
> 

Looks nice ⭐ 

Hmm, good question 🤔 
I also see that we for a few components explicitly name the "icon" properties as well.
Like for [Autocomplete](https://eufemia.dnb.no/uilib/components/autocomplete/properties) that has `icon`, `input_icon`, and `submit_button_icon`. So maybe we could name this `expanded_icon`?

I am fine with all options, I really don't see any major pros and cons, other than that I value that "we do it the same way" in similar situations for other components 🙏

tujoworker:
Thanks for the good feedback 🙌

I have updated the code to use an object:

```jsx
icon={{
  closed: AddIcon,
  expanded: SubtractIcon,
}}
```

tujoworker:
> @tujoworker it looks good :) Maybe the the alternative with object would be more clear on what is the closed and expanded state.

Yes nice 👍

Also – the animation, when used with +/- is still nice to have (Feedback from me and Magnus, UXer) what do you think Pernilla @svedbergp ?

langz (approved):

tujoworker:
> other than that I value that "we do it the same way" in similar situations for other components

Good points! I first considered that, but because both the `size` and the `position` also would need to get their options, then I think its better with one `icon` prop.

svedbergp:
> > @tujoworker it looks good :) Maybe the the alternative with object would be more clear on what is the closed and expanded state.
> 
> Yes nice 👍
> 
> Also – the animation, when used with +/- is still nice to have (Feedback from me and Magnus, UXer) what do you think Pernilla @svedbergp ?

Yes animation looks good 👍
But it is still possible to use only one icon and get the one rotated? If not specifying an icon, the chevron will be used?

tujoworker:
> But it is still possible to use only one icon and get the one rotated? If not specifying an icon, the chevron will be used?

Yes 👍

PR #1399, opened 2022-04-27
fix(HelpButton): ensure Drawer/Dialog/Modal trigger button gets a label

tujoworker:
This PR fixes an issue where the internally used `HelpButton`, via a Modal/Drawer/Dialog did not get a aria-label - when used with icon only.

MicrosoftTeams-image

We also update the usage of the same in the Portal.


langz (approved):
Nice find 👓

PR #1401, opened 2022-04-27
chore(jest): upgrade jest from 27 to 28

tujoworker:
Update Jest to v28 to be in the loop. 

- The biggest changes are that Jest not ships with jsdom by default, so we add it as an dependency.
- Also, the way we "exclude" and run some test (postbuild) in a separate CI step has changed. 
- We now simply exclude them in the main config, but change the config when the command is called from the CLI to not be excluded.
- We also needed to ensure that the `__tests__` folder is not a part of the build, which is anyway a babel bug, that we have not seen until jest did tell us so.


tujoworker:
> Just checked out this branch, and ran a few of our test commands. Seems to work as expected 🙏 ✅

Thank you for checking it out 👍

PR #1404, opened 2022-05-06
fix(HelpButton): Adds test when used inside texts

langz:
I'll use the HelpButton in a similar scenario, and thought it would be fine to have a test for this scenario


tujoworker (approved):
ohh yes, that looks good – fantastic! 👏

PR #1405, opened 2022-05-10
fix(Modal): fix data-testid usage in object defined props

tujoworker:
This PR fixes issues like this:

Screenshot 2022-05-09 at 15 58 11

Screenshot 2022-05-09 at 16 08 02

It should be possible to extend it like this:

```tsx
declare module 'react' {
  interface HTMLAttributes<T> extends React.HTMLAttributes<T> {
    [property: `data-${string}`]: string
  }
}
```

While `tsc --noEmit` does work fine, VSCode does not like it. 

Therefore, we create a custom type `DataPropTypes` and extend `triggerAttributes` and `closeButtonAttributes` with it.

```tsx
[property: `data-${string}`]: string
```

This 👆 is supported after TS v4.4 via [Symbol and Template String Pattern Index Signatures](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-4.html#symbol-and-template-string-pattern-index-signatures)

### DIFF
--- a/packages/dnb-eufemia/src/components/button/Button.d.ts
+++ b/packages/dnb-eufemia/src/components/button/Button.d.ts
@@ -1,6 +1,8 @@
 import * as React from 'react';
 import { SkeletonShow } from '../skeleton/Skeleton';
 import { IconPrimaryIcon } from '../icon-primary/IconPrimary';
+import { DataAttributeTypes } from '../../shared/interfaces';
+
 export type ButtonText = string | React.ReactNode;
 export type ButtonVariant =
   | 'primary'
@@ -72,9 +74,10 @@ export type ButtonOnClick = string | ((...args: any[]) => any);
  */
 export interface ButtonProps
   extends Omit<
-    React.HTMLProps<HTMLElement>,
-    'disabled' | 'size' | 'title' | 'wrap'
-  > {
+      React.HTMLProps<HTMLElement>,
+      'disabled' | 'size' | 'title' | 'wrap'
+    >,
+    Partial<DataAttributeTypes> {
   /**
    * The content of the button can be a string or a React Element.
    */
--- a/packages/dnb-eufemia/src/components/modal/types.ts
+++ b/packages/dnb-eufemia/src/components/modal/types.ts
@@ -1,5 +1,6 @@
 import * as React from 'react'
 import { CloseButtonProps } from './parts/CloseButton'
+import { ButtonProps } from '../button/Button'
 import { ModalRootProps } from './ModalRoot'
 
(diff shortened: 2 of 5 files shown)


langz (approved):
Thanks, that was quick 🙏

packages/dnb-eufemia/src/components/modal/types.ts
```diff
@@ -1,7 +1,30 @@
 import * as React from 'react'
 import { CloseButtonProps } from './parts/CloseButton'
+import { ButtonProps } from '../button/Button'
 import { ModalRootProps } from './ModalRoot'
 
+type DataPropTypes = {
```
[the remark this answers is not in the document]
tujoworker: Yes, true! We should do this 👍

packages/dnb-eufemia/src/shared/interfaces.tsx
```diff
@@ -10,3 +10,23 @@ export interface ISpacingElementProps {
...
+   *
+   * Effects: triggerAttributes, closeButtonAttributes
+   */
+  'data-testid'?: string
+
+  /**
+   * In future we want to use this below.
+   * But its supported from TS v4.4 - so we may wait some more months.
+   */
+  // [property: `data-${string}`]: string
```
tujoworker, line 31: @langz I have changed to support older TS versions for now. It solves the issue. We may change that then in future.

packages/dnb-eufemia/src/components/button/Button.d.ts
```diff
@@ -72,9 +74,10 @@ export type ButtonOnClick = string | ((...args: any[]) => any);
  */
 export interface ButtonProps
   extends Omit<
-    React.HTMLProps<HTMLElement>,
-    'disabled' | 'size' | 'title' | 'wrap'
-  > {
+      React.HTMLProps<HTMLElement>,
+      'disabled' | 'size' | 'title' | 'wrap'
+    >,
+    Partial<DataAttributeTypes> {
```
tujoworker, line 80: @langz By adding it to the button props, we do support it in many places in one catch. It does not hurt either. But what do you think?

packages/dnb-eufemia/src/shared/interfaces.tsx
langz: Great 👍

packages/dnb-eufemia/src/components/button/Button.d.ts
langz: Sounds smart 👍 
Yeah, I don't think that it will hurt.

PR #1412, opened 2022-05-18
fix(Dialog): fix Safari issue not able to scroll

tujoworker:
This PR fixes an issue where the Dialog component did not scroll when there was content that overflows the Dialog on the Safari iOS browser.

### DIFF
--- a/packages/dnb-eufemia/src/components/dialog/Dialog.tsx
+++ b/packages/dnb-eufemia/src/components/dialog/Dialog.tsx
@@ -90,7 +90,7 @@ function Dialog(
       preventOverlayClose !== undefined ? preventOverlayClose : true
   }
 
-  if (fullscreen === undefined) {
+  if (fullscreen === undefined && fullscreen !== false) {
     currentFullscreen = variant === 'information' ? 'auto' : false
   }
 
--- a/packages/dnb-eufemia/src/components/dialog/DialogContent.tsx
+++ b/packages/dnb-eufemia/src/components/dialog/DialogContent.tsx
@@ -8,6 +8,7 @@ import classnames from 'classnames'
 import {
   isTrue,
   findElementInChildren,
+  validateDOMAttributes,
 } from '../../shared/component-helper'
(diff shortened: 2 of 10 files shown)


langz:
I tried to run this on my phone, iPhone 12 Pro, and then I am not able to scroll the content of the Dialog.

I added more texts in the Paragraph of example https://eufemia-fixdialogsafariscrolling.gtsb.io/uilib/components/dialog/demos#basic-dialog, and this was the result:

https://user-images.githubusercontent.com/1359205/169250482-bb3d4837-9ca4-4c96-abff-b694c0daa3f8.mp4

It seems to just scroll "in the background".

langz (approved):
After the latest changes, I did the same "test" on my phone.
Seems to work perfectly now 👏

tujoworker:
Thank You @langz for the good response and finding the issue 🙌 I will think if we can make some sort of test for this one. But its really hard, because CSS has so big influence, basically, it relies 100% on CSS – and then we need also to do a drag simulation, which is not that easy as well 🤦‍♂️

langz:
> Thank You @langz for the good response and finding the issue 🙌 I will think if we can make some sort of test for this one. But its really hard, because CSS has so big influence, basically, it relies 100% on CSS – and then we need also to do a drag simulation, which is not that easy as well 🤦‍♂️

Haha, I see, yeah seems complicated.

tujoworker:
Yeah, too complicated! 

We would need to simulate a `touchmove` event, and that seems to to be supported in puppeteer yet. Only a `touchstart` and `touchend` is supported at the moment:

https://puppeteer.github.io/puppeteer/docs/puppeteer.touchscreen

Also, one step future we could (would) have to "hack" in to make this CSS only check positive:

```css
  @supports (-webkit-touch-callout: none) {
    @content;
  }
```

I'm not aware of any solutions there.

Changing the user agent alone is easy, but not enough:

`await global.__PAGE__.setUserAgent('iPhone OS 15')`

PR #1413, opened 2022-05-18
fix(Dialog): make Dialog not overflow iOS Safari viewport

tujoworker:
This PR fixes an issue so the Dialog centers in the middle of the screen, instead of overflowing the viewport. See screenshots.

Now:

Screenshot 2022-05-18 at 11 09 12

Before:

Screenshot 2022-05-18 at 11 12 22


langz (approved):
Great 👍 
Is there anyway for us to add some kind of specific tests for these browser specific fixes?
We probably can't use screenshots, as I assume we can't i.e open safari in the build server 🤔

PR #1417, opened 2022-05-19
Remove invalid spacing attributes

tujoworker:
Some components did not "remove" the spacing props when set. They just got added as HTML attributes. But when e.g. `space={{ }}` or `top={true}` was used, React throws an error warning. 

With that PR we add some tests to verify the feature.


langz (approved):
🙇

PR #1422, opened 2022-05-27
feat(Table): move to components and rewrite to TypeScript 

tujoworker:
In order to add more features to the Table, we move it over to be a component (with backwards compatibility) and rewrite it:

- Rewrite to TypeScript
- Rewrite to use React functional component
- Rewrite Sticky Header to a independent Hook
- Move Table to be a component for future features
- Make it backwards compatible with /elements imports
- Add integration tests
- Keep visual tests as before
- Add stories
- Move and extend docs


thaytharma (approved):
Looks good to me! 🙏🏼

packages/dnb-eufemia/src/components/table/Table.tsx
```diff
@@ -0,0 +1,125 @@
...
+  /**
+   * The variant of the component.
+   * Default: primary.
+   */
+  variant?: TableVariants
+}
+
+export const defaultProps = {
+  size: 'large',
+  variant: 'primary',
```
langz, line 60: Not sure if it should default to `basis`, `basic` or `primary`.
Would be clearer to me if the variant use the same naming throughout.

packages/dnb-eufemia/src/components/table/Table.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+  /**
+   * Skeleton should be applied when loading content
+   * Default: null
+   */
+  skeleton?: SkeletonShow
+
+  /**
+   * The size of the component.
+   * Default: medium.
```
langz, line 47: Seems to default to large 
https://github.com/dnbexperience/eufemia/pull/1422/files#diff-13aedba1d448a098b045ee2378c951fb0966404a9047dd4f119885532bfaab3fR59

packages/dnb-eufemia/src/components/table/TableTd.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react'
+import classnames from 'classnames'
+
+type TableTdProps = {
+  /**
+   * Tde content of the component.
```
langz, line 6:
```suggestion
   * The content of the component.
```

langz (approved):
Looks good, I just added a few minor comments 💬 
Most comments is about the table's variant prop and its values. 
Not sure if it is `basis`, `basic` or `primary`, doesn't seem consistent.
I think `primary` would be fine, but as long as it's consistent 🙏 

I'll try running it locally once I get off the train to Voss 🚋

packages/dnb-eufemia/src/components/table/Table.tsx
tujoworker: Thank You ✨

packages/dnb-eufemia/src/components/table/Table.tsx
tujoworker: Thank You ✨ yes, it was a mess! I try to avoid primary, as some parts in DNB will use a different variant as their "main" style. So not sure what to call it actually. `basis` feels not 100% right. But "mint-green-50" either, as it can/will change to grey at some point.

PR #1425, opened 2022-05-31
chore(Demos): add SSR support for all react-live demos

tujoworker:
The advantages of SSR (SSG) support is that the build of the portal will run all the code examples – and if there is a JavaScript/React issue, the build will break.

We use `react-live` for showing demos/examples. It used to support SSR before v2.3.0 (a year ago) but missed it when they converted it to React Hooks. But they want to support it like this [issue](https://github.com/FormidableLabs/react-live/issues/321) shows. So I made a [PR](https://github.com/FormidableLabs/react-live/pull/322). But until they eventually release it, I made a quick build of it and use it directly inside our repo.


packages/dnb-design-system-portal/src/docs/uilib/extensions/payment-card/Examples.js
```diff
@@ -0,0 +1,55 @@
...
+
+export function PaymentCardAllCards() {
+  return (
+    <ComponentBox
+      scope={{ PaymentCard, getCardData }}
+      data-visual-test="all-cards"
+      useRender
+    >
+      {() => /* jsx */ `
+const demoCards = [
```
langz, line 28: Could this use the actual `cardData` from packages/dnb-eufemia/src/extensions/payment-card/utils/cardProducts.js, to ensure that we test all cards?
Not sure if it's a good idea or if it's even possible, just a thought.

langz (approved):

PR #1433, opened 2022-06-07
Facilitate remove-content class for content when Accordion (with prerender) is collapsed

rodynenko:
- [x] look into possible tests

## Summary

Solve the issue about Intersection observer trigger actions on elements hidden inside collapsed Accordion with prerender option.

## Test plan

https://codesandbox.io/s/eufemia-starter-accordion-intersection-e6vr0e?file=/src/App.tsx
Intersection Observer will trigger callback only when accordion is open for first time. Check console log for event details.


tujoworker (approved):
Thank you 🙌

PR #1438, opened 2022-06-08
feat(DatePicker): add size property to component

langz:
This PR adds the `size` property to the DatePicker component, similarly as for components like Input, etc.

image


packages/dnb-eufemia/src/components/date-picker/style/_date-picker.scss
```diff
@@ -374,6 +402,34 @@
       border: 1px solid var(--color-black-border);
       background-color: var(--color-white);
     }
+
```
langz, line 405: Not exactly done/happy with styling/placing the triangle based on the size.

https://github.com/dnbexperience/eufemia/pull/1438/files#diff-5338f2d71ff0969ae92fc20d4dc1c940915daef694c4e800eb05196be74499e1R406-R432

I don't feel like it looks all correct/perfect.
See the deploy preview.

packages/dnb-eufemia/src/components/date-picker/DatePicker.js
```diff
@@ -310,12 +312,12 @@ export default class DatePicker extends React.PureComponent {
           .querySelector('.dnb-input__submit-button__button')
           .getBoundingClientRect().width
         if (align_picker === 'right') {
-          const distance = buttonWidth / 2 - 8
```
langz, line 313: 8 is 1/4 of the default size 32px.
So instead of using the number 8 which only works for the default size, I've changed this to be dynamic to cater for other sizes as well.

packages/dnb-eufemia/src/components/date-picker/style/_date-picker.scss
tujoworker: Really nice work so far @langz  🙌

Two things I suggest: 
- We should not change the size of the triangle. 
- We should probably not show the other sizes in the documentation, but show them for visual tests. And with that, we could keep them in only one visual snapshot test. The reason is, in future we only want to set the size in a spacing "Theme" that changes the component sizes. This way, we can avoid wrong usage of components. But also align better on overall consistency.

I hope you are fine with my arguments so far. Let me know if you disagree or have any questions.

langz: Sounds smart, I'm fine with that :)
I'll look into it sometime today 🕐

OddKMS (commented):
I like what I see, love the tests.

I'm putting an approve as soon as you've implemented the changes suggested by Tobias :)

packages/dnb-eufemia/src/components/date-picker/style/_date-picker.scss
langz: I've addressed these things now 😺

OddKMS (approved):

tujoworker (approved):
YesYesYes – very very nice work 👌

tujoworker:
I rebased this branch with the latest from main – hope it helps for the continues failing visual tests. Not sure why they have failed randommely right in this branch. I feel they are else pretty stable.

langz (commented):
Thanks for the help here @tujoworker :)

PR #1439, opened 2022-06-09
fix(Table): simplify sticky table hook

tujoworker:
In regards to PR #1422 this fixes some issue and makes the sticky functionally more robust.

### DIFF
--- a/packages/dnb-eufemia/src/components/table/__tests__/TableStickyHeader.test.tsx
+++ b/packages/dnb-eufemia/src/components/table/__tests__/TableStickyHeader.test.tsx
@@ -92,25 +92,28 @@ describe('useStickyHeader', () => {
     expect(getTrClasses()).toEqual(['dnb-table__tr', 'sticky'])
   })
 
-  it('should use default stickyOffset when not given', () => {
+  it('should use default header height in rootMargin when stickyOffset is not given', () => {
     render(
       <Table sticky>
         <BasicTable sticky />
       </Table>
     )
 
-    expect(screen.queryByRole('table').querySelector('tr').style.top).toBe(
-      ''
-    )
+    expect(
+      screen
+        .queryByRole('table')
+        .querySelector('tr')
+        .style.getPropertyValue('--table-top')
+    ).toBe('')
     expect(window.IntersectionObserver).toHaveBeenCalledTimes(1)
     expect(window.IntersectionObserver).toHaveBeenCalledWith(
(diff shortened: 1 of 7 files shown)


packages/dnb-eufemia/src/components/table/__tests__/TableStickyHeader.test.tsx
```diff
@@ -99,14 +99,17 @@ describe('useStickyHeader', () => {
...
+        .querySelector('tr')
+        .style.getPropertyValue('--table-top')
+    ).toBe('')
     expect(window.IntersectionObserver).toHaveBeenCalledTimes(1)
     expect(window.IntersectionObserver).toHaveBeenCalledWith(
       expect.any(Function),
-      // Formula: thHeight + tdHeight + offsetTopPx = -(sum)px
-      { rootMargin: '-144px 0px 0px 0px' }
+      // Formula: thHeight + offsetTopPx = -(sum)px
+      { rootMargin: '-80px 0px 0px 0px' }
```
langz, line 112: Not sure if it would be possible to make this "easier to understand".
By quickly scanning this, I am not 100% sure what or why `rootMargin` and its value is important for the test "should use default stickyOffset when not given".

Might be that it's not possible to dumb down/simplify this though 🧠

langz (approved):
Demo of this looked smooth 🏄

packages/dnb-eufemia/src/components/table/__tests__/TableStickyHeader.test.tsx
tujoworker: Hmm. yeah not sure how to make that more clear either – but I can explain it here :P the default height of the table header (`thHeight`) is 80px so when we call the interaction observer, we send in this option, because that's the offset we want in order to show/hide the shadow.

tujoworker: The offset is meant to push down the whole thing, so the sticky part starts later during scrolling. Useful when a overlay header is used, like we have in our Portal.

tujoworker: We do not test the intersection observer by itself, but its integration and the effect. So its very abstract.

I tried to make some comments and enhance the test caption/description. Hope this helps. Let me know if you have any idea to make this more clear.

PR #1441, opened 2022-06-09
fix(Modal): fix flaky overlay opacity color issue happening in Chrome browser

tujoworker:
Here's a [reprod](https://codesandbox.io/s/eufemia-drawer-background-color-bug-ui417b?file=/src/App.tsx): (Shows a black overlay when opened)

Because it happen not on all Chrome browsers (all on latest) its hard to make a test that verifies the issue.


langz (approved):
Thanks for the quick and important fix 🥇

PR #1445, opened 2022-06-10
release of v9.27

tujoworker:


langz (approved):
I get the same error as in https://github.com/dnbexperience/eufemia/pull/1438#issuecomment-1152266031 when running screenshot tests locally for this branch `main`. Doesn't seem like the check "Visual Tests / Visual Test Action (push)" is ran as part of this "release build", so doesn't really matter for this exact PR. 

Not sure if this is a big concern or not.
Maybe it could be related to a few minor changes we've done with our tests, like bumping jest, etc.

tujoworker:
> Maybe it could be related to a few minor changes we've done with our tests, like bumping jest, etc.

Yes, because we do not force a rebase/updated branch. I will look into it asap. 
I also have a branch ready where I update puppeteer to its latest version.

tujoworker:
@langz we can enable visual tests on the `main` branch if we want to – before, when we ran them on linux, they where faster, but flaky. When we think we are ok with it failing 0% - 5% of the time, then we can enable them. 
(But on the `release` and other release branches we should not.)

PR #1448, opened 2022-06-10
chore(CI): speed up build time when making build for visual tests

tujoworker:
Omit image processing when creating a build for visual tests.

This step takes several minutes: `Running gatsby-plugin-sharp.IMAGE_PROCESSING jobs - 345.638s` but we do not test images (I think).


langz (approved):
Ohh, that would be nice to save a few minutes 🕐 👏

PR #1450, opened 2022-06-11
chore(CI): speed up build time when making build for visual tests (#2)

tujoworker:
Omit creating pages that are not needed for visual tests.


tujoworker:
We do not safe that much of time with that approach.

PR #1453, opened 2022-06-13
chore(repo): declare deps correctly towards Yarn PnP

tujoworker:
This PR only fixes missing, wrong configured deps when using PnP

Prettier (and ESLint) in VSCode does still not work with PnP

Therefore, we still may have to wait.

Why is PnP interesting? Because of the "zero" install on CI (and locally) – but most for our visual tests on CI, because the macOS form GitHub Actions are very slow on that step.


packages/dnb-design-system-portal/scripts/version.js
```diff
@@ -5,7 +5,7 @@
 
 const fs = require('fs-extra')
 const path = require('path')
-const { isCI } = require('ci-info')
+const { isCI } = require('repo-utils')
```
langz, line 8: 👍

langz (approved):

PR #1454, opened 2022-06-13
chore(repo): update Node.js version to latest v14

tujoworker:
This version includes a fix so we can/would run jest with Yarn PnP: https://github.com/nodejs/node/pull/41221

### DIFF
--- a/.github/workflows/build-check.yml
+++ b/.github/workflows/build-check.yml
@@ -45,7 +45,7 @@ jobs:
       - name: Use Node.js
         uses: actions/setup-node@v2
         with:
-          node-version: 14.17.0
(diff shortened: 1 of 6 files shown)


tujoworker:
ICU does not work properly. It got updated/changed in this [Node.js commit](https://github.com/nodejs/node/commit/6035492c8f) – now `node -p 'process.config.variables.icu_small'` returns `false`, which means, it won't respect `NODE_ICU_DATA` according to the [docs](https://github.com/nodejs/full-icu-npm#am-i-using-small-icu).

tujoworker:
The solution would be to upgrade to v16 of Node.js

PR #1455, opened 2022-06-13
chore(deps): update jest deps to get correct types package ( 27=>28)

tujoworker:


langz (approved):
Looks good 👀 
I've tried comparing this branch vs main branch locally when running `yarn test:screenshots`.
I get a bit different results depending on the branch(but also for the same branch). 8, 9 or 10 failing visual tests.
As the screenshot tests is naturally a bit "random", I would say it works as expected ✅

PR #1459, opened 2022-06-14
fix(Icons): import new version of stopwatch icon

langz:
Mats updated the icon in Figma, seems like it was "corrupted" because it was not correctly aligned in the 16x16px-box(see image).
image

It should not be "corrupted"/cut now.


tujoworker:
In order to make the request etc. the branch needs to be named: `eufemia-icons`
This is defined [here](https://github.com/dnbexperience/eufemia/blob/61c9c119085c380227bf6546ee51f8273a7529a6/.github/workflows/icons-lib.yml#L6).

We could change it to include other rules if we want. Any recommendations from your side?

We have some [docs](https://eufemia.dnb.no/contribute/deploy/icons) about the current setup.

PR #1461, opened 2022-06-15
chore(CI): refactor visual test run

tujoworker:
I have hopes that this can at least shape off some minutes.

With this PR we build the portal on Linux and test with macOS. 

What is the reason behind this? macOS is very much slower in Github Actions, so we now use it for the only one reason – testing visually.


langz (approved):
Cool idea 👍

PR #1463, opened 2022-06-15
fix(Dialog): fix Modal backdrop / overlay false-positive click issue while e.g. selecting

tujoworker:
Reported by @sivran via Slack:

> Could we ignore the onMouseUp event when releasing the mouse outside the modal?
It’s an issue when e.g. selecting text inside an input field and releasing the mouse outside the modal by accident.

https://user-images.githubusercontent.com/1501870/173918168-63f2125a-d0f6-475a-9c42-b63093bfc901.mov


langz (approved):
Works great.
Just gave it a test run in Chrome, Edge, and Safari 💻

tujoworker:
> Gave it a test run

Thank you for checking it out! I also had a positive test run on iOS just to be sure 👍

PR #1464, opened 2022-06-16
feat(Timeline): date supports a list of dates

langz:
This PR adds support for passing a list of dates to the date property of a timeline item.

Screenshot is an example of how it will look:
image

Sidenote: I believe @thaytharma or myself(?) will sometime soon open a PR to change the property names for a few props.
`infoMessage` to `content`, and `date` to `subtitle`.


packages/dnb-eufemia/src/components/timeline/TimelineItem.tsx
```diff
@@ -159,17 +159,32 @@ const TimelineItem = (localProps: TimelineItemProps) => {
     )
   }
 
+  const getDate = () => {
+    const TimelineItemDate = ({ date }: { date: React.ReactNode }) => (
+      <span
```
langz, line 164: Instead of using a span with styling display block, maybe I should just use a div or so?

tujoworker (approved):
Nice 👌

packages/dnb-eufemia/src/components/timeline/TimelineItem.tsx
tujoworker: I'm fine with both. But they are much welcome. Because the rendered height of the `.dnb-timeline__item` element gets then a correct value of e.g. `64px` and not `64.75px`.

langz: Thanks, I'll switch to a `div` then 🙏

PR #1466, opened 2022-06-16
feat(Input): add inner_element property for internal usage (as of now)

tujoworker:


langz (approved):
Nice, will be interesting to see how we'll use this 👍 😄

PR #1468, opened 2022-06-20
feat(Timeline): rename name to title & date to subtitle with backwards compatibility

thaytharma:
Rename name/dare-props in`Timeline`-component:
- `name` is changed to `title` in TimelineItem
- `date` is changed to `subtitle` in TimelineItem

### DIFF
--- a/packages/dnb-eufemia/src/components/timeline/style/_timeline.scss
+++ b/packages/dnb-eufemia/src/components/timeline/style/_timeline.scss
@@ -28,7 +28,7 @@
     }
     &__content {
       padding-bottom: var(--spacing-small);
-      &__date {
+      &__subtitle {
         cursor: text;
         font-size: var(--font-size-x-small);
         font-weight: var(--font-weight-basis);
--- a/packages/dnb-eufemia/src/components/timeline/style/_timeline.scss
+++ b/packages/dnb-eufemia/src/components/timeline/style/_timeline.scss
@@ -22,7 +22,7 @@
         justify-content: center;
         padding: 0;
       }
-      &__name {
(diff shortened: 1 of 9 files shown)


packages/dnb-eufemia/src/components/timeline/style/_timeline.scss
```diff
@@ -22,13 +22,13 @@
...
       }
-      &__name {
+      &__title {
         cursor: text;
       }
     }
     &__content {
       padding-bottom: var(--spacing-small);
-      &__date {
+      &__subtitle {
```
langz, line 31: I think we need to update the snapshot tests for a change like this.
`yarn test:update`

langz (approved):
Looks great 👍 

Not sure which release this should be included in if it's "breaking", based on semantic versioning 🤔

packages/dnb-eufemia/src/components/timeline/style/_timeline.scss
tujoworker: Me and @thaytharma have agreed to serve backwards compatibility and release it as a fix or feature. 

This PR needs to be squashed to a single commit with another git prefix. 

Sidenote: for a breaking change, we need more in the commit message than just the break prefix. More details [here](https://eufemia.dnb.no/contribute/style-guides/git).

tujoworker (approved):
Thanks again for making this PR happen 🙏

I have added backwards compatibility (+ some more) in this commit: https://github.com/dnbexperience/eufemia/pull/1468/commits/a7cd72b0bbc56391bbadea63a2926e0c8a192e7e

PR #1469, opened 2022-06-20
fix(Input): enhance vertical centering for Safari 

tujoworker:
This change should not influence Firefox or Chromium based browser, but enhance Safari (Webkit based browsers).

It fixes an issue occurred in PR #1467

Image from iOS

### DIFF
--- a/packages/dnb-eufemia/src/components/autocomplete/__tests__/__snapshots__/Autocomplete.test.js.snap
+++ b/packages/dnb-eufemia/src/components/autocomplete/__tests__/__snapshots__/Autocomplete.test.js.snap
@@ -2048,9 +2048,6 @@ legend.dnb-form-label {
     word-break: normal;
     position: relative;
     z-index: 2; }
-  .dnb-input input.dnb-input__input {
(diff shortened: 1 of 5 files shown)


langz (approved):
Code seems good ✅ 
I tried comparing https://eufemia.dnb.no/uilib/components/autocomplete#default-autocomplete vs running this branch locally, both in Safari, but wasn't able to see any difference. I might be looking at the "wrong" demo/example though 🙏

tujoworker:
> Code seems good ✅ I tried comparing https://eufemia.dnb.no/uilib/components/autocomplete#default-autocomplete vs running this branch locally, both in Safari, but wasn't able to see any difference. I might be looking at the "wrong" demo/example though 🙏

Yeah, its mainly a iOS Safari issue. There we need the system in order to set our root font-size (for auto sizing when the users has defined a larger font-size) – but the default system font-size on iOS is 17px and not 16px like on all other browsers. 

The fix:
Screenshot 2022-06-20 at 14 00 53

Before:
Screenshot 2022-06-20 at 14 01 52

PR #1470, opened 2022-06-20
feat(DrawerList): use CSS Grid for placing list content to enhance flexibility

tujoworker:


packages/dnb-design-system-portal/src/docs/uilib/components/dropdown/Examples.js
```diff
@@ -300,7 +304,7 @@ export const DropdownMoreMenu = () => {
   more_menu="true"
   size="small"
   title="Choose an item"
-  data={() => [<Link href="/">Go to this Link</Link>, 'Or press on me', <>Custom component</>]}
+  data={() => [<Link href="/" key="item-1">Go to this Link</Lik>, 'Or press on me', <>Custom component</>]}
```
langz, line 307:
```suggestion
  data={() => [<Link href="/" key="item-1">Go to this Link</Link>, 'Or press on me', <>Custom component</>]}
```

`Lik` to `Link`

packages/dnb-eufemia/src/fragments/drawer-list/style/_drawer-list.scss
```diff
@@ -44,7 +44,7 @@
     }
     html[data-visual-test] &,
     &--no-animation {
-      animation: drawer-list-slide-top-up 1ms ease-out 1 forwards;
+      animation-duration: 1ms !important;
```
langz, line 47: Not sure if we could have done something else, to avoid adding `!important`(I just assume we don't like !important). On the other hand, adding more selectors just to add css specificity is also not nice 🤷

packages/dnb-eufemia/src/fragments/drawer-list/style/_drawer-list.scss
```diff
@@ -165,9 +165,22 @@
...
+        justify-content: center;
+      }
+
+      @supports (display: grid) {
+        display: grid;
+
+        // Use two columns
+        grid-template-columns: 1fr minmax(2rem, auto);
+
+        // use a hight number in order to center the check icon and suffix with "grid-row: 1 / -1;"
```
langz, line 181:
```suggestion
        // use a height number in order to center the check icon and suffix with "grid-row: 1 / -1;"
```

`hight` to `height`? Not sure 🤔

langz (approved):

packages/dnb-eufemia/src/fragments/drawer-list/style/_drawer-list.scss
tujoworker: I very much agree. But on the "no animation" its the shortest path to success without extra code and potential future css specificity misalignment.

packages/dnb-eufemia/src/fragments/drawer-list/style/_drawer-list.scss
tujoworker: yeah, it should be "high" 🤦‍♂️ like 10 rows is much for that little cell, but we need a number here in order to let the second column be vertically centered. So 10 seems to me a ok number. Its basically, like if you would this in your data:

```js
const data = [{
    content: ['line 1', 'line 2', 'line 3', ... , 'line 10']
  },
  ...
 ]
```

PR #1471, opened 2022-06-21
fix(Autocomplete): ensure "no options" is read out in aria-live

tujoworker:


langz (approved):

packages/dnb-eufemia/src/components/autocomplete/Autocomplete.js
```diff
@@ -228,6 +228,7 @@ export default class Autocomplete extends React.PureComponent {
 
     custom_element: PropTypes.object,
     custom_method: PropTypes.func,
+    ariaLiveDelay: PropTypes.number,
```
langz, line 231: I assume this is just some kind of "internal prop", meaning that we don't need to add it to `packages/dnb-design-system-portal/src/docs/uilib/components/autocomplete/properties.md` 🙏

PR #1483, opened 2022-06-28
chore(Portal): add support for iOS restore scroll position 

tujoworker:
Released in [v1.2.x](https://github.com/tujoworker/gatsby-plugin-scroll-position/releases/tag/v1.2.0)

For more context: PR #1481


langz (approved):
I tried out the deploy preview on my iPhone, and I am not sure if I experienced any difference compared to https://eufemia.dnb.no/.

tujoworker:
> not sure if I experienced any difference compared to https://eufemia.dnb.no/

true, its more theoretical when it comes to iOS and small screens. I see the improvement on iPad when I test it, because there the sidebar menu is visible.

PR #1485, opened 2022-07-13
docs: Remove 'new' status for components older than 6 months

thaytharma:
Removed '`new`' status for components older than 6 months, including 
`InfoCard`, `Avatar`, `Breadcrumb`, `Timeline`, `Tag`.


dinarosv (approved):
This makes sense 🙌

PR #1496, opened 2022-08-09
fix(Internal): avoid running componentDidMount twice

tujoworker:
This PR fixes an issue where components, supporting temporary snake_case and camelCase props would run componentDidMount twice, given they are a class component (effects for now only Modal).


langz (approved):
Hmm, interesting that we have to add a componentDidMount to avoid running componentDidMount twice 🤔 😂

tujoworker:
> Hmm, interesting that we have to add a componentDidMount to avoid running componentDidMount twice 🤔 😂

yeah, the extended class will then not set this method, because it exists already.

PR #1501, opened 2022-08-10
fix(Drawer): ensure it removes scroll possibility when opened

tujoworker:
The Drawer component had a missing ref property (`ref={context?.scrollRef}`), so the body scroll lock did not work. The added tests do verify this now.


langz (approved):
🙇

PR #1502, opened 2022-08-10
chore(Docs): simplify Examples syntax to only use a string instead of a function

tujoworker:
We have used a function, only to make the VSCode [extension](https://marketplace.visualstudio.com/items?itemName=mgmcdermott.vscode-language-babel) Babel Template Literal to work properly. But VSCode with [this extension](https://marketplace.visualstudio.com/items?itemName=Razio.es6-string-jsx) just works fine.

We have it already as a [recommendation](https://github.com/dnbexperience/eufemia/blob/8ee83266742208b74a364f7ead1f9004f9699864/.vscode/extensions.json#L9).

The code change is simply done with a search (`() => /* jsx */`) and replace (`/* jsx */ ``).


langz (approved):
Looks good too me, less code :)

PR #1503, opened 2022-08-11
fix(Slider): convert Slider from Class to Function component

tujoworker:
- convert Slider from Class to Function component
- Deprecated `on_init` and `raw_value`.
- Remove `use_scrollwheel` with warning (why? Too much complexity for a feature not used much – is was optional before already)
- Rewrite tests from Enzyme to Testing Library


langz (approved):
I have nothing to complain about 🥂

PR #1506, opened 2022-08-12
docs(Slider): rename prop thump_title to thumb_title

langz:
Adds info about changing prop thump_title to thumb_title in release docs


tujoworker (approved):
Yeah! 🕺

Let's get this in. 

Soon I start to add multi-button support. This prob will then most probably go away, as we then only support an array of "values", with an object, containing such a prop. But lets see how it goes ;)

PR #1510, opened 2022-08-15
fix(Drawer): fix Safari text selection + drag/scroll issue

tujoworker:
Reported in this [thread](https://dnb-it.slack.com/archives/CMXABCHEY/p1660052015576469).

There was a pseudo-element that cause in Safari Desktop a wired scroll effect when selecting text and dragging the mouse to the right side. That pseudo-element seems to not be used, as there where not other visible styles attached to it (whooot? yes 🫣) so we remove it for now. It may be some helper for scrolling or something. But it should then be declared as so.

### DIFF
--- a/packages/dnb-eufemia/src/components/dialog/__tests__/__snapshots__/Dialog.test.tsx.snap
+++ b/packages/dnb-eufemia/src/components/dialog/__tests__/__snapshots__/Dialog.test.tsx.snap
@@ -2035,14 +2035,6 @@ button.dnb-button::-moz-focus-inner {
     width: 100%;
     height: 100%;
     overflow-x: hidden; }
-    .dnb-drawer__inner::after {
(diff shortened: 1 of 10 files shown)


langz (approved):
Hehe, seems magic that the pseudo-element was not used 🪄 
Any scenarios on any browsers or devices you would want me to assist testing on?

tujoworker:
> Any scenarios on any browsers or devices you would want me to assist testing on?

I did a wider research and found that there is outdated color related stuff. For me it looks that the pseudo-element was used for a "filled" background color before, and after the large refactor, things got even more split up. The screenshot attached shows how it was before the Modal=>Drawer/Dialog split: 

Screenshot 2022-08-15 at 12 53 25

We only use the "style_type" on the `Drawer.Body` today, which again is documented.

PR #1517, opened 2022-08-18
feat(Slider): add multi thumb button support

tujoworker:
This is a larger rewrite of the Slider component in order to support multiple thumb buttons. Check out this [demo](https://eufemia-featmultibuttonslider.gtsb.io/uilib/components/slider/#slider-with-multiple-thumb-buttons).

Screenshot 2022-08-19 at 10 05 19

- [x] Have a look if it makes sense to use memo somewhere.


packages/dnb-eufemia/src/components/slider/SliderHelpers.tsx
```diff
@@ -0,0 +1,144 @@
...
+    height,
+    top: 0,
+    left: 0,
+    right: width,
+    bottom: height,
+  })
+  return div
+}
+
+export const getValueByIndex = (
```
ahmetacer5, line 99: I think function name misleading, should we replace it with something else like `getUpdatedValues`?

langz (approved):
Looks very nice, big job 🏗️ 
I tested it in a few different browsers locally, and on my iPhone, and it worked great :)

packages/dnb-eufemia/src/components/slider/SliderProvider.tsx
```diff
@@ -0,0 +1,233 @@
...
+
+      if (isMulti) {
+        const currentIndex = handleCurrentIndex(currentValue)
+
+        newValue = getValueByIndex(
+          currentIndex,
+          currentValue,
+          value as Array<number>
+        )
+      }
```
ahmetacer5, line 155: what you think about this suggestion? 

```suggestion
      if (isMulti) {
        const currentIndex = handleCurrentIndex(currentValue)
       
        newValue = [...value]
        newValue.splice(
          currentIndex,
          1,
          currentValue
        )
      }
```

packages/dnb-eufemia/src/components/slider/SliderProvider.tsx
```diff
@@ -0,0 +1,233 @@
...
+        const obj: onChangeEventProps = {
+          value: newValue,
+          rawValue,
+          raw_value: rawValue, // deprecated
+          event,
+          number: null,
+        }
+
+        if (number_format) {
+          obj.number = formatNumber(currentValue, number_format)
```
ahmetacer5, line 167: This logic does not respect array values when isMulti is true.

ahmetacer5:
I think when there is more than 2 element in the array the slider track disappears.
image

packages/dnb-eufemia/src/components/slider/SliderProvider.tsx
tujoworker: It does. Because `currentValue` will always be a number. But in contrast to the `value`, the `number` will always be a string of the thumb you just change, and not an array with all numbers, like the `value`. But I think, this is ok, because it "complicates" the usage and is maybe not even needed. If you want, you can easily map over the `value` array and format it to your needs. The reason why we also support the number format, is because of the internal labels and for screen readers.

tujoworker:
> I think when there is more than 2 element in the array the slider track disappears.

That is on purpose. Because, where should it show up? Also, I guess we later need to support to avoid showing (by a property) it for even two thumbs. Because, sometimes it does not make sense to show it.

What do you think?

packages/dnb-eufemia/src/components/slider/SliderProvider.tsx
ahmetacer5: I agree it complicates more, I also see that we never display formatted value within the slider component we just format it and return it to user then this makes me think of removing number_format from Slider component. What do you think?

ahmetacer5:
> 

I think we can show it between first and the last item passing below the thumbs in between. Like in the screenshot below.

image

if we are applying a track style it should support all the cases. Highlighting range being selected makes it more accessible and also that is the only thing that indicates where is the start and the end.

tujoworker:
> between first and the last item

Yeah, maybe. I check with UX 👍

tujoworker:
The feedback is, that this is not used in DNB . So, for now, I made it like you suggest [with this change](https://github.com/dnbexperience/eufemia/compare/b2785fd7c5d4494c47835e810db630a6ba7a7871..490cf3f9cb49a4b068989785b6b12bc8d7c0b88b)

ahmetacer5:
> 

looks good thank you 👍

ahmetacer5:
I cannot approve it anymore, because it's merged already, But I approve by commenting 😄 ✅

tujoworker:
> I cannot approve it anymore, because it's merged already, But I approve by commenting 😄 ✅

👍 ohh sorry – I need to watch out better next time! Thanks for all your feedback and let me have a new change in future – getting feedback from you 🙏

PR #1518, opened 2022-08-19
feat(Anchor): rewrite to TypeScript

### LINKED ISSUE #1516: Error using Anchor element in TypeScript file
## 🐛 Bug Report

TypeScript error when trying to use the Anchor component.
When trying to use the Anchor element in a TypeScript file, both locally(in eufemia) and externally(other apps).
I get the following error:

## Error screenshot

image

## Error in text:

(alias) module Anchor
(alias) const Anchor: React.ForwardRefExoticComponent<React.RefAttributes<any>>
import Anchor
Type '{ children: string; href: string; }' is not assignable to type 'IntrinsicAttributes & RefAttributes<any>'.
  Property 'children' does not exist on type 'IntrinsicAttributes & RefAttributes<any>'.ts(2322)
No quick fixes available

## To Reproduce

Try to use the Anchor element in a TypeScript file

## Eufemia Version

v.9.28.0

tujoworker:
This is a very quick JS to TS rewrite. Let me know if you find things we should double check or improve.

fixes #1516


packages/dnb-eufemia/src/components/skeleton/SkeletonHelper.d.ts
```diff
@@ -23,8 +23,10 @@ export interface SkeletonContextProps {
   };
 }
 
+export type SkeletonMethodes = 'shape' | 'font' | 'code';
```
langz, line 26:
```suggestion
export type SkeletonMethods = 'shape' | 'font' | 'code';
```

I think `methodes` could work as well, but it is an ancient way of spelling it 👴

langz (approved):
Thanks, that went fast 🐎 
The rewrite looks good to me :)
One thing I'm a bit confused about, not related to the rewrite itself, is why some components have props in both _snake_case_ and _camelCase_.
Probably a task we can do in the future sometime 🕦

packages/dnb-eufemia/src/components/skeleton/SkeletonHelper.d.ts
tujoworker: Sorry for the confution! The `inner_ref` is an very old thing. Later I needed `internalClass` and because we prefer now camelCase (and Web Components support is not needed anymore) it mate it in as camelCase.

Anyway, I think, also the Anchor, and all other re-written TS components (Slider and Modal) an get an additional internal re-write, where we rather rewrite them to camelCase, but provide a `withSnakeCaseProps` – instead of `withCamelCaseProps` as we do now.

PR #1520, opened 2022-08-22
chore(withSnakeCaseProps): add camelCase to snake_case converter

tujoworker:
This can be used to re-write components to camelCase, but provide snake_case backwards compatibility.


langz (approved):
Nice nice 👍

PR #1522, opened 2022-08-23
fix(Drawer): fix possible Safari selection causing selection scroll (horizontal) issue

tujoworker:
Reported [here](https://dnb-it.slack.com/archives/CMXABCHEY/p1661162150422519?thread_ts=1660052015.576469&cid=CMXABCHEY) 🐛

Here is a [demo build](https://eufemia-fixdrawerstickyheader.gtsb.io/uilib/components/drawer/demos/#drawer-with-custom-content).

It only happens in Safari Desktop (v15). And its a Safari bug. But this PR provides a fix for it.

## How to reproduce?

When scrolling down, and selecting text from left to right, the content of the drawer, moves to the left.

## The reason?

The shadow of the header bar, which is again in a pseudo-element, makes the content moving to the left. If we remove the `overflow: hidden` also a Safari scroll-bar (horizontal) will show up.

## The fix

We can avoid that the pseudo-element is placed "outside of the view" to the right, and use rather a negative margin with a padding to get the correct spacing and placement in place.

### DIFF
--- a/packages/dnb-eufemia/src/components/dialog/__tests__/__snapshots__/Dialog.test.tsx.snap
+++ b/packages/dnb-eufemia/src/components/dialog/__tests__/__snapshots__/Dialog.test.tsx.snap
@@ -2142,15 +2142,16 @@ button.dnb-button::-moz-focus-inner {
     position: sticky;
     top: 0;
     z-index: 99;
-    margin: var(--drawer-spacing) 0; }
+    margin: var(--drawer-spacing) calc(var(--drawer-spacing) * -1);
+    padding: 0 var(--drawer-spacing); }
   .dnb-drawer--spacing .dnb-drawer__navigation.dnb-section.dnb-drawer__navigation--sticky {
     z-index: 2999; }
   .dnb-drawer .dnb-drawer__navigation--sticky::before {
     content: '';
     position: absolute;
     z-index: -1;
-    left: -50%;
-    right: -50%;
+    left: 0;
+    right: 0;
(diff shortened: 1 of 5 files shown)


packages/dnb-eufemia/src/components/dialog/__tests__/__snapshots__/Dialog.test.tsx.snap
```diff
@@ -2142,15 +2142,16 @@ button.dnb-button::-moz-focus-inner {
     position: sticky;
     top: 0;
     z-index: 99;
-    margin: var(--drawer-spacing) 0; }
+    margin: var(--drawer-spacing) calc(1px - 1px - var(--drawer-spacing));
```
dinarosv, line 2145: Why `1px - 1px` here?

dinarosv (approved):
Looks good! :)

packages/dnb-eufemia/src/components/dialog/__tests__/__snapshots__/Dialog.test.tsx.snap
tujoworker: To get minus. Should we add a comment? (we have this a couple of places)

dinarosv: Hmm I see, is it maybe possible to use `var(--drawer-spacing) * -1`?

PR #1523, opened 2022-08-23
fix(Slider): (multi-thumbs) ensure correct index is set when thumb positions switches

tujoworker:
When pressing a second time on a thumb, after it has changed position, the internal index was not updated.

Bug:

https://user-images.githubusercontent.com/1501870/186253079-4fae245f-430d-411a-b1bb-bc9610dc196a.mov

Fix:

https://user-images.githubusercontent.com/1501870/186253105-4df1e666-33f8-48de-8198-73955c2a9dc9.mov


langz (approved):
Nice find, I didn't notice this the last time I gave it a try 👍

PR #1525, opened 2022-08-24
docs(Lists): add Definition List example using CSS Grid

tujoworker:
The question is, should this be possible to achieve within the "component", without having these CSS lines? I think so as of now. Will have to check with UX.

The only thing is the "minimum width" that should be editable.

Screenshot 2022-08-24 at 15 31 28


AadneRo (dismissed):

AadneRo (commented):
It looks like the `Remove list styles` appear twice in the portal. Unless there is something I missed

tujoworker:
> appear twice in the portal

OMG, they are! 🫣 Thank You 🙏

tujoworker:
Closing this PR in favor of #1536

PR #1526, opened 2022-08-24
feat(Slider): add multiThumbBehavior property

tujoworker:
In addition to the new property;

- Add a test + some more
- Add type docs

https://user-images.githubusercontent.com/1501870/186597285-13a41a28-de1e-4bc4-ae20-b08a58b567b9.mov


ahmetacer5:
> In addition to the new property;
> 
> * Add a test + some more
> * Add type docs
> 
>  slider.swap.mov

What you think about this behaviour?

https://user-images.githubusercontent.com/16961900/186603193-b027f7e7-8042-43c6-9d39-511a42e77119.mov

tujoworker:
> What you think about this behaviour?


yeah, we may add this as well. Do you have any property name recommendation? Or should it do so as default? It differs if the use-case I would say, so both can be interesting. Probably as a separate feature, or as an alternative:

`omitThumbSwap: 'push'`

ahmetacer5:
> 

What you think about this?
thumbBehaviour: 'omit' | 'push' | 'swap' = 'push'

packages/dnb-eufemia/src/components/slider/__tests__/Slider.test.tsx
```diff
@@ -290,6 +310,60 @@ describe('Slider component', () => {
...
+
+      expect(onChange.mock.calls[1][0].value).toEqual([10, 50, 50])
+      expect(getThumbElements(0).getAttribute('style')).toBe(
+        'z-index: 3; left: 10%;'
+      )
+      expect(getThumbElements(1).getAttribute('style')).toBe(
+        'z-index: 3; left: 50%;'
+      )
+      expect(getThumbElements(2).getAttribute('style')).toBe(
+        'z-index: 4; left: 50%;'
```
langz, line 361: Perfect, now it makes sense, was wondering why it was 55% earlier 🙏

langz: Great test by the way :)

tujoworker:
Yeah, we can do that 👍 

Regarding the push behavior – I see, material UI is [recommending](https://codesandbox.io/s/tmp3br?file=/demo.tsx) doing it in your app instead. But the code looks so terrible to have in your app code, just to get the push behavior  🤦‍♂

Screenshot 2022-08-25 at 09 44 15
️

langz (approved):
Nice feature 💯 
If we think this will be used a lot, it could be beneficial to add an example to the docs, so that it will be easier to see/find out that we have this feature for devs.

ahmetacer5:
> 

Yea I agree better to have the logic within the component.

tujoworker:
> it could be beneficial to add an example to the docs

yeah, made an example there as well now 👍

https://user-images.githubusercontent.com/1501870/186615626-55bcda49-05ba-41cb-bd5a-a6958f89fca2.mov

PR #1529, opened 2022-08-25
feat(Slider): add tooltip on active thumb button

tujoworker:
Its optional to show a Tooltip. To show it, use the `tooltip` property. Also, `alwaysShowTooltip` will ensure, it always is visible, and not only on `mouseOver`, `touchStart` and `focus`.

Check out the docs and the [demos](https://eufemia-featslidertoolbar.gtsb.io/uilib/components/slider/).

Here is a [CSB](https://codesandbox.io/s/eufemia-slider-tooltip-cnl7be?file=/src/App.tsx) and a 👉 [demo to test out](https://cnl7be.csb.app/).

- [x] Should we only show a toolbar, if a formatting is given? Or should it be prop based? `tooltip={true}`

Relies on PR #1547

Screenshot 2022-08-25 at 12 34 35

Screenshot 2022-08-25 at 12 46 14

Screenshot 2022-08-25 at 12 50 50


langz:
I'll take a look at this, and test it a bit when I get time tomorrow 🕙

langz:
A small nitpick that I saw when using my iPhone, both Safari and Chrome.

There's a small gap in the tooltip, between the "circle" and the "arrow", but it only happens at certain times/values:
302283051_801310464399567_2180611951952021715_n
302345667_859426375038146_4013779052565230971_n

tujoworker:
yeah, I agree – we should fix this in another PR 👍

langz:
Some other thoughts 💭 

I can see that we are able to click/drag the tool tip, which will also move/drag the slider. 
Should it be like that, or would clicking on the tooltip rather make it possible to select the text in the tooltip, like here https://eufemia-featslidertoolbar.gtsb.io/uilib/components/tooltip/#button-with-active-tooltip ?

Do we have to consider what will happen if the tooltip has more text than what fits in the container/screen size? Will the text in the tooltip break into multiple lines?

langz (approved):
🥇

tujoworker:
> Should it be like that, or would clicking on the tooltip rather make it possible to select the text in the tooltip

Yeah, nice finding! And yes, we should change that and make it select able.

When it comes to line wrapping, I will have a look at it as well.

tujoworker:
Regarding the wrapping. Because of UX, we should not wrap numbers. And because the Slider, in theory, only will show numbers, I think, its OK to have `nowrap` for now.
I made some changes to the Tooltip behavior. Now the text is selectable, but there are some small issues with the position. We need to fix this in the Tooltip component in another PR later.

PR #1535, opened 2022-08-29
fix(Autocomplete): make DrawerList direction observer work

tujoworker:
Reported in this [Slack Thread](https://dnb-it.slack.com/archives/CMXABCHEY/p1656411051734679).

The not documented, and not used `min_height` inside Autocomplete, which again was forwarded to the `DrawerList`, did make the calculation to be `NaN`, so it resulted in not reacting and aligning the opening direction.

The added test do fail, when `min_height: null` is defined in the default props (like before).

The test and the mocks are 100% the same, so I think its fine to share them.

You can [click on this link](https://eufemia-fixautocompletedirectionoberse.gtsb.io/uilib/components/autocomplete/#default-autocomplete) to test it out. Simply focus the Autocomplete Input and scroll up & down.

### DIFF
--- a/packages/dnb-eufemia/src/components/autocomplete/Autocomplete.js
+++ b/packages/dnb-eufemia/src/components/autocomplete/Autocomplete.js
@@ -221,7 +221,6 @@ export default class Autocomplete extends React.PureComponent {
       PropTypes.string,
       PropTypes.node,
     ]),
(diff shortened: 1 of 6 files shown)


langz (approved):
Great investigation 🕵️ 
Couldn't be too easy to find, even the though the fix is "just removing a prop"  🤯

PR #1536, opened 2022-08-29
feat(DefinitionList): add horizontal direction (inline) support

tujoworker:
Replaces PR #1525

This is how it looks now ([demo](https://eufemia-featinlinedefinitionlist.gtsb.io/uilib/elements/lists/#definition-list-in-horizontal-direction)):

Screenshot 2022-08-30 at 13 10 50


AadneRo (approved):
Nice! I'll use this in the Upload component🎉

PR #1537, opened 2022-08-30
fix(Slider): make it optional to provide an updated prop value

tujoworker:
This change is based on feedback & experience from Ilnur.

So we also change the docs to make it clear that the `max` property is important to use.


packages/dnb-eufemia/src/components/slider/SliderProvider.tsx
```diff
@@ -102,6 +102,8 @@ export function SliderProvider(localProps: SliderProps) {
   } = allProps
 
   const [value, setValue] = React.useState<ValueTypes>(_value)
+  const [exterValue, updateExternValue] =
```
langz, line 105:
```suggestion
  const [externValue, updateExternValue] =
```

langz (approved):

packages/dnb-eufemia/src/components/slider/SliderProvider.tsx
tujoworker: Thank you 🙏

PR #1538, opened 2022-08-30
docs(RangeSlider): remove the docs and deprecate the styles to be removed in v10

tujoworker:


langz (approved):
👍 
Just curious, but how do we actually remember to remove these comments and lines of codes by v10 release? Some kind of tooling that helps us? Or is it pure mental power 🧠

tujoworker:
> 👍 Just curious, but how do we actually remember to remove these comments and lines of codes by v10 release? Some kind of tooling that helps us? Or is it pure brain power 🧠

If we note that something is deprecated in the v10 release notes, but also note it in every place it should be changed/removed, then we should search though manually and make some PR. But it often even does not matter if it happens later in v10 or in the first release.
I'm not aware of any smart tools helping us with that. But I assume there are some 🤔😀

PR #1541, opened 2022-08-31
fix(StepIndicator): change chevron icon to pointing to the right

tujoworker:
[Slack Thread](https://dnb-it.slack.com/archives/CMXABCHEY/p1656588844209229) with more info about the change.

Here is how it behaves now:

https://user-images.githubusercontent.com/1501870/187615866-de8c87af-f76f-4ce4-9a2b-a04c4f3ac2fc.mov


langz (approved):
▶️

PR #1542, opened 2022-08-31
fix(DrawerList): remove unused white area on the right side

### LINKED ISSUE #1531: Let Autocomplete's options use the full width, before it breaks to a new line
It would be awesome if Autocomplete's options could use the full width in the drawer before breaking into a new line.

Attached is a screenshot from a form I'm making where we use Autocomplete to help the user to select country code for a phone number:

image

It's clearly more space available on the right side, not sure why it's not available for use 🤔 

I think the reason for it to "break early" is something something related to https://github.com/dnbexperience/eufemia/blob/fe6fbb73fdba4ebf2166710d6ff48bfba36a5876/packages/dnb-eufemia/src/fragments/drawer-list/style/_drawer-list.scss#L179

Here's a screenshot of the same example, when I just remove the line of styling above:

image

An other example of when it doesn't "use the full width" can also be seen in the docs, https://eufemia.dnb.no/uilib/components/autocomplete/demos/#custom-width, pressing the top autocomplete example, the option "The Godfather" could have enough space to not wrap the text "The Godfather" over two lines.

image

tujoworker:
Fixes #1531


langz (approved):
Thanks, I haven't had time to look at improving this yet, but now you've done it 😸

PR #1543, opened 2022-08-31
fix(Autocomplete): ensure value is not visible behind the trigger button

tujoworker:
Before:
Screenshot 2022-08-31 at 10 51 27

After:
Screenshot 2022-08-31 at 10 51 57


packages/dnb-eufemia/src/components/autocomplete/style/_autocomplete.scss
```diff
@@ -134,8 +134,8 @@
       cursor: text;
     }
 
-    &--has-submit-element .dnb-input__placeholder,
-    &--has-submit-element .dnb-input__input {
```
langz, line 138: Are we sure that we don't need this styling for `&--has-submit-element `as well?

Reason for asking is just because I find it a bit strange that these existing lines of code didn't actually do anything useful 🤔

langz (approved):

packages/dnb-eufemia/src/components/autocomplete/style/_autocomplete.scss
tujoworker: We do actually not remove it, but rather use `--has-inner-element`  👍 because its only then we want to cahnge/remove the padding:

Screenshot 2022-08-31 at 14 25 37

PR #1544, opened 2022-08-31
fix(Input): ensure dnb-input--null class will not be set

tujoworker:
It has no visual or negative effect. Just something I saw in the Autocomplete component.


langz (approved):
💯

PR #1549, opened 2022-09-05
fix(Tooltip): convert to TypeScript

tujoworker:
Next, we can convert it to use React Hooks along with testing lib 👍


packages/dnb-eufemia/src/components/tooltip/Tooltip.tsx
```diff
@@ -4,101 +4,36 @@
...
-    hide_delay: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
-    target_selector: PropTypes.string,
-    target_element: PropTypes.oneOfType([
-      PropTypes.object,
-      PropTypes.node,
-    ]),
-
-    ...spacingPropTypes,
-
-    class: PropTypes.string,
```
langz, line 67: Is it safe to remove this, not sure if found it in [TooltipProps](https://github.com/dnbexperience/eufemia/blob/452e889307c7420aafc179b0287ebc6de5a08987/packages/dnb-eufemia/src/components/tooltip/types.ts)

packages/dnb-eufemia/src/components/tooltip/TooltipContainer.tsx
```diff
@@ -4,62 +4,33 @@
...
-    fixed_position: null,
-    no_animation: null,
-    useHover: true,
-    style: null,
-    attributes: null,
-    children: null,
-  }
+type TooltipContainerState = {
+  w?: number
+  h?: number
```
langz, line 23: At a glance, I am not 100% sure what `w` and `h`, could we give these props any more meaningful names?

packages/dnb-eufemia/src/components/tooltip/Tooltip.tsx
tujoworker: `spacingPropTypes` is basically the same as `Props & ISpacingProps` – while `class` is not documented, and is a Web Component feature, which is, from what I know, not used in prod.

langz (approved):

packages/dnb-eufemia/src/components/tooltip/TooltipContainer.tsx
tujoworker: yes 🙏 I will rename them to bodyWidth, because that is what it is.

PR #1552, opened 2022-09-05
fix(Tooltip): use Eufemia cubic-bezier for animations

tujoworker:


langz (approved):
Gave it a spin in a few browsers, and on my phone, tried to compare it with the existing animations on https://eufemia.dnb.no/, but wasn't actually able to see the visual difference 👀 👓

tujoworker:
> but wasn't actually able to see the visual difference 👀 👓

yeah, its probably only noticeable when the position is animating:

Old
https://user-images.githubusercontent.com/1501870/188561969-4f80c85f-0c29-4877-b283-0b5b6948d1f7.mov


New
https://user-images.githubusercontent.com/1501870/188562000-dac5fd24-467b-42df-9253-c88e24e8e0d2.mov

PR #1555, opened 2022-09-12
fix(Tooltip): rewrite to functional components with React Hooks

tujoworker:
[Preview](https://eufemia-fixtooltiphoosrewrite.gtsb.io/uilib/components/tooltip/demos/)


packages/dnb-eufemia/src/components/tooltip/TooltipWithEvents.tsx
```diff
@@ -4,145 +4,123 @@
...
-      'current'
-    )
-      ? props.target
-      : React.createRef()
-  }
+function TooltipWithEvents(props: TooltipProps & TooltipWithEventsProps) {
+  const {
+    children,
+    target,
+    // internalId,// NB: Do not remove internalId from props!
```
langz, line 22: Maybe it would be even better/easier to have this comment closer to the type, `type TooltipWithEventsProps`.
Like on the following line? https://github.com/dnbexperience/eufemia/blob/09142b1d7a7339b2fab4224075cb342a049e58f4/packages/dnb-eufemia/src/components/tooltip/TooltipWithEvents.tsx#L15

langz (approved):

packages/dnb-eufemia/src/components/tooltip/TooltipWithEvents.tsx
tujoworker: yeah, that would make sense. But maybe I just have stated the sentence too poorly. What I want to say is, do not "destruct" `internalId` – because we want it to keep it in the "rest" props, and send it along. Any recommendation of a better sentence?

tujoworker: We also can remove it, and ensure this behavior with a test 🤭

langz: What you wrote, _do not "destruct" internalId – because we want it to keep it in the "rest" props, and send it along_, described it perfectly for me. Now I understood it, thanks 🙇

PR #1561, opened 2022-09-14
feat(useHeightAnimation): add hook to make height (auto) animations easy

tujoworker:
Will be used in #1484

PR #1566, opened 2022-09-16
feat(HeightAnimation): add new component

tujoworker:
This replaces the previews PR #1561
I think we rather simplify the usage and make it a opt-in  element replacement, instead of dealing with a Hook.

Questions:
- Is it overkill to have it as a own component?


packages/dnb-eufemia/src/components/space/Space.js
```diff
@@ -33,15 +33,15 @@ export default class Space extends React.PureComponent {
...
-    element: PropTypes.string,
+    element: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
     inline: PropTypes.oneOfType([PropTypes.string, PropTypes.bool]),
     no_collapse: PropTypes.oneOfType([PropTypes.string, PropTypes.bool]),
 
     ...spacingPropTypes,
 
     stretch: PropTypes.oneOfType([PropTypes.string, PropTypes.bool]),
     skeleton: PropTypes.oneOfType([PropTypes.string, PropTypes.bool]),
-    class: PropTypes.string,
```
langz, line 44: Is it safe to remove this?
I see we have this in defaultProps, and destructures it from props.

packages/dnb-eufemia/src/components/space/__tests__/__snapshots__/Space.test.js.snap
```diff
@@ -3,29 +3,29 @@
 exports[`Space component have to match snapshot 1`] = `
 <Space
   bottom={null}
-  class="class"
-  className="className"
+  class={null}
```
langz, line 6: As far as I understood, we removed this prop on X 🤔

langz (approved):

langz:
Should we remove the hooks tab that was added to helpers doc as part of #1561 ?
https://github.com/dnbexperience/eufemia/pull/1561/files#diff-552d4d07787de8a0fcb5b900d1cbbbf7f6eb27707968cdb4a828431192cf06b8

langz:
Seems like the component(left side of the video) has a bit more "choppy animation" compared to the hook(right side of the video), at least in Chrome on my Mac.


https://user-images.githubusercontent.com/1359205/190922126-7ab65687-5598-4bcc-8c93-b2c8cf0f8909.mov

packages/dnb-eufemia/src/components/space/Space.js
tujoworker: Its a Web Component thing. Its not documented else, should be safe.

packages/dnb-eufemia/src/components/space/__tests__/__snapshots__/Space.test.js.snap
tujoworker: Yes, true 👍

tujoworker:
> Seems like the component(left side of the video) has a bit more "choppy animation" compared to the hook(right side of the video), at least in Chrome on my Mac. Might be that we don't expect the same result, but I thought so 🤔
> Screen.Recording.2022-09-18.at.20.09.39.mov

Nice findings! The example has had a margin top inside the content. When it gets removed form the DOM, it suddenly jumps. We can't do much about it if that would happen in a prod app. But its always possible to use padding or a wrapper outside of the animated children instead.

PR #1568, opened 2022-09-19
chore(EventEmitter): add useEventEmitter

tujoworker:
- For internal usage only as of now
- Rewrite `EventEmitter` to TypeScript
- Move it into `/helpers`
- add `useEventEmitter` React Hook + Story Book example

Here is a [CSB](https://codesandbox.io/s/eufemia-event-emitter-example-t2dtut?file=/src/App.tsx)


packages/dnb-eufemia/src/shared/helpers/__tests__/useEventEmitter.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  it('has update function', () => {
+    const { result } = renderHook(() => {
+      return useEventEmitter('unique-id')
+    })
+
+    expect(typeof result.current.update).toBe('function')
+  })
+
+  it('has will set data', () => {
```
AadneRo, line 25: Can we rename this test description? maybe "It stores the value when updating"

AadneRo (commented):
Except from my one comment this looks good to me, I don't understand everything though 😂

packages/dnb-eufemia/src/shared/helpers/__tests__/useEventEmitter.test.tsx
tujoworker: 🙌 Yes, very good point – I added some more tests as well.

PR #1569, opened 2022-09-20
feat(HeightAnimation): adjust height with animation when content changes

tujoworker:
This PR is a small adjustment in order to adjust the height automatically when its opened and the content changes.

While nested content will disappear immediately (by React), we make the height animation smooth, especially for the content below.

[Preview](https://eufemia-featheightanimationtransitions.gtsb.io/uilib/components/height-animation/demos/#heightanimation).

Some thoughts: Here we enter a grey-zone for when this will be needed and useful. There are may other ways to solve the whole animation stagey. Like make a copy of effected DOM elements and use these for the animation, especially during disappearance.

- [x] Add integration test

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/components/height-animation/Examples.tsx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/components/height-animation/Examples.tsx
@@ -12,6 +12,8 @@ export function HeightAnimationDefault() {
         /* jsx */ `
 const Example = () => {
   const [openState, setOpenState] = React.useState(false)
+  const [isOpen, setIsOpen] = React.useState(openState)
+  const [contentState, setContentState] = React.useState(false)
 
   const onChangeHandler = ({ checked }) => {
     setOpenState(checked)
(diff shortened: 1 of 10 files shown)


langz:
I do find it a bit difficult to evaluate the different ways of solving the animation strategy.
I think this approach looks alright, and think it could be a good strategy going forward with 🔮

langz:
Here's a video of me clicking a bit around in the [preview](https://eufemia-featheightanimationtransitions.gtsb.io/uilib/components/height-animation/demos/#heightanimation)

Not sure if it's the code or the "example code" that makes it sometimes feel a bit "yanky"(the "change height as well"-button sort of flashes in or out, etc. See video around 13/14 sec and 22/23 sec.)

Also it seems like the edges of the "change height as well"-button is also cut at times.

Screenshot 2022-09-21 at 11 36 47




https://user-images.githubusercontent.com/1359205/191469764-63d433ea-f42c-4e1a-a6e4-12785b39a70b.mov

langz (approved):

tujoworker:
> Here's a video of me clicking a bit around in the [preview](https://eufemia-featheightanimationtransitions.gtsb.io/uilib/components/height-animation/demos/#heightanimation)
> 
> I am not sure if it's the code of the component or the "example code" that makes it sometimes feel a bit "janky"(the "change height as well"-button sort of flashes in or out, etc. See video around 13/14 sec and 22/23 sec.)
> Screen.Recording.2022-09-21.at.11.31.45.mov
> 
> Also it seems like the borders/edges of the "change height as well"-button is also cut at times.
> 
> Screenshot 2022-09-21 at 11 36 47

Regarding the cut edges, this happens due to the fact that [we want to have](https://github.com/dnbexperience/eufemia/pull/1569/files#diff-3b5fb3be52818733ebcbc3ae0a3029a6a278c4a0c01d8794781b260b4a919f63R10) `overflow-y: hidden` – but you can not overflow only in once direction – yet.

PR #1576, opened 2022-09-21
chore: simplify usage of extendPropsWithContext

tujoworker:
The motivation is to simplify the `extendPropsWithContext` and make it clear when to use what (function vs class components). We also do not need a Hook for it, so it replaces `usePropsWithContext`.


packages/dnb-eufemia/src/components/form-row/FormRow.js
```diff
@@ -147,10 +147,10 @@ export default class FormRow extends React.PureComponent {
 
   render() {
     // use only the props from context, who are available here anyway
-    const props = extendPropsWithContext(
+    const props = extendPropsWithContextInClassComponent(
       this.props,
       FormRow.defaultProps,
-      this.context.FormRow
+      this.context.FormRow // do not run "includeValidProps" here
```
langz, line 153: How come?
Is it because we have other components that run includeValidProps which use this component FormRow? 🤔

packages/dnb-eufemia/src/shared/helpers/filterValidProps.tsx
```diff
@@ -27,4 +27,19 @@ function filterValidProps<Props>(
...
-export { filterValidProps }
+/**
+ * Filters out props from a given object/context
+ * It returns a new object, with keys defined in validKeys
+ *
+ * @param props object of component properties
+ * @param excludeKeys object with keys that should be excluded
+ * @returns filtered properties
+ */
+function filterInvalidProps<Props>(
```
langz, line 38: Does it make sense to rename the filename? 
Not important, but it's a bit funny to `import { filterInvalidProps } from '../filterValidProps'` 🃏

langz (approved):
Nice nice, a lot of changes.
Any way I can assist in validating/testing this?

packages/dnb-eufemia/src/components/form-row/FormRow.js
tujoworker: ohh, yes, I think its not valid anymore. Will use `includeValidProps` there as well and check it out.

packages/dnb-eufemia/src/shared/helpers/filterValidProps.tsx
tujoworker: Yes, I totally agree, its rarely used, so we rather can only use the `filterValidProps` 👍

tujoworker:
It's hard to test this manually. We should probably add always a [provider test](https://github.com/dnbexperience/eufemia/blob/bf3769f8d6dc8bd38f672ac99fa6525e5092204d/packages/dnb-eufemia/src/components/breadcrumb/__tests__/Breadcrumb.test.tsx#L44) in every component. But not only testing locale, but some essential component properties.

PR #1578, opened 2022-09-22
feat(Logo): add inherit_color prop

tujoworker:


langz (approved):
Looks great 💚 

A thought...
Should the DNB Logo actually be able to inherit any color?
I assume the Brand Center would have a few rules regarding which colors the logo can have. But we shouldn't probably have those rules in our code 🤔 😹

tujoworker:
> Looks great 💚
> 
> A thought... Should the DNB Logo actually be able to inherit any color? I assume the Brand Center would have a few rules regarding which colors the logo can have. But we shouldn't probably replicate those rules in our code 🤔 😹

Black and white is often used 🤷‍♂️

PR #1579, opened 2022-09-22
fix(Avatar): don't overwrite SVG color

tujoworker:
Due to SVG path changes, the [card icon](https://github.com/dnbexperience/eufemia/pull/1575/files#diff-eac78f1448dd7a3bd40fe11a60945a1412cc91bc75027e8ad08aebefeec994f8) can not get colored with `fill` color anymore:

Screenshot 2022-09-22 at 12 58 41

By default, the Icon component does inherit colors, so we may rather relay on that. The only thing left (with a negative result) is the Logo component. This PR #1578 will fix our tests.


langz (approved):
Nice find, thanks for fixing this :)
Any other places where this problem could occur, where we style/overwrite svg?

tujoworker:
> Nice find, thanks for fixing this :) Any other places where this problem could occur, where we style/overwrite svg? 🤔 I didn't find any more though.

I could not find any places 👍

PR #1580, opened 2022-09-23
chore(Icons): add colored icons visual test

tujoworker:
With that test, we can catch icons, that are build in a different way, or that our CSS code for inheriting colors does not work properly.


langz (approved):
Super neat tests 🎸

PR #1582, opened 2022-09-26
fix(Modal): Safari Desktop fullscreen video issue

tujoworker:
Here is a reprod [CSB](https://codesandbox.io/s/eufemia-modal-fullscreen-safari-bug-3sv139?file=/src/App.tsx). 

# The bug 

Run [this here](https://3sv139.csb.app) in Safari Desktop, click on the full-screen button, even without playing the video:

Screenshot 2022-09-26 at 18 26 34

You see this here, instead of the full-screen video:

Screenshot 2022-09-26 at 18 30 12

# Reason and solution

This is because of the CSS animation and the `forwards` usage. But happily, we do not need in on the "show" animation, because, that's anyway our "default". So we can remove it. Full-screen video works now fine without it.


langz (approved):
Interesting 🤔 ✅

PR #1581, opened 2022-09-26
feat(Upload): add new Upload component

AadneRo:
Add Upload component mvp.

Still missing some features:
- Drag and drop to upload file

I also removed the validation, we can add validation later if needed

PR #1583, opened 2022-09-27
chore(Icon): replace data-test-id with data-testid

langz:
This PR renames the property `data-test-id` to  `data-testid`, so that we can use the `ByTestId` query from [Testing Library](https://testing-library.com/docs/queries/bytestid/).

The suggested improvement has come as a result of this discussion.
https://github.com/dnbexperience/eufemia/pull/1581#discussion_r980069352


tujoworker (approved):
✨ That looks 😻  🕺 `removeIconTitle` 🕺

PR #1584, opened 2022-09-28
docs(Icon): replace data-test-id with data-testid

langz:
I found out that the changes as part of #1583 is to some extent "breaking" for a few applications in the monorepo, as I see a few tests that rely on the `data-test-id`, that will need to be changed to `data-testid`.


tujoworker:
I personally would like to not use `data-testid` insdie of Eufemia. When you use SCSS modules, you are almost forced to use it. But we use BEM, and there I do not see it as a good lib pattern, because apps use it. And with that, the "refactor" argument you have with testid usage, also goes away.

I do not really see as a breaking change either, because the app works fine still, its just the tests do fail, and needs to be updated. 

But we can ask/or mention this to the team, using it, so they know it and can act when updating to this version.

Its not documented, so it should not really be an issue to "change" from our side.

tujoworker (approved):

PR #1585, opened 2022-09-28
chore(CI): bail on first failing visual test

tujoworker:
Not sure if this is something that makes things better. But we can safe potentially save some time, if it fails early.


langz (approved):
Let’s go for it🤞

Issue #1411, opened 2022-05-16
Tooltip: Disappears after a brief time with controlled `active` prop


dagerikhl:
## 🐛 Bug Report

When controlling a tooltip by setting the `active` property based on whether or not there is content for the tooltip, the tooltip sometimes disappears right after appearing.

In my application, I'm reliant on controlling the active prop.

It looks like the active prop is sent in correctly to the top-level Tooltip component, but somewhere down in the component tree it suddenly changes to `false`.

Did not have this problem in Eufemia v9.24.0, but I did have it in 9.26.0. I suspect this commit is the culprit: fb0fe332ee05c5ee6bee66b9b401106f96066b1f

Not sure if is also present in 9.25.0, but I expect so.

## To Reproduce

https://codesandbox.io/s/sharp-star-3x53k3

1. Try to hover the Button a couple of times.
2. Observe that sometimes (often), the tooltip disappears almost immediately after appearing.

## Expected behavior

The tooltip remains active, as the active prop sent to the Tooltip component is truthy.

## Run `npx envinfo`

Paste the results here:

```bash
  System:
    OS: Windows 7 6.1.7601
    CPU: (4) x64 Intel(R) Xeon(R) CPU E5-2650 v4 @ 2.20GHz
    Memory: 17.73 GB / 32.00 GB
  Binaries:
    Node: 14.16.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.17 - C:\Program Files\nodejs\yarn.CMD
    npm: 6.14.12 - C:\Program Files\nodejs\npm.CMD
  Managers:
    Gradle: 6.6 - C:\Gradle\gradle-6.6\bin\gradle.BAT
    Maven: 3.6.3 - C:\Programs_Portable\apache-maven-3.6.3\bin\mvn.CMD
  Utilities:
    Git: 2.36.1. - /mingw64/bin/git
  IDEs:
    VSCode: 1.63.0 - C:\Users\ADXXXXX\AppData\Local\Programs\Microsoft VS Code\bin\code.CMD
  Languages:
    Bash: 4.4.23 - C:\Program Files\Git\usr\bin\bash.EXE
    Java: javac 18 - /c/Program Files/Common Files/Oracle/Java/javapath/javac
    Perl: 5.34.0 - C:\Program Files\Git\usr\bin\perl.EXE
  Browsers:
    Internet Explorer: 11.0.9600.20296
```

## Eufemia Version

9.26.0

[...]

PR #1588, opened 2022-09-29
fix(Tooltip): fix React Portal handling

tujoworker:
Enhance usage of portal, so we not use `createPortal` and `render`.
This fixes an issue where `mouseOver` and `mouseOut` was not given the wanted/same result:

https://user-images.githubusercontent.com/1501870/193398634-55d2b1ca-1bbd-45a7-92f4-5affd5c40a67.mov

I can imagine that this issue  #1411 can be related.


langz (approved):
Tested the change in different browsers(Safari, Edge, Chrome) on my MacBook, and a bit on my iPhone, seems to work great :)

Let's pray that #1411 is related then 🙏

PR #1591, opened 2022-09-30
fix(Tooltip): merge style property with internal

tujoworker:
+ Add some more tests
I will rebase this branch when PR #1590 is merged.


langz (approved):
Aha, nice find.
I can guess that overwriting the internal style can result in a few interesting results 😹

PR #1595, opened 2022-10-03
fix(Accordion): support nested accordions

tujoworker:
When nesting accordions, the header CSS styles got inherited, which led to a wrong styling state when opening/closing.

Also add `width: 100%;` to the content and inner content, so nested accordions to stretch, like the main one does do.

**Not supported nested accordions ❌**

Reproduction [CSB](https://codesandbox.io/s/eufemia-nested-accordion-086cc9?file=/src/App.tsx).

(👇 only the first should be expanded)

Screenshot 2022-10-03 at 12 30 09

**Supported nested accordions ✅**

Screenshot 2022-10-03 at 13 56 43


langz (approved):
Looks okay to me, great that the children/nested accordions take 100% width now :)

I don't fully understand the sentence in the description _"Only the first should be expanded."_.
Should it only be possible to expand one "child/nested accordion" at the time?
When running the new storybook example [locally](http://localhost:8002/?path=/story/eufemia-components-accordion--nested-accordion) now, I'm able to expand both.

Further, I am not able to reproduce the image/screenshot in description under "Supported nested accordions ✅" when both arrows is pointing upwards, but their background color is white: https://user-images.githubusercontent.com/1501870/193556907-0380fac8-090c-45ff-8208-f2fca4d00c9f.png
Not sure if this is too important or not though.

tujoworker:
@langz Thank You testing and finding the mentioned issues 🙏 Yes, very true! 🤦‍♂️ The icons should not point up. I'll find the cause and fix it.

PR #1603, opened 2022-10-04
chore(deps): update Gatsby to latest because of useful fix

tujoworker:
Because of this [fix](https://github.com/gatsbyjs/gatsby/pull/36623) which I think if have seen happening in our Portal as well.

Run lint with NODE_ENV=test in `lint:ci` to prevent babel parsing error (because of babel config change it gets picket up now)


langz (changes requested):
I gave it a `yarn reset`, and a `yarn start`.
This is what happens locally here when I visit http://localhost:8000/:
image

Could of course just be something with my computer setup 🤔

tujoworker:
> This is what happens locally here when I visit

Renaming the babel config file the how its state in the [documentation](https://www.gatsbyjs.com/docs/how-to/custom-configuration/babel/#how-to-use-a-custom-babelrc-file) (`babel.config.cjs` => `.babelrc`) seems to help.

langz (commented):
This time around, I did the same:`yarn reset`, and a `yarn start`.
This is what happens locally here when I visit http://localhost:8000/:

https://user-images.githubusercontent.com/1359205/194820626-01d05805-dc4d-4f1d-ae17-4e5f917b59b1.mov

tujoworker:
aww, this time also the CI suffers with the same result as you got. Anyway, it looks like it works fine now. At least for me when I run the same reset command as you did.

tujoworker:
no does not work properly yet, I will have to come with an update.

tujoworker:
Now it should work properly. There where some small changes in the internal Webpack config from Gatsby. So made an additional commit.

langz (approved):
Thanks, looks and works great locally now :)

tujoworker:
We may rather rebase this PR once #1623 has landed.

PR #1605, opened 2022-10-04
fix(Breadcrumb): make animation work properly when declared as JSX

tujoworker:
When a Breadcrumb was declared like so:

```jsx
    <Breadcrumb>
      <Breadcrumb.Item
        text="Page item 1"
      />
      <Breadcrumb.Item
        text="Page item 2"
      />
    </Breadcrumb>
```

https://user-images.githubusercontent.com/1501870/193920026-7900d0ce-95fb-46fa-bf91-cde6e67434c1.mov

The animation did not work nicely then. Why? We do not know the count of each item.

[CSB](https://codesandbox.io/s/eufemia-breadcrumb-animation-issue-thsgsm?file=/src/App.tsx)

This PR ensures we always know the number of each item during the each re-render phase.


langz (approved):
Worked great on my devices and browsers

PR #1606, opened 2022-10-05
fix(Breadcrumb): ensure correct spacing when collapsed

tujoworker:
This PR ensures correct spacing. It should always be used when defined.


langz (approved):
👾

PR #1607, opened 2022-10-05
feat(HeightAnimation): add --hidden class when closed and keepInDOM is true

tujoworker:


packages/dnb-eufemia/src/components/height-animation/HeightAnimation.tsx
```diff
@@ -80,6 +80,10 @@ export default function HeightAnimation({
         isVisible && 'dnb-height-animation--is-visible',
         isVisibleParallax && 'dnb-height-animation--parallax',
         isAnimating && 'dnb-height-animation--animating',
+        !isVisible &&
+          !isAnimating &&
+          !open &&
+          'dnb-height-animation--hidden',
```
langz, line 86: Why do we need to add this CSS class?
It's not like we've added any styling for this class.
Or should the changes done as part of this PR in `packages/dnb-eufemia/src/components/height-animation/stories/HeightAnimation.stories.tsx` rather have been done in `packages/dnb-eufemia/src/components/height-animation/style/_height-animation.scss` ?

langz (approved):

langz (approved):
Other than my one comment, being unsure about why we add a CSS class that we don't have styling for, it looks great 💯  
I tested the existing HeightAnimation examples and the ones for Breadcrumb which uses HeightAnimation, and those examples still works as expected 🙏

packages/dnb-eufemia/src/components/height-animation/HeightAnimation.tsx
tujoworker: yeah, I think we then should add a comment. Will do so.
The reason is, when using this in the Accordion, we probably want to continue to use `display: none` – when `prerender` (keepInDOM) is true. Right now this component uses only `visually: hidden`. When we have this `--hidden` class at the correct time, other use cases can make such changes. 
Another option could be to support this directly in this component as a property. But I feel, that having this class, we make it more flexible, as of now.

langz: Aha, then I understand, thanks 👍

PR #1610, opened 2022-10-05
fix(Lists): rewrite Dl, Ol and Ul (and Li) to TypeScript

tujoworker:
Also, added release notes about it + some more release notes.


packages/dnb-eufemia/src/elements/Dl.tsx
```diff
@@ -34,12 +30,15 @@ const Dl = ({
...
 export type DlLineProps = {
-  children: React.ReactNode
+  //
 }
 
-Dl.Line = ({ className, children }: DlLineProps & HTMLSpanElement) => {
+Dl.Line = ({
+  className,
+  children,
+}: DlLineProps & React.AllHTMLAttributes<HTMLSpanElement>) => {
```
langz, line 39:
```suggestion
}: DlItemProps & React.AllHTMLAttributes<HTMLSpanElement>) => {
```

langz (approved):

langz:
Just curious, but why do we do:
```
export type LiProps = {
  //
}
```
and 
```
export type DlLineProps = {
  //
}
```

Is it because we know that we'll eventually create a few props in these types in the future? 👨‍🚀

packages/dnb-eufemia/src/elements/stories/Lists.stories.tsx
```diff
@@ -75,7 +75,7 @@ export const ListSandbox = () => (
             <Li>Item</Li>
             <Li>
               Item
-              <Ol type="I" start="3">
+              <Ol type="I" start={3}>
```
langz, line 78: In the light of this change, do/should we also change this to use a number and not a string?
https://github.com/dnbexperience/eufemia/blob/main/packages/dnb-design-system-portal/src/docs/uilib/elements/lists.md?plain=1#L134

tujoworker:
> 

> So therefore it makes sense creating the types now, even though they are empty

I have removed them. It was more that I copy the `PropTypes` and convert them over to TS manually. And because we inherit the types from the HTML Elements, we suddenly do not need to define children etc. in addition.

PR #1614, opened 2022-10-07
docs(Components): add description to all component docs

langz:
This PR makes a few improvements in the following page https://eufemia.dnb.no/uilib/components/:
- Adds missing descriptions to components.
- Edits a few descriptions for components that had poor description.
- Sorts the components/mdx's by their title. 
I think it's better than what it is today, but would be even better if the order would be the same(and shared), using levels like for fragments, as in the sidebar menu. Not sure how easy/quick that would be to do, and it's also not very important 😹 

For comparison: [existing](https://eufemia.dnb.no/uilib/components/) vs. [changes introduced in this PR](https://eufemia-docsadddescriptiontoallcompone.gtsb.io/uilib/components/)


tujoworker (approved):
This looks awesome ✨🎉

PR #1623, opened 2022-10-11
fix(gatsby-plugin-eufemia-theme-handler): support Gatsby > v4.24

tujoworker:


langz (approved):
Works on my machine 🤖

PR #1625, opened 2022-10-11
fix(Dl): remove span as direct sibling in horizontal layout direction definition lists

tujoworker:
axe (automatic a11y tests) did fail when span was used as direct sibling of a dl list. 

Instead, we use an empty dd and set the width to 100% to ensure the next dt will wrap to a new line.

cc @AadneRo


langz (approved):
Looks great to me 👍  
I'm however a bit unsure of the usage of `Dl` vs. `dl`, but I assume it's intended that we use `dl` and not `Dl` [here](https://github.com/dnbexperience/eufemia/pull/1625/files#diff-d7df987749598fee369106a766e6ebf8ba81512894c4adef3cd1871beb1ddd8c).

packages/dnb-eufemia/src/elements/Dl.tsx
```diff
@@ -37,9 +37,10 @@ Dl.Item = ({
   children,
 }: DlItemProps & React.AllHTMLAttributes<HTMLSpanElement>) => {
   return (
-    <span className={classnames(className, 'dnb-dl__item')}>
+    <>
       {children}
-    </span>
+      <dd aria-hidden className={classnames(className, 'dnb-dl__item')} />
```
langz, line 42: Should this actually be `dd` and not `dl`?

tujoworker: It needs to be `dd` because only a `dd` is allowed inside a `dl` – that's what axe tells us at least. When I tested it with a span, VoiceOver did work fine. So this PR is more about to avoid noise in the automated a11y tests.

tujoworker:
> I assume it's intended that we use `dl` and not `Dl` [here](https://github.com/dnbexperience/eufemia/pull/1625/files#diff-d7df987749598fee369106a766e6ebf8ba81512894c4adef3cd1871beb1ddd8c).

I looks strange/chaotic, yes. We should use `Dd` consequent.

PR #1626, opened 2022-10-12
fix(Dl): ensure correct spacing

tujoworker:
Using CSS gap did also include our "hidden" wrapper element. So we need to use margin instead.

cc @AadneRo


packages/dnb-eufemia/src/elements/__tests__/Lists.test.tsx
```diff
@@ -11,6 +11,36 @@ import Dt from '../Dt'
 import Dd from '../Dd'
 
 describe('Dl', () => {
+  it('should support spacing props', () => {
```
langz, line 14: Could/should we add this test to Element.test.tsx as well, or is it just redundant?

langz (approved):

packages/dnb-eufemia/src/elements/__tests__/Lists.test.tsx
tujoworker: I'll add it 😉

PR #1632, opened 2022-10-13
fix(Drawer): ensure ScrollView is used for scrollable content

tujoworker:


langz (approved):

tujoworker:
Found a way of verifying the removal of `overflow-x: hidden;` so added a test for that.

PR #1636, opened 2022-10-13
fix(Section): make style_type property accept any string

tujoworker:
Also, clean up types structure to avoid later problems of inheriting HTMLElement props.


packages/dnb-eufemia/src/components/section/Section.tsx
```diff
@@ -56,14 +56,15 @@ export type SectionProps = {
...
-  React.HTMLProps<HTMLElement>
+}
 
 const defaultProps = {
   element: 'section',
 }
 
-export default function Section(localProps: SectionProps) {
+export default function Section(
+  localProps: SectionProps & SpacingProps & React.HTMLProps<HTMLElement>
```
dagerikhl, line 66: Is there a strategy for exposing prop (types) of Eufemia components? I've mostly used to import the props for the component, e.g. `import { SectionProps } from "@dnb/eufemia/components/section";` (path may be slightly incorrect), but lately I've also seen cases of `ComponentAllProps` which includes all possible props in the union used for `localProps`. Here, there's not exposed type for "all viable props for `Section`".

I find it useful if I'm allowing one of my components to take inn all props used by its top-level Eufemia component, e.g.:
```ts
interface OwnProps {
  isLoading?: boolean;
}

export type MyComponentProps = OwnProps & SectionProps;

export const MyComponent = ({ isLoading, ...rest }: MyComponentProps) => {
  // ...
  return <Section {...rest}>...</Section>;
};
```

It's of course possible to add the `& SpaceProps & React.HTMLProps<HTMLElement>`, but I found the `ComponentAllProps` to be a very nice pattern :D Like in Tooltip: https://github.com/dnbexperience/eufemia/blob/b9f7a6d4f747771f0d40d76ca9cb5c2bb1ee23bc/packages/dnb-eufemia/src/components/tooltip/types.ts

Just wondering if there's a strategy or we should expect to union the collection of props ourselves sometimes? :)

dagerikhl (approved):
Very nice! Looks like a good solution to me, and solves the TS vs. feature problem :D 

Just had a question about the exposure of prop types, but it's not a blocker for this issue.

packages/dnb-eufemia/src/components/section/Section.tsx
tujoworker: Yes! I agree, we should keep this pattern and document it on the [contribution pages](https://eufemia.dnb.no/contribute/).

I have pushed this change to this PR. Let me know if you have other suggestions than using "All*" 😉

dagerikhl: Glorious, love it! 😍 Can't think of a better naming convention... Complete? But All is better I think :)

tujoworker: PR #1639

PR #1637, opened 2022-10-13
fix(Icon): fix override of data-testid property when provided

langz:
The changes I made in https://github.com/dnbexperience/eufemia/pull/1583 make it pointless to pass a value for `data-testid`to the Icon component, as we always overwrite the `data-testid`.

For example, the following code will end with a `data-testid` that will have the value `question icon`:
`<Icon icon={question} data-testid="custom-data-testid" />,` but in this scenario I would want it to be `custom-data-testid`.

In general, I think I made a a slight mistake with #1583 and that we shouldn't override the value of `data-testid` when provided.


tujoworker (approved):
Nicely explained PR and good tests 💛

PR #1638, opened 2022-10-14
chore(Portal): fix src vs build usage during development

tujoworker:
It could happen that the some `build` styles where used alongside with styles from `src` when run in dev, if it existed. This PR ensures we do not get this anymore.


langz (approved):
💯

PR #1639, opened 2022-10-14
docs(Contribution): small types enhancements

tujoworker:
Following up this comment: https://github.com/dnbexperience/eufemia/pull/1636#discussion_r995370817


langz (approved):
Looks like a nice pattern to me 🥇 
Not 100% decided if we should name it `AllComponentProps` or `ComponentAllProps` 🤔

dagerikhl (approved):
Wonderful! :D Though I think I might agree with @langz on the naming order, I think it could be a bit more intuitive with the component name first, as that often is what we start writing when needing to import something from this component.

PR #1641, opened 2022-10-17
fix(Drawer): make spacing of header and nav robust

tujoworker:
Ensure header and navigation is rendered outside of inner content to avoid dealing with negative margins, which again causes often a horizontal scrollbar.

Related changes are in this [commit](https://github.com/dnbexperience/eufemia/pull/1641/commits/35dd8bce7be53008907f759eb8b6d0ca5d171885).


langz (approved):
Looks good, I wasn't able to see any visual differences 🤓

PR #1644, opened 2022-10-18
fix(Space): rewrite Space component to TypeScript

tujoworker:


langz (approved):
Great 💯 
Seems to be a few minor TS errors popping up, that we have to fix 🔨 
But that's just nice, shows that TS works, forcing us to use the correct types 🙏

PR #1646, opened 2022-10-18
feat(HeightAnimation): make the view open by default

tujoworker:
I rethought the purpose of this component and think now that this component will probably mostly be used as a "height change" component, and not always from nothing to auto. 

When someone needs the first render to be nothing/0 then he anyway will send in `false` on the first render cycle. And then change it when needed.

When we have open by default, we do not need to define it anymore to use this component for only a height animation.

### DIFF
--- a/packages/dnb-eufemia/src/components/height-animation/__tests__/useHeightAnimation.test.tsx
+++ b/packages/dnb-eufemia/src/components/height-animation/__tests__/useHeightAnimation.test.tsx
@@ -6,6 +6,7 @@
 import React from 'react'
 import classnames from 'classnames'
 import { render, act, fireEvent } from '@testing-library/react'
+import { renderHook } from '@testing-library/react-hooks'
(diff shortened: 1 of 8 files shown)


packages/dnb-eufemia/src/components/height-animation/__tests__/useHeightAnimation.test.tsx
```diff
@@ -175,6 +176,40 @@ describe('useHeightAnimation', () => {
...
+      isAnimating: false,
+      isInDOM: true,
+      isOpen: true,
+      isVisible: true,
+      isVisibleParallax: true,
+      open: true,
+    })
+  })
+
+  it('should be clsoe by if open is false', () => {
```
langz, line 196:
```suggestion
  it('should be closed if open is false', () => {
```

langz (approved):

PR #1647, opened 2022-10-18
chore(Upload): add drop zone

tujoworker:
Putting this in as a chore, because its not released yet.

This PR adds support for file drag&drop to the browser window.

### DIFF
--- a/packages/dnb-eufemia/src/components/upload/Upload.tsx
+++ b/packages/dnb-eufemia/src/components/upload/Upload.tsx
@@ -53,6 +53,7 @@ const Upload = (localProps: UploadProps & SpacingProps & LocaleProps) => {
     buttonText,
     loadingText,
     errorLargeFile,
+    errorUnsupportedFile,
     deleteButton,
     fileListAriaLabel,
     ...props
--- a/packages/dnb-eufemia/src/components/upload/Upload.tsx
+++ b/packages/dnb-eufemia/src/components/upload/Upload.tsx
@@ -7,7 +7,6 @@ import P from '../../elements/P'
 import Dl from '../../elements/Dl'
 import Dt from '../../elements/Dt'
 import Dd from '../../elements/Dd'
-import HeightAnimation from '../height-animation/HeightAnimation'
 
 // Shared
 import { createSpacingClasses } from '../space/SpacingHelper'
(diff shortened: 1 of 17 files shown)


AadneRo (commented):
The animation looks awesome! 🎉😍

packages/dnb-eufemia/src/components/upload/Upload.tsx
```diff
@@ -7,27 +7,29 @@ import P from '../../elements/P'
...
+import { verifyFiles } from './UploadVerify'
+
+import type { UploadFile, UploadAllProps } from './types'
 
 export const defaultProps = {
   fileMaxSize: 5000,
 }
 
-const Upload = (localProps: UploadProps & SpacingProps & LocaleProps) => {
+const Upload = (localProps: UploadAllProps) => {
```
AadneRo, line 32: Why don't we add the `SpacingProps` and `LocaleProps` inside the `UploadProps`?

tujoworker: Yes, this is also fin, but not so flexible. We now also return the needed `React.HTMLProps<HTMLElement>` so we also support vanilla HTML attributes. Also, with that, it would get pretty much in there.

tujoworker: We have concluded in this [PR](https://github.com/dnbexperience/eufemia/pull/1639) – you can view the docs from this PR [here](https://eufemia-docsenhancecontributiondocs.gtsb.io/contribute/getting-started/#typescript-types).

packages/dnb-eufemia/src/shared/locales/nb-NO.js
```diff
@@ -132,6 +132,8 @@ export default {
       loadingText: 'Laster opp',
       errorLargeFile:
         'Filen du prøver å laste opp er for stor, vi støtter ikke filer større enn %size MB.',
+      errorUnsupportedFile:
+        'Filen du prøver å laste opp blir ikke støttet.',
```
langz, line 136:
```suggestion
        'Filen du prøver å laste opp er ikke støttet.',
```

AadneRo (approved):

langz (approved):

langz:
On a side note, should `multipleFiles` default to `true`?
Personally as a user, I like the possibility of selecting multiple files in "one go", instead of having to open the "file selector" multiple times.

tujoworker:
I agree! Is it ok to say, if you want the user to upload only one file, then use `multible={false}`?

langz:
I would say so, or maybe we'll have to rename the prop, `singleFile`. We don't have to do anything regarding this now, I'll give it some thought 💭

packages/dnb-eufemia/src/components/upload/__tests__/UploadDropzone.test.tsx
```diff
@@ -0,0 +1,106 @@
+import React from 'react'
+import {
+  act,
+  // act,
+  fireEvent,
+  render,
+  // screen,
+  // waitFor,
```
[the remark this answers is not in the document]
tujoworker: Yes! Sure 🙏

PR #1648, opened 2022-10-19
chore(Upload): default to multiple files by changing prop from multileFiles to filesAmountLimit

tujoworker:
This PR is regarding [this input](https://github.com/dnbexperience/eufemia/pull/1647#issuecomment-1283628816).


packages/dnb-design-system-portal/src/docs/uilib/components/upload/demos.md
```diff
@@ -27,7 +27,7 @@ By using the `Upload.useUpload` you can remove or add files or the status displa
 
 ### Upload multiple files
```
langz, line 28:
```suggestion
### Upload single file - Limits to selecting only one file at a time in the file explorer
```

This is just a suggestion, so that "single file" is not misinterpreted as that you can only upload/add one file.

packages/dnb-design-system-portal/src/docs/uilib/components/upload/properties.md
```diff
@@ -7,8 +7,8 @@ showTabs: true
 | Properties                                  | Description                                                                          |
 | ------------------------------------------- | ------------------------------------------------------------------------------------ |
 | `acceptedFileTypes`                         | _(required)_ List of accepted file types.                                            |
-| `fileMaxSize`                               | _(optional)_ fileMaxSize is max size of each file in MB.                             |
-| `multipleFiles`                             | _(optional)_ if set true, accepting multiple files is allowed.                       |
+| `singleFile`                                | _(optional)_ If set `true`, only one file will be accepted.                          |
```
langz, line 10:
```suggestion
| `singleFile`                                | _(optional)_ If set `true`, only one file will be accepted at a time.                          |
```

langz (approved):
Thanks for improving this 🙏 

Only thing I'm a bit unsure about is that `singleFile` could be misinterpreted as that the component will only support one file. So like if you select a new file, you'll override the existing/previous selected file 🤔 
But in our scenario, we can always add multiple files.

AadneRo (approved):

packages/dnb-design-system-portal/src/docs/uilib/components/upload/properties.md
tujoworker: You are right! also `multipeFiles` was in that case misleading. 

So I propose; when `singleFile` is true, then we should "replace" the existing file. Because we want the user to "force" to choose one – to avoid confusion.

tujoworker:
You are right! also `multipeFiles` was in that case misleading.

So I propose; when `singleFile` is true, then we should "replace" the existing file. Because we want the user to "force" to choose one – to avoid confusion.

tujoworker:
Maybe we should replace singleFile with filesLimit og limitFiles as well.

langz (approved):
Thanks for the great work here with setting a limit to number of files allowed in the file uploader @tujoworker 👏 

I really like it 💯 

I somehow am also left with some thoughts:

- If we limit the user to only upload a certain amount of files, let's say filesAmountLimit={2}, wouldn't it also be nice to inform the user with some kind of info text about this limitation, so that he will know of this limitation before he sees the error message when exceeding the limitation unknowingly.
- There's also something that I can't really put my finger on in regards to how we handle the limitation, and the error message displayed. In the given scenario I have filesAmountLimit={2} and when adding/selecting my 3rd file, I see the following message:
image
I'm thinking, I know that I just tried adding a 3rd file, but it didn't seem to have been added as I can only see 2 files. Also, now that I only have 2 files(which is an allowed number of files) why would I "need" to remove one of my two files to proceed/remove the error message.

- I'm also trying to think about how this limitation and the error message would "play along" when the file uploader is part of a form, let's say that my form requires 2 files. And I would like to enable/disable the submit button of my form based on this. Would I enable my submit button if `files` is 2 then? But `files` is also 2 when displaying the error message "Du kan laste opp 2 filer. Fjern noen og prøv på nytt.", but I guess my submit button should be disabled when we display an error 🤔 

I feel like I'm overcomplicating this stuff now, so please go ahead and ignore my comments, and deliver what we think is the necessary MVP that we later can build upon.

tujoworker:
You are so right! I will check with Mats. 

Also, changing the warning to something like:

> Det er begrenset hvor mange filer du kan laste opp (%amount).

will maybe help. 

But we should also inform the user beforehand.

PR #1373, opened 2022-03-28
break: re-define Media Query Breakpoints and layout width

tujoworker:
Why?

We need to change/align our breakpoints in code. This will help everyone work with confidences that components and page templates cross team will work together.

In code the new breakpoints will be:

- small: 640px / `40em`
- medium: 960px / `60em`
- large: 1152px / `72em`

For now, we skip `x-large` until someone needs it.

BREAKING CHANGE

PR #1650, opened 2022-10-20
feat(useMedia): add React Hook to handle Eufemia layout breakpoints

tujoworker:
This PR adds a new React Hook that is based on the already existing internals of the existing [`useMediaQuery`](https://eufemia.dnb.no/uilib/usage/layout/media-queries/#usemediaquery-hook-usage) React Hook.

The new `useMedia` hooks is a fixed and simplified version of it. 

Before you would have to do this:

```jsx
  const isSmall = useMediaQuery({ when: { max: 'small' } })
```

While with this hook, you can do this:

```jsx
  const { isSmall, isMedium, isLarge } = useMedia()
```

In v10 of Eufemia the breakpoints will be changed and aligned in this PR #1373 – if you want to use this hook once released, and with v9 – then you may want to define the internally used breakpoints by yourselves.

**NB:** It should be done only temporary, because DNB should align on set of breakpoints to best UX and consistency.

```jsx
<Provider
  value={{
    breakpoints: {
      small: '40em',
      medium: '60em',
      large: '72em',
    },
  }}
 >
 <App /> {/* if app uses useMedia */}
</Provider>
```

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/usage/layout/Examples.tsx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/usage/layout/Examples.tsx
@@ -1,0 +1,92 @@
+/**
+ * UI lib Component Example
+ *
+ */
+
+import React from 'react'
+import ComponentBox from 'dnb-design-system-portal/src/shared/tags/ComponentBox'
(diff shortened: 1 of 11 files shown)


langz (approved):
Nice, seems to be simplifying a lot 🙏

PR #1652, opened 2022-10-21
chore(Jest): upgrade jest form v28 to v29 for faster executions

tujoworker:
- Here is the [PR](https://github.com/facebook/jest/pull/13421)
- Update snapshots regarding the [snapshotFormat change](https://jestjs.io/blog/2022/08/25/jest-29)


langz (approved):
Good good, 
I guess if the tests pass, then it's working as expected 😹 🙏

PR #1654, opened 2022-10-21
feat(icons): hierarchy, layout_card and unlock

tujoworker:
It works 👍

Screenshot 2022-10-21 at 13 59 55

Also, the colorized snapshot is now blue, so its we actually can see eventual diffs 😃


langz (approved):
🥇

PR #1655, opened 2022-10-21
chore(Housekeeping): rename prop form "is" to "as"

tujoworker:


langz (approved):
💯 
I see a few references to `is` in `custom-element-polyfill.js`, are those references related to this change? 🤔

PR #1657, opened 2022-10-21
feat(Upload): highlight existing files visually

tujoworker:
New UX changes to highlight files which where already added. Also, do not/never disable the browse files button.

### DIFF
--- a/packages/dnb-eufemia/src/components/upload/Upload.tsx
+++ b/packages/dnb-eufemia/src/components/upload/Upload.tsx
@@ -61,7 +61,8 @@ const Upload = (localProps: UploadAllProps) => {
 
   const spacingClasses = createSpacingClasses(props)
 
-  const { files, setFiles, setInternalFiles } = useUpload(id)
+  const { files, setFiles, setInternalFiles, existsInFiles } =
+    useUpload(id)
 
   return (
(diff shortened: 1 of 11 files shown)


langz:
I don't think the highlighting works exactly as intended, or that I'm misunderstanding how it should work.

When only adding one file, and trying to add the same file again, I don't see any highlighting:

https://user-images.githubusercontent.com/1359205/197730209-e3ae152f-10dd-47af-9c86-5803cd61c0d9.mov

When adding two files, I see highlighting when adding one of the the same files again, but adding it again for the 3rd time I don't see any highlighting(should we only see it the 1st time the file was added again?):

https://user-images.githubusercontent.com/1359205/197730867-30c5eb65-0f5c-4c36-bda2-eec204f5cce8.mov

tujoworker:
@langz Can you refresh the page an try again?

langz:
> @langz Can you refresh the page an try again?

Sure sure, I'm experience the same though. Running the branch locally. Chrome Version 106.0.5249.119 (Official Build) (x86_64)

langz (approved):
I'm still not able to test/view this properly on my laptop, but tested it on my phone, and it works great there 👏

PR #1658, opened 2022-10-22
chore(Upload): set correct max file-size

tujoworker:
Ohh, we ended up allowing 5.000 MB per file. That's very large files 🤭


langz (approved):
🤣

PR #1661, opened 2022-10-24
fix(Breadcrumb): support null as children

langz:
After the following [change](https://github.com/dnbexperience/eufemia/pull/1605/files#diff-105824c101ceb4b6537e4898deff813846530866c5e7a1040807ba39a7f07d75R47), where we now use React.cloneElement, the following code snippet would result in an error:
```
<Breadcrumb>
   <Breadcrumb.Item />
   <Breadcrumb.Item />
   {null}
</Breadcrumb>
```

The error:
`React.cloneElement(...): The argument must be a React element, but you passed null.`

The suggested fix is just to remove/filter out not valid React elements. before trying to clone them.

This should fix it when passing children, but should we add support for this when passing `null` in `data` as well? 
Or is it just normal(not expected by the consumer) to not be able to pass null in this situation 🤔 
```
<Breadcrumb data={[ { text: "A" }, { text: "B" }, null ]} />
```

### DIFF
--- a/packages/dnb-eufemia/src/components/breadcrumb/BreadcrumbMultiple.tsx
+++ b/packages/dnb-eufemia/src/components/breadcrumb/BreadcrumbMultiple.tsx
@@ -43,9 +43,11 @@ export const BreadcrumbMultiple = ({
           )
         })}
 
-        {items?.map((item, i) =>
-          React.cloneElement(item, { key: i, itemNr: i })
-        )}
+        {items
+          ?.filter((item) => React.isValidElement(item))
+          .map((item, i) =>
+            React.cloneElement(item, { key: i, itemNr: i })
+          )}
       </Section>
     </HeightAnimation>
   )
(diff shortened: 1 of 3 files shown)


packages/dnb-eufemia/src/components/breadcrumb/BreadcrumbMultiple.tsx
```diff
@@ -43,9 +43,11 @@ export const BreadcrumbMultiple = ({
           )
         })}
 
-        {items?.map((item, i) =>
-          React.cloneElement(item, { key: i, itemNr: i })
-        )}
+        {items
+          ?.filter((item) => React.isValidElement(item))
+          ?.map((item, i) =>
```
tujoworker, line 48: The second `?` is not needed. When `items` is null/undefiend, we will not hit the map. So up to you if you want to change it or not.

tujoworker (approved):
🫶

packages/dnb-eufemia/src/components/breadcrumb/BreadcrumbMultiple.tsx
langz: Thanks, removed 🙏

PR #1667, opened 2022-10-25
feat(Tabs): make tabs content animate its height

tujoworker:
Here is a [CSB version](https://codesandbox.io/s/eufemia-tabs-animation-65147j?file=/src/App.tsx) that does NOT animate.

And here [with the change](https://mhu0bh.csb.app/).


langz (approved):

langz:
I like the animation a lot :)

However when comparing the two CSBs, [CSB version that does NOT animate](https://codesandbox.io/s/eufemia-tabs-animation-65147j?file=/src/App.tsx) vs. [with the change](https://mhu0bh.csb.app/). 
I do feel like the transitioning is not so "subtle", and sort of "blinks", I guess it's because the transitioning of the color from white to "green-ish". But it could be a good thing that I don't feel it's very subtle, as I will as a user at least understand that there's something that's changing.

https://user-images.githubusercontent.com/1359205/197778075-5618cc20-50bc-457c-8bad-9271ce3f77d0.mov

tujoworker:
Yep – I see – that happens when a section is used inside AND `prerender` is used. When no section is used (that has its own background) – then I think its fine with the opacity effect. Here's an [example](https://eufemia-feattableheightanimation.gtsb.io/uilib/components/tabs/demos/#tabs-and-prerender).

langz:
> Yep – I see – that happens when a section is used inside AND `prerender` is used. When no section is used (that has its own background) – then I think its fine with the opacity effect. Here's an [example](https://eufemia-feattableheightanimation.gtsb.io/uilib/components/tabs/demos/#tabs-and-prerender).

Looks great 💯

PR #1668, opened 2022-10-25
fix(Breadcrumb): support a single child

langz:
Types and code did not support passing a single child as children to Breadcrumb component.

When only passing a single breadcrumb.item as children the type would report a problem:
image

Also, when only passing a single breadcrumb.item as children, the filter I added [here](https://github.com/dnbexperience/eufemia/pull/1661/files#diff-105824c101ceb4b6537e4898deff813846530866c5e7a1040807ba39a7f07d75R47) will make the code break, as an object can't do "filter".

This PR should fix these two issues.


tujoworker (approved):
Yes! Awesome – I'm 🤩

PR #1669, opened 2022-10-25
fix(Table Tr): support a single child

langz:
In light of https://github.com/dnbexperience/eufemia/pull/1668, I was looking for other scenarios where our types should also support a single react element, and not only an array/list of react elements.


tujoworker (approved):
👏

PR #1679, opened 2022-10-28
chore(release): make "next" releases possible

tujoworker:
Make it possible to release versions to the `next` NPM tag. Its very the same as beta and alpha. So we may have a discussion on what should run of tests and if this version "info" should be pushed back to GitHub, like here: https://github.com/dnbexperience/eufemia/releases/tag/v9.35.0


langz (approved):
🥇

PR #1680, opened 2022-10-28
fix(StepIndicator): add example on how to use it with a router (browser location)

tujoworker:
I think it could be good to have this listed in the change logs, because we have found examples in prod where this is not used. So when the user presses the back button, they will be relocated to another page, instead to the previews step.


langz (approved):
Looks great 👍 

Is there anyway we can word ourselves even better, to prevent people thinking they must have to use `@reach/router`. People tend to take the examples quite literally 🔤

tujoworker:
> Is there anyway we can word ourselves even better, to prevent people thinking they must have to use `@reach/router`. People tend to take the examples quite literally 🔤

Yes, I agree.  Thought the same during the communication afterwards I had with the dev. 👍

PR #1684, opened 2022-11-01
docs(Upload): polish the info and examples

tujoworker:
I think, before announcement, the docs needs some more info about how we want it to be used.

Any thoughts and suggestion that can improve it further?

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/components/upload/info.md
+++ b/packages/dnb-design-system-portal/src/docs/uilib/components/upload/info.md
@@ -4,6 +4,39 @@ showTabs: true
 
 ## Description
 
-The Upload should be used in scenarios where the user has to upload any kind of files.
+The Upload component should be used in scenarios where the user has to upload any kind of files.
+
+## How to use the Upload component
+
+- Files selected by the user should be uploaded immediately (temporary location).
+- The user should be able to remove them (files) during the session.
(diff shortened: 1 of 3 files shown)


packages/dnb-design-system-portal/src/docs/uilib/components/upload/info.md
```diff
@@ -4,6 +4,39 @@ showTabs: true
...
 ## Description
 
-The Upload should be used in scenarios where the user has to upload any kind of files.
+The Upload component should be used in scenarios where the user has to upload any kind of files.
+
+## How to use the Upload component
+
+- Files selected by the user should be uploaded immediately (temporary location).
+- The user should be able to remove them (files) during the session.
+- If the Upload component is shown in a submit form, then a [GlobalStatus](/uilib/components/global-status) should be a part of the form.
```
langz, line 13: Maybe we could/should provide an example? 🤔

langz (approved):
Great 👍 
I can try to go over all the documentation for the upload component, and see if I find any possible improvements tomorrow 🙏

packages/dnb-design-system-portal/src/docs/uilib/components/upload/info.md
tujoworker: Yes! That would be nice 👍 we have to check that out, and find a system where we can make a demo.

PR #1693, opened 2022-11-02
chore(Portal): convert styling from Emotion to SCSS

tujoworker:
This is the first step in offloading the CPU usage during development/build. We will with that, move work from Node.js (Emotion) to Dart (SASS).

It effects only the usage of Emotion in the Portal styles and not the Emotion usage in Examples/Demos.

## My thoughts about this change

Is this measurable? Me, as a human, can't see any difference. Maybe it saves some milliseconds here and there.

Has it negative impact on DX? yes and no. I really like the API of styled-components, the directness and declarative syntax it brings to the code base. 

I miss-like the additional API of `:global` – and that the same name is used to include class names  and a real "global scope" when used without a selector.

But on the other side, we have all the additional tooling it needs to make auto-complete and linting possible. And also that we could not use existing Eufemia mixins, like `allBelow` etc. 

## Future thoughts

- Later we probably want to move the portal illustrations to SVG sprites instead of having them as JSX based React components. 
- And also the Eufemia icons will need a better solution (there is a task for that already).


langz:
Hopefully I have time to start testing this and looking into it by tomorrow 🕚

tujoworker:
Sure, np 🕺 – and by the way, the styles are copy&paste for the very most part. The grid generation got some larger rewrite, because we did generate them with JS, but now with a mixin.

langz:
I did some quick visual comparisons.
Leftside: https://eufemia-chorescssportalstyles.gtsb.io/
Rightside: https://eufemia.dnb.no/
Screenshot 2022-11-06 at 10 59 26

tujoworker:
ohh, nice findings 🙏 here is [the diff](https://github.com/dnbexperience/eufemia/compare/d05935af5a57624da403245ab47462686665d110..3618ae4189c9b782cfae05d72de6e4c18b763466).

langz:
image

image

langz:
https://eufemia.dnb.no/

image

https://eufemia-chorescssportalstyles.gtsb.io/

image

tujoworker:
ohh, again! Nice finding 🙏 here is [the first fix](https://github.com/dnbexperience/eufemia/pull/1693/commits/e143c9e12cb4f2c239946fc8ecc55e314ed438f1) and here the [second](https://github.com/dnbexperience/eufemia/pull/1693/commits/8185bffeb48ddff68eebe852e797b9d907181598).

packages/dnb-design-system-portal/src/shared/parts/icons/ListAllIcons.module.scss
```diff
@@ -0,0 +1,53 @@
...
+  display: flex;
+  flex-wrap: wrap;
+  flex-direction: row;
+  align-items: flex-end;
+
+  padding: 0;
+  margin: 0;
+}
+
+.listitemStyle {
```
langz, line 13:
```suggestion
.listItemStyle {
```

langz (approved):
Looks really good 🥳 

I've added a few comments about renaming `listitemStyle` to `listItemStyle`.
Other than that, the last minor nitpick I have is that in Safari on my iPhone 12(haven't been able to reproduce it in desktop yet), it seems like the DNB logo is slightly cut off at the B, in DNB:

image

tujoworker:
> * Doesn't seem like we have animation when pressing the home button

yes, that was intentional. This saves a lot of extra complexity. At least how it was solved with Emotion. 
Also, there exists a v2 design. So we may rather spend time on some "nice to have" things, when this will come to life.

tujoworker:
Thank you so much for all the feedback! It has been very very helpful 🙏

PR #1695, opened 2022-11-03
docs(Upload): Upload loading state example

langz:
Replaces `checkedonChange` with `on_change`.
Reason for doing so is because the example did not work, and I got the following error in the console:
`Warning: React does not recognize the `checkedonChange` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `checkedonchange` instead. If you accidentally passed it from a parent component, remove it from the DOM element.`


tujoworker (approved):
👌

PR #1706, opened 2022-11-07
feat(Th): add table header sortable props

tujoworker:


packages/dnb-design-system-portal/src/docs/uilib/components/table/Examples.js
```diff
@@ -120,36 +120,36 @@ export const TableLongHeader = () => (
 <Table>
   <caption className="dnb-sr-only">A Table Caption</caption>
   <thead>
-    <tr className="dnb-table--small">
-      <th scope="col" colSpan="2">
+    <Tr className="dnb-table--small">
```
langz, line 123: Should this example, `TableLongHeader`, still use `className` for the `Tr` and `Th` components?
Or should the usage of className rather be represented by the usage of the component's properties instead?

packages/dnb-design-system-portal/src/docs/uilib/components/table/Examples.js
```diff
@@ -180,61 +180,61 @@ export const TableSticky = () => (
...
   <thead>
-    <tr>
-      <th scope="col" colSpan="2">
+    <Tr>
+      <Th scope="col" colSpan="2">
         Header
-      </th>
-      <th scope="col" className="dnb-table--sortable dnb-table--reversed">
+      </Th>
+      <Th scope="col" className="dnb-table--sortable dnb-table--reversed">
```
langz, line 187: Should this example, `TableSticky`, still use `className` for the `Th` components?
Or should the usage of className rather be represented by the usage of the component's properties instead?

packages/dnb-eufemia/src/components/table/TableTd.tsx
```diff
@@ -7,12 +7,6 @@ export type TableTdProps = {
    * Default: null
    */
   children?: React.ReactNode
-
-  /**
-   * Custom className on the component root
-   * Default: null
-   */
-  className?: string
```
langz, line 15: As the prop `className` is removed, should we also remove the usage of it in this component?

packages/dnb-eufemia/src/components/table/TableTh.tsx
```diff
@@ -8,10 +8,28 @@ export type TableThProps = {
   children: React.ReactNode
 
   /**
-   * Custom className on the component root
-   * Default: null
+   * Defines the table header as sortable
+   * Default: false
    */
-  className?: string
```
langz, line 14: As the prop `className` is removed, should we also remove the usage of it in this component?

packages/dnb-eufemia/src/components/table/TableTr.tsx
```diff
@@ -15,12 +15,6 @@ export type TableTrProps = {
   children:
     | React.ReactElement<TableTdProps>
     | Array<React.ReactElement<TableTdProps>>
-
-  /**
-   * Custom className on the component root
-   * Default: null
-   */
-  className?: string
```
langz, line 23: As the prop `className` is removed, should we also remove the usage of it in this component?

langz (commented):
Looks good 🙏 

The main thing I didn't completely get is why we remove the prop `className` from all components Td, Th, Tr, but that we don't remove the usage of it? 🤔 
Is it because we only want to use it internally, and not "publicly promote it"? 🤔

packages/dnb-eufemia/src/components/table/TableTh.tsx
tujoworker: `className` is still supported, but via the `HTMLTableCellElement` 👍

packages/dnb-eufemia/src/components/table/TableTr.tsx
tujoworker: `className` is still supported, but via the `HTMLTableRowElement` 👍

packages/dnb-design-system-portal/src/docs/uilib/components/table/Examples.js
tujoworker: Yes, I will remove this one in another PR 👍

packages/dnb-design-system-portal/src/docs/uilib/components/table/Examples.js
tujoworker: This will also be solved in a different PR 😁

tujoworker:
> The main thing I didn't completely get is why we remove the prop `className` from all components Td, Th, Tr, but that we don't remove the usage of it? 🤔 Is it because we only want to use it internally, and not "publicly promote it"? 🤔

Eufemia started with only table styles. But now that we have a component, we should go away from supporting "all features" with classes as well. This way we have less documentation to maintain.

langz (approved):
Thanks for looking into my comments, I think I understand it all now 🧠 💪

packages/dnb-eufemia/src/components/table/TableTr.tsx
langz: Of course, my bad 😹

PR #1710, opened 2022-11-08
chore(ReactLive): add Babel/Gatsby Plugin to transform code before hitting ReactLive

tujoworker:
With that change, we can rewrite our demos/examples to normal React function components and also get type safe checks by TypeScript. See [this commit](https://github.com/dnbexperience/eufemia/pull/1710/commits/1b244a84467b4bb4d9b54c0313961d8fe187a38a).

After:

Screenshot 2022-11-08 at 20 04 07

Or with a render function:

Screenshot 2022-11-08 at 15 29 46

Before it was inside back ticks/template literals:

Screenshot 2022-11-08 at 15 30 33

## Test plan

As of now, not integrations tests where written. But on the other hand, we run this code both during build and the visual tests. So we should cover at least "that it works".

We may consider to release this as a NPM package – and maybe move it to another repo as well?, and then we can make individual code coverage/integration tests. Just some thoughts.


langz (approved):
Great improvement 👏 
I was not able to use the "IntelliSense"(suggesting props, etc) for the StepIndicator in packages/dnb-design-system-portal/src/docs/uilib/components/step-indicator/Examples.tsx , but it worked for the example in packages/dnb-design-system-portal/src/docs_dummy/dummy/Examples.tsx

Might just be something with my setup 🤔

tujoworker:
> I was not able to use the "IntelliSense"(suggesting props, etc) for the StepIndicator

Yeah, I guess because we do not have types for StepIndicator inside the repo 😁 I will drop this commit btw. And we than can make the actual rewrites in other PRs, but probably only for components that exists with types.

tujoworker:
... and by the way, the we need to use a new "major" version from "emotion" in order to make this work. So we get a warning in the dev/build logs:

> warn Plugin gatsby-plugin-emotion is not compatible with your gatsby
version 4.24.4 - It requires gatsby@^5.0.0

But it should work fine. The reason is that the new (next) versions will not insert import helpers twice (via Babel): 

> ModuleParseError: Module parse failed: Identifier '___EmotionJSX' has al
  ready been declared (19:16)
  File was processed with these loaders:

And that's the same reason why it does not work with JavaScript files, but for another Babel transform (can remember right now which one).
There is for sure a way to omit this, and make it work. But I could not find any for a couple of hours. And I also think this is good enough for the improvement we want (and get).

PR #1711, opened 2022-11-08
feat(Table): add Th.HelpButton to be used in Table Headers

tujoworker:


packages/dnb-eufemia/src/components/table/style/_table-sort.scss
```diff
@@ -176,4 +176,11 @@
   & &__th#{&}--no-wrap {
     white-space: nowrap;
   }
+
+  & &__th,
+  & > thead > tr > th {
+    .dnb-help-button {
```
langz, line 182: It's fine, but I do sometimes feel that as specific selectors like this, .dnb-help-button, is a bit strange for a component that can take "anything"(React.ReactNode) as children. 

"You can pass anything, but when passing this or that specific component, we add some styling to it" 🤔

langz (approved):

packages/dnb-eufemia/src/components/table/style/_table-sort.scss
tujoworker: Yes – true, we can/should add this to the docs – right now with "words" as well. In this PR, its kind of documented because we use it in demos/examples. 
The background is that this is an UX pattern. But to make it more stream lined, we now have a "wrapper" component called `Th.SortButton` – so it will feel less "random" for the developer, but rather meant to be used.

PR #1716, opened 2022-11-10
chore(Table): convert table examples to TypeScript and the new format

tujoworker:


langz (approved):
I don't fully get the "TypeScript features" to work when running this branch locally in VSCode.
image

tujoworker:
> I don't fully get the "TypeScript features" to work when running this branch locally in VSCode.

Yes, very good point – I did not check that.

We have two choices, I think:

1. make the import so that TypeScript understands that this component is written in TS:

```jsx
import Table from '@dnb/eufemia/src/components/table'
import Th from '@dnb/eufemia/src/components/table/TableTh'
etc.
```

2. Or create index files in TypeScript – but that will complain for components not written in TypeScript (e.g. the Code element):
3. 
Screenshot 2022-11-10 at 20 47 05

> (alias) module Code
(alias) const Code: React.ForwardRefExoticComponent<React.RefAttributes<any>>
import Code
Type '{ children: string; }' has no properties in common with type 'IntrinsicAttributes & RefAttributes<any>'.ts(2559)

So we would then probably have to build all types and disable the auto generation.

I mean, at some point of time, we want to do this anyway. Its just a question of when.

langz:
> > I don't fully get the "TypeScript features" to work when running this branch locally in VSCode.
> 
> Yes, very good point – I did not check that.
> 
> We have two choices, I think:
> 
> 1. make the import so that TypeScript understands that this component is written in TS:
> 
> ```js
> import Table from '@dnb/eufemia/src/components/table'
> import Th from '@dnb/eufemia/src/components/table/TableTh'
> etc.
> ```
> 
> 2. Or create index files in TypeScript – but that will complain for components not written in TypeScript (e.g. the Code element):
> 
> Screenshot 2022-11-10 at 20 47 05
> 
> > (alias) module Code
> > (alias) const Code: React.ForwardRefExoticComponent<React.RefAttributes>
> > import Code
> > Type '{ children: string; }' has no properties in common with type 'IntrinsicAttributes & RefAttributes'.ts(2559)
> 
> So we would then probably have to build all types and disable the auto generation.
> 
> I mean, at some point of time, we want to do this anyway. Its just a question of when.

Thanks for looking into this, I'm not sure if we have to do anything in regards with this in this current PR.
We should probably do alternative 3 as you say, at some point 🙏

tujoworker:
@langz I [added](https://github.com/dnbexperience/eufemia/compare/611e3f28defc6425b27c3b544ef1ff39714aaf19..768e689fc1dea2a0498b548f989ad16673e2cf77) the imports to work with TypeScript. For examples, the import does not need to look pretty. And its helpful to get the TS checks while I continue the table enhancements.

PR #1717, opened 2022-11-11
chore(easing): move easing to properties

tujoworker:


packages/dnb-eufemia/src/style/core/properties.scss
```diff
@@ -106,6 +106,9 @@
   --shadow-default-blur-radius: 16px;
   --shadow-default-color: rgba(51, 51, 51, 0.08);
 
+  // Easing
+  --easing-default: cubic-bezier(0.42, 0, 0, 1);
```
langz, line 110: During which type of build will this property be added to packages/dnb-eufemia/src/style/properties.js?

langz (approved):

packages/dnb-eufemia/src/style/core/properties.scss
tujoworker: The same places `helper-classes.scss` is imported, but "before". 

Screenshot 2022-11-11 at 14 42 02

So it should be fine, or better even.

PR #1719, opened 2022-11-14
feat(Upload): support files dropped on the document body

tujoworker:
I think its better to make this not optional. 

But then the question is, should only the first support it? 

If yes, then we probably want a way to support via a property.
Or should it be possible to disable?

But probably, its fine as of now. Having several upload components on the page is probably something that not should happen.

### DIFF
--- a/packages/dnb-eufemia/src/components/upload/UploadDropzone.tsx
+++ b/packages/dnb-eufemia/src/components/upload/UploadDropzone.tsx
@@ -5,6 +5,8 @@ import HeightAnimation from '../height-animation/HeightAnimation'
 import { UploadContext } from './UploadContext'
 import type { UploadAllProps, UploadFile, UploadProps } from './types'
 
+export type UploadDragEvent = React.DragEvent | DragEvent
+
 export default function UploadDropzone({
   children,
   className,
@@ -17,7 +19,7 @@ export default function UploadDropzone({
 
   const { onInputUpload } = context
 
-  const getFiles = (event: React.DragEvent | DragEvent) => {
+  const getFiles = (event: UploadDragEvent) => {
     const fileData = event.dataTransfer
 
     const files: UploadFile[] = []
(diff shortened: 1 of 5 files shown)


packages/dnb-eufemia/src/components/upload/UploadDropzone.tsx
```diff
@@ -17,45 +17,76 @@ export default function UploadDropzone({
 
   const { onInputUpload } = context
 
-  const getFiles = (event: React.DragEvent) => {
+  const getFiles = (event: React.DragEvent | DragEvent) => {
```
langz, line 20: We could consider creating `type UploadDragEvent = React.DragEvent | DragEvent` or so, and reusing this everywhere in this file where we have `React.DragEvent | DragEvent`

langz (commented):
Some of the existing functionality don't seem to work flawlessly now:
- No yellow animation when uploading the same file multiple times by dropping files in other places than the "upload zone".
- Seems like it's replacing the existing file, and not adding new/more files.

No yellow animation:

https://user-images.githubusercontent.com/1359205/201651296-f69a663e-bb69-427e-a60f-290d1f50399e.mov



Replacing existing file instead of adding files:

https://user-images.githubusercontent.com/1359205/201651280-6bc17ac1-1c36-49cf-a6e5-6ead42d68820.mov



macOS: 12.6 (21G115)
Chrome: Version 107.0.5304.110 (Official Build) (x86_64)

langz:
> 

Good questions.
I'm trying to think of a scenario where apps don't want to get this functionality by default, but can't think of any 🤔 

_Having several upload components on the page is probably something that not should happen._
I could think of large forms, where it probably makes sense to have multiple upload components on a page, to separate which category/section the files belong to.  Or maybe I've just created a scenario for it in my head now 😂

packages/dnb-eufemia/src/components/upload/UploadDropzone.tsx
tujoworker: Agree very much!

tujoworker:
@langz also, now only the first one will accept files from the "body" – what do you think? Is that better or worse? We could maybe add an property instead `dropFilesSelector` and default to body? This way, its possible to define a HTML selector – or disable it, if null/undefined was given. Just an idea.

langz:
> @langz also, now only the first one will accept files from the "body" – what do you think? Is that better or worse? We could maybe add an property instead `dropFilesSelector` and default to body? This way, its possible to define a HTML selector – or disable it, if null/undefined was given. Just an idea.

Thanks for improving this. In the possible scenario of where there's multiple upload components in a single page, I do think it will cause less confusion/damage for the user to only add the files dragged to "body" to one of the upload components, than to all of the upload components.

In general, if there's multiple upload components in a single page, I do think it would be favourable that the developer could at least select which upload component that get the files dragged to "body". If it's somehow super important that the 2nd upload component on my page should get the files dragged to "body", then I would have to make sure that the 2nd upload component would mount 1st to make this happen. That wouldn't result in meaningful code 😂 
I would say that this is probably an edge case, and that we can handle this later if we wish.

langz:
Works on iPhone as well👌https://user-images.githubusercontent.com/1359205/201878970-22ff3870-1a73-4621-841d-c66fa92b1080.MOV

tujoworker:
> Works perfectly fine on iPhone as well👌

ohh, I didn't know that this was possible 😁

PR #1720, opened 2022-11-14
chore(Housekeeping): rename typography example

tujoworker:


langz (approved):
Looks good 🌮 
Just noticed that the background of this page is now a grid.
Doesn't seem to be done as part of this PR though. Just an obsvervation.

Take a look at the following screenshot and compare it to https://eufemia.dnb.no/uilib/typography/font-weight/

image

PR #1725, opened 2022-11-15
chore: rename screenshot tests

langz:


tujoworker (approved):
Nice finding! We need to remove all png files inside the snapshots folder for the renamed tests – and generate them again. Because they use the test names for the file names. But sadly there is no automation. Yes, I agree, its a little bit cumbersome.

We need anyway to consider to mover over to [jest-image-snapshot](https://github.com/americanexpress/jest-image-snapshot) – they support `customSnapshotIdentifier`. Even if ours is 10x faster, the [docs say](https://github.com/Prior99/jest-screenshot) – it does not support a [custom file name](https://github.com/Prior99/jest-screenshot/issues/71).

langz:
> Nice finding! We need to remove all png files inside the snapshots folder for the renamed tests – and generate them again. Because they use the test names for the file names. But sadly there is no automation. Yes, I agree, its a little bit cumbersome.
> 
> We need anyway to consider to mover over to [jest-image-snapshot](https://github.com/americanexpress/jest-image-snapshot) – they support `customSnapshotIdentifier`. Even if ours is 10x faster, the [docs say](https://github.com/Prior99/jest-screenshot) – it does not support a [custom file name](https://github.com/Prior99/jest-screenshot/issues/71).

I'll remember to also remove the "old files" as well 🙏

PR #1727, opened 2022-11-15
docs(Paragraph): move docs from /typography to /elements

langz:
Moved Paragraph docs from /typography to /elements.
I've not adjusted any of the contents of the docs yet, could be worth revising this now that it's been moved.

PR #1728, opened 2022-11-15
docs(Heading): move docs from /typography to /elements

langz:
Moved Heading docs from /typography to /elements.
I've not adjusted any of the contents of the docs yet, could be worth revising this now that it's been moved.


tujoworker (approved):
Nice 👍 We need to add `redirect_from` as well, like in #1727.

tujoworker:
> Why do we add the `redirect_from` property?

Yes, or if we miss an internal link to this page, the it will not get 404. But we should actually double check the internal linking anyway.

PR #1731, opened 2022-11-16
fix(Provider): rewrite to functional component

tujoworker:
This fixes also an issue; when the root context changes locale, it did not update its nested contexts, because we stored them in a state.

Here is a [reprod](https://codesandbox.io/s/eufemia-table-context-locale-2nbgkp) that shows the issue. When the locale changes in the "root" context, it did not update its nested contexts with that change.


langz (approved):
Looks good, is there any way I can assist with testing if this works as expected(not sure if it's needed though)?

jorgenholme (approved):
Awesome 💯

tujoworker:
> is there any way I can assist with testing if this works as expected(not sure if it's needed though)?

We can make a beta release and test it in the mono-repo 👍

PR #1738, opened 2022-11-21
feat(Table): enhance handling of spacing in TDs

tujoworker:
Also, split up styles and re-arrange examples.


packages/dnb-eufemia/src/components/table/style/_table-td.scss
```diff
@@ -0,0 +1,26 @@
...
+@import './_table-header-buttons.scss';
+
+.dnb-table {
+  & > tr > td,
+  & > tbody > tr > td,
+  &__td,
+  & > tr > th[scope='row'] {
+    padding: var(--spacing-small);
+    padding-top: calc(var(--spacing-small) * 1.25);
+    padding-bottom: calc(var(--spacing-small) * 1.188); /* 19/16 */
```
langz, line 15: I don't initially understand why we multiple with `1.188` to figure out the bottom padding.
Seems complicated, and am not able to see the context/reason, but it's probably needed 🙏

langz (approved):

packages/dnb-eufemia/src/components/table/style/_table-td.scss
tujoworker: ahh, yes! Just move this block. This is outdated and should be 1.25 (to get 20px) instead of 1.188 so we get the correct height (for the soon, outdated styles).

PR #1749, opened 2022-11-23
fix(HeightAnimation): avoid usage of useLayoutEffect during SSR

tujoworker:
React warns about the usage of `useLayoutEffect` when React renders on the server (SSR).

This behavior can be enabled with the `FAST_DEV` flag in gatsby-config.

One "quick" fix is to use useEffect on the server, which again does nothing there.

The React "team" has [some options](https://gist.github.com/gaearon/e7d97cdf38a2907924ea12e4ebdf3c85) to that "problem". We basically use option 1, with this approach.

Any other suggestions?

PS: I tried to create a test. But invalidating both the document or the window object (`delete globalThis.window`) makes the testing lib not rendering the component or hook.

### DIFF
--- a/packages/dnb-design-system-portal/src/shared/menu/ToggleGrid.tsx
+++ b/packages/dnb-design-system-portal/src/shared/menu/ToggleGrid.tsx
@@ -25,7 +25,7 @@ function isGridVisible() {
 }
 
 export function GridActivator() {
-  React.useLayoutEffect(() => {
+  React.useEffect(() => {
     if (isGridVisible()) {
       makeGridVisible()
     }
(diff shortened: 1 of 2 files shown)


langz (approved):
Thanks for the explanation.
I also think option 1 seems to be the the best, as long as the _[effect isn't important for first render](https://gist.github.com/gaearon/e7d97cdf38a2907924ea12e4ebdf3c85#option-1-convert-to-useeffect)_.

tujoworker:
I did check it and could not find any 👍 But I also think it is just fine, because on SSR, there is no difference and on the browser, with re-hydration, `useLayoutEffect` will still run before any `useEffect`, which is important for that specific use case.

PR #1754, opened 2022-11-29
fix(NumberFormat): use `dnb-sr-only` instead of `dnb-sr-only--inline`

tujoworker:
I can't find any reasons anymore why not just use `dnb-sr-only`. Its the only place also, when `--inline` was used.


langz (approved):
🤝

PR #1756, opened 2022-11-29
feat(Table): add sticky header support for overflow hidden on parent elements

tujoworker:
The new sticky handling (with JavaScript) is opt-in. I'm not sure if we should maybe make this as the default/only way? Because the position: sticky support is almost never possible, because of overflow usage. See description change in info.md file.

What do you think?


langz (approved):
Looks good 👍 

A couple observations when running the branch locally:
- I did not fully understand which values I can/should provide to [stickyOffset](https://eufemia-fixtablesticky.gatsbyjs.io/uilib/components/table/properties/)? 10, "10px", "10rem", "10em"? It says it "defines the offset (top) in rem", does it mean that when I provide 10, it will have offset of 10rem?
- In the example [Table with sticky header](https://eufemia-fixtablesticky.gatsbyjs.io/uilib/components/table/demos/#table-with-sticky-header), I don't think I was able to get sticky using only `sticky={true} `

tujoworker:
> I did not fully understand which values I can/should provide

Yes! You are right. Will fix that!

> I don't think I was able to get sticky using only `sticky={true}`

Before, we did [remove our](https://github.com/dnbexperience/eufemia/pull/1756/files#diff-d4e53e843594dd5f7da5e8b7b37204027d58812d66ded4676034884469d7c0c7L566) overflow. Even the one from the scroll view. Just to make the `sticky={true}` work.

And because it is so hard to ensure, no "overflow" is used in your app (parent children) we should consider to deprecate the support for it. On the other hand, the JavaScript solution does flicker some times and for now, it only works on the body "scroll". We should actually add support for having a sticky header inside a modal/drawer as well. Right now, it would not work with the JavaScript solution. 🧐

NancyBolstad:
Thanks for providing these features. After testing around, I have a couple of comment/questions:

1. Would be nice to have a case for TableContainer in storybook. I have created one story for TableContainer locally for testing, I can make a PR if this is something you would like to include. Felt it was nice to have for testing the sticky header behaviour when there are serval tables on the same page, for example.
2. Curious to know why `<th>` is hardcode to be  `role = “columnheader”` ? What if the scope is row? 
3. It seems like the sticky header doesn’t work well when we are using TableContainer (see the recording below). 

Screenshot 2022-12-01 at 15 06 21

tujoworker:
Thank you for your very good input! @NancyBolstad 

I will iterate and see what we can already improve in this PR.

1. See point 3.
2. [this PR](https://github.com/dnbexperience/eufemia/pull/1764) will fix it.
3. I will add support for it in this PR as well, including a Story to test it on.
4. Is actually intended to look like. But we should probably rename Row 3 to something more understandable. It uses both `noSpacing` and `align="right"`.

```jsx
<Td noSpacing align="right">
  <Code>noSpacing + align="right"</Code>
</Td>
```

packages/dnb-eufemia/src/components/table/__tests__/TableContainer.test.tsx
```diff
@@ -68,6 +68,15 @@ describe('TableContainer', () => {
     expect(Array.from(element.classList)).toContain('custom-class')
   })
 
+  it('should include ectra classe in scrol-view', () => {
```
langz, line 71:
```suggestion
  it('should include extra classes in scroll-view', () => {
```

packages/dnb-design-system-portal/src/docs/uilib/components/table/demos.md
```diff
@@ -40,24 +41,13 @@ A `small` sized table is only for special circumstances, where a lot of data nee
...
-
 <TableSticky />
 
-### Table with long header text (wrapping)
+### Table with a max height
 
-<TableLongHeader />
+A sticky table header with `sticky="css-position"` and `max-height` on the `Table.ScrollView`.
+
+<TableStickyMaxHeight />
```
langz, line 50: Nice👌

langz (approved):

PR #1762, opened 2022-12-01
docs(Radio Group): add example for radio group with label above

langz:
I find this example nice to have in the docs.

I've based the naming around "Radio group with label above", is there perhaps a better word/term I could use than "above"? 😂


tujoworker (approved):
👏 I'm fine with the above usage 💪

PR #1767, opened 2022-12-05
chore: add dependabot deps updater

tujoworker:
We may try to test this out. Not sure yet if it will be practical useful.


langz (approved):
What's different/new with this change, compared to the regular dependabot PRs we get from time to time, like https://github.com/dnbexperience/eufemia/pull/1765 ?

tujoworker:
> What's different/new with this change

This one will not only make PRs fro security vulnerabilities, but for regular feature updates.

PR #1769, opened 2022-12-05
chore(react-router): remove this deps including example

tujoworker:
We had only one example using `react-router-dom`. We also did link to a CSB with an example. So I think we should remove this example and dependency.


langz (approved):
Smart smart , cleaning up 💯

PR #1794, opened 2022-12-08
fix(InputPassword): hide password reveal in Edge

langz:
## Summary

This PR hides the "password reveal" button in Microsoft Edge.

Here's a screenshot of how it looked when both our internal password reveal button, and the default one in Microsoft Edge was displayed simultaneously:

image


tujoworker (approved):
Nice! And away goes the 👻

PR #1796, opened 2022-12-09
fix(scrollToLocationHashId): omit calling scrollTo when element is not found

tujoworker:
This is a documented helper class. 

But when there was not an element, `scrollTo` was still called. Which has negative effects on default browser location hash handling.

### DIFF
--- a/packages/dnb-eufemia/src/shared/__tests__/helpers.test.js
+++ b/packages/dnb-eufemia/src/shared/__tests__/helpers.test.js
@@ -3,6 +3,9 @@
  *
  */
 
+import React from 'react'
+import { render } from '@testing-library/react'
+
 import {
   setPageFocusElement,
   applyPageFocus,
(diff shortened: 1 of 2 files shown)


packages/dnb-eufemia/src/shared/__tests__/helpers.test.js
```diff
@@ -67,21 +70,95 @@ describe('"applyPageFocus" should', () => {
...
+        hash: '#unique-id',
+      },
+    })
+
+    const scrollTo = jest.fn()
+    jest.spyOn(window, 'scrollTo').mockImplementation(scrollTo)
+
+    render(<div id="unique-id">content</div>)
+
+    scrollToLocationHashId({ offset: 100 })
```
langz, line 90: Based on the it statement, shouldn't the offset be 0 or less? Might be something different though 🤔 

https://github.com/dnbexperience/eufemia/pull/1796/files#r1044462555

tujoworker: yes, that was my bad! I was thinking of the offside, inside the function 🫣 very good you found that 💯

PR #1809, opened 2022-12-13
feat(Table): add useHandleSortState hook

tujoworker:
This PR adds a little React Hook that could be very handy when dealing with three stage sorting, which should be naturally the default behavior.

But we also ensure that the `Th.SortButton` actually reacts correctly as well. This is handled by CSS. Also, for the CSS states, [this CSB](https://codesandbox.io/s/bold-field-glyxq6) was provided.

What do you think about all the option names and the name of the React Hook?

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/components/table/Examples.tsx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/components/table/Examples.tsx
@@ -21,64 +21,96 @@ import Th from '@dnb/eufemia/src/components/table/TableTh'
 import Td from '@dnb/eufemia/src/components/table/TableTd'
 import Tr from '@dnb/eufemia/src/components/table/TableTr'
 import TableContainer from '@dnb/eufemia/src/components/table/TableContainer'
+import useHandleSortState from '@dnb/eufemia/src/components/table/useHandleSortState'
 
 export const TableVariantBasic = () => (
-  <ComponentBox hideCode data-visual-test="table-default">
-    <Table.ScrollView>
-      <Table>
-        <caption className="dnb-sr-only">A Table Caption</caption>
-        <thead>
-          <Tr>
-            <Th>Column</Th>
-            <Th>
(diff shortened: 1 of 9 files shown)


langz:
I think the name `useHandleSortState` is appropriate 👍 

To try to learn/understand the "option names" I tried playing with the table in https://eufemia-feattableusehandlesortstate.gatsbyjs.io/uilib/components/table/demos/#basic-table, and I wasn't really able to understand how it all works 🤔  Maybe it's just a bug or so in the example.

tujoworker:
haha, yeah, this Hook helps only for the "visual" state. But the it can then be used for the sorting logic. Should we maybe provide that too in the Example? Not in the Hook, because this is very individual by every app and based on the data they have.

It's just, then we need to put our table data in an data container, and that will maybe destroy the "vanilla" HTML feeling you have when looking on the example code.

Would a nota bene be an idea?

tujoworker:
> Not sure if the change of the positioning of the "sort icon"(arrow) is expected or not, when changing "alignment".

Yes, it is on purpose. Because of the strange alignment:

Screenshot 2022-12-14 at 10 18 13

Let me know if there are other good arguments for not doing it 😉

langz (approved):

tujoworker:
Here is [the "enhancement"](https://github.com/dnbexperience/eufemia/pull/1809/commits/a31da8b53ea011fb1ec295c07f652c3170938aa5) 🔥

tujoworker:
> * It seems like the focus/border is a bit "cut off" on the right side.

This happens because the table is inside a `ScrollView` that has `overflow: auto` – the only thing we could do is to "move" the last and the first one, 2px inside. What do you think?

langz:
> 

Thanks for checking, as it's not a "problem" with the Table itself, it's all good 🙏

tujoworker:
Added `activeSortName` based on feedback from Emiliana 👍

PR #1812, opened 2022-12-14
fix(ScrollView): ensure an ancestor scroll area (body) can still be scrolled with mouse wheel

tujoworker:
When using mousewheel or touchpad while holding the mouse over e.g. a table, you could not scroll the "body" down anymore. With this change, you can do it as expected.

I think, this should be the default behavior.


langz (approved):
Great find👌

PR #1817, opened 2022-12-15
fix(Button): support target property

langz:
This PR is adding documentation and "TypeScript support" for the `target` property.

Here's an example of when consumers using TypeScript, tries to use the Button component with the `target` property as of today(pre this PR):

image


langz:
Was it so that I need to generate/add this change to `packages/dnb-eufemia/src/components/button/Button.d.ts` as part of this PR as well, or will the "build server" automagically take care of this? I can't remember 😊

tujoworker (approved):
Nice. We need soon to rewrite it to TypeScript 😉

As of now, there is a type definition file for the button. So it will not "overwrite" it actually with the "new" docs.
I pushed a commit with it, so should be fine.

PR #1822, opened 2022-12-15
chore(Tag): rewrite tests w/o using data-testid

langz:
Removes the following `data-testid`'s:
- tag
- tag-group
- tag-group-label

I checked in the monorepo, and it exists only a few occurrences of this over there, and have already opened a PR for it 🙏

### DIFF
--- a/packages/dnb-eufemia/src/components/tag/__tests__/Tag.test.tsx
+++ b/packages/dnb-eufemia/src/components/tag/__tests__/Tag.test.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { fireEvent, render, screen, within } from '@testing-library/react'
+import { fireEvent, render, screen } from '@testing-library/react'
 import Tag from '../Tag'
 import nbNO from '../../../shared/locales/nb-NO'
 import {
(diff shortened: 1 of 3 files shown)


packages/dnb-eufemia/src/components/tag/__tests__/Tag.test.tsx
```diff
@@ -173,8 +146,8 @@ describe('Tag', () => {
       </Tag.Group>
     )
 
-    expect(screen.queryByTestId('tag').className).toMatch(
-      skeletonClassName
+    expect(document.getElementsByClassName(skeletonClassName).length).toBe(
```
tujoworker, line 149: Nice, you use `getElementsByClassName` – here we could probably also use `toHaveLength(1)` instead of getting the lengths first. But just keep it as is. I do not bother 😉

tujoworker (approved):
I like that change a lot 🙏

packages/dnb-eufemia/src/components/tag/__tests__/Tag.test.tsx
langz: I've updated this for all PRs related to the removal of data-testid now 🙏

PR #1829, opened 2022-12-19
chore(Timeline): rewrite tests w/o data-testid

langz:
Removes the following `data-testid`'s:
- timeline
- timeline-item-label-icon
- timeline-item-label-title
- timeline-item-content-subtitle
- timeline-item-content-info
- timeline-item

I checked in the monorepo, and it exists only a few occurrences of this over there, will have to change there as well.


tujoworker (approved):
Awesome work ✅

PR #1830, opened 2022-12-19
chore(Portal): fix sidebar on small view ports

tujoworker:


langz (approved):
Thanks for the quick fix 🎊 
Works great here 🙏

PR #1835, opened 2022-12-19
fix(Table): make sticky header work when used in combination with accordion rows

tujoworker:
Reported bug.

👉Before
Screenshot 2022-12-19 at 20 20 13

👉After
Screenshot 2022-12-19 at 21 30 10


langz (approved):
✅

PR #1841, opened 2022-12-20
fix(Anchor): use text-decoration instead of box-shadow for chromium breaking underline support

tujoworker:
Chromium based browser have no longer support for Anchor underline based on `box-shadow`. At least when I test it. 

Screenshot 2022-12-21 at 06 21 20

That said, we can still support it (with `inset` and negative offset), but the underline offset is then not enough by 1 or 2px.

Screenshot 2022-12-21 at 06 22 25

vs

Screenshot 2022-12-21 at 06 23 20

But now that we do not need to support IE anymore, we can make use of newer CSS properties, like:

- `text-decoration-color`
- `text-decoration-thickness`
- `text-underline-offset`

Out puppeteer chromium version does not support `thickness` from what I can see, so it is 1px thick:

Screenshot 2022-12-20 at 20 29 12

I'm not 100% sure if we should release this in v10 or v9 as per now. But I think, it should be ok in v9 as well.


langz (approved):
Looks great 👍 

One observation/thought is that the "alignment" of the letters/line following the "line break" is a bit further to the left than the 1st/"starting line". Not sure if it should be like this or not 🤔. 
This was in Chrome using the [Deploy preview](https://eufemia-fixanchorunderline.gatsbyjs.io/):

Screenshot 2022-12-21 at 09 17 13

tujoworker:
Yeah, this is because of the tiny padding (left&right). UX wants even more. We should check if we can remove it entirely.

PR #1849, opened 2023-01-04
fix(Anchor): allow rel property in TS

langz:
Seems like `HTMLProps` includes more props than `HTMLAttributes`, `rel` being one of them.

Or maybe we should use `React.ComponentPropsWithoutRef<"a">`?  🤔 
https://github.com/typescript-cheatsheets/react/blob/main/docs/advanced/patterns_by_usecase.md#componentprops

### DIFF
--- a/packages/dnb-eufemia/src/elements/Anchor.tsx
+++ b/packages/dnb-eufemia/src/elements/Anchor.tsx
@@ -33,7 +33,7 @@ export type AnchorProps = {
 }
 
 export type AnchorAllProps = AnchorProps &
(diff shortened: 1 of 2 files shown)


tujoworker (approved):
That's interesting 🤔 we may actually switch over to use  `ComponentPropsWithoutRef`. It sounds easier to use/read than HTMLPropa etc.

tujoworker:
Do you found the commit that made the change (from prev version) by the way?

langz:
> Do you found the commit that made the change (from prev version) by the way?

https://github.com/dnbexperience/eufemia/pull/1715

PR #1851, opened 2023-01-04
release of v9.45.1

langz:


tujoworker (approved):
It will go out as v9.45.1 as long as only a fix: and not a feat: git tag is listed under effected commits. But the title does not really matter 😉

langz:
> It will go out as v9.45.1 as long as only a fix: and not a feat: git tag is listed under effected commits. But the title does not really matter 😉

Thanks, I updated the title from `release of v9.46` to `release of v9.45.1` anyways 😹

Issue #1853, opened 2023-01-05
InputMasked with Slider not working with decimals


oggez:
## 🐛 Bug Report

When using a InputMasked with Slider the values will not update if it is less than 1 (e.g. 0,55) .

https://user-images.githubusercontent.com/17780527/210816769-2195d7c7-aa65-44d0-b281-cf68f6424b39.mov

## To Reproduce

Create a InputMasked and Slider component that uses the same values.
Drag the value to lower than 1 and see that the InputMasked value is not being updated.

Demo:
https://codesandbox.io/s/elated-sun-rpkc0p

## Expected behavior

The value of InputMasked should be updated when values are lower than 1

## Eufemia Version

9.45.1

PR #1854, opened 2023-01-05
fix(InputMasked): add replace decimal in parseFloat

oggez:
## Summary

Solves issue #1853

## Test plan

Tested in storybook by adding slider

https://user-images.githubusercontent.com/17780527/210823380-c5ebc43a-e368-4432-a2cd-55318baca095.mov

### DIFF
--- a/packages/dnb-eufemia/src/components/input-masked/InputMaskedUtils.js
+++ b/packages/dnb-eufemia/src/components/input-masked/InputMaskedUtils.js
@@ -148,7 +148,8 @@ export const correctNumberValue = ({
     if (
       localNumberValue !== '0' &&
       localNumberValue.startsWith('0') &&
-      parseFloat(numberValue) === parseFloat(localNumberValue)
+      parseFloat(numberValue.replace(decimalSymbol, '.')) ===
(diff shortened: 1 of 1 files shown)


tujoworker (approved):
Thank you for the contribution!

I sent you an invite to the team, so you don't need to use a Fork next time. You need to accept the invitation.

That said, could you add the following test to `InputMasked.test.js`? It basically tests that issue.

```jsx
  it('should handle decimal values less than 1', () => {
    const MockComponent = () => {
      const [value, setState] = React.useState(1.5)

      return (
        <>
          <Component
            value={value}
            number_mask={{
              allowLeadingZeroes: true,
              allowNegative: false,
              allowDecimal: true,
            }}
          />

          <button
            onClick={() => {
              setState(value - 0.4)
            }}
          />
        </>
      )
    }

    render(<MockComponent />)

    const element = document.querySelector('input')
    const button = document.querySelector('button')

    expect(element.value).toBe('1,5')

    fireEvent.click(button) // 1,1
    fireEvent.click(button) // 0,7
    fireEvent.click(button) // 0,3

    expect(element.value).toBe('0,30')
  })
```

PR #1855, opened 2023-01-06
feat(P): deprecate style_type property

langz:


packages/dnb-eufemia/src/elements/Ingress.js
```diff
@@ -6,6 +6,6 @@
 import React from 'react'
 import P from './P'
 
-const Ingress = (p) => <P style_type="ingress" {...p} />
```
langz, line 9: Not sure if it's safe to "just remove this" 🤔

packages/dnb-eufemia/src/elements/Lead.js
```diff
@@ -6,6 +6,6 @@
 import React from 'react'
 import P from './P'
 
-const Lead = (p) => <P style_type="lead" {...p} />
```
langz, line 9: Not sure if it's safe to "just remove this" 🤔

packages/dnb-eufemia/src/elements/Ingress.js
tujoworker: It looks like some example code uses `ingress` then? Because the visual test did fail. The new changes is to the left. Less "bold".

langz: Thanks, I'll take a look 👀

tujoworker (approved):
Nice! I see that `/Ingress` is not documented. And I think we should remove it in another PR. Its basically the same as `medium` font-weight. 

Screenshot 2023-01-06 at 14 51 05

Will you follow that up @langz?

langz:
> Nice! I see that `/Ingress` is not documented. And I think we should remove it in another PR. Its basically the same as `medium` font-weight.
> 
> Screenshot 2023-01-06 at 14 51 05
> 
> Will you follow that up @langz?

Yes yes, #1872

PR #1860, opened 2023-01-06
fix(Button): remove left and right padding of tertiary button

tujoworker:
The tertiary button has space to its left and right to ensure space to other elements, much like a text link. But our other buttons do not have a space, so this PR is about to align to the other button variants and remove the space entirely.

The main commit is [this one here](https://github.com/dnbexperience/eufemia/pull/1860/commits/37ee3c94df51177bdf0d0ac4a514650c19ba677d).

**For reference:**

Screenshot 2023-01-06 at 10 23 42

**Without padding (proposed):**

Screenshot 2023-01-06 at 10 23 33

**With padding (currently):**
Screenshot 2023-01-06 at 10 23 10


langz:
Good job :)
I'll help do some testing and reviewing this 🙏

langz:
My 1st observation was the screenshot tests, seems to be some alignment issue and/or spacing issues(see red lines):

Screenshot 2023-01-06 at 10 48 16

tujoworker:
> My 1st observation is the screenshot tests, seems to be some alignment issue and/or spacing issues(see red lines):

Very good observations! Looks like we have to fix these padding's. 
That makes me think of it should rather be a v10 fix? I initially thought this should be ok 😰

langz:
> > My 1st observation is the screenshot tests, seems to be some alignment issue and/or spacing issues(see red lines):
> 
> Very good observations! Looks like we have to fix these padding's. That makes me think of it should rather be a v10 fix? I initially thought this should be ok 😰

Hmm, good point.
We can and should of course fix our components to look the same as before, like fix the spacing in the breadcrumbs(see image).

Screenshot 2023-01-06 at 12 24 42


But most likely the same visual changes as in breadcrumb can/will happen in other applications using the tertiary button as well, and that's perhaps a breaking change? 🤔

tujoworker:
I'm thinking of removing the padding on the first item? What do you think?

**Proposed**

Screenshot 2023-01-06 at 16 05 23

**Current**

Screenshot 2023-01-06 at 16 05 42

langz:
> I'm thinking of removing the padding on the first item? What do you think?
> 
> **Proposed**
> 
> Screenshot 2023-01-06 at 16 05 23
> 
> **Current**
> 
> Screenshot 2023-01-06 at 16 05 42

I like it :)
Must be easier for apps using this to align it with other stuff now when there's no left padding/spacing on the first item.

tujoworker:
Lets make it a part of v10 then 🥤

packages/dnb-design-system-portal/src/docs/uilib/about-the-lib/releases/eufemia/v10-info.md
```diff
@@ -51,6 +51,10 @@ v10 of @dnb/eufemia contains _breaking changes_. As a migration process, you can
 
 1. Find the `target_blank_title` property and replace it with `targetBlankTitle`.
 
+### [Button](/uilib/components/button)
+
+1. The padding of the `teriary` button is removed. Please, check your application and add back the padding of `0.5rem` if needed.
```
langz, line 56:
```suggestion
1. The padding of the `tertiary` button is removed. Please, check your application and add back the padding of `0.5rem` if needed.
```

langz (approved):

langz:
Nice nice, the padding/spacing in the breadcrumb got improved I will say. Seems like the chevron is more "centered" now than previously:

image

PR #1861, opened 2023-01-06
chore: change P.test.js to use testing library

AudunBeck:
## Summary
Changing the use of enzyme to react testing library.

## Test plan
Changed testing to use testing library standards and some refactoring in the process.
Ran 
`yarn test /P.test.js`
to see that test now works.

### DIFF
--- a/packages/dnb-eufemia/src/elements/__tests__/P.test.js
+++ b/packages/dnb-eufemia/src/elements/__tests__/P.test.js
@@ -11,6 +11,7 @@ import {
   axeComponent,
 } from '../../core/jest/jestSetup'
 import Component from '../P'
+import { render } from '@testing-library/react'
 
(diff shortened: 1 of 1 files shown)


tujoworker (approved):
🎊 Thank you for the contribution 🙏

I think, because the p tag probably never will be "rewritten" with nested elements, its fine to use `firstChild` here. Other places we use `document.querySelector`, which is easy to use/replace instead of Enzymes `find` method.

AudunBeck:
> 🎊 Thank you for the contribution 🙏
> 
> I think, because the p tag probably never will be "rewritten" with nested elements, its fine to use `firstChild` here. Other places we use `document.querySelector`, which is easy to use/replace instead of Enzymes `find` method.

I will do a quick rewrite to match the way it has been done in the FormLabel.test.js

PR #1867, opened 2023-01-06
chore(Portal): add position relative to example-box

tujoworker:
I can't remember, why we not did set position relative after the Emotion => SCSS conversion. But right now, I can't find any negative issues having it.

This fixes an issue a few places, where the class `example-box` is used, like on this page: https://eufemia.dnb.no/uilib/getting-started/demos/

Screenshot 2023-01-06 at 14 31 47


langz (approved):
Perfecto 🇮🇹

PR #1871, opened 2023-01-06
chore(MediaQueryUtils): remove deprecated addListener

langz:


tujoworker (approved):
yep, that's an IE11 fallback 👍

PR #1873, opened 2023-01-06
chore: make Slack step not fail when run from Fork

tujoworker:
Honestly, I thought we have fixed the issue by checking the "owner". But with the two new fork PRs, I clearly see we did not.

Screenshot 2023-01-06 at 15 20 16

I don't think the issue is gone with that PR. But I don't know what we can do more about it either. 

The [docs](https://docs.github.com/en/actions/learn-github-actions/expressions#always) says clearly, that the step should not fail, when `always` is used. Do they not?

If we would add a env check `if: ${{ always() && secrets.SLACK_WEBHOOK }}`, then we are probably on the same page as before? What do you think? @langz 

I also wander, how we can get the [rest of the actions](https://github.com/dnbexperience/eufemia/pull/1854) to start running?

Screenshot 2023-01-06 at 15 26 06

### DIFF
--- a/.github/workflows/build-check.yml
+++ b/.github/workflows/build-check.yml
@@ -75,4 +75,4 @@ jobs:
         env:
           SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK }}
           MATRIX_CONTEXT: ${{ toJson(matrix) }}
-        if: ${{ always() && github.repository_owner == 'dnbexperience' }}
+        if: ${{ always() && !github.event.pull_request.head.repo.fork }}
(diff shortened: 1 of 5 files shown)


langz (approved):
Looks good, let's give this a try 🥇 
I don't really have much experience with configuring this.
I can of course try to familiarize myself with how this works, and come back to you 🙏

tujoworker:
Lets try this change first before we try more 👍 Even I bet it will still fail, because it probably has a reason we added `&& github.repository_owner == 'dnbexperience'` 😇

tujoworker:
What we could else try is this [**How to skip steps or jobs in GitHub Actions for PRs from forks**](https://til.cazzulino.com/devops-ci-cd/how-to-skip-steps-or-jobs-in-github-actions-for-prs-from-forks) in addition.

PR #1874, opened 2023-01-06
chore: rewrite button.test.js to testing library

AudunBeck:
## Summary
Changing the use of enzyme to react testing library.

## Test plan
Changed testing to use testing library standards and some refactoring in the process.
Run
`yarn test /Button.test.js`
to see that test now runs using testing library

### DIFF
--- a/packages/dnb-eufemia/src/components/button/__tests__/Button.test.js
+++ b/packages/dnb-eufemia/src/components/button/__tests__/Button.test.js
@@ -13,6 +13,7 @@ import {
 } from '../../../core/jest/jestSetup'
 import Component from '../Button'
 import IconPrimary from '../../IconPrimary'
+import { fireEvent, render } from '@testing-library/react'
 
 const props = fakeProps(require.resolve('../Button'), {
   optional: true,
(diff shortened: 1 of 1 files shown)


packages/dnb-eufemia/src/components/button/__tests__/Button.test.js
```diff
@@ -48,218 +49,241 @@ describe('Button component', () => {
...
+    render(
       <Component
         {...props}
         href="https://url"
         target="_blank"
         icon={null}
       />
     )
-    expect(Comp.find('svg').exists()).toBe(true)
+    expect(document.querySelector('svg') !== undefined).toBe(true)
```
tujoworker, line 109:
```suggestion
    expect(document.querySelector('svg')).toBeTruthy()
```

Is probably easier to "read".

tujoworker (approved):
Awesome 💯

I think, using `toBeTruthy` will be easier to "read" when it comes to your question. It applies on both  `!== null` and `!== undefined`. I commented it just a few places, but see there are many more 🤓

PR #1885, opened 2023-01-09
chore(Shared): remove export of types from interfaces

langz:


tujoworker (approved):
We probably should write about this change in the v10 docs (a new section about types?).

PR #1892, opened 2023-01-09
feat(SpacingUtils): deprecate createStyleObject

langz:


tujoworker:
`createStyleObject` was never documented, so I think, we are fine with not listing it as a v10 thing.

PR #1922, opened 2023-01-14
fix(Table): keep paragraph font-size when table size is medium or small

tujoworker:
Text inside tables do not need to be inside a paragraph. Therefore they should not inherit the table sizes.

Thanks Emiliana for reporting 🙏


langz (approved):
👌

PR #1925, opened 2023-01-16
feat(Modal, Dialog, Drawer): remove hideCloseButton

langz:
Just a beginning.

Here's a few of the the visual differences, which doesn't look too good, just by removing this "deprecated prop":
image

image


langz:
I'm closing this, as I think the removal of `hideCloseButton` makes more sense to do together with removing the `mode` property

PR #1928, opened 2023-01-16
feat: build pure component css packages (without deps)

tujoworker:
This PR will change how the styling package are build/handled. 

Before we included all needed style deps and the needed CSS "reset" to make them look good.

- New is that we do not include deps and the reset anymore. This makes it possible to import only some components, without the extra code load.
- For snapshot style comparison, we use the these deps.scss files. Let me know if you have any better name than `deps.scss`.


langz (approved):
Nice nice, great work 👍 

I've ran this PR locally, and did not find any major concerns, works great 💪

PR #1934, opened 2023-01-18
feat(build): replace node-sass with dart-sass

tujoworker:
The motivations is to get closer to be able to use Node.js v18.

- dart-sass has issues when run in jest with jsdom, because jsdom exposes `window` – so dart-sass thinks it runs inside a browser environment, and wants to use http://, instead of file://. 
- That's the reason we have to tweak the window location object. Also, we use still the deprecated API, because the compile API does still not work with jsdom.
- I have compared the css output inside the build folder, and could not find any difference, except for the different formatting. Which we also see in the changed snapshots.

Side-note: dart-sass has a misleading name. Its actually JavaScript based, and not faster than node-sass – the only difference is that the source is written in Dart. But node-sass is not maintained anymore. So no choice.


langz (approved):
🙇

PR #1944, opened 2023-01-20
chore: upgrade yarn version to v3.3.1

tujoworker:
Our current version is a year old. So I thought, its time to make an update on that front.


langz (approved):
💯

PR #1563, opened 2022-09-14
feat(Breadcrumb): add animation when in collapse mode

tujoworker:
This PR depends on #1561 

Check out the [demos](https://eufemia-featbreadcrumbanimation.gtsb.io/uilib/components/breadcrumb/demos/).

PR #1949, opened 2023-01-26
fix(Breadcrumb): ensure focus state is not partially hidden

tujoworker:
🙏 Tack Micaela 🇸🇪

With PR #1563 we got this issue when focusing the item:

Screenshot 2023-01-26 at 09 31 07

By adding a visual test, simulating the focus state, we should avoid this in the future.

We fix the issue by removing the `overflow: hidden` – its not used anyway, because our `HeightAnimation` will add&remove it during the animation anyway.

We also rename the visual test file (for v10 migration) and refactor the Example file.


langz (approved):
🙏

PR #1945, opened 2023-01-20
feat: replace jest visual test driver to jest-image-snapshot

tujoworker:
Because `jest-screenshot` is not really maintained anymore, and it lacks Node.js v18 support and does still not work with Node.js v16 on A1 chips, we need another approach.

[jest-image-snapshot](https://github.com/americanexpress/jest-image-snapshot) looks to be a good fit. It supports 

- Node.js v18
- jest watch CLI command to update outdated snapshots on the fly
- jest.retryTimes
- and it's maintained.

PR #1951, opened 2023-01-26
chore: upgrade to Node.js v18 and Rollup.js v3

tujoworker:
**This PR depends on #1945** but will be rebased later onto the v10 branch.

The motivation is to bee in the loop when it comes to Node.js version support.

**Why upgrade Rollup.js at the same time?**

I first tried to upgrade only Rollup.js to v3, but it requires us to upgrade Node.js v14.18 – but this Node.js version has problems with localization, so it won't work with all our NumberFormat parts. While Node.js v18 does work just fine – like before.

We also can't upgrade to Node.js v18 without upgrading Rollup.js, as it breaks in v2 – so we need to upgrade both at the same time.

**Background info:** We use Rollup.js for creating UMD and MJS bundles. Even they are probably not used by any project, I think, we still keep them. They have been requested several years back, when things where unsure, where the tech setup headed towards.

More info here: https://eufemia.dnb.no/uilib/usage/first-steps/bundles/


langz (approved):
Looks great 👍 
Do/should we verify that the bundles look/work as expected in any other ways than in our tests, or is this good enough? 🙏

tujoworker:
I have checked them out manually 👍 and they looked identical. Except, now we also remove comments.

PR #1969, opened 2023-02-02
fix(TableContainer): ensure bottom border on all tables

tujoworker:
This PR fixes an issue introduced in #1956 where the first table in a container, has no bottom border.

Screenshot 2023-02-02 at 11 39 54

PR #1970, opened 2023-02-02
fix(TableContainer): enhance handling of empty head and foot area

tujoworker:
This PR is based on #1969 

It corrects the height of the foot/head area when they are empty. Now they even do not need to exists.


NancyBolstad (approved):
Nice! 👍

PR #1976, opened 2023-02-08
feat(ProgressIndicator): add submit type

tujoworker:
Screenshot 2023-02-08 at 09 12 47
Screenshot 2023-02-09 at 10 24 41

What about calling this new variant/type for `submit`? Any other/better arguments/ideas?

PS: This PR needs a UX RFC before release.


packages/dnb-design-system-portal/src/docs/uilib/components/progress-indicator/demos.md
```diff
@@ -4,322 +4,102 @@ redirect_from:
...
 
-### Default ProgressIndicator is Circular
+### Submit form activity indicator
 
-<ComponentBox>
-	{/* jsx */ `
-<ProgressIndicator />
-`}
-</ComponentBox>
+Use this indicator in forms, side by side a submit button. Its a light weight and dense way to tell the user whats happening at the moment.
```
langz, line 33:
```suggestion
Use this indicator in forms, side by side a submit button. It's a lightweight and dense way to tell the user what's happening at the moment.
```

langz (approved):
Great addition👏 

Sometimes when adding functionality like this, to an existing component, I'm a bit afraid that we and/or consumers of the component possibly lose the understanding of the component.

As of pre this PR, I believe the existing component had two "variants", which was circular or linear progress indicator. And in both these "variants" I believe all props could been used to make necessary changes in both variants. By now adding a prop `state` which can only be used when `type="submit"`, and doesn't make sense when circular or linear, I feel we complicate stuff a bit. Both for the consumer, and us developers. 
It's not super clear to me or maybe for the consumer that the only props applicable for the new "submit type" is `label`, `type`, `state`, and no other of the 10+ props listed in the [properties tab](https://eufemia-featprogressindicatorformsubmi.gatsbyjs.io/uilib/components/progress-indicator/properties/). Is there some alternative approach where us as developers could actually see which props thats required for the "submit type progress indicator", like i.e making an internal component (SubmitProgressIndicator) with it's own props `label`, `type`, `state`.
And a way where consumers more easily could see which props that's applicable for the "submit type progress indicator"? 

Reason for me thinking like this now, is because of my recent experiences with the "old Modal component" which supported dialog and drawer as two "modes", but then had a lot of props which was only applicable for one of the modes. I found it very difficult to see/understand which props that can be used in the component when mode is dialog and which props that can be used when mode is drawer. Both from a consumer perspective and a developer perspective.

Not sure if it's easy to make this change though, or what our alternatives are 🤔

PR #1977, opened 2023-02-08
fix(InfoCard): enhance accessibility + set max width of 60ch

tujoworker:
- avoid use a h3, because we not know the heading level outside of the component. And screen readers do not need to consume this in their index of all headings.
- set a max width for [better readability](https://www.uutilsynet.no/veiledning/tekst-og-struktur/226). 60ch is on the smaller side, but we also use a very small font size there.
- refactor `data-testid`.
- also, fix the skeleton view:

Screenshot 2023-02-08 at 11 48 44


langz (approved):
Looks good :) 
I'm however starting to think that v10 will be "quite breaking" for some of our consumers, now that we'll include the data-testid removal as well. But I guess there's no better time now that we'll release a new major version 🙏

tujoworker:
Yeah, it felt right on this one at least. We need to find out the strategy on the rest of them.

PR #1982, opened 2023-02-09
chore: remove role="main" because of IE11 drop

tujoworker:


langz (approved):
Didn't know that we would have to do `<main role="main">` for IE11.
You would think that the `<main>` element had a role of main, but probably not for IE 🤯 😂

PR #1984, opened 2023-02-09
feat(ScrollView): add interactive=auto to observe the content

tujoworker:
When `interactive=auto` is used, and there is a scrollbar, make the scrollview tabbable. 
The screenshot shows, that I was able to focus it:

Screenshot 2023-02-09 at 22 06 07

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/components/fragments/scroll-view/properties.md
+++ b/packages/dnb-design-system-portal/src/docs/uilib/components/fragments/scroll-view/properties.md
@@ -4,7 +4,7 @@ showTabs: true
 
 ## Properties
 
-| Properties                                  | Description                                                                                            |
-| ------------------------------------------- | ------------------------------------------------------------------------------------------------------ |
-| `interactive`                               | _(optional)_ set to `true` to make the content accessible to keyboard navigation. Defaults to `false`. |
-| [Space](/uilib/components/space/properties) | _(optional)_ spacing properties like `top` or `bottom` are supported.                                  |
+| Properties                                  | Description                                                                                                                                                                                                           |
+| ------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `interactive`                               | _(optional)_ to make the content accessible to keyboard navigation. Use `true` or `auto`. Auto will detect if a scrollbar is visible and make the ScrollView accessible for keyboard navigation. Defaults to `false`. |
+| [Space](/uilib/components/space/properties) | _(optional)_ spacing properties like `top` or `bottom` are supported.                                                                                                                                                 |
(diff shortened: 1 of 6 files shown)


packages/dnb-eufemia/src/fragments/scroll-view/ScrollView.tsx
```diff
@@ -61,19 +61,69 @@ function ScrollView(localProps: ScrollViewAllProps) {
...
+    if (interactive === 'auto') {
+      setAsInteractive(hasScrollbar())
+    }
+  }, [interactive, children]) // eslint-disable-line react-hooks/exhaustive-deps
+
+  React.useLayoutEffect(() => {
+    /**
+     * Safari Desktop adds one pixel "on zoom" level 1
+     * therefore we just remove it here
+     */
```
langz, line 95: I'm not sure if this comment adds much meaning, when we already have the same comment here: https://github.com/dnbexperience/eufemia/pull/1984/files#diff-c4e193e7296352891f7c623b7c0bdf0f56ad24178c6ae7a20f578718dfc1f939R116-R119

packages/dnb-design-system-portal/src/docs/uilib/components/fragments/scroll-view/properties.md
```diff
@@ -4,7 +4,7 @@ showTabs: true
...
-| ------------------------------------------- | ------------------------------------------------------------------------------------------------------ |
-| `interactive`                               | _(optional)_ set to `true` to make the content accessible to keyboard navigation. Defaults to `false`. |
-| [Space](/uilib/components/space/properties) | _(optional)_ spacing properties like `top` or `bottom` are supported.                                  |
+| Properties                                  | Description                                                                                                                                                |
+| ------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `interactive`                               | _(optional)_ to make the content accessible to keyboard navigation. Use `true` or `auto`. Auto will detect if a scrollbar is visible. Defaults to `false`. |
```
langz, line 9: _Auto will detect if a scrollbar is visible._, and do what? What will the "auto" value actually do?
Set the `interactive` prop to `true` if a scrollbar is visible?(Just me guessing)

tujoworker: yeah, nice finding. It will make it reachable (interactive, so the user can use their arrow keys) by setting tabindex=0. Do you have a better text suggestion?

tujoworker: Maybe something like that?

> Auto will detect if a scrollbar is visible and make the ScrollView accessible for keyboard navigation.

PR #1990, opened 2023-02-14
fix(Button): support rel property

langz:
Sort of a continuation of #1817 

This PR is adding documentation and "TypeScript support" for the `rel` property.

Here's an example of when consumers using TypeScript, tries to use the Button component with the `rel` property as of today(pre this PR):

Skjermbilde 2023-02-14 kl  14 42 47


thaytharma (approved):
Thanks a lot Anders!🤘🏼🤘🏼

tujoworker (approved):

PR #1994, opened 2023-02-17
chore(CodeBlock): remove unused parts

tujoworker:
- I checked the logs during SSR for these props. None where used.
- Remove `react-markdown` dependency.


langz (approved):
👑

PR #1997, opened 2023-02-21
docs(ProgressIndicator): removes props min_time and variant

langz:


packages/dnb-design-system-portal/src/docs/uilib/components/progress-indicator/events.md
```diff
@@ -6,6 +6,6 @@ redirect_from:
 
 ## Events
 
-| Events        | Description                                                    |
-| ------------- | -------------------------------------------------------------- |
-| `on_complete` | _(optional)_ will be called once the `min_time` has timed out. |
+| Events        | Description                                                |
+| ------------- | ---------------------------------------------------------- |
+| `on_complete` | _(optional)_ will be called once it's no longer `visible`. |
```
langz, line 11: I'm not 100% sure about this statement 🤔

tujoworker: I did not check that out. But could it be that the on_complete is called when its finished a "round" of animation? Right before it would start from 0 again?

langz: I tried this, and waited for quite some time, but I never got any .logs in the console:

image


But in the same example, if I manually set visible to true, then false, then I get a log statement, so I think it's related to visible 🤔

tujoworker: You are right 👍

tujoworker (approved):
👌

PR #2000, opened 2023-02-22
chore(Portal): add Cypress tests

tujoworker:
We had some cy tests, but have not run them on the CI. This PR enables these and runs them on the build we already make. There we rename the visual test GH Actionto e2e and run it there as a job.

We also show the job name, with a link in the Slack notification:

Screenshot 2023-02-22 at 08 59 16

The total time will not take longer, because we run it in parallel:

Screenshot 2023-02-22 at 08 52 05


langz (approved):
💯

PR #2004, opened 2023-02-22
fix(Autocomplete): pass status props to SubmitButton

langz:
From:
image

To: 
image


tujoworker (approved):
I would love if the new tests would use testing lib instead of enzyme. I know, you prefer consistency. But we need to do the transformation because of React v18 very soon anyways 😉

langz:
> I would love if the new tests would use testing lib instead of enzyme. I know, you prefer consistency. But we need to do the transformation because of React v18 very soon anyways 😉

Done :)

PR #2009, opened 2023-02-22
docs(NumberFormat): fix info docs

langz:
From:
image

To:
image


tujoworker:
ohh, do we use `pre` other places as well?

langz:
> ohh, do we use `pre` other places as well?

I think so, there's a few references to `<pre` in different places.

tujoworker (approved):

tujoworker:
It only effects when `pre` was used in .md files. So all other places should be fine from what I saw.

PR #2010, opened 2023-02-23
chore: fix autocomplete visual tests

tujoworker:
Also, check if `message` is in `matcherResult`. It happened locally during local reproduction, so I got no report.


langz (approved):
❤️

PR #2021, opened 2023-02-27
chore(FormSet): fix vertical is a non-boolean attribute

langz:
Should fix the following "docs issue":

Screenshot 2023-02-27 at 10 55 13

Screenshot 2023-02-27 at 10 55 02


langz:
@tujoworker, I see that in packages/dnb-eufemia/src/components/form-set/FormSet.d.ts there's around 30 different props for the FormSet, but  in the [docs](https://eufemia.dnb.no/uilib/components/form-set/properties/) we only present 10ish of them? Do you think we should add more of these props to the `[docs](https://eufemia.dnb.no/uilib/components/form-set/properties/)`, or is it fine as it is?

tujoworker (changes requested):
I think the opposite is what we should use. At least in the React world.

langz:
> I think the opposite is what we should use. At least in the React world.

Sure sure, though I'm not sure what the opposite would be? I would think that the opposite would be to only allow a boolean value for the `vertical` prop, but that would result in the same error I believe 🤔

PR #2024, opened 2023-02-27
chore(CI): continue on Slack error

tujoworker:
Here are [the docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) about the `continue-on-error: true` command.


langz (approved):
Smart smart 🗣️ 
It's not like a failing Slack integration is so important that we want the workflow to fail 👍

PR #2031, opened 2023-03-01
fix: remove FormRow context support from /elements

tujoworker:
This PR solves #2021

We can safely do so, because none of our elements do (will) use any of the "FormRow" props.

The internal Element HOC is used by:

- Anchor
- Blockquote
- Code
- Dd
- Dl
- Div
- Td
- H
- Hr
- Img
- Li
- Ol
- P
- Span
- Ul

Also, its not documented that it should support it.


joakbjerk (approved):
👍

langz (approved):
Great, thanks for looking into this 👍

PR #2034, opened 2023-03-01
docs: remove `properties.scss` import from docs

tujoworker:
Because it's actually not valid and needed to import these into SASS. They are import by the core package anyway and will not give any value in the SASS world. now that we do not need to use a postcss to support IE11 fallbacks.


langz (approved):
🙇

PR #2036, opened 2023-03-01
Test espen

eliland:


tujoworker:
Nice!

Two things:

1. I see you used v10 as the basis. That means, you would need to change the target base in a PR to v10 as well. You can change that in the PR.
2. I see you probably do not use prettier on saving files. I recommend it to do so. And if you don't want that, then we still need to ensure, the prettier config is picked up correctly. Right now it's not.

PR #2043, opened 2023-03-03
chore: require Node.js v18 for the repo

tujoworker:
When a new dev uses a different version, he will quickly get a failure message when trying to run Node/yarn install.

Should we mention the change somewhere?

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "gatsby/browserslist": "4.21.4"
   },
   "engines": {
-    "node": "14.17.0",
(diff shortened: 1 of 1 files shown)


langz:
> When a new dev uses a different version, he will quickly get a failure message when trying to run Node/yarn install.
> 
> Should we mention the change somewhere?

If the "failure message" is enough(impossible to not see), then I think that's enough 🙏

langz (approved):

PR #2045, opened 2023-03-03
chore(Types): removes comment about not changing the docs

langz:


tujoworker (approved):
It's actually still true (as of now). But I'm ok with removing it.

langz:
> It's actually still true (as of now). But I'm ok with removing it.

Aha, okay, is it?
I thought now that we do manual changes in the .d.ts files, that we don't want to do auto stuff 🤔

PR #2047, opened 2023-03-03
chore(Types): import SpacingProps in d.ts files

langz:
This PR imports and extends SpacingProps instead of "duplicating" props like `space`, `top`, `right`, `left`, `bottom` in each d.ts file.


tujoworker (approved):
This is a solid enhancement 🙏

PR #2052, opened 2023-03-06
chore(Space): remove PropTypes

langz:
I'm not sure if this is the correct way of fixing this.

But the problem I want to fix is that in v10 as of now, for `Space` component in VSCode, we get that it has two definitions(PropTypes and TypeScript), but looks to default to PropTypes.
Is this correct? To me, I would guess we should only use TS?
This is how it looks in VSCode:
Screenshot 2023-03-06 at 13 34 55

Screenshot 2023-03-06 at 13 35 07

After this attempt of an fix(it shows the TS definition):
Screenshot 2023-03-06 at 13 33 52

### DIFF
--- a/packages/dnb-eufemia/src/components/space/Space.tsx
+++ b/packages/dnb-eufemia/src/components/space/Space.tsx
@@ -4,7 +4,6 @@
  */
 
 import React from 'react'
-import PropTypes from 'prop-types'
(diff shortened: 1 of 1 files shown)


langz:
I don't think this is correct at all, making an issue instead.
https://github.com/dnbexperience/eufemia/issues/2053

PR #2054, opened 2023-03-06
chore(Portal): change imports to relative and add ESLint rule

tujoworker:
This is just a quick draft/RFC – in response to PR #2051. I think, it could be just as fine, when we import things "inside" its own package with relative imports. To ensure that, we could use the added ESLint plugin. It also has auto-fix and is actively maintained. 

What do you think?


langz (approved):
💚

Issue #2056, opened 2023-03-07
Input component, is onKeyDown and onChange real properties?


langz:
In the following example/docs(https://eufemia.dnb.no/uilib/components/input/events/#prevent-setting-a-new-value) we refer to `onKeyDown` and `onChange`, are these real events? I think perhaps `onChange` should be `on_change`, and I don't see `onKeyDown` in the events of the component https://eufemia.dnb.no/uilib/components/input/events/#events

<img width="1790" alt="image" src="https://user-images.githubusercontent.com/1359205/223386694-b2c22bbf-1146-4501-9b96-9651baee11ca.png">

PR #2057, opened 2023-03-07
chore(TypeScript): Change stories.js to stories.tsx

langz:
What's been done so far:
- Renamed stories.js files to stories.tsx.
- Fixed minor TS issues in stories and/or in .d.ts files.

Motivation:
- TypeScript all over the place.
- With TypeScript it's easier that both examples and stories are "in sync" with the real component. Until now, I feel like both examples and stories could contain faulty/wrong usage of the real component.


packages/dnb-eufemia/src/components/input/Input.d.ts
```diff
@@ -216,6 +208,8 @@ export interface InputProps
    */
   on_change?: (...args: any[]) => any;
 
+  on_key_down?: (...args: any[]) => any;
```
langz, line 211: I added this `on_key_down` prop/event, as we had a few "stories" where we used it.
I think I can add documentation for this `on_key_down` prop/event as part of https://github.com/dnbexperience/eufemia/issues/2056

tujoworker (approved):
This was a really nice "cleaning in shelves" 💪

PR #2058, opened 2023-03-07
chore(Portal): make Examples imports relative in mdx files

tujoworker:
I'm not sure if this is needed. And beside this change in here, we still have 201 imports like this left: `from 'Docs/` – I still think, maybe it makes sense some places to change, and other places probably not.


langz:
Why would it make sense to _some places to change, and other places probably not_?
How would I as a developer know if I should use relative path or Docs/uilib/components... ?

tujoworker:
Yeah, good thoughts! I would personally prefer relative paths, just because, its the default behavior in the Node.js ecosystem. Alias is something that is invented on top of bundlers.
I remember, that the UX designer, Kevin wanted to have them explicit, like we have today. And I guess, its more logical for none developers.

But, because we have used it almost all places as such, we may stick with it as of now and change it rather later – everywhere.

PR #2065, opened 2023-03-09
chore(TypeScript): type form status props

langz:


tujoworker (approved):
And another nice improvement 🙏

PR #2071, opened 2023-03-09
chore(TypeScript): type icons

langz:
These components all use some kind of icon, but some use `IconPrimary` as they pass the props to the Button component(which uses `IconPrimary`),  and others use `Icon` directly in their component.

As of now, I've referred to the correct types from either `IconPrimary` or `Icon`, for the given property, but could/should it only be one `Icon` type, or does it make sense to have both `IconPrimary` and `Icon`?


tujoworker (approved):
Internally we can use Primary, but in the docs, we should try to not document too much about it, as you can always use Icon with imports, and just use IconPrimary, when you know what you are doing.

PR #2072, opened 2023-03-09
chore(Portal): upgrade to MDX v2

tujoworker:
With this, we are ready for a Gatsby v5 upgrade, as soon as we are ready for React v18.
And MDX file is with this, like inside a JSX. Before it was vanilla MD with some extra flaviour.


langz (approved):
Looks to work on my machine 🤖

PR #2074, opened 2023-03-10
chore(TypeScript): type label & label_direction types

langz:


packages/dnb-eufemia/src/components/progress-indicator/ProgressIndicator.d.ts
```diff
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import type { SectionSpacing, SectionStyleTypes } from '../Section';
 import type { SpacingProps } from '../space/types';
+import type { FormLabelLabelDirection, FormLabelText } from '../FormLabel';
```
langz, line 4: I'll revert this, as ProgressIndicator  is not actually using FormLabel

langz: Reverted.

PR #2076, opened 2023-03-10
chore: add sb theme files

tujoworker:
In this PR, we add a new theme file, that looks for component files. Simply run `yarn build` to collect future theme files.

We also:
- Give all properties a prefix of `sb-` so we never overwrite the existing.
- Add an example/WIP theme file for button.
- Add a single, first visual test for the button.
- Make a properties snapshot test.


langz (approved):
💯

joakbjerk (approved):
🏦

PR #2077, opened 2023-03-10
chore(Portal): fix layout re-render issue

tujoworker:
The page did scroll to top when other page assets where loaded on hover (e.g. hover in the sidebar menu). But that was just a side-effect of this issue.


langz (approved):
🔍 ✅

PR #2082, opened 2023-03-10
chore(TypeScript): type locale

langz:
`locale` prop of DatePicker looks to be something else than a string: https://eufemia.dnb.no/uilib/components/date-picker/properties/#properties, hence I'm not changing that :)


tujoworker (approved):
💯 Nice!

PR #2083, opened 2023-03-10
chore(Portal)move browser assets to dnb subfolder

joakbjerk:


tujoworker (changes requested):
🎉 Nice work 🙏

There are a couple of file we should not move. I hope you can move them back again without too much effort 🙏

We should not relocate the following files:
- CNAME 
- LICENSE.txt
- robots.txt
- google4f78509f2ca83a08.html
- google59b639c8cd158d31.html
- We probably don't need to relocate everything inside `/static/images` either, as they are no brand specific .

joakbjerk:
> 🎉 Nice work 🙏
> 
> There are a couple of file we should not move. I hope you can move them back again without too much effort 🙏
> 
> We should not relocate the following files:
> 
>     * CNAME
> 
>     * LICENSE.txt
> 
>     * robots.txt
> 
>     * google4f78509f2ca83a08.html
> 
>     * google59b639c8cd158d31.html
> 
>     * We probably don't need to relocate everything inside `/static/images` either, as they are no brand specific .

I'll move them out again then :D

tujoworker (approved):
Nice work 🥰

PR #2090, opened 2023-03-13
chore(CI): API rate limit exceeded for installation ID 15060380

tujoworker:
[This](https://github.com/dnbexperience/eufemia/actions/runs/4402679787/jobs/7710112359) is an issue we can't do much about it. It looks like the GitHub Action we use, uses an GitHub API for checking if its ready to go. But there is a rate limit. Its a [known issue](h).

For nor we can try to increase the interval from 3 sec to 10 sec.


langz (approved):
Thanks for looking into this 🙏 📿

PR #2092, opened 2023-03-13
chore(TypeScript): use IconProps in IconPrimary

langz:


packages/dnb-eufemia/src/components/breadcrumb/BreadcrumbItem.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react'
 
 // Components
 import Button from '../button/Button'
-import IconPrimary, { IconPrimaryIcon } from '../icon-primary/IconPrimary'
+import IconPrimary, { IconIcon } from '../icon-primary/IconPrimary'
```
langz, line 5: Need to fix this

packages/dnb-eufemia/src/components/icon/Icon.d.ts
```diff
@@ -11,6 +11,16 @@ export type IconHeight = string | number;
...
 
+type IconColorHexString = `#${string}`;
+type IconColorRGBString = `rgb(${number},${number},${number})`;
+type IconColorRGBAString = `rgba(${number},${number},${number},${number})`;
+type IconColorCSSPropertiesString = `var(--color-${string})`;
+type IconColorColorString =
+  | IconColorHexString
+  | IconColorRGBString
+  | IconColorCSSPropertiesString;
+export type IconColor = IconColorColorString | string;
```
langz, line 22: I'm not too sure about how much value all of this string stuff does.
Maybe I should just do `export type IconColor = string;`.
My main thing that I would like to accomplish is just to share `IconColor` so that we know it's the "same thing" across different components, etc.

tujoworker: I think you have a good point in your thoughts. `string` may be good enough.

tujoworker (approved):

packages/dnb-eufemia/src/components/icon/Icon.d.ts
langz: Updated to `export type IconColor = string;` then :)

PR #2094, opened 2023-03-13
chore(TypeScript): use StepIndicatorItem in StepIndicator

langz:
This PR:
- Tries to reuse the `StepIndicatorItemProps` as a potential type in `StepIndicator`'s `data` prop.
- Removes some undocumented and what I believe is faulty props for the `StepIndicator` which seems to actually belong to `StepIndicatorItem`: `title`, `is_current`, `inactive`, `disabled`, `status`, `status_state`, `on_render` (not sure how/why they was being generated for the StepIndicator though 🤔 )

### DIFF
--- a/packages/dnb-eufemia/src/components/step-indicator/StepIndicator.d.ts
+++ b/packages/dnb-eufemia/src/components/step-indicator/StepIndicator.d.ts
@@ -1,28 +1,23 @@
 import * as React from 'react';
 import type { SpacingProps } from '../../shared/types';
-import { FormStatusText } from '../FormStatus';
 import type { SkeletonShow } from '../Skeleton';
+import type { StepIndicatorItemProps } from './StepIndicatorItem';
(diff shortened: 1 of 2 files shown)


packages/dnb-eufemia/src/components/step-indicator/StepIndicator.d.ts
```diff
@@ -1,28 +1,23 @@
...
-
-      /**
-       * Will be called once the user clicks on the current or another step. Will be emitted on every click. Returns an object `{ event, item, current_step }`.
-       */
-      on_click?: (...args: any[]) => any;
-      on_render?: (...args: any[]) => any;
-    }[];
-export type StepIndicatorTitle = string | React.ReactNode;
-export type StepIndicatorStatusState = 'warn' | 'info' | 'error';
+  | Pick<
```
langz, line 10: Picking props, as it seems like `StepIndicatorItemProps` also includes some "internal props" like `currentItemNum` which is required.

PR #2095, opened 2023-03-13
feat(Skeleton): removes style_type prop

langz:
Doesn't look like `style_type` is used, couldn't even find styles(.scss code) matching ``dnb-skeleton--${style_type}``.
If we think we'll add skeleton styles in the near future, we may of course keep this though 🙏


tujoworker (approved):
Yes, true, for consistency, we should remove this feature.

We did use it to set a CSS modifier: `dnb-skeleton--${style_type}`

There are a couple of extra theme files, that are related to this too:

https://github.com/dnbexperience/eufemia/tree/v10/packages/dnb-eufemia/src/components/skeleton/style/themes

langz:
> Yes, true, for consistency, we should remove this feature.
> 
> We did use it to set a CSS modifier: `dnb-skeleton--${style_type}`
> 
> There are a couple of extra theme files, that are related to this too:
> 
> https://github.com/dnbexperience/eufemia/tree/v10/packages/dnb-eufemia/src/components/skeleton/style/themes

Thanks, I'll remove these files as part of this PR as well then 🙏

tujoworker:
> Do you think we should list this as a breaking change or not @tujoworker ?

This component was created with some flexibility in mind. But we have since seen the news to rather foster consistency.

I don't think it would have been UX driven to use a different variant/style. And if someone would have used "shine", it would not be "critical" if the style changes for those in v10. It's not a breaking change in my eyes, just because of the circumstances mentioned. But you could mention it in the EUFEMIA_CHANGELOG file. Maybe?

langz:
> > Do you think we should list this as a breaking change or not @tujoworker ?
> 
> This component was created with some flexibility in mind. But we have since seen the news to rather foster consistency.
> 
> I don't think it would have been UX driven to use a different variant/style. And if someone would have used "shine", it would not be "critical" if the style changes for those in v10. It's not a breaking change in my eyes, just because of the circumstances mentioned. But you could mention it in the EUFEMIA_CHANGELOG file. Maybe?

Great, thanks, I'll add it to the EUFEMIA_CHANGELOG file 🙏

PR #2096, opened 2023-03-13
chore(TypeScript): only export the shared types

langz:
This is just a draft/suggestion as of now(didn't mark it as draft as I wanted the builds/CI to run), and will need some more work to fix TS errors, etc.

The question I'm thinking about is do we want to export all of our types or not?

If we `export` everything I think the eslint rule `no-unused-vars` will not be able to find types that's not used internally, or maybe it's a way of configuring this to assist us a bit more, even if we still `export` all types 🤔 

Or is it super important for the client's(users of eufemia) to be able to build their own types on top of the types we export to them? The more types we export to consumers, the more stuff can also potentially break 🤔

### DIFF
--- a/packages/dnb-eufemia/src/components/accordion/Accordion.d.ts
+++ b/packages/dnb-eufemia/src/components/accordion/Accordion.d.ts
@@ -7,8 +7,8 @@ import type { SpacingProps } from '../space/types';
 import AccordionContent from './AccordionContent';
 import AccordionHeader from './AccordionHeader';
 import AccordionProvider from './AccordionProvider';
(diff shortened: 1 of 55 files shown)


tujoworker:
Its very common to export all types, just so types can be re-sued – but also "forced" in a more custom scenario. Like the usage of `x as ButtonProps`.

And as long as we say, we do not think that changing a TS type is a breaking change, but rather a "dev" enhancement, then I don't see the problem. But we should probably document this, because its not written down.

That said, I agree with your thoughts and this PR! Its just that I think, we may maybe, just maybe, create more work down the road.

langz:
Thanks for feedback @tujoworker, closing this for now then 🙏

PR #2100, opened 2023-03-15
docs: crate main doc for how to deal with brand

joakbjerk:
Added a draft for the main documentation for "How to deal with the brand". Still not entirely sure where the most optimal home for it would be, so this is still subject to change, and open to better suggestions 😄

As of now it's place under "usage" in the "uilib" portal. The choice was between there or in the "brand" portal, but I'm not sure if either of those are the right choice either 😅

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/contribute/contact.mdx
+++ b/packages/dnb-design-system-portal/src/docs/contribute/contact.mdx
@@ -12,9 +12,10 @@ import { Button } from '@dnb/eufemia/src'
 
 For all discussion topics, ask in Slack:
 
-<Button href="https://dnb-it.slack.com/archives/CMXABCHEY">
-  #eufemia-web
-</Button>
+<Button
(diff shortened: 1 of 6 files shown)


tujoworker (approved):
Nice! Lets start from there 👍

packages/dnb-design-system-portal/src/docs/contribute/contact.mdx
```diff
@@ -12,9 +12,10 @@ import { Button } from '@dnb/eufemia/src'
 
 For all discussion topics, ask in Slack:
 
-<Button href="https://dnb-it.slack.com/archives/CMXABCHEY">
-  #eufemia-web
-</Button>
+<Button
+  href="https://dnb-it.slack.com/archives/CMXABCHEY"
+  text="#eufemia-web"
+/>
```
langz, line 19: This change changed the looks of the button from 
image
 to 
image

PR #2102, opened 2023-03-15
style: move :root vars into their component itself

langz:
There's still a few :root variables left, as I didn't find it easy to move them or "duplicate them" as I feel like they should refer to the same source of truth.

I also kept :root variables of modal, as when I tried removing them, all modals(drawer and dialog) didn't work/look as expected. Not exactly sure why 🤔 

The changes have been tested in the portal locally, let's see if the CI and our screenshot tests will see anything I've not discovered 👀 

Suggestion to breaking change message:

BREAKING CHANGE:

Multiple CSS variables have been removed from :root, see https://github.com/dnbexperience/eufemia/pull/2102


tujoworker (approved):
Looks good – very nice work 🙏

PR #2103, opened 2023-03-15
fix: remove FormRow context support from components not using FormRow

tujoworker:
FormRow is only used for components that have a label. We have [docs about it](https://eufemia.dnb.no/contribute/getting-started/#formrow-support-with-includevalidprops).

- This fix will be released in v9
- It will be cherry-picked to v10
- This fix is related with PR #2031
- Tests are added in PR #2106


langz (approved):
Great great 👏 
I've only tested that the warnings/errors in the console of the Form Status -> Toggle Animation story doesn't show up, and it doesn't 🥳 🎆 🍾

PR #2104, opened 2023-03-15
chore(TypeScript): use HProps in H1->H6

langz:


packages/dnb-eufemia/src/elements/H.js
```diff
@@ -50,14 +47,13 @@ export default class H extends React.PureComponent {
...
       level, // eslint-disable-line
       size,
-      className,
       ...props
     } = this.props
 
     return (
       <E
         as={as || is}
-        internalClass={classnames(size && `dnb-h--${size}`, className)}
```
langz, line 60: I removed the `className` property of `H` element, as I wasn't able to see it being used anywhere...

tujoworker: True, we had no tests or usage for it. But we should keep it. I've added a test to this PR to confirm this.

tujoworker (approved):
Very nice improvement  👍

packages/dnb-eufemia/src/elements/H.d.ts
```diff
@@ -12,12 +12,14 @@ export type HSize =
 export interface HProps
   extends React.HTMLProps<HTMLElement>,
     SpacingProps {
-  className?: string;
```
langz, line 15: I'll probably add this property again then?

packages/dnb-eufemia/src/elements/H.d.ts
```diff
@@ -12,12 +12,14 @@ export type HSize =
...
   as?: string;
   level?: string;
   size?: HSize;
   children?: React.ReactNode;
 }
+
+export interface SharedHProps extends Omit<HProps, 'as'> {}
```
langz, line 21: If/when adding `className` again to interface `HProps`, should we omit it here?
Do we want the possibility/TS-support to give H1/H2/H3/H4/H5/H6 the `classname` property?

packages/dnb-eufemia/src/elements/H.d.ts
tujoworker: I don't its needed, `React.HTMLProps<HTMLElement>` should take care of `className`.

langz: Aha, thanks 🥇

PR #2105, opened 2023-03-15
chore: refactor integration tests to TypeScript

tujoworker:
I want to add tests for PR #2103 – but I want to do it in TS files. So this is a simple refactor of almost all tests. But not all. Just the easy once 😁


langz (approved):
Great work 👍 
Feels super nice adding more TypeScript, which results in us removing old/outdated properties from tests, etc which shouldn't be supported 👏

PR #2109, opened 2023-03-17
feat: use CSS vars in focus-ring (`fakeFocus`)

tujoworker:
This change makes things more flexible in terms of theming.

The actual changes are in [this commit](https://github.com/dnbexperience/eufemia/pull/2109/commits/32e5fc616f8a077e567e4fcda2d4cc464f0c6089).

I play with some thoughts of renaming `fakeFocus` to `focusRing`. Should we do that?


langz (approved):
Great 👍 

Yeah, I do also think it makes sense to rename, so that the CSS vars sort of match the mixin, I would vote for that 👨‍🚀

tujoworker:
> I vote for that 👨‍🚀

Here is a PR #2110

PR #2110, opened 2023-03-17
feat: rename SCSS mixin `fakeFocus` to `focusRing`

tujoworker:
This is a following up PR of #2109

BREAKING CHANGE:

SCSS mixin was renamed from `fakeFocus` to `focusRing`.


langz (approved):
🙇

PR #2111, opened 2023-03-17
chore(Portal): add new fonts and remove preload

tujoworker:
As a temporary solution, we add these fonts inline as base64. Its a copy&paste basically.


langz (approved):
Seems to work locally here 🤖

eliland (approved):

PR #2112, opened 2023-03-17
feat: add shared Theme component and useTheme hook

tujoworker:


packages/dnb-eufemia/src/shared/__tests__/Theme.test.tsx
```diff
@@ -0,0 +1,90 @@
+import React from 'react'
+import { render } from '@testing-library/react'
+import Theme from '../Theme'
+
+describe('Theme', () => {
+  it('does not render a dive, when no name was given', () => {
```
langz, line 6: _Does not render a dive_?
I wasn't easily able to understand this it statement, and the actual test 🤔

langz (approved):

packages/dnb-eufemia/src/shared/__tests__/Theme.test.tsx
tujoworker: Yeah, true. I removed it, because originally I thought we may not wrap things in a "div" when no name was given. So but that's not the case anymore. So this test is a left over I removed now 👍

joakbjerk (commented):
🎡

PR #2115, opened 2023-03-17
chore(TypeScript): use AccordionProps in AccordionGroupProps

langz:
On a side note, I see the file name is AccordionProvider, but the component is named AccordionGroup 🤔


tujoworker (approved):

tujoworker:
Nice fining! We should rename it. It can be useful for debuging reasons to have correct names.

langz:
> 

I agree, and I tried to give it a quick look now, but I'm actually not sure what to rename to what 😂 
I'm not sure if the components and or files, with types, should be named `AccordionProvider` or `AccordionGroup`.
When searching for both I find multiple comments, files, types, examples(before-started.mdx) with references to both, so I'll have to actually figure out what is what 🤔 
I can take a deeper look tomorrow or on the plane home after todays meeting ✈️

PR #2063, opened 2023-03-08
chore(Portal): refactor SidebarMenu to TypeScript and FC with hooks

joakbjerk:

PR #2120, opened 2023-03-21
 chore(SidebarMenu): refactor menu items to be grouped and given the correct highlighting when active

joakbjerk:
Before:
Screenshot 2023-03-21 at 10 50 41

After:
Screenshot 2023-03-21 at 10 50 06


langz:
Looks nice, I’ll take a look at this tomorrow morning🌅

langz:
This is just an observation, and I'm not 100% sure how or if we would want to improve upon this observation:

Seems like when clicking info, demos, etc, for components who have "child components" or a layer below, that the "item" doesn't stay "active" in the menu.

Examples:

https://deploy-preview-2120--eufemia.netlify.app/uilib/components/pagination/infinity-scroller/info/
https://deploy-preview-2120--eufemia.netlify.app/uilib/components/fragments/drawer-list/info/
https://deploy-preview-2120--eufemia.netlify.app/uilib/components/fragments/scroll-view/info/

 

https://user-images.githubusercontent.com/1359205/226849460-6213f226-11fe-4eef-b5bc-d617cbb33d3d.mov

tujoworker:
@langz Nice catch! 🤩

langz:
I reported the following issue https://github.com/dnbexperience/eufemia/issues/2134 that I found when testing this PR.
Reason for reporting it as an issue is because I thought it was related to https://github.com/dnbexperience/eufemia/pull/2063 which already exists in v10.

But when running v10 locally, I wasn't able to reproduce this issue, so maybe it's actually related to some changes in this PR?

packages/dnb-design-system-portal/src/shared/menu/SidebarMenu.tsx
```diff
@@ -277,62 +248,93 @@ function ListItem({
...
-          'dnb-anchor--no-underline',
-          'dnb-anchor--no-radius',
-          'dnb-anchor--no-hover',
-          icon && graphics[icon] ? 'has-icon' : null
+          'dnb-sidebar-menu',
+          `l-${level}`,
+          isActive && 'is-active', // use anchor hover style
+          isInsideActivePath && 'is-inside-active-path',
+          isInsideActiveCategory && !isInsideActivePath && 'is-inside',
+          status ? `status-${status}` : null,
```
langz, line 264:
```suggestion
          status && `status-${status}`,
```

Would this give the same result?
Reason for suggesting it is just to "follow the same style/way" as the lines above https://github.com/dnbexperience/eufemia/pull/2120/files#diff-80240cf1f7728b7dff84548e0d34eb958ce8b4ce677e7fd8ffa320a7c83241fdR261-R263

tujoworker: `false` is just fine – we use false a lot inside classNames – so I would rather use it because the `null` return is unnecessary.

PR #2122, opened 2023-03-21
docs(ColorsTable): fixes color and hover effect

langz:


tujoworker (approved):
That was fast and it works perfectly 🤩

PR #2129, opened 2023-03-21
docs(Naming conventions): fix console warning when using class

langz:


tujoworker (approved):
@joakbjerk will also go thought all places and do the same 👍

joakbjerk (approved):

PR #2135, opened 2023-03-22
chore: move Anchor to subfolder (for now)

tujoworker:


langz (approved):
Great great 👍 
Ran it locally, and it still seems like the Anchor works on my machine 🤖 😸

PR #2141, opened 2023-03-22
docs(ToggleButton): removes outdated label_position docs

langz:
This seems to just be a copy+paste from `Radio`'s properties docs, https://eufemia.dnb.no/uilib/components/radio/properties/#radio-group-context, as `ToggleButton` doesn't have a property named `label_position`. Hence, I removed this documentation.

If the documentation "actually is true" in terms of it actually "_passing through some(which?) of ToggleButton properties to the Group. This way all nested ToggleButton's will get the properties._", then we should rather replace the `label_position` with an actual property `x` that does what it says. But I wasn't quickly able to find "one of these properties".


tujoworker (approved):
Yes agree 😝

PR #2143, opened 2023-03-22
Chore(Blockquote): move blockquote to subfolder

joakbjerk:


packages/dnb-design-system-portal/src/shared/menu/SidebarMenu.tsx
```diff
@@ -90,7 +90,7 @@ export default function SidebarLayout({
 
   function scrollToActiveItem() {
     if (!scrollRef?.current) return
-
+    console.log('Yall here?')
```
langz, line 93: I'm here 🙌

joakbjerk: Rebase gone wrong? 😅 no idea where that came from 😝

tujoworker (changes requested):
Moving styles should not cause the visual snapshots to change 🤔

tujoworker (approved):
Nice work – good to get the last commit in, so we don't set types for `className` and `children` 👍

PR #2167, opened 2023-03-29
chore(Portal): ensure correct heading level on re-render

tujoworker:


langz:
Seems like I still get the same issue, but I wouldn't say it's a "critical issue":


https://user-images.githubusercontent.com/1359205/228530178-7efeaf98-7306-41c0-997a-cdfbb56bf9f7.mov

tujoworker:
@langz Thanks for testing this out 🙏 I think the latest commit should "force" it to be right 😜

langz (approved):

PR #2170, opened 2023-03-29
feat(Icons): remove iOS (PDF) icons library support

tujoworker:
BREAKING CHANGE:

The support for icons with the PDF format got discontinued


tujoworker:
To make the xml generation work, I needed to remove `svg2vectordrawable/svgo` resolution. But that's fine, as we already serve optimized SVGs – svg2vectordrawable is using v2, but it does not need to pick up our config – witch is only compatible for our svgo deps version – v3.

langz (approved):

PR #2175, opened 2023-03-30
chore: make it possible to overwrite `failureThreshold` per test

tujoworker:


langz (approved):
I ran `yarn test:screenshots` 3 times, with the new `failureThreshold` locally on my machine, and all tests pass ✅

tujoworker:
yeah, its the new/old threshold. And I wonder, if we could go down even further. 
This PR is motivated by a icons library update, where the snapshots did not trigger a change, because of the to small changes in the comparison.

langz:
> yeah, its the new/old threshold. And I wonder, if we could go down even further. This PR is motivated by a icons library update, where the snapshots did not trigger a change, because of the to small changes in the comparison.

I can try out a few different lower thresholds tomorrow and report back 🔙

tujoworker:
@langz I made some more enhancements and could lower the threshold even more. You may check it locally as well. But I'm confident that it will run thought.

langz:
> > yeah, its the new/old threshold. And I wonder, if we could go down even further. This PR is motivated by a icons library update, where the snapshots did not trigger a change, because of the to small changes in the comparison.
> 
> I can try out a few different lower thresholds tomorrow and report back 🔙

I've tried out a `failureThreshold` of `0.001` and `0.005`, and both thresholds seems to work here.

tujoworker:
Yes. It pretty good now. We could probably run it on 0 as well. 
But we also should have a low retry amount. So we should look at the logs for some upcoming PRa and evaluate if we should do anything more about it.

PR #2176, opened 2023-03-30
chore(docs): remove unnecessary fragment in code examples

### LINKED ISSUE #2172: Should we remove the React.Fragments (<>) in our examples/docs?
## 🐛 Bug Report

Related to #2007

In v10 we display a lot more React.Fragments(`<>`) in our examples than what we did in v9:

Screenshot 2023-03-30 at 12 01 23

As part of v10  #2007 we refactored our examples from `.md` and `.jsx` to `.tsx`. In that process we changed our examples from:

```
<ComponentBox>
{`
<P>Hello World</P>
<P>Hello World</P>
`}
</ComponentBox>
```

To:

```
<ComponentBox>
   <P>Hello World</P>
   <P>Hello World</P>
</ComponentBox>
```

This leads to React(?) automatically wrapping the `<P>`'s in a React.Fragment `<>`, so the output of the example in our docs(LiveCode) will look like:

```
<>
   <P>Hello World</P>
   <P>Hello World</P>
</>
```

Previously in v9, it would look like:
```
<P>Hello World</P>
<P>Hello World</P>
```

Following is one of many many example in v10 today:
Screenshot 2023-03-30 at 12 01 23

## To Reproduce

Take a look at one of the many examples in v10s portal/docs.

## Expected behavior

Should we just keep it as it is in v10 or try to somehow fix this, to make it look more like it did in v9?
Following is a screenshot from v9:
Screenshot 2023-03-30 at 12 00 54

## Thoughts

I don't think it's a possible solution to wrap each example as done in v9, as we'll lose TS-functionality I believe:
```
  <ComponentBox>
    {
      /* jsx */ `
<P>Hello World</P>
`
    }
  </ComponentBox>
```

tujoworker:
closes #2172

Its fixed in this PR https://github.com/dnbexperience/babel-plugin-react-live/pull/5

### DIFF
--- a/packages/dnb-design-system-portal/package.json
+++ b/packages/dnb-design-system-portal/package.json
@@ -72,7 +72,7 @@
     "fs-extra": "10.0.0",
     "gatsby": "4.24.4",
     "gatsby-plugin-algolia": "0.26.0",
-    "gatsby-plugin-babel-react-live": "1.2.1",
+    "gatsby-plugin-babel-react-live": "1.3.0",
     "gatsby-plugin-catch-links": "4.24.0",
     "gatsby-plugin-emotion": "8.0.0",
     "gatsby-plugin-eufemia-theme-handler": "workspace:*",
--- a/yarn.lock
+++ b/yarn.lock
@@ -10868,10 +10868,10 @@ __metadata:
   languageName: node
   linkType: hard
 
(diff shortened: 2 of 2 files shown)


langz (approved):
That was fast, looks great, thanks 👏 🐎 🏎️

PR #2179, opened 2023-03-31
chore(Form): remove redundant form styling

joakbjerk:
Remove the form.scss stylesheet, since all it does is setting the default font-size and line-height that is already applied.

copilot:summary

copilot:walkthrough


tujoworker (approved):

langz (approved):
I agree, doesn't seem like this adds much/any value today 🥇 
Maybe we could also remove the `"dnb-form"` classnames from here: 
https://github.com/dnbexperience/eufemia/blob/v10/packages/dnb-design-system-portal/src/docs/uilib/components/form-label/Examples.tsx#L35-L40

PR #2181, opened 2023-03-31
chore: cancel CI runs on new git pushes

tujoworker:
I think we should try that for now. Because it does most of the time not make sense to continue runs when somethings in the branch changes. I have see that other CI environments, like Gatsby Cloud, do the same.


langz (approved):
Smart, I also think it makes sense to cancel existing/in progress builds when there's a new push 👍

PR #2183, opened 2023-03-31
fix(ToggleButton): fixes error when pressing enter

### LINKED ISSUE #1696: Pressing enter/return key to click/toggle a ToggleButton results in Uncaught TypeError: Cannot read properties of undefined (reading 'persist')
## 🐛 Bug Report

When pressing the enter/return key to click/toggle a ToggleButton, it results in the following error in the console:
`Uncaught TypeError: Cannot read properties of undefined (reading 'persist')`

## To Reproduce

- https://eufemia.dnb.no/uilib/components/toggle-button/#unchecked-togglebutton
- Tab once to focus the toggle button
- Press enter/return key
- Check the console

My setup:
MacBook Pro Version 12.6 (21G115)
Chrome Version 107.0.5304.87 (Official Build) (x86_64)

langz:
Quick fix as it seems like the `onClickHandler` expect an object with a property `event`.

Fixes #1696

### DIFF
--- a/packages/dnb-eufemia/src/components/toggle-button/ToggleButton.js
+++ b/packages/dnb-eufemia/src/components/toggle-button/ToggleButton.js
@@ -205,7 +205,7 @@ export default class ToggleButton extends React.PureComponent {
   onKeyDownHandler = (event) => {
     switch (keycode(event)) {
       case 'enter':
-        this.onClickHandler(event)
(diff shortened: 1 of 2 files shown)


tujoworker (approved):
Nice! Added a small test and could confirm the error exception happening 👍

PR #2192, opened 2023-04-11
chore: spelling previews -> previous

langz:


tujoworker (approved):
We should have a git bot that alerts when this word is written wrong 😆

PR #2199, opened 2023-04-12
docs(Space): refactor jsx examples to tsx

langz:


tujoworker (approved):
Nice 🚀

packages/dnb-design-system-portal/src/docs/uilib/components/space/Examples.tsx
```diff
@@ -440,3 +440,47 @@ VisualSpace.defaultProps = {
 }
 
 export { MagicBox, VisualSpace }
+
+export const SpaceExamplesSameResult1 = () => (
+  <ComponentBox hidePreview hideToolbar>
+    /** All of these methods will result in the same spacing */
```
tujoworker, line 446: These comments needs to be wrapped inside `{}` – not sure we should ignore this some how, because in this case, it will look not so natural. But I'm file with it, if you ask me.

tujoworker (approved):

packages/dnb-design-system-portal/src/docs/uilib/components/space/Examples.tsx
langz: I agree, however I just wrapped it inside braces for now, I still think it looks good enough:

image

PR #2200, opened 2023-04-12
feat: move DNB icons to subfolder (with v9 compatibility)

tujoworker:
This change should be 100% backwards compatible.

I have split up all the changes in different commits, so you may check out the relevant pats only.

## What are the overall changes?

1. Move JSX icon files into DNB subfolder.
2. Use tsx as file extension (no good reasons, but why not).
3. Create fallback index files, that exports the icons from the subfolder.
4. Refactor and expand the tests.


joakbjerk (approved):
Nice! Works on my machine 💻 🔥

tujoworker:
@langz yeah, it does not change how SVGs are fetched, just how they get converted. But you are right, we could add some docs about how I run it locally when making such a change. Even its more a "core" thing. 

I simply run `yarn workspace @dnb/eufemia dev:icons` to run the conversion from SVGs to JSX. Should we document that, do you think?

tujoworker:
@langz I'v [added](https://github.com/dnbexperience/eufemia/pull/2200/commits/901d94a3688135657969642062a488f2280c8b33) some info as mentioned. Does it look OK?

langz (approved):

PR #2201, opened 2023-04-12
chore: make it possible to include themes from elements and extensions

tujoworker:
By having this extra "comment" we now get themes inside elements and extensions included. 
The factory looks for the comment – and because it was missing, they (if some would exist) got not added.


joakbjerk (approved):
🍾

langz (approved):

PR #2202, opened 2023-04-12
docs(GlobalStatus): refactor jsx examples to tsx

langz:


packages/dnb-design-system-portal/src/docs/uilib/components/global-status/Examples.tsx
```diff
@@ -294,3 +295,85 @@ export const GlobalStatusUpdate = () => (
...
+
+      statusOne.update({
+        text: 'Updated Text',
+      })
+
+      // 3. and removes "custom-id-1" again if needed
+
+      statusOne.remove()
+
+      return <></> //ignore this line
```
langz, line 343: Not sure if there's a better way of "not returning any react components"/children to `ComponentBox`

tujoworker, line 343:
```suggestion
      return <GlobalStatus id="other-global-status" />
```
We could return the component like so. Even in theory, we would have to put `GlobalStatus.create` example coe in a useEffect or outside of the React tree.

tujoworker (approved):

packages/dnb-design-system-portal/src/docs/uilib/components/global-status/Examples.tsx
```diff
@@ -294,3 +295,85 @@ export const GlobalStatusUpdate = () => (
...
+export const GlobalStatusInfoExampleManipulate1 = () => (
+  <ComponentBox hidePreview hideToolbar>
+    {() => {
+      // 1. Update / extend the the status like so:
+
+      const statusOne = GlobalStatus.create({
+        id: 'other-global-status', // or main
+        status_id: 'custom-id-1',
+        text: 'New Text',
+        item: 'Item from status #1',
```
langz, line 330: I'm having a few TS-errors related to this example, and I'm not really sure myself if the examples are incorrect(using the wrong props and/or combination of props) or if the type definitions should be edited/updated to make these examples "legal". 

I don't know too much about these GlobalStatus functions, do you have a few minutes one day to take a look at this @tujoworker?

PR #2203, opened 2023-04-12
docs(Skeleton): refactor jsx examples to tsx

langz:


packages/dnb-design-system-portal/src/docs/uilib/components/skeleton/Examples.tsx
```diff
@@ -3,8 +3,10 @@
  *
  */
 
-import React from 'react'
+import React, { Suspense } from 'react'
+import { createBrowserHistory } from 'history'
 import ComponentBox from '../../../../shared/tags/ComponentBox'
+import Context from '@dnb/eufemia/src/shared/Context'
```
langz, line 9: I wasn't able to do `import Context from '@dnb/eufemia/shared/Context'`.
Not sure if we should be able to do so or not?
I do see multiple references to "non working/correct" jsx examples using `import Provider from '@dnb/eufemia/shared/Provider'`:
-  https://eufemia.dnb.no/uilib/usage/customisation/provider/
- https://eufemia.dnb.no/uilib/usage/customisation/localization/#localization

tujoworker: What happens if you do so?

tujoworker (approved):

tujoworker:
I hope its ok that I added an commit. I got a TS error in VSCode.

tujoworker:
ohh, I see, you pushed the same fix 😆

packages/dnb-design-system-portal/src/docs/uilib/components/skeleton/Examples.tsx
tujoworker: In docs we should always use it without `src`, while in code we always should include `src`. I think it should just work fine when we have a strict rule there. You screenshot looks like to have a wrong import path, in that sense, that it not includes `src`.

PR #2207, opened 2023-04-12
docs(InputMasked): refactor jsx examples to tsx

langz:


packages/dnb-design-system-portal/src/docs/uilib/components/input-masked/Examples.tsx
```diff
@@ -225,3 +225,68 @@ const Wrapper = styled.div`
...
+        value="1234.50"
+        on_change={({ numberValue }) => {
+          console.log(numberValue) // type of float
+        }}
+      />
+    </ComponentBox>
+  </Wrapper>
+)
+
+export const InputMaskedInfoDecimalsProvider = () => (
```
langz, line 259: I don't fully understand this example.
As the InputMasked has `as_currency="USD"`but I don't see any dollar sign 💵  behind the value 🤔 

TODO: I'll add a link to the example from the deploy preview here.

tujoworker: Yeah, its an "fictive" example. To make it work, you need to use just one of these mask variants: e.g. remove `number_mask`. Also using `use_currency` does not make any difference in this case. You may find a nice way to show the "Decimals" functionality.

langz: Thanks, I've split the example into to examples, one for currency and one for number 💯

PR #2212, opened 2023-04-12
chore(Elements): split visual identity styles

joakbjerk:
## Summary

Moving DnB specific styling to own theme files for all elements


packages/dnb-eufemia/src/elements/anchor/style/themes/dnb-anchor-theme-ui.scss
```diff
@@ -0,0 +1,44 @@
...
+  &:not(:active) {
+    color: var(--color-sea-green);
+  }
+}
+
+// Load anchor contrast again here, so that the green color set in color: var(--color-sea-green);
+// does not override the .dnb-anchor-contrast set in dnb-anchor.scss
+.dnb-anchor--contrast {
+  @include anchor-mixins.useAnchorContrastStyle();
+}
```
joakbjerk, line 44: This does not feel like the most optimal way to solve this, as the contrast is already set in the dnb-anchor.scss file. But if I did not set the dnb-anchor--contrast here, then all the DnB themed contrast anchors would recieve the --color-sea-green from the .dnb-anchor class at the top of this file.

Can you think of a better way of solving this @tujoworker  @langz 🙏

packages/dnb-eufemia/src/elements/img/Img.tsx
```diff
@@ -47,7 +47,7 @@ const Img = ({
         as="img"
         alt={alt}
         internalClass={classnames(
-          'dnb-img__img',
+          'dnb-img',
```
tujoworker, line 50: I think, it was on purpose to have `dnb-img__img` here, because the Element sets this class, based of `as="img"`.

See comment below.

packages/dnb-eufemia/src/elements/img/Img.tsx
```diff
@@ -47,7 +47,7 @@ const Img = ({
...
           hasError && 'dnb-img__img--error'
```
tujoworker, line 51: But maybe, we can remove it, and also change the modifier to dnb-img--error' instead?

```suggestion
          hasError && 'dnb-img--error'
```

tujoworker (approved):
Nice work 👌 Thank you 🙏

packages/dnb-eufemia/src/elements/anchor/style/themes/dnb-anchor-theme-ui.scss
tujoworker: I would like to argument like so; because the mixin `useAnchorContrastStyle` does contain almost 100% color changes, it should only be set in the Anchor theme file. What do you think about this?

packages/dnb-eufemia/src/elements/img/Img.tsx
joakbjerk: Hmm, I really have no Idea of what is best to do here 😄

packages/dnb-eufemia/src/elements/anchor/style/themes/dnb-anchor-theme-ui.scss
joakbjerk: Yeah that would work better actually 😃

packages/dnb-eufemia/src/elements/img/Img.tsx
tujoworker: I would suggest doing the latter. I can't find any good reasons to have a extra identifier. Its an internal style also.

PR #2216, opened 2023-04-13
docs(Space): refactor jsx properties to tsx

langz:


tujoworker (approved):

tujoworker:
Nice! If you wish, we can add the Provider as a default scope. Then we would not need to add it every time its needed.

langz:
> Nice! If you wish, we can add the Provider as a default scope. Then we would not need to add it every time its needed.

It's fine, doesn't bother me too much 💯

PR #2217, opened 2023-04-13
chore: sync type definition docs from props docs during build

tujoworker:
Check out the two first commits. 

The third one is simply after running: `yarn build` – sadly, we remove a good amount of newlines. But I have investigated this before and could not find any solution. 

Its because we convert the file to AST and back to code again. And sadly we then get it back without the extra newlines. Prettier does not put a newline back again.


langz (approved):
🤝

PR #2220, opened 2023-04-14
chore: change theme class name to `eufemia-theme`

tujoworker:
I think we need to distinguish this particular class naming.


joakbjerk:
> I think we need to distinguish this particular class naming.

I agree, It would lead to less confusion and make it easier to distinguish the base Eufima styling from the DnB specific things 🔥

langz (approved):
Looks good, a few failing tests 🛑

PR #2226, opened 2023-04-17
feat(PaymentCard): rename type Metalic to Metallic

langz:
I think `Metallic` is a an actual word, and not `Metalic`.

It's of course not a important breaking change, but perhaps a nice small improvement 🙏


joakbjerk:
> Nice! @joakbjerk could you double check if the new design does include a "breaking" change? Like if a card type is removed? Or the name is changed? I guess not.

Will take a look 🤙

joakbjerk (approved):

PR #2144, opened 2023-03-23
feat: simplify CSS packages/themes structure

tujoworker:
This changes where styles are placed and from what CSS package they will be consumed.

Its a larger change. But I think its worth it, because its much simpler and yet needed for our future theming capabilities.

BREAKING CHANGE:

- The package `dnb-theme-ui` was renamed to `dnb-theme-basis`.
- The package `dnb-ui-components` was renamed and moved inside a theme `/style/themes/theme-ui/dnb-theme-components.*`.
- `dnb-ui-tags` was renamed and moved form `/style/dnb-ui-tags.*` to `/style/themes/theme-ui/dnb-theme-tags.*`.

PR #2227, opened 2023-04-17
chore: refactor theme factory to include only existing theme files

tujoworker:
This PR adds a new feature to the theming setup. 
A theme can now add a fallback inside a SASS file:

```scss
$THEME_FALLBACK: 'ui';
```

The theme factory will then also add all non existing theme files to this file (component based), instead of importing a whole theme set as a fallback.

In return, this PR adds some breaking changes based on this PR #2144 

We need to amend/edit the git commit message as soon as possible with the list below:

- The package dnb-theme-ui was renamed to ui-theme-basis.
- The package dnb-ui-components was renamed and moved inside a theme /style/themes/theme-ui/ui-theme-components.*.
- dnb-ui-tags was renamed and moved form /style/dnb-ui-tags.* to /style/themes/theme-ui/ui-theme-tags.*.


langz (approved):
Ran it a few times locally, and it seems to work 👍 

I did find a few references to `dnb-theme` when searching in our codebase, in `06-css-packages.mdx`, etc.
Not sure if we should rename these to `ui-theme` as well.

joakbjerk (approved):
Seems to be working on my 💻 ✅

tujoworker:
@langz True! I found a few more. Pushed these as well now 🙏 (there are more with that name, but they belong to the JavaScript world.

PR #2237, opened 2023-04-19
feat(Typography): add `--line-height-lead` property

tujoworker:
This PR adds a new property, and replaces with that the used calc method. It does not change anything visually (except for the wip sb).

It also corrects wrong defined sb properties and uses the new lead property there as well.


joakbjerk (approved):
lookie good💪

PR #2239, opened 2023-04-20
fix(Button): align large size of "icon only button" to be same as in Figma

tujoworker:
This PR fixes a wrong sized icon button we have in our code. The size is different than a `large` button, when no text is given (40px vs 48px).

Screenshot 2023-04-20 at 10 16 11


langz (approved):
👍

tujoworker:
Another example of why visual tests are awesome 🔥

autocomplete-have-to-match-disabled-state snap-diff

Looks like this PR needs some more changes.

joakbjerk (approved):

PR #2240, opened 2023-04-20
fix(NumberFormat): hide screen reader only text from being copied as HTML

### LINKED ISSUE #2238: Copy from NumberFormat
## 🐛 Bug Report

Hi, when copying values from a number format and pasting into another application (e.g. outlook) the resulting formatting is bad. I assume this being a problem only in application where it is html that is pasted and not raw text - but ufortunately a case we find our users tries to do quite often.

## To Reproduce

1. Go to https://eufemia.dnb.no/uilib/components/number-format/demos/
2. Select the rendered demos and Ctrl-C:
image

3. Paste the copied content into e.g. Microsoft outlook and compare the result (all numbers are shown 3 times!)
image

## Expected behavior

I understand that outlook has a bad html rendering engine so we do not expact all styles to match perfectly - I would however expect each number only to be rendered / showed once.

```

## Eufemia Version
Latest

tujoworker:
This PR fixes #2238

- Tested with VoiceOver and NVDA that the hidden "screen reader only parts" still works as before.
- Rewrite of all the integration tests to find the attribute `data-text` instead of `textContent`.


joakbjerk (approved):
tested and compared to previous version of eufemia on chrome, looks good now📈

PR #2243, opened 2023-04-21
chore: split out sharable parts of tertiary button

tujoworker:
The interesting parts are in the first commit.
This can/will make it easier when theming for sb.

### DIFF
--- a/packages/dnb-eufemia/src/components/button/style/themes/dnb-button-theme-ui.scss
+++ b/packages/dnb-eufemia/src/components/button/style/themes/dnb-button-theme-ui.scss
@@ -6,15 +6,11 @@
 @import '../../../../style/core/utilities.scss';
 @import './button-mixins.scss';
 
+// Because the tertiary button has so much code,
+// we slice the shareable parts out in a separate file
+@import '../dnb-button--tertiary.scss';
+
 .dnb-button {
-  border: none;
-  box-shadow: none;
-
-  text-align: left;
-
-  // no animation yet
-  // transition: background-color 300ms var(--easing-default), box-shadow 300ms var(--easing-default);
-
   &--primary {
     color: var(--color-white);
     background-color: var(--color-sea-green);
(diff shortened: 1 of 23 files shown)


snorrekim:
If it's sharable then shouldn't it be imported in the dnb-button.scss file, and not the dnb-button-theme-ui.scss file? Or is the thought to make it more optional?

packages/dnb-eufemia/src/components/button/style/themes/dnb-button-theme-ui.scss
```diff
@@ -6,15 +6,11 @@
...
-  border: none;
-  box-shadow: none;
-
-  text-align: left;
-
-  // no animation yet
-  // transition: background-color 300ms var(--easing-default), box-shadow 300ms var(--easing-default);
+// Because the tertiary button has so much code,
+// we slice the shareable parts out in a separate file
+@import '../dnb-button--tertiary.scss';
```
snorrekim, line 11: Is the idea that we place sharable code in the `style` folder as optional imports for themes?

snorrekim (approved):

packages/dnb-eufemia/src/components/button/style/themes/dnb-button-theme-ui.scss
tujoworker: In this case, it would be optional to include by you, yes. But you also can more pars around so if you need to. 

It's just that these styles are so massive, compared to the other themes parts. 

Let me know if you have a better idea.

snorrekim: The only question is if these styles should just be placed in the dnb-button.scss file or not, if we're going to use them in every theme anyway. But that's something I can figure out as I continue working on the button. So please go ahead and merge it in.

snorrekim: Also, I will have some time to work on button again going forward, as the work for May 2nd are mostly done.

PR #2245, opened 2023-04-21
chore(Typography): create visual tests for all typography theming

joakbjerk:


tujoworker (approved):
It looks good to me. Even, I think, Eiendom has no changes to paragraph styles. So I would argue;

> we should not test what we don't change. 

Just because it creates parts we need to maintain in future.

langz (approved):

tujoworker:
The snyk CI seems to be hanging. Should I merge it, or do you want to re-run the CI?

joakbjerk:
> The snyk CI seems to be hanging. Should I merge it, or do you want to re-run the CI?
 
You can merge it if you want 🙏

PR #2249, opened 2023-04-24
chore: ensure SB theme elements gets imported

tujoworker:
## Test plan
Because we will run visual tests on these styles, we will ensure the existence of the import of elements in future. 

Update: I will add a postbuild test as well.


joakbjerk (approved):
🙏

Issue #2190, opened 2023-04-04
DatePicker submit_button_text


joakbjerk:
## 🐛 Bug Report

The submit_button_text props does not seem to be used at all in the DatePicker component, leaving developers unable to change the submit button text from OK

PR #2254, opened 2023-04-25
fix(DatePicker): fix text prop for submit and cancel buttons

joakbjerk:
## Summary

Solves [this issue](https://github.com/dnbexperience/eufemia/issues/2190) where developers were unable to change the submit and cancel button text

### DIFF
--- a/packages/dnb-eufemia/src/components/date-picker/DatePicker.js
+++ b/packages/dnb-eufemia/src/components/date-picker/DatePicker.js
@@ -509,6 +509,8 @@ export default class DatePicker extends React.PureComponent {
       mask_order,
       mask_placeholder,
       align_picker,
+      submit_button_text,
+      cancel_button_text,
       reset_button_text,
 
       hide_navigation: _hide_navigation, // eslint-disable-line
@@ -703,6 +705,8 @@ export default class DatePicker extends React.PureComponent {
                         onSubmit={this.onSubmitHandler}
                         onCancel={this.onCancelHandler}
                         onReset={this.onResetHandler}
+                        submitButtonText={submit_button_text}
+                        cancelButtonText={cancel_button_text}
                         resetButtonText={reset_button_text}
                       />
                     </>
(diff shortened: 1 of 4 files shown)


tujoworker (approved):
Looks good to me 👌

joakbjerk:
> Nice, this looks great 👏 Should we perhaps add tests for these "new" props as well? 🤔

Made tests for the custom text props @langz  @tujoworker please double check if they are done correctly 🙏

langz:
> > Nice, this looks great 👏 Should we perhaps add tests for these "new" props as well? 🤔
> 
> Made tests for the custom text props @langz @tujoworker please double check if they are done correctly 🙏

Thanks, looks great.
Only "improvement" or possible change I would consider is to have one render for each it statement, where we only include the props needed for the it statement.
Like for the it statement, `should show custom text for submit button`, I would consider only doing:
`  render(
    <Component
      opened
      label="Custom submit text"
      show_input
      submit_button_text="Yes"
      show_submit_button
    />
  )`
  
  Assuming these are the only props needed to test the it statement 🙏

langz (approved):

joakbjerk:
> > > Nice, this looks great 👏 Should we perhaps add tests for these "new" props as well? 🤔
> > 
> > 
> > Made tests for the custom text props @langz @tujoworker please double check if they are done correctly 🙏
> 
> Thanks, looks great 🚀 Only "improvement" or possible change I would consider is to have one render for each it statement, where we only include the props needed for the it statement. Like for the it statement, `should show custom text for submit button`, I would consider only doing:
> 
> ```
> render(
>     <Component
>       opened
>       label="Custom submit text"
>       show_input
>       submit_button_text="Yes"
>       show_submit_button
>     />
>   )
> ```
> 
> Assuming these are the only props needed to test the it statement 🙏
> 
> This is not important at all, just a suggestion as I would rather prefer smaller/easier to read "renders" containing only what's needed for the given it statement, instead of having one large render that covers many different tests/it statements.

Great suggestion, will change it now 🔥

PR #2260, opened 2023-04-26
chore: fix demo example

### LINKED ISSUE #2174: Fix Card complaint demo
## 🐛 Bug Report

Should we fix "Card complaint demo" before release of v10?
Are all the changes(marked in red) intentional, or not? 
If they are "breaking" for our demo app, it might be breaking for other apps that use eufemia as well 🤔 
I couldn't find info in `v10-info.mdx` about `Tabs` having breaking changes, but it will probably be generated documentation of this based on a commit containing "breaking change:" or so, I guess/hope.

image

tujoworker:
I could not see the issues anymore. I guess there was an issue with our CSS setup, which has been resolve in that case.

Closes #2174

This PR fixes this issue:

---

Screenshot 2023-04-26 at 09 51 04

---


langz (approved):
🏁

PR #2263, opened 2023-04-27
chore: refactor theme handling

tujoworker:
This PR enhance theme handling.

- The Eufemia Portal will load themed CSS always, and not inline. This would probably result in a lower google lighthouse score, just because a CSS file is requested. But I think that's fine for our needs. On the positive side, when switch the brand theme, only the new brand theme styles are loaded (as long as there is JavaScript), and not the default styles in addition – like before. 
- Also, its a step further for actually making this a custom NPM package (or a part of the main package), so a project could use it in prod.
- Use object in local storage instead of theme name only. This gives us more flexibility to fine-tune special needs.


joakbjerk (approved):
Nice 👍  Looks good too me 🍾

PR #2265, opened 2023-04-27
chore(Portal): add FilterByTheme as a global component

tujoworker:
With that, we can show demos/text based on the choose theme. So we can put e.g. this in a MDX file, without importing it:

```mdx

## Demos

<FilterByTheme name="sbanken">

## Sbanken related docs

Text and Demos

</FilterByTheme>

... more shared stuff

```

What component name would you else recommend? (instead of `FilterTheme`)

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/brand/development.mdx
+++ b/packages/dnb-design-system-portal/src/docs/brand/development.mdx
@@ -4,4 +4,18 @@ title: 'Sbanken Development'
 
 # Development with the Sbanken brand
 
-WIP
+## Write dedicated documentation
+
+If you need to show some documentation only for when the Sbanken theme is choose, you can do so:
+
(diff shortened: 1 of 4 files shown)


langz (approved):
Good good, I can't come up with a better name than `FilterTheme` as of now.

joakbjerk (approved):
✅ 

I can't think of anything right now either, `EufemiaTheme` or just `Theme` ?

tujoworker:
Would FilterByTheme or ShowByTheme be better?

joakbjerk:
> Would FilterByTheme or ShowByTheme be better?

Yeah, the intent is clearer to me then atleast 😊🤌

eliland (approved):
FilterByTheme sounds best to me :)

PR #2267, opened 2023-04-27
chore: add support for batch testing themes (visual)

tujoworker:
Instead of having a duplication of a visual test, we can use the jest "each" feature.

Some questions:

- Is this a good alternative? Or maybe any other alternatives?
- Is it a big disadvantage to have theme name not in the file name? 

Thanks @snorrekim for requesting a better solution than duplicating a file 🙏


joakbjerk (approved):
Looks good to me 😄 should be flexible enough to work with too, without having to create a test file per theme 👍

eliland (approved):

PR #2269, opened 2023-04-28
chore(InputMasked): on custom mask – avoid interaction stall after focus

tujoworker:
Add as `chore` as we release the original in v9 with PR #2270


joakbjerk (approved):
Looks good 🔥

PR #2270, opened 2023-04-28
fix(InputMasked): on custom mask – avoid interaction stall after focus

tujoworker:
The intention is to release this fix in v9.
The tests are added in PR #2269 to avoid git conflicts.


joakbjerk (approved):
Looks good too! 🙏

PR #2271, opened 2023-05-02
feat(Logo): Add brand and variant params for Sbanken logo

eliland:
## Summary

Added parameters to Logo component to enable it to show any of the three Sbanken logo variants. Defaults to the DNB logo as before, so changes should be non-breaking.

copilot:summary


tujoworker (approved):
thank you for the good work on getting these new logos in!

I think we should get this in as is and iterate:

- Rewrite to using React Hooks and TypeScript.

Because, how things where structured in the first place, was not an optimal starting point. 
The only thing I could se we do, is the same approach for batch testing visual tests, like in #2267

I hope its ok for you if I push a rebase+commit?

tujoworker:
In [addition](https://github.com/dnbexperience/eufemia/pull/2271/commits/d8f5b81b5efeeebf146dc38e7168979d47620227), we now have an example of the compact variant + visual test.

PR #2273, opened 2023-05-03
feat(Button): sbanken styling

snorrekim:
## Summary

The button component has been updated with the sbanken styling needed for now. Primary, secondary and tertiary button, as well as their large and icon button equivalents.


tujoworker:
We have two conflicting files at the moment. Hope its OK for you if I rebase ✨

tujoworker (approved):
Looks good to me 👌 Nice work 🙏

We may look into excluding signal button examples when this PR #2280 is in.

langz (approved):
Looks great 👏 

@tujoworker, do we have a strategy in place for when a theme doesn't have a certain `variant`? Like removing the props from the prop list and/or removing the examples?
Reason for saying this is because it seems like the sbanken themed button don't have a "signal" variant:
image

tujoworker:
@langz Yes, when PR #2280 is in, we can do that 👍

PR #2275, opened 2023-05-03
feat(Anchor): move from elements to components

tujoworker:
This PR moves the Anchor from /elements to /components.

BREAKING CHANGE:

Anchor was moved form `/elements` to `/components`


langz:
Looks like we should rebase with v10 🙏

langz (approved):
Looks good, 
Just need to export Anchor from `packages/dnb-eufemia/src/elements/index.ts` I believe 👍 

I do also see the following warning in the terminal when running(yarn start) this branch locally:

```
warn There are routes that match both page and redirect. Pages take precedence over redirects so the redirect will not work:
 - page: "/uilib/elements/anchor/" and redirect: "/uilib/elements/anchor" -> "/uilib/components/anchor"
 - page: "/uilib/elements/anchor/" and redirect: "/uilib/elements/anchor/" -> "/uilib/components/anchor"
```
 
 
 
image

langz:
Maybe we can also remove the `Events`-tab in the docs for Anchor, as it seems to not have any events:

image

PR #2280, opened 2023-05-04
feat(Theming): add `VisibilityByTheme` as a shared component

tujoworker:


snorrekim:
Insted of `FilterByThem.Exclude`, we could have the props `include` and `exclude` instead of  `name`

snorrekim:
Or perhaps `visible` and `hidden` would be better names, then there's no confusion about what "Filter" means, are we filtering out or in?

`<FilterByTheme name="sbanken">` doesn't really say if sbanken is filtered out or in.

tujoworker:
I think your suggestion is very good! For better TypeScript support, we should use arrays. So moved this component to be a part of Eufemia, added tests and docs 👍

packages/dnb-design-system-portal/src/docs/contribute/style-guides/documentation.mdx
```diff
@@ -13,13 +13,13 @@ The documentation is written in enhanced Markdown, called MDX. It allows us to i
 If you need to show some documentation only for when a certain theme is choose, you can do so:
```
langz, line 13:
```suggestion
If you need to show some documentation only for when a certain theme is chosen, you can do so:
```

langz, line 13:
```suggestion
If you need to show some documentation only for when a certain theme is selected, you can do so:
```

packages/dnb-design-system-portal/src/docs/uilib/components/button/info.mdx
```diff
@@ -33,10 +33,14 @@ It is **not** recommended to use the tertiary button without an icon. Looking fo
 
 <TertiaryButtonSizes />
 
+<VisibilityByTheme hidden="sbanken">
```
langz, line 36: Nice nice, we should remember to add this to the demos for variant signal as well 🙏 
Maybe in a upcoming PR

langz (approved):

packages/dnb-design-system-portal/src/docs/uilib/usage/customisation/theming.mdx
```diff
@@ -46,11 +46,48 @@ render(
...
+  </Theme>
+)
+```
+
 `<Theme>` Will create a `div` wrapper by default, when no custom element is defined (`element="span"`). It sets these CSS classes:
 
 - `eufemia-theme__{theme-name}`
 - `eufemia-theme__{theme-name}--{theme-variant}`
 
+###
```
langz, line 89:
```suggestion
```

langz (approved):

joakbjerk (approved):
Looks good to me 👍

PR #2285, opened 2023-05-05
chore: disable diff limit on postbuild test

tujoworker:
During the release of a beta, we get a [git error](https://github.com/dnbexperience/eufemia/actions/runs/4892791322/jobs/8734965454) on the postbuild step. This PR disables the limit. That results that the test takes about a minute locally. But we don't run this locally normally, and it only happens now that we have so many file changes from the v10 branch.

Screenshot 2023-05-05 at 13 45 59


langz (approved):
✅

PR #2290, opened 2023-05-05
feat(Anchor): add `scrollToHash` feature

### LINKED ISSUE #2286: Certain links in our documentation doesn't work in v10 anymore
## 🐛 Bug Report

Certain links in our docs doesn't seem to work properly anymore as part of v10.

One example is the following link to `/uilib/helpers/tools/#code-editor-extensions`:
Screenshot 2023-05-05 at 13 46 01

They do however work in v9 / what's deployed to https://eufemia.dnb.no/uilib/getting-started/

Initially I'm not sure what's causing this, hence me reporting it instead of fixing it.

tujoworker:
This PR/RFC is regarding Anchor animated scrolling.

Chromium/Blink based browser to not support smooth scrolling by default. And on macOS/iOS it's not supported at all.

Screenshot 2023-05-08 at 14 37 56

The Eufemia Portal did have code that makes it possible to get this feature. Its a relative small fix. And most of the code has existed as a Portal anchor link for several years already. 

The question is, should we get this in to the Anchor component? I guess yes. But then, how?

**Opt-in**

But this feature may be interesting for projects as well. But we may debate if it should not be available via a prop, be imported and spread to all anchors via a context.

**Opt-out**

On the other hand, it may be interesting to have this as an opt-out feature, as it solves and issue most are not aware of.

Closes #2286

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/components/anchor/properties.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/components/anchor/properties.mdx
@@ -11,6 +11,7 @@ showTabs: true
 | `to`                                        | _(optional)_ use this prop only if you are using a router Link component as the `element` that uses the `to` property to declare the navigation url. |
 | `target`                                    | _(optional)_ defines the opening method. Use `_blank` to open a new browser window/tab.                                                              |
 | `targetBlankTitle`                          | _(optional)_ the title shown as a tooltip when target is set to `_blank`.                                                                            |
+| `scrollToHash`                              | _(optional)_ When set to true, a click will attempt to use `window.scroll` to support the missing Chromium support.                                  |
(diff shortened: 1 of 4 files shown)


langz (approved):
Thanks, I like this :)

Do we know why this change is required to make the links work in v10 as they did in v9?

joakbjerk (approved):
🍾

tujoworker:
No, it did not work in v9 either. We used the [same approach](https://github.com/dnbexperience/eufemia/blob/main/packages/dnb-design-system-portal/src/shared/tags/Anchor.js#L47) there as well.

tujoworker:
But that's from our side. And I thought this was fixed by Chrome. But me and @joakbjerk could reproduce it a month ago. It needs still to be enabled with a flag. Bu its also missing in mac/iOS. 

Screenshot 2023-05-08 at 14 37 56

PR #2293, opened 2023-05-08
docs(v10): prop change for FormStatus of status to state

langz:
Here's the actual PR, https://github.com/dnbexperience/eufemia/pull/1914/files

Commit of https://github.com/dnbexperience/eufemia/pull/1914/files should perhaps have included "Breaking change" or so? 🤔


joakbjerk (approved):
✅

tujoworker (approved):
Yes, I agree, we should add it to the v10 docs 👍

PR #2294, opened 2023-05-08
feat(Dialog): adds ReactNode as type for texts

langz:
Reason for adding `React.ReactNode` as a type is because a lot of applications sends in a [<FormattedMessage/>](https://formatjs.io/docs/react-intl/components/#formattedmessage) as a text/translation:

image


joakbjerk (approved):
🔥

tujoworker (approved):

PR #2307, opened 2023-05-09
chore(Docs): mention deprecated style type prop paragraph

joakbjerk:


langz (approved):
Looks good to me, I can't remember we deprecated it though. It's been too long 🧠

PR #2308, opened 2023-05-09
chore(Autocomplete): update autocompleteinputref type

joakbjerk:
## Summary

This is just a proposal to solve the typescript error shown in the image below
image (4)
 
It might be to correct way to solve this, or it might not need to be solve at all. But I feel that this approach is more specific than using ` Record<string, unknown>`.
`
React.MutableRefObject<HTMLInputElement | undefined>` vs ` Record<string, unknown>`.


tujoworker:
I accidentally pressed the close button because it was on the same place – behind the approve button (at this screen height I had). Sooo000o0oOooO bad UX 😝

And I can't reopen it either.

Screenshot 2023-05-09 at 14 17 12

PR #2310, opened 2023-05-09
chore(Modal): add spacing props to modal and dialog header

joakbjerk:


langz (dismissed):
I think we should only add `SpacingProps` to `DialogHeaderProps`, and not `ModalHeaderProps`.
I wasn't able to see any difference in `Drawer` when adding bottom/top etc, so therefore I don't think we need the `SpacingProps` in `ModalHeaderProps` which is used in `DrawerHeaderProps`.

langz:
I updated this PR, just so that we'll be able to include it in https://github.com/dnbexperience/eufemia/pull/2317 and release it today 👍

PR #2315, opened 2023-05-09
chore: fix button types in import from /components/Button

tujoworker:
The problem:

Screenshot 2023-05-09 at 14 56 52

Where we can solve it:
Screenshot 2023-05-09 at 14 56 25

✅ How we can solve it:
Screenshot 2023-05-09 at 14 56 34


langz (approved):
🎸 👨‍🎤

PR #2318, opened 2023-05-09
feat(Anchor): Sbanken styling, added icon props

snorrekim:
## Summary

Sbanken will be using an icon either in front or after the link. Other use cases should work, but haven't been the focus of these changes.

I've added props for icons so that we have can have better control over the styling. These should work for dnb style as well


snorrekim:
I can try pushing a bump commit

snorrekim:
That worked, checks are now running

snorrekim:
Can one of you try to update the screenshot tests? Neither me or Espen seem to get correct results.

tujoworker (changes requested):
Very good work so far @snorrekim!

I think, we need to look into couple of things.

- [ ] Eufemia Icon properties (such as `iconLeft` and `iconRight`) do often handle icons as direct imports, so the user does not need to import the icon component in addition.
- [ ] Right now, the target blank icon is placed in the CSS, which is only so we can use the Anchor styles, without a component. But I think, we should do both, or rather only in the component. Because its more flexible in terms of colors. And also, we can use the `iconRight` property internally then.
- [ ] Change all component examples to use the properties `iconLeft` and `iconRight` instead of inline.
- [ ] Should we provide a visual example with a chevron and target blank? I guess you added it so we can test such a case?
- [ ] I tested the text wrapping behavior with the before/after (non-breakable-space) method vs padding and could not find any difference. Did you manage to compare this?
- [ ] Add integration tests for the new properties `iconLeft` and `iconRight` and how they behave.

I can contribute with all of these if that would be preferable?

snorrekim:
@tujoworker 
What exactly do you mean with "Eufemia Icon properties (such as iconLeft and iconRight) do often handle icons as direct imports, so the user does not need to import the icon component in addition." Are you thinking we should do the same thing as `<Button>`? Where the prop can also accept a string like `iconLeft="chevron_right"`?

As for the "_blank" links. The only thing new is that if you use `iconLeft` then the _blank icon will still appear, that was not possible before, any icon inside the link would hide the _blank icon, in fact, any html element (like a `<span>`) would hide it. The example I added was to show that the _blank icon could be replaced.

As for the before/after (non-breakable-space) method vs padding, the difference is the width of the underline, Sbanken styling requires that the underline is one space wider than the word. Though I might be open to completely changing the styling so we don't use text-underline, but a pseudo-underline with background gradient, similar to what we do with button and border using outline.

I'll change the icon props to accept strings as well, and adding tests for the new props. The rest I think should be a seperate task. This PR is about adding sbanken styling and adding two new optional props.

tujoworker:
Yes yes 🙌 sorry for the imprecise request of comments! In this PR should just add related things for sure.

snorrekim:
Ready for tests for review again, as long as someone runs the screenshot tests again

PR #2321, opened 2023-05-10
chore: fix VisibilityByTheme React hydration issue

tujoworker:
This fixes temporarily an React hydration issue seen in PR #2318 and #2319

PR #2326, opened 2023-05-10
chore: test React v18 on Portal

tujoworker:
**NB:** This is an early draft and we may close it as soon as we find out the the time scope may be to large. 

Because we face issues like in #2321 we may upgrade React to v18 on the Portal.

Known issues: 

- our type checks do fail
- visual tests run does fail


tujoworker:
The scope of work is too hight for goring forwards with an upgrade right now.

PR #2329, opened 2023-05-11
chore: run React hydration fix only on prod build

tujoworker:
Right now we get a React warning during development about using useEffect in an unmounted component. 
This PR should fix this.


langz (approved):
🚀

PR #2330, opened 2023-05-11
chore: show URL while running visual tests

tujoworker:
Not sure if that is useful. But sometimes, its good to quickly check how the page behaves during the slightly different behavior. Like no animations etc.

Screenshot 2023-05-11 at 10 42 52


joakbjerk (approved):
I like this idea 😃

langz (approved):

PR #2334, opened 2023-05-12
chore(HelpButton): make render example ts

langz:
Reason why I like this, is only because of TypeScript. So that if we change HelpButton or render prop in the future, it's easier to find necessary changes in our docs.


tujoworker (approved):
haha nice 🙏

PR #2357, opened 2023-05-19
chore(FormStatus): remove globalStauts prop from HTML attributes

tujoworker:
Reported as such in the CLI when running the tests:

> Warning: React does not recognize the `globalStatus` prop on a DOM element.


langz (approved):
💯

PR #2359, opened 2023-05-19
resolve conflicts with the `main` branch

tujoworker:
NB: The **ProgressIndicator** `huge` size is only for testing purposes.


tujoworker:
We have not. Its also for "manual" testing 😜 so maybe we remove it or add it to the prop table, but mention, its only for testing. What do you think? I mean, we probably don't need it.
But a larger size is interesting, because it could so easily be used as a bar like this:

Screenshot 2023-05-19 at 10 36 51

langz:
> We have not. Its also for "manual" testing 😜 so maybe we remove it or add it to the prop table, but mention, its only for testing. What do you think? I mean, we probably don't need it. But a larger size is interesting, because it could so easily be used as a bar like this:
> 
> Screenshot 2023-05-19 at 10 36 51

Let's ignore it, if we don't need it then 🤝

PR #2360, opened 2023-05-19
chore(DatePicker): close on reset button click

joakbjerk:


langz (approved):
Looks okay too me 👍 
Perhaps a "designer" or so could look at this and approve this approach.
I could see both options(closing and not closing) after clicking reset to be viable, so I'm not sure what's best 🤔 

If it's decided to close, we could add a unit test for testing that it closes after pressing reset.
We could use the test "will close the picker after selection" as example/starting point.

joakbjerk:
> Looks okay too me 👍 Perhaps a "designer" or so could look at this and approve this approach. I could see both options(closing and not closing) after clicking reset to be viable, so I'm not sure what's best 🤔
> 
> If it's decided to close, we could add a unit test for testing that it closes after pressing reset. We could use the test "will close the picker after selection" as example/starting point.

Yeah, there was a discussion about it the slack thread linked in the [jira task](lhttps://jira.tech.dnb.no/browse/EDS-413). Interpreted it as  having the`DatePicker` close on reset by default to be the desirable thing to have happen. Both the submit and cancel button works like that, so I guess that would make the reset button behavior be more in line with those buttons too. But I could be entirely wrong about this 😄

tujoworker (approved):
Yes, a test would be good to have 🙏

PR #2362, opened 2023-05-19
docs(ProgressIndicator): Update size properties

thaytharma:
Update size properties (again).
Added the 'default' size to docs, which is `0.5 rem`.

## Summary

We need this component with the `default` size in our app, so we tried playing around with the sizes. Looking at the [code](https://github.com/dnbexperience/eufemia/blob/main/packages/dnb-eufemia/src/components/progress-indicator/ProgressIndicator.js#L56), 'default' seems to be the default option for the `size`-prop. Take a look at this example: https://eufemia.dnb.no/uilib/components/progress-indicator/demos/#default-linear-progressindicator. 

copilot:summary

## Details

copilot:walkthrough


tujoworker:
Thank you for caring and creating this PR 🙏 If you don't mind (and don't need the update asap), I would recommend to make this change to v10 (#1832) instead. The docs have the file extension `.mdx` and each time we change something in main, we need to fix a conflict 😜

Let me know if you want to update the PR or letting me/us doing this.

langz:
Thanks, I'll make the change in the v10 branch 👍

langz:
This is now included in #2372

thaytharma:
Thank you @langz 😊

PR #2365, opened 2023-05-22
feat(Anchor): remove horizontal padding

tujoworker:
Ideally, we should avoid having invisible padding on components, so they perfectly align by and- or fall in to the "grid" as long as no spacing is given.

We have fixed this with the tertiary button a while ago, and now, its Anchor's turn.

The only down-side is the focus state/ring. We are as of now, not able to add a 2px fake padding there. But the designers are OK with that. It's also only an aesthetic question, and should now lower a11y very much.

**New**
Screenshot 2023-05-22 at 18 57 59

**Old**
Screenshot 2023-05-22 at 18 57 39


packages/dnb-eufemia/src/components/anchor/__tests__/__image_snapshots__/anchor-for-ui-have-to-match-breaking-lines.snap.png
joakbjerk, line 1: Looks like there's an embedded white space in front of bottom `Newline` in the Example.tsx. Has it always been like this, as it does not seem to have been introduced in this PR? And is it something we should "correct" or is there a reason behind this? 🤔

joakbjerk (approved):

packages/dnb-eufemia/src/components/anchor/__tests__/__image_snapshots__/anchor-for-ui-have-to-match-breaking-lines.snap.png
tujoworker: Yes, it has been there to show that there's still an underline. But I think, this rather confusing. Its also not a "real-world" example. Can remove it another PR i have up asap.

PR #2368, opened 2023-05-23
fix(Icon): enhance alignment when used in paragraphs

tujoworker:
Before:
Screenshot 2023-05-23 at 20 44 35

After:
Screenshot 2023-05-23 at 20 44 12


joakbjerk (approved):
💯

PR #2373, opened 2023-05-25
fix(Dialog): make action buttons responsive

joakbjerk:
## Summary

A proposal to make the action buttons inside the `Dialog` responsive. 
Not sure this is the best approach though 😄


packages/dnb-eufemia/src/components/dialog/style/dnb-dialog.scss
```diff
@@ -118,6 +118,15 @@
     > :not(:last-child) {
       margin-right: 1rem;
     }
+
+    @media screen and (max-width: 40rem) {
```
tujoworker, line 122:
```suggestion
    @include allBelow(small) {
```

packages/dnb-eufemia/src/components/dialog/style/dnb-dialog.scss
```diff
@@ -118,6 +118,15 @@
...
+      flex-flow: column;
+      gap: 0.5rem;
+
+      & > .dnb-button {
```
tujoworker, line 126: We should rather remove the spacing props from the buttons I would say. Is that possible?

tujoworker (changes requested):

packages/dnb-eufemia/src/components/dialog/style/dnb-dialog.scss
joakbjerk: Not sure if I'm answering your question here, but I'll give it a go 😄 
The buttons does not seem to be using the space props.

Screenshot 2023-05-25 at 13 41 33

It looks to be set by a `> :not(:last-child)` selector instead.

Would this solution be better?
Screenshot 2023-05-25 at 13 42 58

tujoworker: Yes, just remove all the margins and use flex row but with gap and also flex wrap.

tujoworker (approved):
Nice 👌 🙏

PR #2422, opened 2023-06-02
style(Button): remove tertiary button padding sbanken

snorrekim:
## Summary

Removed tertiary button padding because the intended use is often in line with text above or below the button. This will bring it in line with DNB tertiary button.


snorrekim:
A note, one of the screenshot tests was behaving weirdly because of the order the tests were run in (so i fixed it). If we are testing both parent and child component, then we need to test the parent component before the child, because the tests changes the css.

tujoworker (approved):
Nice, thanks you for the work 🎉

PR #2426, opened 2023-06-04
feat(Portal): copy icon name to clipboard

thaytharma:
## Summary
A suggested change that allows the user to copy the icon name to the clipboard by clicking on it.
This does not wrap the text element with the `<code>`-tag which would result in a styling change.

copilot:summary

## Details
Skjermbilde 2023-06-04 kl  07 10 13

copilot:walkthrough


langz (approved):
Thanks, looks like a great improvement 👏

tujoworker (approved):
Thank you for the nice improvement 🙏

I tested it with VoiceOver, and found no negative side-effects 🙌

PR #2427, opened 2023-06-04
Draft: feat(Portal): Copy color values to the clipboard

thaytharma:
## Summary

A suggested change that allows the user to copy the colors values from the color table to the clipboard by clicking on it.
Unfortunately, this not seem to work with the `td`-tag as parent.
Also tried using the `<Copy>`-tag which also does seem to work.

Let me know if you have any suggestions!

copilot:summary

## Details
Skjermbilde 2023-06-04 kl  07 33 26

copilot:walkthrough


thaytharma:
@tujoworker Do you have any suggestions on how to solve the issue mentioned in the description? 🤞🏼

tujoworker:
The Markdown support for tables is handled with the [remark-gfm](https://github.com/remarkjs/remark-gfm) dependency. It translates markdown tables, inkl. the inner markdown, to an AST and then via the [mdx](https://github.com/mdx-js/mdx) dependency to JSX->HTML.

But I can't find any options in `remark-gfm` to replace e.g. a `code` tag with a different element, so I pushed a different solution in a new commit.

tujoworker:
I'm not sure whats best. Right now we "mix" JSX with markdown: `<td>`#00343E`</td>` –  But I'm thinking if we should  rather rewrite the whole table to markdown? or maybe keep things as JSX, but then provide a prop (`copy`), instead of mixing? Or maybe its OK to mix?

tujoworker:
It would look like this:

Screenshot 2023-06-05 at 11 51 33

I think, its easier to "read" in code.

thaytharma:
> It would look like this:
> 
> Screenshot 2023-06-05 at 11 51 33
> I think, its easier to "read" in code. For sure, we would make this change in a another PR.

I agree, it looks better 🙌🏼

thaytharma:
Feel free to merge the PR @tujoworker

PR #2428, opened 2023-06-05
chore: removed fuzzy border in screenshots

snorrekim:
## Summary

Screenshots test randomly have fuzzy borders where the background grid is visible. Most likely due to sub-pixles. This can cause a lot of false positives. Adding a white box-shadow removes the fuzziness without changing the size of the screenshots.


snorrekim:
I also tried rounding up the width and height of the screenshot wrapper, but it had mixed results. It might make it more consistent than now, but the fuzzy borders still showed up.

snorrekim:
Input-have-to-match-input-with-clear-button-in-hover-state.snap-diff.png was the only screenshot with oddness that I could see, but it seems the new screenshot is the correct one, the old screenshot was wrong, but it was also so large that the error wasn't noticed.

snorrekim:
I added a small feature that lets us zoom images on keyboard focus. Makes it a bit easier to compare details.

PR #2441, opened 2023-06-07
chore: make ErrorHandler work with TypeScript

tujoworker:
- Don't use `new` when we expect a string, as it does not work with TS.
- So go purely with a function, I think is just as  fine.


langz (approved):
🐎

PR #2455, opened 2023-06-09
fix(Autocomplete): fix stretched form status when using suffix

langz:
## Summary

First time reported in Slack: https://dnb-it.slack.com/archives/CMXABCHEY/p1686124109235509

Using `status_props={{ stretch: true }}` in a Autocomplete which uses suffix results in a not fully stretched form status message.

## Details

Today, we calculate some kind of `maxWidth` for the `FormStatus`, based on ["the "linked mother"".](https://github.com/dnbexperience/eufemia/blob/f435633c73bb8d9e57c6663d8207fb95e0de00dc/packages/dnb-eufemia/src/components/form-status/FormStatus.js#L308)
For the `Autocomplete` with a `FormStatus`, it always uses the width of the `<input/>` to calculate the max width of the `FormStatus` which is applied in the `style` prop. 

This usually works perfectly fine, as the `<input/>` used/rendered in `Autocomplete` is "usually" as wide as the parent`Autocomplete` itself.

However when using `Autocomplete` with `suffix_value` the `<input/>` is NOT as wide as the `Autocomplete`, as it takes away space/room from the `<input />` to display the suffix value. Which of course results in a too small of a "stretched global status". See the image:

Screenshot 2023-06-09 at 13 57 22

Because of this, I've changed the `width_selector`(selector for which element to use for calculating the maxWidth of the FormStatus) of `FormStatus` in `Autocomplete` to be a different parent element, which includes the `<input/>` and the `suffix`. But only when there's a form status and it exists a suffix value for the Autocomplete. The reason for making it conditional as of now, is because I'm a bit unsure about the side-effects of doing this for all autocompletes.

Things to consider: 
- [x] Evaluate if this should still be conditional(as implemented now) or if we should always use the given `width_selector`.
- [x] Add a test?

Would be great if we could include this in release of v10.1 -  #2463

### DIFF
--- a/packages/dnb-eufemia/src/components/autocomplete/Autocomplete.js
+++ b/packages/dnb-eufemia/src/components/autocomplete/Autocomplete.js
@@ -1882,10 +1882,11 @@ class AutocompleteInstance extends React.PureComponent {
       this.context.drawerList.original_data
     )?.suffix_value
 
+    const innerId = suffixValue && showStatus ? `${id}-inner` : null
+
     // also used for code markup simulation
     validateDOMAttributes(null, mainParams)
     validateDOMAttributes(null, shellParams)
-
     return (
       <span {...mainParams}>
         {label && (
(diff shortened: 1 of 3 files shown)


tujoworker (approved):
I think, your thoughts a right 🙏

PR #2460, opened 2023-06-12
chore: removes enzyme snapshot tests

langz:
I've removed all snapshots that compared against snapshot files, and kept/rewritten inline snapshot tests to work with @testing-library


langz:
Not sure why the builds fail during installation of dependencies 🤔 
https://github.com/dnbexperience/eufemia/actions/runs/5254622346/jobs/9493466568?pr=2460

PR #2470, opened 2023-06-14
chore(tests): removes react-fake-props

langz:
TODO:
- [x] A lot of TypeScript errors now that our props(default props) is typed using `ComponentProps`. Most of it seems to be related to "ref".
- [x] Further try to remove unnecessary props(finding the needle in the haystack).
- [x] Investigate why DrawerProps type gives strange errors.
- [x] Fix postbuild test.


tujoworker (approved):
💪 Wow! What an achievement for the whole testing setup and its future 🎉

PR #2472, opened 2023-06-14
chore: fix existing "/build" check

tujoworker:
Locally, when running a prebuild, no package.json is generated. Therefore, we should check for an index file instead.


langz (approved):
🧙‍♂️

PR #2488, opened 2023-06-21
chore(Portal): fix type errors for react v18

joakbjerk:


packages/dnb-eufemia/src/components/modal/parts/ModalHeader.tsx
```diff
@@ -84,7 +88,7 @@ export default class ModalHeader extends React.PureComponent<
             {usedTitle}
           </h1>
         )}
-        <div>{children}</div>
+        <div>{children as React.ReactNode}</div>
```
joakbjerk, line 91: The casting is done since Section component only takes `ReactNode` as children and not `React.ReactNode | ((...args: any[]) => any)` as all the Modal parts do. I feel telling `Typescript` that children is `ReactNode` by casting it like this might be a little risky, as it also could be `(...args: any[]) => any)`.

Not sure how to best solve this, but would it make sense to allow Section children to be `React.ReactNode | ((...args: any[]) => any)` like the modal parts?

PR #2490, opened 2023-06-21
fix(Button): disabled DatePicker button

langz:
Reported: https://dnb-it.slack.com/archives/CMXABCHEY/p1687179613562739

From:
Screenshot 2023-06-21 at 13 40 21

To:
Screenshot 2023-06-21 at 13 39 45

@tujoworker, could you take a quick view/test of this PR to help evaluate if this fix has any unwanted results or not 🤔


tujoworker:
Nice! We should also add a (maybe hidden) visual test for this.

langz:
> Nice! We should also add a (maybe hidden) visual test for this.

Yeah, that would be nice.
I've now added a visual test, and updated a code to something I believe work 🙏

tujoworker (approved):
I think that's a perfect solution 💅

PR #2493, opened 2023-06-22
fix(InfoCard): set max-width on text content

tujoworker:
EDS-469


kimroen:
Suggestion: Perhaps `InfoCard` could support `stretch`? I'm not sure what would be involved in making that happen, so treat that as a very light suggestion 😄

joakbjerk (approved):
🔥

PR #2481, opened 2023-06-16
feat: add support for React v18

langz:
image

Tasks left:
- [x] Search for `React v18` in our code, it exists a few comments about what to do when we get v18, and do what's needed.
- [x] Evaluate or Fix Minified React errors ([JIRA](https://jira.tech.dnb.no/browse/EDS-476)) https://github.com/dnbexperience/eufemia/pull/2481#issuecomment-1601630753
- [x] Fix linting run in CI https://github.com/dnbexperience/eufemia/pull/2481#issuecomment-1601606984
- [x] Fix remaining TS errors in our stories.
- [x] Fix screenshot tests, it seems like screenshot tests are more unstable(?), now that we are using React v18.
- [x] Look into supporting React v17 as well, this is related to the "v18 changes" done in [TooltipPortal](https://github.com/dnbexperience/eufemia/blob/1e8021ccea7b03f9a60c78cc6760388ac36b55b1/packages/dnb-eufemia/src/components/tooltip/TooltipPortal.tsx#L8) which uses `react-dom/client` - @tujoworker 👀  #2511

PR #2504, opened 2023-06-27
chore: remove usage of `@reach/router`

tujoworker:
- removing one story and the deps
- part of PR #2481


langz (approved):
Bye bye @reach/router 👋👋

PR #2510, opened 2023-06-28
feat(Avatar): sbanken styling

snorrekim:
## Summary

Simple enough restyle: colors and fonts.


packages/dnb-eufemia/src/components/avatar/__tests__/__image_snapshots__/avatar-for-sbanken-src-have-to-match-png-image-of-local-src.snap.png
tujoworker, line 1: Maybe we could change the logo as well based on the brand theme?

tujoworker (approved):
Very good work 💫

packages/dnb-eufemia/src/components/avatar/__tests__/__image_snapshots__/avatar-for-sbanken-src-have-to-match-png-image-of-local-src.snap.png
snorrekim: Seems like the wrong kind of example to auto theme, the image needs to be selected by the user. We could make two different examples, but then we'd have to make two different tests.

PR #2511, opened 2023-07-03
feat(Tooltip): remove group property to add React v18 compatibility

tujoworker:
Not ideal to remove a prop after v10 release, but as it is relative harmless in terms, that the group feature is mostly to enhance a Tooltip with an position animation. The usage must be extremely low, or non existing.

This change is needed in PR #2481

Before:

https://github.com/dnbexperience/eufemia/assets/1501870/06e6828b-3cbf-40b1-a229-b12f962beb2d

After:

https://github.com/dnbexperience/eufemia/assets/1501870/8b68d308-c6b1-4501-b261-a5015f02486c


langz (approved):
🚀

PR #2512, opened 2023-07-03
feat(Button): sbanken tertiary icon button hover and focus

snorrekim:
## Summary

Tertiary icon button (dnb does not use this variant) for sbanken needs to have a hover and focus effect. The focus effect for sbanken tertiary button was also missing background and text color.


tujoworker (approved):
Looks good 💅

PR #2514, opened 2023-07-03
feat(ScrollView): add sbanken styling to scrollbar

joakbjerk:


tujoworker (changes requested):
I think we should move also:

```css
--scrollbar-track-width: 0.5rem;
--scrollbar-thumb-width: 0.5rem;
```

as it can be a property to change by themes as well.

tujoworker:
I think, we also can remove the properties from the portal sidebar menu, as they should be covered with this change. 

@joakbjerk Can you double check that?

joakbjerk:
> I think, we also can remove the properties from the portal sidebar menu, as they should be covered with this change.
> 
> @joakbjerk Can you double check that?

can do

PR #2520, opened 2023-07-07
fix: add external link icon to anchor with element as children

joakbjerk:
Proposal to make the external link icon show up inside `Anchor` if `children` is set to be an `Element`.

This change does however require developers to set the styling of the child to have some sort of `inline` display to not throw off/break the icon and tooltip alignment, but this might be a non issue. 

See examples provided in screenshot below
Screenshot 2023-07-07 at 14 49 00


packages/dnb-eufemia/src/components/anchor/__tests__/__image_snapshots__/anchor-legacy-icon-usage-for-ui-have-to-match-anchor-with-target-blank-legacy-icon.snap.png
joakbjerk, line 1: Allowing element as children whilst also showing the external link button seems to have changed the screenshot test, not sure if this change is acceptable?

tujoworker:
Nice! A jest test that confirms the changes would be good to have 🙏

tujoworker (approved):

PR #2521, opened 2023-07-07
feat(Anchor): focus-visible and test coverage

snorrekim:
### Screenshot 'active' simulation fix and better state (hover, active, focus) coverage
The current tests did not properly test hover, focus and active. I've fixed a bug in the screenshot 'active' simulation (it broke tests run after it) and extended our tests to cover states. This uncovered some bugs in sbanken active state and all themes' contrast anchor active, focus, hover states.

There was also a bug where `display: flex` changed the line-height of many of the examples and screenshots.

### Contrast link bug
The screenshot simulation fixes uncovered some bugs in contrast anchor active, focus, hover states. I've fixed those bugs, with the assumption that the style for contrast anchor is supposed to match what can be found in the footer at dnb.no

### :focus-visible
Refactored Anchor to use `:focus-visible`, this means we avoid many complex css specificity problems when theming.


tujoworker (approved):
Very nice work 🙏 Thanks you!

PR #2523, opened 2023-07-07
feat(Button): sbanken hover and error

snorrekim:
## Summary

- Updated the hover effect for sbanken buttons. 
- Updated sbanken error button styles and added screenshot tests for error states.

### Missing error button hover, active, focus styles?
The new error button screenshots uncovered that these buttons do not have their own style for hover, active, focus in DNB/Eiendom theme. They just look the same as a regular button when interacted with. I don't know if that is intended or not, but might be a bug. (Sbanken theme has styles for these states).


tujoworker (approved):
🔥 Great work, Snorre

PR #2524, opened 2023-07-07
chore(Portal): use Gatsby Slices for a couple of global components

tujoworker:
- Use Gatsby (v5) [Slices](https://www.gatsbyjs.com/docs/reference/built-in-components/gatsby-slice/) for faster builds. On CI we go from 3:30 to 2:30 for a Gatsby builds (warm). A local warm rebuild takes with this enhancement from 6 to 10 seconds on my Mac.
- Also, move "core" parts into "shared" so all the components are collected in one place.


tujoworker:
It looks like it depends more on the available cache, than an actually speed difference. Locally I could not find any difference. 
We do probably have way to little pages with an amount of around 300.

tujoworker:
It looks like it depends more on the available cache, than an actually speed difference. Locally I could not find any difference. 
We do probably have way to little pages with an amount of 300.

PR #2529, opened 2023-07-21
chore(Accordion): refactor to functional component and typescript

joakbjerk:


packages/dnb-eufemia/src/components/accordion/AccordionContent.tsx
```diff
@@ -15,14 +14,19 @@ import {
...
+export type AccordionContentProps = React.HTMLProps<HTMLElement> &
+  SpacingProps & {
+    instance?: React.LegacyRef<any>
+    className?: string
+    children?: React.ReactNode | ((...args: any[]) => any)
+  }
 
-export default function AccordionContent(props) {
-  const context = React.useContext(AccordionContext)
+export default function AccordionContent(props: any) {
```
langz, line 28: Could this be typed using `AccordionContentProps`?

```suggestion
export default function AccordionContent(props: `AccordionContentProps`) {
```

joakbjerk: Yeah, nice catch 😄 🙏

langz:
I've tried to compare the deploy preview of this PR https://eufemia-git-chore-refactor-accordion-class-component-eufemia.vercel.app/ with our latest release https://eufemia.dnb.no/

Here's an observation:
Left side is v10, right side is the PR

Screenshot 2023-08-01 at 13 49 41


I've marked with yellow what I've seen that's different.

langz:
I've tried to compare the deploy preview of this PR https://eufemia-git-chore-refactor-accordion-class-component-eufemia.vercel.app/ with our latest release https://eufemia.dnb.no/

Here's an observation:
Left side is v10, right side is the PR
I've marked with yellow what I've seen that's different

Screenshot 2023-08-01 at 13 50 37

Seems like different accordions is expanded, even though I believe the "example code" is the same.

langz:
Seems like the Accordion doesn't want to expand after a few "expands".

https://github.com/dnbexperience/eufemia/assets/1359205/21e2c7d7-b398-44e2-8a7e-cd11440613a3

joakbjerk:
> I've tried to compare the deploy preview of this PR https://eufemia-git-chore-refactor-accordion-class-component-eufemia.vercel.app/ with our latest release https://eufemia.dnb.no/
> 
> Here's an observation: Left side is v10, right side is the PR I've marked with yellow what I've seen that's different
> Screenshot 2023-08-01 at 13 49 41
> 
> Seems like there's a difference in if the second accordion is expanded or not, even though I believe the "example code" is the same. Not sure if this is intended or not.

I think this is caused by the `remember_state` property, if you opened it by accident, then it will remain opened even after refreshing, that's at least what I found when testing. So I don't think it has anything to to with the changes, but I might be wrong 😅

joakbjerk:
> Seems like the Accordion doesn't want to expand after a few "expands".
> Screen.Recording.2023-08-01.at.14.00.08.mov

Dang, I thought I had fixed that 😅

langz:
I just did a new round of testing, and it looks very very good now 👏 
The only "strange behavior" I found was that it seems like the enter button by the numlock keys(see what's marked in green), somehow doesn’t work when tabbing to the Accordion and wanting to expand it. It does however work here https://eufemia.dnb.no/uilib/components/accordion/#single-accordion

image

joakbjerk:
> I just did a new round of testing, and it looks very very good now 👏
> 
> The only "different behavior" I found compared to what's on https://eufemia.dnb.no/ was that it seems like the enter button by the numlock keys(see what's marked in green), somehow doesn’t work when tabbing to the Accordion and wanting to expand it. It does however work here https://eufemia.dnb.no/uilib/components/accordion/#single-accordion
> 
> image
> 
> EDIT: This is not a picture of my actual keyboard, I do have a more up-to-date keyboard 🤡

NIce catch 😅 

` const keyPressed = event.code

    if (keyPressed === 'Enter' || keyPressed === 'Space') {
      event.preventDefault()
      onClickHandler(event)
    }`
    
    Is most likely the culprit, so I'll change that 🙏

packages/dnb-eufemia/src/components/accordion/AccordionContent.tsx
```diff
@@ -157,7 +162,7 @@ export default function AccordionContent(props) {
   }
 
   if (!expanded || disabled) {
-    innerParams.disabled = true
+    innerParams['disabled'] = true
```
langz, line 165: Not sure if it makes sense to type `innerParams` instead of using the bracket notation here.
I get why we use it for `'aria-hidden'`, etc, but for disabled it feels more like a "type hack".

But if this is very difficult or doesn't make sense, just ignore this comment 🙏

langz (approved):
Great work, 
I've given it a new round of testing, and I was not able to find any "discrepancies" 💚

packages/dnb-eufemia/src/components/accordion/AccordionContent.tsx
joakbjerk: Yeah this was done to mostly get rid of the "`disabled does not exist on innerParams`" type error. But we can circumvent this by casting innerParams as `HtmlProps<HtmlElement>`

PR #2538, opened 2023-07-31
chore(StepIndicator): refactor to functional component and typescript

joakbjerk:


packages/dnb-design-system-portal/src/docs/uilib/components/step-indicator/properties.mdx
```diff
@@ -10,23 +10,28 @@ showTabs: true
...
 | `on_item_render`                            | _(optional)_ callback function to manipulate or wrap every item. Has to return a React Node. You receive an object you can use in your custom HOC `{ StepItem, element, attributes, props, context }`.                                                     |
+| `on_click`                                  | _(optional)_ Will be called once the user clicks on the current or another step. Will be emitted on every click. Returns an object `{ event, item, current_step, currentStep }`.                                                                           |
+| `on_change`                                 | _(optional)_ Will be called once the user visits actively a new step. Will be emitted only once. Returns an object `{ event, item, current_step, currentStep }`.                                                                                           |
```
langz, line 20: I see that we have and use both `current_step` and `currentStep`.
Do we need both? Or would just one suffice?

joakbjerk: I feel like sticking to `currentStep` would be the best, since it follows the camelCase format that we are going to be using when we rid ourselves of `Webcomponent` legacy code. The only reason I have left `current_step` there is because it might lead to a breaking change if we remove it 😅

joakbjerk: Maybe adding a deprecation warning to `current_step` in the callback function could be a first step in that direction in this PR?

langz: I see, makes sense 👍 
For this PR, to include it in v10, maybe we could do as @tujoworker did in https://github.com/dnbexperience/eufemia/pull/2585/files ?
Like add info about it in v11 docs, and add a lot of comments in our code that `current_step` will be deprecated as of v11?

Thoughts @tujoworker?

packages/dnb-eufemia/src/components/step-indicator/StepIndicatorContext.tsx
```diff
@@ -0,0 +1,295 @@
...
+import { onMediaQueryChange } from '../../shared/MediaQueryUtils'
+import {
+  StepIndicatorData,
+  StepIndicatorDataItem,
+  StepIndicatorMode,
+  StepIndicatorProps,
+} from './StepIndicator'
+import { StepIndicatorItemProps } from './StepIndicatorItem'
+
+// We use this array to filter out unwanted
```
langz, line 19:
```suggestion
// We use this array to filter out unwanted properties
```

packages/dnb-eufemia/src/components/step-indicator/StepIndicatorContext.tsx
```diff
@@ -0,0 +1,295 @@
...
+    <StepIndicatorContext.Provider value={contextValue}>
+      {props.children}
+    </StepIndicatorContext.Provider>
+  )
+}
+
+/**
+ * Like "Object.assign" – but safe
+ * A new falsy value will not be used, if it exists already
+ * Also,
```
langz, line 269:
```suggestion
```

langz (approved):
Gave it some testing, and it works like a charm ⭐ 
Great work in the "refactor to functional component and typescript assembly line" 🧑‍🏭 🔨 

We might consider adding a few comments about the "tobe deprecated" current_step property, to make it easier for us to remember to deprecate it in the upcoming v11 release ☺️

packages/dnb-design-system-portal/src/docs/uilib/components/step-indicator/properties.mdx
tujoworker: We could use the situation to integrated `withSnakeCaseProps`. With this we can rewrite the component to camelCase, but still support the old props. We use this a couple of places already (Slider and Tooltip). There is also a helper for types called `ToSnakeCasePartial`.
But should probably be done in a separate PR.

PR #2540, opened 2023-08-01
feat(Input): sbanken styling

snorrekim:
## Summary

Sbanken styling for input. Added screenshot tests for error state with button.


snorrekim:
I used css variables extensively to make the css as lean as possible. 

The biggest hurdle though, was that Input is too dependant on the Button component. There should be a larger degree of separation, the button used in an input does not have enough in common with the regular button. At most it could use some common variables for color. Either we should make it possible to completely strip the Button component of styling, or we should not use it here, and just use a regular <button> element

tujoworker:
I agree with you. I'll create a ticket for that.

tujoworker (approved):
Looks 100% good to me. Wonderful work 🚀

PR #2547, opened 2023-08-07
fix(Modal): removes attr data-dnb-modal-active from document.body

langz:


tujoworker (approved):

tujoworker:
Should we maybe log this PR as a "fix"? So its mentioned in the change log?

PR #2550, opened 2023-08-09
feat(Anchor): inline style sbanken

snorrekim:
## Summary

Added the class `.dnb-anchor--inline` for sbanken anxhors inside text lines. Needed to be accessibility compliant.


tujoworker (approved):
Nice work 🙏

PR #2552, opened 2023-08-09
feat(Dialog): add hideConfirm-prop

thaytharma:
## Summary

Add `hideConfirm`-prop to the `Dialog`-component.

With `hideConfirm` set to `false` (default behavior)
Skjermbilde 2023-08-09 kl  21 57 50

With `hideConfirm` set to `true`
Skjermbilde 2023-08-09 kl  21 56 57


tujoworker (approved):
Nice! Thank you 🙏

PR #2555, opened 2023-08-10
docs: add docs on how to link to a specific theme

### LINKED ISSUE #2374: How to link to docs of specific brand/theme?
## 🚀 Feature Proposal

I would like to link the docs of a given component in a given brand/theme to other designers and devs. Sending a link like: https://eufemia.dnb.no/uilib/components/badge/demos?theme=dnbeiendom
Let's say that a designer is wondering how the anchor looks for brand/theme DNB Eiendom, I would want to be able to link it to him.

As of today, the person receiving the link would have to know to right click the "three dots" in the top right corner and "click change style theme" and select the correct theme.

Potential possible ways of solving this:
- Using a query param?
- When in the portal, we could a query param to the URL when user changes to any theme other than the default theme, so that the user quickly can copy the URL displayed in his browser's address bar and share it with his friends.

tujoworker:
Closes #2374

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/contribute/style-guides/documentation.mdx
+++ b/packages/dnb-design-system-portal/src/docs/contribute/style-guides/documentation.mdx
@@ -1,7 +1,7 @@
 ---
 title: 'Documentation'
 order: 3
-status: 'wip'
+status: 'new'
 ---
 
 # Documentation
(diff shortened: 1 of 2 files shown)


langz (approved):
Aha, so it was already possible 🤔 
Thanks for adding the docs 🙇

PR #2577, opened 2023-08-22
portal: fix active sidebar menu items to only show one at a time

tujoworker:
It's almost perfect. But I think, its much better as before. Now we do not get the double active (current) items. 

Before:

Screenshot 2023-08-22 at 10 35 46

After:
Screenshot 2023-08-22 at 10 36 08

This also fixes a scroll issue, because it used the first active item as a reference "to be in view":

Before:

https://github.com/dnbexperience/eufemia/assets/1501870/c8bb5863-d7ab-4fb3-8242-61c089ed14b7

After:

https://github.com/dnbexperience/eufemia/assets/1501870/cf8f724f-ff6e-4ce0-8d2b-50dd4034a763

### DIFF
--- a/packages/dnb-design-system-portal/src/shared/menu/SidebarMenu.tsx
+++ b/packages/dnb-design-system-portal/src/shared/menu/SidebarMenu.tsx
@@ -61,6 +61,9 @@ export default function SidebarLayout({
               status
               icon
               showTabs
+              tabs {
+                key
+              }
               theme
             }
           }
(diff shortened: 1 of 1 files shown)


langz (approved):
Looks great 🟢 

This is just "another observation" of the sidemenu, and not directly related to the fix you've done here.
`Input` is highlighted/selected when pressing `InputMasked`:

https://github.com/dnbexperience/eufemia/assets/1359205/1a2563a7-afa5-4d23-a139-56a9f06760ec

PR #2580, opened 2023-08-22
feat(Forms): Field width features upgraded

henit:
## Summary

The field components (`Field.String`, `Field.Selection` etc) had a width prop that worked. However, when crearting custom components in applications using `FieldBlock`, there where no direct way of setting similar widths.

copilot:summary

## Details

The `FieldBlock` component now has two props for controlling width. The `width`-prop controls the width of the outer block element of the component, controlling how much space it takes up. Setting it to `large` or `medium` sets that block to the same width as for instance `Field.String` is by default, so field features inside it using the whole available width would have that same width. However, this limits the width of labels as well, so if the area where the component is placed has more space available, you might want the label to span wider, while the actual field component has that specific width. To achieve that, the `contentsWidth` prop can be used, which sets the given width on a wrapper around the actual field (children of `FieldBlock`) while not limiting the possible width of the label above or the error/warning/info message below. Both elements use 100% width if the props are omitted.

copilot:walkthrough

## Test plan

FieldBlock in the portal now has several examples with both `width` and `contentsWidth`, while the field-pages still have all possible widths as examples (such as `Field.String`).

Screenshot 2023-08-22 at 14 03 03

Screenshot 2023-08-22 at 14 03 13


packages/dnb-eufemia/src/extensions/forms/FieldBlock/FieldBlock.tsx
```diff
@@ -20,6 +20,10 @@ export type Props = ComponentProps &
   > & {
     forId?: string
     children: React.ReactNode
+    /** Width of outer block element */
+    width?: 'medium' | 'large'
```
langz, line 24: I see we have changed multiple types of `width?: false | 'medium' | 'large'` to be `width?: 'medium' | 'large'`. Should we do this for `width` property in `Number` and `String` as well?
Or should these two still support `false`?

packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber.tsx
```diff
@@ -14,7 +14,7 @@ export type Props = ComponentProps &
     countryCodeFieldClassName?: string
     numberFieldClassName?: string
     // Styling
-    width?: false | 'medium' | 'large'
+    width?: 'medium' | 'large'
```
langz, line 17: I see we have changed multiple types of `width?: false | 'medium' | 'large'` to be `width?: 'medium' | 'large'`. Should we do this for `width` property in `Number` and `String` as well?
Or should these two still support `false`?

langz (approved):

packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber.tsx
henit: I see this is a bit unintuitive. Here is the reason why `false` remained in `Field.String` and `Field.Number`: Some components have default widths when no props is given, to make them as "plug & play" as possible. Most text fields in design sketches seem to have the large-width for text inputs and the medium for numbers, so String and Number use `large` and `medium` when no `width`-prop is given to avoid the need to set `width` for most usecases. The result is that in this example, the first one, which has no `width`-prop set, is the same width as the one set to `large`:

Screenshot 2023-08-22 at 14 56 11

However, when these components are used inside compositions, like Postal code and city, the outer component need to be able to control the widths without "competing" with the default with. So I needed a way for `PostalCodeAndCity` to be able to say "skip setting the default width" to the internal render of `Field.String`, to produce this result:

Screenshot 2023-08-22 at 14 56 44

with the postal code set to specific widths in the css of this component. Maybe we can find a more intuitive way of supporting both feature-props (for implementations to override the defaults) and "internal control" like this in the future.

PR #2582, opened 2023-08-23
feat(Section): Sbanken styling and new variant prop

Sundfjord:
Added `variant` property to Section.
Added Sbanken styling.

PR #2585, opened 2023-08-23
docs(Section): move deprecated styles into a seperate docs page

tujoworker:
This is an addition to PR #2582

It's not given that this approach is the best solution. So let me know if we should change something about it or if there is a better approach.


packages/dnb-design-system-portal/src/docs/uilib/about-the-lib/releases/eufemia/v11-info.mdx
```diff
@@ -22,4 +23,18 @@ $ npm i @dnb/eufemia@11
 $ yarn add @dnb/eufemia@11
 ```
 
+## Section
+
+1. Most of the `type_style` where removed. Instead use the new `variant` property:
```
langz, line 28:
```suggestion
1. Most of the `style_type` variants were removed. Instead use the new `variant` property:
```

langz (approved):

PR #2592, opened 2023-08-24
chore: hide visual-test examples declaratively

tujoworker:


langz (approved):
Great work, I really like this approach 🥇 
I'm however not 100% sure what we could name this component.
When I read `VisualTest` for the 1st time, I assumed that a component would need to be wrapped in `<VisualTest>` to be able to visual test it. Which is not the case, as that would lead to much much code as we would almost wrap all of our examples in the `<VisualTest>` 😂 

Also, with the current naming `<VisualTest>` where I assume `<VisualTest>`'s main purpose is to only render children for visual testing, it feels a bit strange for me to provide `hideOnTest`(not render children for visual test) to a component whose sole purpose is to only render children for visual testing. It's almost like I would expect `<VisualTest hideOnTest>` to always be hidden(not rendered).

Here are just some suggestions / alternatives / food for thought:
1. `VisualTestOnly` and `DocsOnly`
`<VisualTestOnly><X/><VisualTestOnly>`
`<DocsOnly><X/><DocsOnly>`

2. `Hidden`
`<Hidden in="docs"><X/><Hidden>`
`<Hidden in="tests"><X/><Hidden>`

3. `Display`
`<Display in="docs"><X/><Display>`
`<Display in="tests"><X/><Display>`

4. `OnlyDisplay`
`<OnlyDisplay in="docs"><X/><OnlyDisplay>`
`<OnlyDisplay in="tests"><X/><OnlyDisplay>`

But I guess the naming sort of depends on what we "feel like we do" in these situations.
When rendering X only for visual tests - do we feel like we are "hiding X in docs" or only making sure that we "only display it in visual test". I feel like I think that I "only want do display x in visual test", so perhaps option 1 is the better of my many suggestions? 🤔 😂 

Any other suggestions is also very much welcome 🙏

tujoworker:
@langz Is maybe [VisibleWhenVisualTest](https://github.com/dnbexperience/eufemia/pull/2592/commits/447d8ae19ebb4572d3446721417b3141829acb6a) good?

PR #2595, opened 2023-08-25
feat(Checkbox): add innerRef prop to link a React Ref to the input element

joakbjerk:


packages/dnb-eufemia/src/components/checkbox/Checkbox.js
```diff
@@ -170,6 +171,13 @@ export default class Checkbox extends React.PureComponent {
     }
   }
 
+  initializeRefs = (ref) => {
+    this._refInput.current = ref
+    if (this.props.innerRef) {
+      this.props.innerRef.current = ref
```
langz, line 177: I'll let @tujoworker take a look at this PR, as I'm not 100% sure about refs and if `_refInput` can or will "collide" with `innerRef` in this situation.

How I read the code is that we'll replace `innerRef.current` with some other ref if `innerRef` exists.
So if I as a consumer provide a `innerRef` prop to this component, the component will actually replace parts of it, `innerRef.current`? 🤔

tujoworker (approved):
Nice 🚀

packages/dnb-eufemia/src/components/checkbox/Checkbox.js
tujoworker: yes, the internal ref gets replaced. But we now use a simpler method, and just set it in the constructor, like we don in a various other components already 👍

PR #2599, opened 2023-08-29
Feat/typography size and mobile sbanken

snorrekim:
## Summary

Some changes to the sizes of Sbanken headings. Sbanken also want headings to be smaller on mobile devices. 

Discussion:
- Do we have a standard breakpoint for "small devices"? Currently using 40rem (640px)
- Can we avoid `useTheme` in heading components?
- Should we reconsider the separation of the `<Headings>` and `<H1> <H2> ... etc` components? (causes duplicate/different screenshots and documentation)
- Should we define a clear relation between font-sizes and heading sizes? Currently a bit confusing. `xx-large` is only used by headings, but is presented as a general typography.


tujoworker (approved):
Looks good 👌 

I pushed a commit where we move `useTheme` to H as you suggest in the PR description.

PR #2610, opened 2023-09-06
fix(MaskedInput): make cleanedValue contain leading zeroes

joakbjerk:
More [info](https://dnb-it.slack.com/archives/G01RK37BMR8/p1693399952666769).


tujoworker (changes requested):
We need a test on this 😜🙏🏻

tujoworker (approved):

PR #2613, opened 2023-09-06
fix(Drawer): render Drawer.Navigation when hideCloseButton is true

tujoworker:
I'm actually not sure why we had this check to prevent it from rendering. Sadly, there was no comment in the code. Also, I can not find any tests proving this behavior. So in worst case, we "break" a layout by adding an additional spacing on top. 
If there is a need to prevent that, we should accommodate that with a more logical approach.


joakbjerk (approved):
🙏

PR #2617, opened 2023-09-06
feat(PaymentCard): update height and weight to reflect Figma sketches

langz:
## Summary

Updates height and weight to reflect the [latest/only Figma sketches we have of PaymentCard](https://www.figma.com/file/461cAN5Qc3Nks4ztZ9pjtM/Eufemia---Payment-Cards?mode=dev).

## History

- [Year 2020](https://github.com/dnbexperience/eufemia/pull/667) height was `53.98mm`≈`204.02px` and width was `85.6mm`≈`323.52px`. 
I can't find any Figma sketch documenting this, only in [our code](https://github.com/dnbexperience/eufemia/blob/main/packages/dnb-eufemia/src/extensions/payment-card/style/dnb-payment-card.scss#L58C1-L59). 
Maybe it never existed any Figma sketches of this?

- [Year 2023](https://github.com/dnbexperience/eufemia/pull/2264) there was a feature request to add more cards.
As part of this feature request, there was created a [Figma sketch](https://www.figma.com/file/461cAN5Qc3Nks4ztZ9pjtM/Eufemia---Payment-Cards?node-id=4%3A505&mode=dev)(only consisting the new cards, and not including compact variants). In this file the height was `216px` and width was `343px`.

## Motivation

As we link to the [Figma sketch](https://www.figma.com/file/461cAN5Qc3Nks4ztZ9pjtM/Eufemia---Payment-Cards?node-id=4%3A505&mode=dev) from our docs, and that other designers use the Figma sketch as a base/master for other designs/sketches, it only makes sense that the coded/implemented version of PaymentCard in Eufemia has the same height and width as what's in the Figma sketch.

## What's actually the correct height and width? 
The one implemented in year 2020, or the one implemented in the [sketches in 2023](https://www.figma.com/file/461cAN5Qc3Nks4ztZ9pjtM/Eufemia---Payment-Cards?node-id=4%3A505&mode=dev)(same as in this PR)?
I'm not sure, and I'm not sure if we'll ever know, but the coded version or the Figma version should be the same, and for now, since I don't know what's correct or not, the coded version gives in and follows the [latest/only Figma sketches](https://www.figma.com/file/461cAN5Qc3Nks4ztZ9pjtM/Eufemia---Payment-Cards?node-id=4%3A505&mode=dev) we have.

## Additional thoughts

Using the following [Figma sketch](https://www.figma.com/file/461cAN5Qc3Nks4ztZ9pjtM/Eufemia---Payment-Cards?node-id=4%3A505&mode=dev) as the main/root documentation for Figma feels a bit narrow and misleading compared to the coded/technical implementation, as the Figma sketch miss multiple cards and how they all look in the compact card variant(all part of the 2020 implementation). Was there ever a Figma sketch documenting what was implemented in 2020, which could be edited/updated as part of the feature request of 2023? If not, then I guess it's better to use the Figma sketches of the "feature request of 2023" as the main/root documentation instead of having none 🤷 

## Technical thoughts
I considered using "style" decorator, but did not do it since _they won't be included in the [releases change log](https://github.com/dnbexperience/eufemia/releases)_ - https://eufemia.dnb.no/contribute/style-guides/git/#commit-messages


tujoworker (approved):
Nice Work 👌

Frikki:
The 2020 millimeter sizes are correctly modeled from a physical card. This PR’s millimeter sizes don’t reflect reality.

langz:
> The 2020 millimeter sizes are correctly modeled from a physical card. This PR’s millimeter sizes don’t reflect reality.

Aha, that's of course the reason for doing the sizing in mm, thanks 👍 
I'll try to check again with the designer what he wants or not.

Frikki:
@langz Yes, but depending on a device’s pixel density (PPI), such a physical measurement translates differently on each device. We can’t just say that 85.6 mm is 323.52 px. It doesn’t make sense when we specify styles for device screens.

langz:
> @langz Yes, but depending on a device’s pixel density (PPI), such a physical measurement translates differently on each device. We can’t just say that 85.6 mm is 323.52 px. It doesn’t make sense when we specify styles for device screens.

Thanks for feedback :)

So what you are saying is that the existing unit(`mm`) used for expressing length in `PaymentCard` is inherently wrong for web. So to improve upon the existing code, we should change to use a different unit?

As part of https://github.com/dnbexperience/eufemia/pull/2617/commits/d8b0e18ea9f1804e1769e5a05ae0f3f5f1b24697 I've now changed to use `px`, and aligned it with the values used in the [Figma design](https://www.figma.com/file/461cAN5Qc3Nks4ztZ9pjtM/Eufemia---Payment-Cards?node-id=4%3A505&mode=dev).

PR #2620, opened 2023-09-07
chore(Portal): add env variable "filter=/component-name" for faster dev builds

tujoworker:
This PR:

- adds the possibility to start the portal with only some pages e.g.: `filter=/button yarn start`
- and it speeds up our e2e and visual tests by simply not building some docs pages.

The CI speed improvement is for sure only in the "build" step. There we safe 50% of the duration. We go from 16 min to 8 min. Also because the bundle is smaller, the upload and the download process of the artifacts is faster.

### DIFF
--- a/packages/dnb-design-system-portal/gatsby-config.js
+++ b/packages/dnb-design-system-portal/gatsby-config.js
@@ -166,18 +166,18 @@ if (
   !global.pagesPath.includes('_dummy')
 ) {
   const queries = require('./src/uilib/search/searchQuery')
-  if (queries) {
-    plugins.push({
-      resolve: 'gatsby-plugin-algolia',
-      options: {
-        appId: process.env.ALGOLIA_APP_ID,
-        apiKey: process.env.ALGOLIA_API_KEY,
-        indexName: process.env.ALGOLIA_INDEX_NAME, // for all queries
-        queries,
-        chunkSize: 10000, // default: 1000
-      },
-    })
-  }
+  // if (queries) {
+  //   plugins.push({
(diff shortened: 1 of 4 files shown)


packages/dnb-design-system-portal/gatsby-node.js
```diff
@@ -118,6 +118,69 @@ exports.createPages = async (params) => {
   await createRedirects(params)
 }
 
+exports.onPostBuild = async (params) => {
+  await createRedirects(params)
+
+  if (deletedPages.length) {
+    params.reporter.warn(
+      `❗️ These pages where deleted:\n${deletedPages
```
langz, line 126:
```suggestion
      `❗️ These pages were deleted:\n${deletedPages
```

packages/dnb-design-system-portal/gatsby-config.js
```diff
@@ -166,18 +166,18 @@ if (
...
+  //     resolve: 'gatsby-plugin-algolia',
+  //     options: {
+  //       appId: process.env.ALGOLIA_APP_ID,
+  //       apiKey: process.env.ALGOLIA_API_KEY,
+  //       indexName: process.env.ALGOLIA_INDEX_NAME, // for all queries
+  //       queries,
+  //       chunkSize: 10000, // default: 1000
+  //     },
+  //   })
+  // }
```
langz, line 180: Should we leave it commented like this, include it, or remove it? 🤔

langz (approved):
Nice 🐎 
This will make things faster ⚡

PR #2623, opened 2023-09-07
fix(ProgressIndicator): ensure correct positioning inside its boundary

tujoworker:
Re-prod: https://codesandbox.io/s/busy-haibt-fwyzqg?file=/src/App.js

Because of the `position: absolute` used in the `svg` elements. When using special app CSS like `text-align: center`, we need to ensure the position gets not influenced/changed.


langz (approved):
❤️ 🚀

PR #2624, opened 2023-09-07
chore(Portal): fix linting, enable FAST_DEV and enable Algolia seach indexing

tujoworker:
`FAST_DEV` speeds up the cold startup-time of the portal. With that we reduce the total amount of time to around 50%.


langz (approved):
🎸

PR #2625, opened 2023-09-07
chore(Modal): safeguarding focusing element

langz:
When debugging/using the Modal, I often get the following error/warning in the console:

```
Eufemia TypeError: Cannot read properties of null (reading 'focus')
    at ModalContent.tsx:259:1
```

This PR tries to safeguard the usage of .focus(), so that we don't see the error/warning all the time


tujoworker (approved):

tujoworker:
Absolutely! 🔏🧨

PR #2626, opened 2023-09-08
feat(Dialog): sbanken theme

snorrekim:
## Summary

sbanken theming of dialog

#### Undefined cssVariables

How do we define cssVariables theme variables when there is no default value (because the value is theme specific)?

In this PR there are several variables in the common css that have no value, but as we can see, it drastically simplifies the theming files.
```
  --dialog-background
  --dialog-icon-color--info
  --dialog-icon-color--warning
  --dialog-icon-background--info
  --dialog-icon-background--warning
```

#### "Protected" cssVariables

In dnb-icon I've "protected" the default value whilst still making it available for customisation, again this leaves an undefined cssVariable `--icon-border-positioning--default` that can be used to override the variable.

```
  --icon-border-positioning: var(--icon-border-positioning--default, -25%);
```

This was a way to protect it from failing if the variable is set to something that does not exist. But the Icon code does not need it itself. Should components make such allowances for others to use, or should they be more self contained?


packages/dnb-eufemia/src/components/dialog/__tests__/__image_snapshots__/dialog-for-sbanken-have-to-match-the-dialog-progressindicator-window.snap.png
tujoworker, line 1: I wonder if we should rebase this branch/PR as the color of this indicator should have Sbanken color now.

tujoworker (approved):
Nice work 👌 good if you have a double-check regarding the progress indicator color.

packages/dnb-eufemia/src/components/dialog/__tests__/__image_snapshots__/dialog-for-sbanken-have-to-match-the-dialog-progressindicator-window.snap.png
snorrekim: I'll make sure to rebase before I merge

PR #2645, opened 2023-09-13
feat(Forms): merge FieldGroup into FieldBlock

henit:
FieldGroup and FieldBlock has been merged into one component named FieldBlock. In addition, I added more example code and description in the Create your own component section.


packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/create-component.mdx
```diff
@@ -9,28 +9,43 @@ breadcrumb:
...
+
+In the example above, you see how you can create your own user input functionality in a standardized context using `FieldBlock`. This allows you to display labels, error messages and other surrounding elements in a consistent manner with the ready-made fields found in Eufemia Forms.
+
+Remember that everything that happens by itself using `useDataValue` and the rest of the available helper functionality, you can override individually to make the component work exactly as you want. If, for example, you need to carry out your own custom validation and cannot use the built-in validation with JSON Schema or by sending in a derivative validator as is done in the example above, you can write your own logic for this, and send the result in as props to `FieldBlock` in the form of `error`. All direct props override standard handling, so you have full control over your component. If you need something that looks even more different than the usual fields, you can drop `FieldBlock` and display surrounding elements in other ways if your project requires it, but still get all the help of data flow logic like `useDataValue` and other help functions in Eufemia Forms offer.
```
tujoworker, line 44:
```suggestion
Remember that everything that happens by using `useDataValue` and the rest of the available helper functionality, you can override the behaviour individually to make the component work exactly as you want. 

If – for example; you need to carry out your own custom validation and cannot use the built-in validation with a JSON Schema or by sending in a derivative validator (as is done in the example above), you can write your own logic for it, and send the result in as props to `FieldBlock` in the form of `error`. All direct props override standard handling, so you have full control over your component. 

If you need something that looks even more different than the usual fields, you can drop `FieldBlock` and display surrounding elements in other ways – but still get all the help of a data flow logic, such as `useDataValue` offers.
```

tujoworker (approved):
Nice merge! We need tests for the FieldBlock. It can be added later for sure. But it would be best to get it in right away.

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Examples.tsx
```diff
@@ -1,51 +1,152 @@
...
+                  required
+                />
+                <Layout.FlexItem width="large">
+                  <Slider
+                    min={1900}
+                    max={new Date().getFullYear()}
+                    step={1}
+                    label="Birth year"
+                    label_direction="vertical"
+                    // @ts-ignore
```
langz, line 124:
```suggestion
```

I'm curious to if we could remove this line, and somehow make TypeScript happy instead? 🤔 
Reason for me thinking this is because I assume most "consumers" use TS, and they would most likely end up doing the same thing as done in the example. Would be great if we could provide a better alternative then @ts-ignore if possible.
But then again, not everyone use TS, so maybe it's not good to fill our examples with a lot of TS specific code 🤔 

Just thoughts 💭

langz (approved):

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Examples.tsx
tujoworker: Right now, TS is not supported. But this [PR](https://github.com/dnbexperience/eufemia/pull/2648) will add support for it.

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/create-component.mdx
```diff
@@ -9,28 +9,43 @@ breadcrumb:
...
+  Iterate,
   Value,
   ValueBlock,
   Visibility,
   useDataValue,
 } from '@dnb/eufemia/extensions/forms'
 ```
 
-Here is a first example on how to compose your own component:
+Here is an example of a custom component. Notice how the props received by your field component are passed through the `useDataValue` hook. This hook does not change the API of the props, so the props returned by the hook share the same typescript type with which it was called. However, it adds a few additional properties to simplify the standardization of field behavior. This in the form of the handler functions `handleFocus`, `handleChange` and `handleBlur`. Even if field components externally have these callback functions named with "on" (eg "onChange"), these will remain untouched, while the "handle" variants add handling that saves you a lot of extra work.
```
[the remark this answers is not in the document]
henit: I was trying to explain it by saying that the "on*" are untouched, and the component code need to call "handle*" to trigger tue `useDataValue` features (like connecting to DataContext and handling error state). More detailed explanations can be found in the code comments in `useDataValue.ts`

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Examples.tsx
henit: It is the example code that does not support TS. In consumer code for this, the type issue for the property that is ts-ignored in that line, could be avoided using a generic in the `useDataValue`-call:
```
const birthYear = useDataValue<number>()
```
but that does not work in the example code since it doesn't currently support typescript syntax. In JS-consumers, this would not matter at all. TS just ensure that the source code, as well as implementations using TS are more secure to avoid issues related to types.

PR #2653, opened 2023-09-13
feat(Theming): refactor the internal structure + align DNB Eiendom theme and Anchor styles

snorrekim:
## Summary

We want to clean up all the different theming structures that have been used into one.

* removed all customization sass mixins
* added and fixed eiendom anchor theme (it didn't work before)
* eiendom and sbanken now has the same theme structure
* {theme}-basis.scss files no longer have code, they only import so it's easier to control what each import does
* theme-mapping seperated from theme properties
* global.scss contains all the "extra" code that isn't part of a structure (this would also be a good place to look for future improvements)


packages/dnb-eufemia/src/style/themes/theme-eiendom/properties.js
tujoworker, line 1: I have double-checked with the devs. They do not use this file in DNB Eiendom. ✅

tujoworker (approved):
What a nice cleanup 👌 Thank you so much for the good afford and work 💪🙏

packages/dnb-eufemia/src/style/themes/theme-eiendom/properties.js
snorrekim: Still, I wonder if all "inherited"-from-ui properties should be available in each theme's properties.js, that way they wouldn't need to import both ui and sbanken properties files if they need both. And that way we could also control `theme-mapping` in the properties file.

Right now, if you use say, the `--font-family-default` property in your react app, you will need both the ui and sbanken properties.js files, but it would deliver the wrong result because we actually map that to something else in sbanken scss.

I'll take a second look and see if we can make the theme mapping a part of each properties.js file.

In fact, as I think on it, if you go from say, sbanken, to eiendom, then you would also lose all the sbanken properties... should the properties file actually be a single file that delivers different values depending on the theme?

PR #2658, opened 2023-09-14
feat(Theme): generate all properties in javascript file

snorrekim:
A suggestion for how to perhaps make sure all properties are available in each theme's `properties.js` file. To avoid potential errors, both crashes and wrong colors.


tujoworker (approved):
Very nice 👌

tujoworker:
It does not really have an effect. But in the test file `makePropertiesFile.test.ts` we still look for `properties.scss` – and this also exists, so the tests do not fail. But maybe we could change it there as well to `properties-js.scss`?

PR #2660, opened 2023-09-15
feat(Forms): expiry field

joakbjerk:
- [x] Add tests
- [x] Add type documentation
- [x] Add portal documentation
- [x] Fine tune toggle functionality 
- [x] Feedback to user / changes to input behavior
- [x] implement `useDataValue` hook and `SharedContext`


packages/dnb-eufemia/src/extensions/forms/Field/Expiry.tsx
```diff
@@ -0,0 +1,256 @@
...
+      on_blur={handleBlur}
+      on_focus={handleFocus}
+      suffix={
+        help ? (
+          <HelpButton title={help.title}>{help.contents}</HelpButton>
+        ) : undefined
+      }
+      input_element={
+        <span className="dnb-date-picker__input__wrapper">
+          <ExpiryDateField
```
henit, line 98: Wouldn't it be possible to achieve the UI in a simpler way using InputMasked?

packages/dnb-eufemia/src/extensions/forms/Field/Expiry.tsx
```diff
@@ -0,0 +1,256 @@
...
+    id: propsId,
+    className,
+    label,
+    error,
+    help,
+    disabled,
+    handleFocus,
+    handleBlur,
+    handleChange,
+  } = useDataValue(props)
```
henit, line 54: Is there any reason for not supporting a value prop? I guess most use cases will have this as an empty input for the user to fill in, but what if for instance the data is saved in local storage, so refreshing the page should have the field filled in because the user is in the process of filling in the form? The architecture of the other field components are created around the pattern that the external value of a field can be received through the value prop, the same type of value are sent out as first argument of onChange, as well as connecting to surrounding contexts using the useDataValue hook.

packages/dnb-eufemia/src/extensions/forms/Field/Expiry.tsx
```diff
@@ -0,0 +1,256 @@
...
+type ExpiryProps = FieldHelpProps &
+  Omit<FieldProps<string>, 'onChange'> & {
+    /**
+     * Fires when input is fully filled out. Has an object as parameter, consisting of `month` and `year` values.
+     */
+    onChange?: ({ month, year }: ExpiryValue) => void
+  }
+
+function Expiry({ ...props }: ExpiryProps) {
+  const [month, setMonth] = useState<string>('')
```
henit, line 34: There might be requirements in the feature of this component that I have not understood the technical implications of yet, but the reusable features of useDataState should make it unnecessary to have local state in the component, and instead just call updates on the value-state held inside the hook. Since this field has two "sub-values" (unless an input masked component can be used instead), the update would need to merge the existing value-object with the sub value (like putting the year onto the existing value-object) which should work and lead to less code and skip the need for the useEffect reaction

packages/dnb-eufemia/src/extensions/forms/Field/Expiry.tsx
```diff
@@ -0,0 +1,256 @@
...
+      id={`${id}__input`}
+      className={classnames(
+        'dnb-date-picker',
+        'dnb-date-picker--show-input',
+        'dnb-forms-field-expiry',
+        className
+      )}
+      label={label ?? sharedContext?.translation.Forms.dateLabel}
+      disabled={disabled}
+      status={error?.message}
```
henit, line 88: Sending only the error message as status limits features that all other field components have, like supporting info and warning props. If the inner component is wrapped in FieldBlock, these props can just be passed through to that component.

packages/dnb-eufemia/src/extensions/forms/Field/Expiry.tsx
```diff
@@ -0,0 +1,256 @@
...
+      })
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [month, year])
+
+  return (
+    <Input
+      id={`${id}__input`}
+      className={classnames(
+        'dnb-date-picker',
```
henit, line 81: Is this copy-paste? Does not seem to be the correct classes for an expiry-component..

packages/dnb-eufemia/src/extensions/forms/Field/Expiry.tsx
henit: The `Field.String` component support mask-prop by the way, leading to the inner implementation using InputMasked instead of Input. Maybe that can be used, and then just wrapped with a parser that "translates" from the "UI value" (like "01/01") to the external value of the Expiry component (the object structure with the two values separate)? That can probably be done in the `fromInput` / `toInput` derivates, getting all the other features automatically (like the distribution of props to FieldBlock etc inside `Field.String`

packages/dnb-eufemia/src/extensions/forms/Field/Expiry.tsx
```diff
@@ -0,0 +1,256 @@
...
+   */
+  month: string
+  /**
+   * Year value from input
+   */
+  year: string
+}
+
+type ExpiryProps = FieldHelpProps &
+  Omit<FieldProps<string>, 'onChange'> & {
```
henit, line 26: Sending `string` to `FieldProps` here states that the value for this field component is a string, so the `value`-prop accepts a string, the validator will expect a validator that expects a string value, while updating the hook with an object in `handleChange`, breaking the validation features. If the `string` value here is replaced by `ExpryValue`, the field will accept that structure of object as the `value`-prop, and you will not need to override the `onChange` prop as it will be set to submit ExpiryValue.

packages/dnb-eufemia/src/extensions/forms/Field/Expiry.tsx
joakbjerk: I did start out with the `Field.String` and a custom mask, but this ended up complicating the process of making the tab-behaviour function like it does in the `DatePicker`-component, and keeping the placeholder text while the user is typing. And as far is I could tell, it did not support labels for the different values either, like the DatePicker component does for `day` `month` `year` etc.

packages/dnb-eufemia/src/extensions/forms/Field/Expiry.tsx
joakbjerk: No, this is done so the components gets the same styling as `DatePicker` which handles the correct highlighting of the input/placeholder based on if there is any input or not

packages/dnb-eufemia/src/extensions/forms/Field/Expiry.tsx
henit: But if Field.String lacks some features you need, could this possibly be simplified at least a bit using InputMasked instead of creating a "custom" mask feature?

packages/dnb-eufemia/src/extensions/forms/Field/Expiry.tsx
joakbjerk: The only reason was due to the values inside the component being split into month and year, so not necessarily a good enough reason 😄

packages/dnb-eufemia/src/extensions/forms/Field/Expiry.tsx
joakbjerk: Hmm not sure if I quite understand? I tried using a `InputMasked` too, but it would no let me keep the placeholder while typing, and then there is the issue with the tab-functionality and missing labels for `month` and `year`. I found the simplified version I tried out, a bit lacking in the accessibility department.

That is why I went with the same approach as done in the `DatePickerInput` component, and the `TextMask` component is also imported from there

henit: There might be specific needs here that makes it hard to use those components, but it just seems to me that this is something the InputMasked component should have features for, so we don't need to write and maintain custom code for this kind of UI details in a component like this one. It should be standardized, so maybe we need to change/extend the features of InputMasked?

packages/dnb-eufemia/src/extensions/forms/Field/Expiry.tsx
henit: I'm not sure how similar cases like this is handled other places in Eufemia, but in my experience, this kind of breaks some of the points behind the BEM-usage, so unless this is a common pattern in the repo, maybe we should achieve this in other ways? Like sass mixins or other ways to reuse the styling without breaking BEM?

packages/dnb-eufemia/src/extensions/forms/Field/Expiry.tsx
joakbjerk: Yeah totally agree! I don't see why the functionality that I needed from `TextMask`, couldn't be integrated into `InputMasked`. As of now, it looks like that `TextMask` is only used internally  for the `DatePickerInput`. @tujoworker  any thoughts?

packages/dnb-eufemia/src/extensions/forms/Field/Expiry.tsx
joakbjerk: Yes, I think that solution would be more desirable

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/feature-fields/Expiry.mdx
```diff
@@ -0,0 +1,27 @@
+---
+title: 'Expiry'
+description: '`Field.Expiry` is a wrapper component for the input of strings, with user experience tailored for payment card Expiry.'
```
langz, line 3:
```suggestion
description: '`Field.Expiry` is a wrapper component for the input of strings, with user experience tailored for expiry dates for payment cards.'
```

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/feature-fields/Expiry/demos.mdx
```diff
@@ -0,0 +1,45 @@
+---
+showTabs: true
+---
+
+import ChangeLocale from 'dnb-design-system-portal/src/core/ChangeLocale'
+import * as Examples from './Examples'
+
+The locale is what determines the components `placeholder` format .e.g. `mm/yy` in norwegian and `yy/mm` in english.
+
+<ChangeLocale bottom label="Locale used in the demos:" showUS={true} />
```
langz, line 10: Nice, great to have this in the docs 👍

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/feature-fields/Expiry/events.mdx
```diff
@@ -0,0 +1,13 @@
...
+showTabs: true
+---
+
+import DataValueWriteEvents from 'Docs/uilib/extensions/forms/data-value-write-events.mdx'
+
+## Events
+
+| Event      | Description                                                                                                                                                     |
+| ---------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `onChange` | _(optional)_ Will be called once the input field is completely filled out. An object containing `month`, `year`, `raw`and `formatted? values as argument        |
```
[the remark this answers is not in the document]
langz: I also think the docs for this event should be revised, as I believe it's called after every change in the input field and not just once the input field is completely filled out.

It seems to "log" after each "change", see this example: https://eufemia-git-feat-expiry-date-field-eufemia.vercel.app/uilib/extensions/forms/feature-fields/Expiry/demos/#empty

langz:
Should it be possible to enter strings/non numeric values to the component?
Here's a short video of me trying to enter some strings together with number(I press "m1" and "å1"):

https://github.com/dnbexperience/eufemia/assets/1359205/d5db5f09-188d-42e2-8b20-8a03d012d62d

packages/dnb-eufemia/src/extensions/forms/Field/Expiry.tsx
```diff
@@ -0,0 +1,209 @@
...
+export type ExpiryValue = {
+  /**
+   * Month value from input
+   */
+  month: string
+  /**
+   * Year value from input
+   */
+  year: string
+}
```
langz, line 23: It's probably correct, but is it so that it's common/normal to represent the month(month number)(maybe the the prop should be named monthNumber), as strings and not numbers? 🤔 
Probably because of 01, 02, 03, 04, 05, 06, 07, 08, 09, etc... 🤔

packages/dnb-eufemia/src/extensions/forms/Field/Expiry.tsx
```diff
@@ -0,0 +1,209 @@
...
+import SharedContext from '../../../shared/Context'
+import { FieldHelpProps, FieldProps, pickSpacingProps } from '../types'
+import { useDataValue } from '../hooks'
+import classnames from 'classnames'
+import { HelpButton, InputMasked } from '../../../components'
+import FieldBlock from '../FieldBlock'
+import { InputMaskedProps } from '../../../components/InputMasked'
+import useHandleCursorPosition from '../../../components/input-masked/hooks/useHandleCursorPosition'
+
+export type ExpiryValue = {
```
langz, line 14: It's probably correct, but is it so that it's common/normal to represent the month(month number)(maybe the prop should be named monthNumber maybe not if it's not a number :P), as strings and not numbers? 🤔
Probably because of 01, 02, 03, 04, 05, 06, 07, 08, 09, etc... 🤔

langz (commented):
Great work, and great tests 👨‍🎤 🎸

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/feature-fields/Expiry/events.mdx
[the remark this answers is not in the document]
joakbjerk: Yeah thats true, forgot to change that part of the text, or should change the onChange functionality again to just fire on fully filled out input? 🤔

packages/dnb-eufemia/src/extensions/forms/Field/Expiry.tsx
```diff
@@ -0,0 +1,209 @@
...
+
+    return handleChange({ ...value, [type]: sanitizedValue })
+  }
+
+  return (
+    <FieldBlock
+      className={classnames('dnb-forms-field-expiry', className)}
+      forId={id}
+      layout={layout}
+      label={label}
```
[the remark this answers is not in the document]
joakbjerk: Ahh, yeah moving it to the input would be better in this case, so we actually keep that label functionality!

joakbjerk:
> Should it be possible to enter strings/non numeric values to the component? Here's a short video of me trying to enter some strings together with number(I press "m1" and "å1"):
> Screen.Recording.2023-10-06.at.09.51.02.mov

I'm thinking just numbers, as that's the value we want the date and year in. The return value of 1m or 2å could be mitigateed by making the onChange only fire on when the input is fully filled out, or empty, since the placeholderCharacters being part of the return value is an unfortunate side effect of the onChange firing on every change (as it should, but maybe not in this use case 😅)

packages/dnb-eufemia/src/extensions/forms/Field/Expiry.tsx
joakbjerk: Yeah, its because the numbers can come with a leading zero, which gets removed if we convert it to a number, We could make it a number, but then the other developers would have to add the leading zero themselves if that is what they need to send with the form data. Not sure what the best cause of action is here 🤔

packages/dnb-eufemia/src/extensions/forms/Field/Expiry.tsx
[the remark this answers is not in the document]
langz: I do also get a few errors/warnings in the console regarding `label`. I think this is also related to this discussion/topic:
image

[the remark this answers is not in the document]
joakbjerk: Feels like another great argument for moving the label to the input element itself 😄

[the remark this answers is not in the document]
joakbjerk: Looks like the input itself is about 3px wider in Chrome, as opposed to firefox 😅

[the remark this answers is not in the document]
joakbjerk: Now that I think about it, its two input fields in one component so we should find a way to wrap them inside a fieldset with a legend as the label 🤔

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/feature-fields/Expiry/Examples.tsx
```diff
@@ -0,0 +1,91 @@
...
+export const ValidationRequired = () => {
+  return (
+    <ComponentBox scope={{ Field }}>
+      <Field.Expiry
+        value={{ month: '08', year: '26' }}
+        label="Label text"
+        onChange={({ month, year }) =>
+          console.log('onChange', { month, year })
+        }
+        required
```
tujoworker, line 87: Does not seem to show an error on blur when no value is given.

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/feature-fields/Expiry/demos.mdx
```diff
@@ -0,0 +1,47 @@
+---
+showTabs: true
+---
+
+import ChangeLocale from 'dnb-design-system-portal/src/core/ChangeLocale'
+import * as Examples from './Examples'
+
+## Demos
+
+The locale is what determines the components `placeholder` format .e.g. `mm/yy` in norwegian and `yy/mm` in english.
```
tujoworker, line 10: Actually, `yy/mm` should be like that only for en_US, not for en_GB (based on what I remember).

packages/dnb-eufemia/src/extensions/forms/Field/style.ts
```diff
@@ -3,4 +3,4 @@
  *
  */
 
-import './style'
+import './style/index.scss'
```
tujoworker, line 6:
```suggestion
```

packages/dnb-eufemia/src/extensions/forms/Field/style/index.scss
tujoworker, line 1: This has changed and should be removed. 

`dnb-expiry.scss` should be added in `dnb-forms-field.scss`.

tujoworker (commented):
Nice work so far 👌

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/feature-fields/Expiry/demos.mdx
joakbjerk: Ahh, yeah that's true!

packages/dnb-eufemia/src/extensions/forms/Field/style/index.scss
joakbjerk: I don't think we actually need `dnb-expiry.scss` anymore now that we have the `MultiInputMask` component

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/feature-fields/Expiry/demos.mdx
joakbjerk: So after doing some research, it seems like the US is also using the `mm/yy` format for credit card expiry, so I will change this brainfart 🙏

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/feature-fields/Expiry/Examples.tsx
joakbjerk: Seems to be the same issue on the `Date` field, so we would need to find a way to trigger the `required` validation when either none or all the inputs in the MultiInputMask is without value

packages/dnb-eufemia/src/extensions/forms/Field/Expiry.tsx
stenburger1: Sorry to un-resolve this: we were the ones who suggested adding the Expiry Date component. https://jira.tech.dnb.no/browse/EDS-571

With regards to @henit's arguments: ironically, the closest we (consumers of eufemia) could "easily" come to the desired behaviour was with an ` <InputMasked />`..

```
      <InputMasked
        className={style.inputMasked}
        mask={[/\d/, /\d/, '/', /\d/, /\d/]}
        mask_options={{
          allowNegative: false,
        }}
        on_change={({ cleanedValue }) => setInput(cleanedValue)}
        show_mask
        /**
         * The character U+2013 "–" could be confused with the ASCII character
         * U+002d "-", which is more common in source code.
         * This is here for screen readers and to allow usage of input.length from
         * the cleanedValue of the on_change.
         */
        placeholder_char='–'
        size='large'
        align='center'
        top='medium'
      />
```

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/feature-fields/Expiry/Examples.tsx
```diff
@@ -0,0 +1,91 @@
+import ComponentBox from '../../../../../../shared/tags/ComponentBox'
+import { Field } from '@dnb/eufemia/src/extensions/forms'
+import { FormError } from '@dnb/eufemia/src/extensions/forms/types'
+
+export const Empty = () => {
+  return (
+    <ComponentBox scope={{ Field }}>
+      <Field.Expiry
+        onChange={({ month, year }) =>
```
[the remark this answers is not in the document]
stenburger1: In general we should aim to reduce boilerplate required by consumers. As this info most likely is going straight into an API request, the simplest for a consumer would be a string omitting the slash,  e.g. `"0924"`.

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/feature-fields/Expiry/Examples.tsx
joakbjerk: It was solved by defining the `emptyValue` on the props for `useDataValue` 👍

langz:
Here's a few observations when testing the component https://eufemia-git-feat-expiry-date-field-eufemia.vercel.app/uilib/extensions/forms/feature-fields/

In Google Chrome, I get the following error/warning in the docs(not sure if this is critical or not):

```
Duplicate form field id in the same form
Multiple form field elements in the same form have the same id attribute value. This might prevent the browser from correctly autofilling the form.
To fix this issue, use unique id attribute values for each form field.
12 resources
```

joakbjerk:
> Here's a few observations when testing the component https://eufemia-git-feat-expiry-date-field-eufemia.vercel.app/uilib/extensions/forms/feature-fields/
> 
> In Google Chrome, I get the following error/warning in the docs(not sure if this is critical or not):
> 
> ```
> Duplicate form field id in the same form
> Multiple form field elements in the same form have the same id attribute value. This might prevent the browser from correctly autofilling the form.
> To fix this issue, use unique id attribute values for each form field.
> 12 resources
> ```

This issue should be fixed now 🙏

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/feature-fields/Expiry/properties.mdx
```diff
@@ -0,0 +1,9 @@
+---
+showTabs: true
+---
+
+import DataValueReadwriteProperties from '../../data-value-readwrite-properties.mdx'
+
+## Properties
+
+<DataValueReadwriteProperties type="string" />
```
tujoworker, line 9: Are there any properties we should `omit`?

packages/dnb-eufemia/src/extensions/forms/Field/Expiry/Expiry.tsx
```diff
@@ -0,0 +1,117 @@
...
+    handleBlur,
+    handleChange,
+  } = useDataValue({
+    ...props,
+    emptyValue: '',
+  })
+
+  if (value.length > 4) {
+    warn(
+      'Expiry field value length should not exceed four, Only the first four characters will be used!'
```
tujoworker, line 46: I don't think we should include this warning, because, we can't take care of dev failures. And this type of issue should be catched before prod.

packages/dnb-eufemia/src/extensions/forms/Field/Expiry/Expiry.tsx
```diff
@@ -0,0 +1,117 @@
...
+    )
+  }
+
+  const expiry: ExpiryValue = {
+    month: value.substring(0, 2) ?? '',
+    year: value.substring(2, 4) ?? '',
+  }
+
+  const id = useRef(propsId || makeUniqueId()).current
```
tujoworker, line 55:
```suggestion
  const idRef = useRef(propsId || makeUniqueId()).current
```

packages/dnb-eufemia/src/extensions/forms/Field/Expiry/Expiry.tsx
```diff
@@ -0,0 +1,117 @@
...
+      className={classnames(className)}
+      labelSecondary={labelSecondary}
+      labelDescription={labelDescription}
+      info={info}
+      warning={warning}
+      error={error}
+      {...pickSpacingProps(props)}
+    >
+      <MultiInputMask
+        id={`${id}__input`}
```
tujoworker, line 70:
```suggestion
        id={`${idRef.current}__input`}
```

packages/dnb-eufemia/src/extensions/forms/Field/Expiry/Expiry.tsx
```diff
@@ -0,0 +1,117 @@
...
+    label = sharedContext?.translation.Forms.expiryLabel,
+    error,
+    info,
+    warning,
+    help,
+    disabled,
+    value = '',
+    labelDescription,
+    labelSecondary,
+    layout,
```
tujoworker, line 34:
```suggestion
    layout = 'vertical',
```

tujoworker: Labels in Forms should be vertical by default.

packages/dnb-eufemia/src/extensions/forms/Field/Expiry/__tests__/Expiry.test.tsx
```diff
@@ -0,0 +1,232 @@
...
+      input.focus()
+    })
+
+    await userEvent.keyboard('1235')
+
+    expect(onChange).toBeCalledTimes(4)
+    expect(onChange.mock.calls[0][0]).toEqual('1')
+    expect(onChange.mock.calls[1][0]).toEqual('12')
+    expect(onChange.mock.calls[2][0]).toEqual('123')
+    expect(onChange.mock.calls[3][0]).toEqual('1235')
```
tujoworker, line 34:
```suggestion
    expect(on_blur).toHaveBeenLastCalledWith('1235')
```

When using "Last", you can remove all the other checks. Should be fine enough. You already did check for `toBeCalledTimes`.

tujoworker (approved):
Nice work! Just some request we need to look at. And some we need to fix before merge.

Also, the `width` of the field should be `width: var(--forms-field-width--small);` while the.
And `justify-content: center;` in the `.dnb-input__shell` I think will be correct.

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/feature-fields/Expiry/properties.mdx
joakbjerk: Hmm, will double check!

PR #2664, opened 2023-09-19
fix(Forms): precede children over title in Form.Selection

Frikki:
If a `<Field.Option>` in a `dropdown` `<Field.Selection>` has `title` and `children` properties, the `title` property precedes the `children` property. That is reversed logic from the `<Field.Option>` component, where the `children` property precedes the `title` property.

This commit aligns the logic with the `<Field.Option>` component, making the `children` property precede the `title` property.


tujoworker (approved):
Thank your for this PR 🙏

PS: Let me know if you have plans to contribute more. Then I add you to the contributors list so you can push directly to the branch, instead of using a fork.

PR #2674, opened 2023-09-22
chore: make CI run on fork PRs

tujoworker:
I'm not sure if this change will actually help. I just observe in this PR #2664 that some tests did run after approval.

And I can't find any info or setting that do explain the difference between the actions.


langz (approved):
Let's give it a go🪃

PR #2675, opened 2023-09-22
docs(Portal): add View Transition

tujoworker:
- Add [View Transition](https://developer.mozilla.org/en-US/docs/Web/API/View_Transitions_API) support. Works only on chromium based browsers. Is a show-case for other DNB projects. The portal does not need such improvements I would argue. 
- Some more link usage alignment to be able to call the `startViewTransition`  APIon just two places.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ completions.json
 .history
 *not_in_use*
 /packages/dnb-ui-lib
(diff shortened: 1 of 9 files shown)


langz (approved):
Nice, looks kinda fancy 💃

PR #2681, opened 2023-09-26
chore(CI): ensure e2e to wait before build is finished

tujoworker:
Our `action-wait-for-check` did not wait anymore. So build did fail right away. To overcome this, we can simply upgrade the the latest version.

More info about the GitHub API change: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


langz (approved):

tujoworker:
Its better. It does wait >3 min now – but it still fails 🤔 So we need to investigate more.

PR #2689, opened 2023-09-27
docs(First Steps): fixes broken link to installation documentation

langz:
The link did not work, see here https://eufemia.dnb.no/uilib/usage/first-steps/ and try to press the `"installation documentation"`-link.

This fix changes the heading from `Installation` to `Installing`, as it somehow makes the link https://eufemia.dnb.no/uilib/usage/first-steps/#installation work. 

I could also rather kept the existing header(Installation) and added a "/" to the end of the link, like https://eufemia.dnb.no/uilib/usage/first-steps/#installation/

Not sure which solution that's preferred to make the link work 🙏


tujoworker (approved):
✍️

tujoworker:
But I think, "installation" is still more correct. Installing is more a "process" you are in. But I may be wrong.

PR #2690, opened 2023-09-27
chore(PostalCodeAndCity): adds support for help prop

langz:
Not 100% sure which commit decoration I should use(chore vs. feat), I would normally use feat, but since this is under the "beta flag" 🤔 ....

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/PostalCodeAndCity.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/PostalCodeAndCity.tsx
@@ -3,8 +3,10 @@ import classnames from 'classnames'
 import SharedContext from '../../../shared/Context'
 import FieldBlock, { Props as FieldBlockProps } from '../FieldBlock'
 import StringComponent, { Props as StringComponentProps } from './String'
+import { FieldHelpProps } from '../types'
 
-export type Props = Omit<FieldBlockProps, 'children'> &
+export type Props = FieldHelpProps &
+  Omit<FieldBlockProps, 'children'> &
   Record<'postalCode' | 'city', StringComponentProps> & {
     width?: 'small' | 'medium' | 'large'
   }
@@ -16,6 +18,7 @@ function PostalCodeAndCity(props: Props) {
     postalCode = {},
     city = {},
     width = 'large',
+    help,
     ...fieldBlockProps
   } = props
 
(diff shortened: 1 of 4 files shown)


packages/dnb-eufemia/src/extensions/forms/Field/PostalCodeAndCity.tsx
```diff
@@ -3,8 +3,10 @@ import classnames from 'classnames'
 import SharedContext from '../../../shared/Context'
 import FieldBlock, { Props as FieldBlockProps } from '../FieldBlock'
 import StringComponent, { Props as StringComponentProps } from './String'
+import { FieldHelpProps } from '../types'
 
-export type Props = Omit<FieldBlockProps, 'children'> &
+export type Props = FieldHelpProps &
```
tujoworker, line 8:
```suggestion
export type Props = 
```

I think, `help` should be a part of `StringComponentProps`, is it not?

langz: Correct, `help` is part of `StringComponentProps`.
But this component `PostalCodeAndCity` doesn't extend/inherit props from `StringComponentProps`.
I could of course do `Pick<StringComponentProps, 'help'>` instead of `FieldHelpProps`, but I think `FieldHelpProps` is "better" or more declarative.

langz: But because of the existing type in `PostalCodeAndCity`: `Record<'postalCode' | 'city', StringComponentProps>`.
Maybe I don't need to add a specific help prop to the component, as I believe they could do the following:
´´´
      <Field.PostalCodeAndCity
        postalCode={{
          help: {
            title: 'Help is available',
            contents:
              'Here is what a team can do for you. . . . It allows you to help others do their best.',
          },
        }}
        city={{
          help: {
            title: 'Help is available',
            contents:
              'Here is what a team can do for you. . . . It allows you to help others do their best.',
          },
        }}
      />
´´´

Maybe I should rather remove the code of this PR where I add the prop `help`, and rather just update the example to showcase the example above?

tujoworker: haha, yeah. Let's go for one help button only here. So we need to set the prop on the component directly and not inside `city` or `postalCode`.

tujoworker (approved):
Nice 🙏

PR #2695, opened 2023-09-28
chore(CI): upgrade playwright and make e2e run shorter in duration 

tujoworker:
- Looks like we need to update our visual snapshots.
- Also, we went down from apr. 32 min to 24 min with the duration of the e2e CI runs by removing the artifact’s step. 
- ⚠️ We need to change the GitHub settings and remove the required build step for PRs.
- 📢 And everyone needs to rebase from main, when this PR is merged.


langz (approved):
🚤

PR #2703, opened 2023-09-29
fix(PaymentCard): removes excess space when formatting card number

langz:
Before this PR the function `formatCardNumber` would return excess/surrounding space around the result(formatted card number).
Like: `formatCardNumber('************1337')` would result in `' **** **** **** 1337 '`.
After this PR, the same function will result in `'**** **** **** 1337'`.

### DIFF
--- a/packages/dnb-eufemia/src/extensions/payment-card/PaymentCard.js
+++ b/packages/dnb-eufemia/src/extensions/payment-card/PaymentCard.js
@@ -181,12 +181,18 @@ export default class PaymentCard extends React.PureComponent {
 
 const formatCardNumberRegex = /(?=(?:....)*$)/g
 
-export const formatCardNumber = (cardNumber, digits) =>
-  digits
-    ? cardNumber
-        .slice(cardNumber.length - digits, cardNumber.length)
-        .replace(formatCardNumberRegex, ' ')
-    : cardNumber.replace(formatCardNumberRegex, ' ')
+export const formatCardNumber = (cardNumber, digits = 8) => {
+  if (!cardNumber) {
+    return cardNumber
+  }
+  if (digits && digits <= cardNumber.length) {
+    return cardNumber
+      .slice(cardNumber.length - digits, cardNumber.length)
+      .replace(formatCardNumberRegex, ' ')
+      .trim()
+  }
+  return cardNumber.replace(formatCardNumberRegex, ' ').trim()
+}
 
 export const getCardData = (productCode) => {
   const card = cardProducts.find(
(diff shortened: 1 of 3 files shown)


packages/dnb-eufemia/src/extensions/payment-card/PaymentCard.js
```diff
@@ -181,12 +181,18 @@ export default class PaymentCard extends React.PureComponent {
...
 
-export const formatCardNumber = (cardNumber, digits) =>
-  digits
-    ? cardNumber
-        .slice(cardNumber.length - digits, cardNumber.length)
-        .replace(formatCardNumberRegex, ' ')
-    : cardNumber.replace(formatCardNumberRegex, ' ')
+export const formatCardNumber = (cardNumber, digits) => {
+  if (!cardNumber) {
+    return null
```
langz, line 186: Perhaps it's better to return back `cardNumber`? 🤔 
Any thoughts?

stenburger1 (commented):
Consider if applying a default of 8 digits is common enough to be applicable. Also I think the `return null` section could pose issues with types depending on https://www.typescriptlang.org/tsconfig#strictNullChecks

langz:
> Consider if applying a default of 8 digits is common enough to be applicable. Also I think the `return null` section could pose issues with types depending on https://www.typescriptlang.org/tsconfig#strictNullChecks

I've updated both the code and docs to doing this now #2543

packages/dnb-eufemia/src/extensions/payment-card/PaymentCard.js
```diff
@@ -181,12 +181,18 @@ export default class PaymentCard extends React.PureComponent {
 
 const formatCardNumberRegex = /(?=(?:....)*$)/g
 
-export const formatCardNumber = (cardNumber, digits) =>
-  digits
-    ? cardNumber
-        .slice(cardNumber.length - digits, cardNumber.length)
-        .replace(formatCardNumberRegex, ' ')
-    : cardNumber.replace(formatCardNumberRegex, ' ')
+export const formatCardNumber = (cardNumber, digits = '8') => {
```
tujoworker, line 184: The default `digits` should be a number. Not a string.

langz: Of course, thanks!
Was making these changes quite fast when dusjing out of the office🤣👌

langz: I've updated this now 🙏

PR #2713, opened 2023-10-01
feat(Forms): enable tree-shaking compatibility

tujoworker:
When importing Field, we import basically everything attached to the object, so we get no tree-shaking. Test results when using Webpack bundle analyser:

```tsx
// 👉 556 KB
import { Field } from '@dnb/eufemia/extensions/forms' // (total 761)
render(<Field.String />

// 👉 325 KB (-231 KB)
import FieldString from '@dnb/eufemia/extensions/forms/Field/String' // (total 530)
render(<FieldString />

// 👉 105 KB (-220 KB)
import { Input } from '@dnb/eufemia' // (total 310)
render(<Input />

```

This PR exports all the parts, but without "objects", so we get the wanted results.

```tsx
import { Field } from '@dnb/eufemia/extensions/forms'

// 👉 325 KB instead of 556 KB
render(<Field.String />
```


langz (approved):
Great👏👏👏

PR #2671, opened 2023-09-21
chore: removes replaceHydrationFunction

langz:
As part of supporting React v18(https://github.com/dnbexperience/eufemia/pull/2481) `replaceHydrationFunction` was added to combat some issues related to SSR in our portal. This PR removes this again.

Related ticket: https://jira.tech.dnb.no/browse/EDS-476

PR #2718, opened 2023-10-03
chore: removes replaceHydrationFunction

langz:
Not really sure about the buikd errors I get in https://github.com/dnbexperience/eufemia/pull/2671, so just trying to put out a new PR, let's see...


langz:
Getting the same error here, so removing this PR, and continuing in https://github.com/dnbexperience/eufemia/pull/2671

PR #2719, opened 2023-10-03
fix(PhoneNumber): remove default +47 value

langz:
Link to Slack thread: https://dnb-it.slack.com/archives/CMXABCHEY/p1695840538402859

I do myself find it a bit strange that the value of is +47 when there's no country code selected.
The following video is from `main` branch running locally, showcasing the scenario I wan't to fix/improve with this PR.

https://github.com/dnbexperience/eufemia/assets/1359205/85e67ba9-8295-4a55-b74a-9a0d08d4dffc

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber.tsx
@@ -23,9 +23,6 @@ function PhoneNumber(props: Props) {
   const sharedContext = useContext(SharedContext)
   const preparedProps: Props = {
     ...props,
-    // Important for the default value to be defined here, and not after the useDataValue call, to avoid the UI jumping
-    // back to +47 once the user empty the field so handleChange send out undefined.
-    value: '+47',
     errorMessages: {
       required: sharedContext?.translation.Forms.phoneNumberErrorRequired,
       ...props?.errorMessages,
@@ -104,7 +101,7 @@ function PhoneNumber(props: Props) {
           countryCodeFieldClassName
         )}
         label={countryCodeLabel}
-        value={countryCode ?? '+47'}
+        value={countryCode}
         disabled={disabled}
         onChange={handleCountryCodeChange}
       />
--- a/packages/dnb-eufemia/src/extensions/forms/Field/__tests__/PhoneNumber.test.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/__tests__/PhoneNumber.test.tsx
@@ -1,0 +1,63 @@
+import React from 'react'
+import { wait } from '../../../../core/jest/jestSetup'
+import { fireEvent, render } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+import PhoneNumber from '../PhoneNumber'
+
+describe('Field.PhoneNumber', () => {
+  it('should return correct value onFocus and onBlur event', async () => {
+    const onFocus = jest.fn()
+    const onBlur = jest.fn()
+    render(<PhoneNumber onFocus={onFocus} onBlur={onBlur} />)
+
+    const phoneElement = document.querySelector(
+      '.dnb-forms-field-phone-number__number input'
+    )
+
+    fireEvent.focus(phoneElement)
+    expect(onFocus).toHaveBeenLastCalledWith(undefined)
+
+    fireEvent.blur(phoneElement)
+    expect(onBlur).toHaveBeenLastCalledWith(undefined)
+
+    await userEvent.type(phoneElement, '99999999')
+
+    fireEvent.focus(phoneElement)
+    expect(onFocus).toHaveBeenLastCalledWith('99999999')
+
+    fireEvent.blur(phoneElement)
+    expect(onBlur).toHaveBeenLastCalledWith('99999999')
+  })
+
+  it('should return correct value onChange event', async () => {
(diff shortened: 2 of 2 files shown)


packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber.tsx
```diff
@@ -23,9 +23,6 @@ function PhoneNumber(props: Props) {
   const sharedContext = useContext(SharedContext)
   const preparedProps: Props = {
     ...props,
-    // Important for the default value to be defined here, and not after the useDataValue call, to avoid the UI jumping
-    // back to +47 once the user empty the field so handleChange send out undefined.
```
langz, line 27: Any other alternative solutions where we don't have to have +47 as default but also avoid "UI jumping"?

Any thoughts @henit @tujoworker

henit: Does it still happen after you removed the fallback +47 in the prop below?

henit: The reason why I defaulted to +47 is that forms I found in the Figma screens had +47 preselected everywhere. But it is kind of against the way the rest of the fields are created (being empty as default unless given an initial value), so from the code, this change makes sense for me. But I guess all implementations will have to provide a
`<Field.PhoneNumber value={source.phoneNumber ?? '+47'} />` now if they want this to be the case..

Example:
Screenshot 2023-10-03 at 14 33 34

tujoworker: I guess it makes sense to default to 47 as mentioned by @henit – but we need tests. I add a quick first test. They would need be aligned if we keep it as is.

packages/dnb-eufemia/src/extensions/forms/Field/__tests__/PhoneNumber.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+        onCountryCodeChange={onCountryCodeChange}
+      />
+    )
+
+    const phoneElement = document.querySelector(
+      '.dnb-forms-field-phone-number__number input'
+    )
+    await userEvent.type(phoneElement, '99999999')
+    expect(onChange).toHaveBeenLastCalledWith('99999999')
+
```
tellnes, line 47: If you empty the field, then the component emits `undefined` and not an empty string. Maybe it is intentional, but I did expect it to be an empty string.

```suggestion

    await userEvent.clear(phoneElement)
    expect(onChange).toHaveBeenLastCalledWith('')

    await userEvent.type(phoneElement, '99999999')
```

packages/dnb-eufemia/src/extensions/forms/Field/__tests__/PhoneNumber.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+    await userEvent.type(codeElement, '+41')
+    fireEvent.keyDown(codeElement, {
+      keyCode: 40, // down
+    })
+    fireEvent.keyDown(codeElement, {
+      keyCode: 13, // enter
+    })
+    await wait(1)
+    expect(onCountryCodeChange).toHaveBeenLastCalledWith('+41')
+    expect(onChange).toHaveBeenLastCalledWith('+41 99999999')
```
tellnes, line 61: Also for the phone country code selector. I expect an empty string and not undefined when you clear it.

```suggestion
    expect(onChange).toHaveBeenLastCalledWith('+41 99999999')

    await userEvent.clear(codeElement)
    await wait(1)
    expect(onCountryCodeChange).toHaveBeenLastCalledWith('')
    expect(onChange).toHaveBeenLastCalledWith('99999999')
```

packages/dnb-eufemia/src/extensions/forms/Field/CountryCode.tsx
```diff
@@ -21,7 +21,7 @@ function CountryCode(props: Props) {
   const {
     className,
     layout = 'vertical',
-    placeholder,
+    placeholder = sharedContext?.translation.Forms.countryCodePlaceholder,
```
tellnes, line 24: In the current UI, this text is longer than the width of the phone country code component. So not all of the placeholder text is not visible to the user.

packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber.tsx
tellnes: Having `+47` as default sounds reasonable to me. But I don't see any issues with having the consumer of the component having to set the default value if you are using it as a controlled component. Just document it.

packages/dnb-eufemia/src/extensions/forms/Field/CountryCode.tsx
langz: Removing the placeholder then 👍

packages/dnb-eufemia/src/extensions/forms/Field/__tests__/PhoneNumber.test.tsx
tujoworker: I agree, but would like to hear @henit what he thinks.

tujoworker:
Closing this. We continue with the default of 47 – but fix also an issue, where `required` not works, as there is already the 47 value.

PR #2721, opened 2023-10-03
fix(Anchor): add --anchor-background-gutter

tujoworker:
Is it a gutter? I'm not sure about the naming.

Keep in mind, you always need to use a unit, to null it out:

```css
--anchor-background-gutter: 0rem;
```


stenburger1 (approved):
Gutter makes sense to me!

PR #2724, opened 2023-10-04
docs(Layout): move layout (spacing and media query) docs to layout component docs

tujoworker:
Does that make sense?

Old:
Screenshot 2023-10-04 at 07 12 13

New:

Screenshot 2023-10-04 at 07 12 24


joakbjerk (approved):
Yeah I feel it makes sense now at least, since we have a `Layout` component. Makes it easier to find all the layout related stuff, since its in one place 👍

PR #2725, opened 2023-10-04
chore(Forms): move main and sub heading styles from layout to form

tujoworker:
These files where missed out on the move in PR #2720


joakbjerk (approved):
🤙

langz (approved):

PR #2727, opened 2023-10-04
chore(CountryCode): adds default placeholder

langz:


tujoworker:
I don't think this will work. There is no room for it as the with is too small. image

langz:
> I don't think this will work. There is no room for it as the with is too small. image

I just formulated the placeholder text in the same manner/style as other placeholder texts we have in the translation files. Is it a problem that it's not room to display the full placeholder text? Surely it must be better than none placeholder? Both for screen users and screenreaders.

I'm open to change the text though, just thought we should have a placeholder for this.

PR #2728, opened 2023-10-04
chore(Forms): Move component files into separate folders

henit:
Moving component files in Eufemia Forms from being in the same folder (Field, Value, DataContext etc) to sub folders per component similar to the rest of Eufemia.


tujoworker (approved):
Very good work 🙏 Thank you!

PR #2729, opened 2023-10-04
fix(Slider): make key navigation work when swapping position

tujoworker:
Sadly there is no support for simulating a keyboard usage. Like so:

```js
await userEvent.keyboard('[ArrowRight]')
```

- userEvent issue: https://github.com/testing-library/user-event/issues/871

So I thought, maybe its time to add an e2e test for a Eufemia component as well.
And I can confirm that the test fails without the "fix".

With mouse:

https://github.com/dnbexperience/eufemia/assets/1501870/ce03fac9-d4ec-405c-b672-ac6e679f4771

With keyboard:

https://github.com/dnbexperience/eufemia/assets/1501870/8a60adc4-01d8-45ac-8977-13ab2129749f

With keyboard after the fix:

https://github.com/dnbexperience/eufemia/assets/1501870/e474091c-3646-4a20-a555-3eac77086872


tujoworker:
Should we name the test files like so maybe?

Slider.e2e.spec.ts

langz (approved):
Great addition and improvement 🙇

PR #2737, opened 2023-10-09
fix(forms): Handle error display when no focus and blur events is called

henit:
When a custom component was build using `useDataValue` without calling focus and blur callbacks, like if forgotten during development, or for interfaces where it is not possible (like based on one time clicks), validation was performed, but the error was never displayed because the hook was waiting for a blur event to actually return the error. Fix: If focus has not been called before a change callback, assume the events are not being used so show the error message as long as the value is invalid (unless `continuousValidation` is false). I added tests for this in a separate test file for the `useDataValue` hook since the existing field compoinents like `Field.String actually call focus/blur.


tujoworker (approved):
Looks good! Great work and good tests 👌

PR #2738, opened 2023-10-10
feat(forms): Session storage on Provider and add defaultData prop

henit:


tujoworker (changes requested):
Nice work so far 👌 would be awesome to have a test coverage for the new prop  / the changes.

I also think about, if its clear enough for a dev to know whats happening, when reading the code and seeing the prop `sessionId` 🤔

henit:
> I also think about, if its clear enough for a dev to know whats happening, when reading the code and seeing the prop `sessionId` 🤔

Do you think the prop should be renamed, or more clearly explained in the documentation?

tujoworker:
Yes, if we name the prop e.g. `sessionStorageId` it helps. Maybe even `storageId`.

henit:
> Yes, if we name the prop e.g. `sessionStorageId` it helps. Maybe even `storageId`.
> 
> Does the store get reset on submit by default? or should we provide a "clean/reset" function in a hook?

I don't know if it will happen, but in case the component would end up supporting multiple ways of storage at some point (like local storage, indexed db, service worker mechanism or some common server side API), maybe we should include the word "session"? AKA: Supporting the suggestion of `sessionStorageId`

PR #2741, opened 2023-10-11
fix(Lead): can now accept className prop without losing styling

snorrekim:
The `Lead` component  can now receive a `className` prop without losing its own `.dnb.p--lead` class. Adding custom classes or using styled components should now work as expected.


tujoworker (approved):
Nice finding and fix! 👌
I hope its OK for you that I pushed a test for it 🙏

snorrekim:
> Nice finding and fix! 👌 I hope its OK for you that I pushed a test for it 🙏

No, that's great. Thanks.

Issue #2691, opened 2023-09-27
Searching provides multiple and erroneous search results


langz:
## 🐛 Bug Report

When trying to use the search field in Eufemia, I get multiple search hits for the same "result" where a few of the results are erroneous(linking to a page that doesn't exist).

This seems to mostly happen when searching for words that are part of the "forms extension".

## To Reproduce

Go to: https://eufemia.dnb.no/, and try searching for "Button" or "PhoneNumber".

See the result when searching on "PhoneNumber":

https://github.com/dnbexperience/eufemia/assets/1359205/c0af4948-ed51-488e-ae80-86997ae570ea

tujoworker:
The reasons is that we move lastly around things without adding a redirect directive (even I never checked if Algolia actually takes this into account).
 
We can bust the index before next release.

PR #2746, opened 2023-10-12
release of v10.10

tujoworker:
Before we can release, we still need to get inn some changes:

- [x] Move Card to the main list of components
- [x] Move Flex to the main list of components
- [x] Move Grid to the main list of components
- [x] Categorize the docs and move all layout docs to its own are
- [x] Deprecated FormRow and FormSet + ensure support for vertical etc. so the shared main Provider can be used instead. 
- [x] Add `_supportsSpacingProps` to all supported components (and rename from _supportsEufemiaSpacingProps)
- [x] Delete the search index (issue #2691)


langz:
Before releasing, I can also do a small check/test of the docs and see if the links still work, now that we've moved a lot out from "Layout.X".

tujoworker:
That would be great @langz 🙏

PR #2757, opened 2023-10-16
fix(StepIndicator): react on changes to `is_current` data property

tellnes:
It looks like the StepIndicator does not react to changes in the `is_current` data property


tujoworker (approved):
Looks good – great work 👍

PR #2764, opened 2023-10-18
chore(Forms): fix types for Form.Handler

tujoworker:
So we are able to use `<form>` element attributes, such as `action`.


langz (approved):
👍

PR #2769, opened 2023-10-18
chore(forms): Eufemia Forms intro page docs updated

henit:


packages/dnb-design-system-portal/src/docs/uilib/extensions/forms.mdx
```diff
@@ -82,55 +46,116 @@ render(
...
+
+## Create your own components
+
+Eufemia Forms consists of helper components and tools so you can declaratively create interactive form components that flawlessly integrates between existing data and your custom form components. This ensures a common look and feel, even when ready-made components are combined with your local custom components.
+
+Read more about on how to [create your own component](/uilib/extensions/forms/create-component).
+
 ## Requirements
 
 Some of the internal logic requires support for importing JSON files. Meta frameworks do often support this by default.
```
langz, line 161: What's a meta framework? 🤔 
Could we add a link to info about what that is, or a short description 🙏

langz (approved):
Great diagrams 👏

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms.mdx
henit: This was written by @tujoworker earlier I think, so he would need to confirm, but I think it is because the AJV library is used internally in Eufemia Forms for validation, and that might require reading of JSON files. We have discussed ways to make the code less dependant on AJV as long as you are not using the JSON Schema validation features, but not concluded on a solution yet. But a few more details like you suggested is probably a good thing. Have you tested this in practice @tujoworker ? Any simple and tested example of how to do that in implementation code?

tujoworker: ajv uses in their code `import 'data.json'`, so they import a json file. But json is not a valid Node.js import path. So a bundles needs to support it. But create react app or Gatsby do support it via Webpack. So there it works as expected. Its just something we need to mention, as we suddenly have ajv in our dependencies and as soon as someone imports `Field.String`, ajv is executed and the json file gets imported.

tujoworker (approved):

PR #2772, opened 2023-10-19
chore: various local build related updates

tujoworker:
- Update outdate docs linking; FormRow and FormSet where moved and do not have the same path, as the component, so they can not be found. But I think thats fine as they are deprecated.
- Correct PaymentCard types as they fail during the docs update routine. We still have a valid syntax and it works fine with TypeScript.

Screenshot 2023-10-19 at 08 44 39

Screenshot 2023-10-19 at 08 46 27


langz (approved):
🖖

PR #2773, opened 2023-10-19
feat(breakpoints): media query mixin offset options

snorrekim:
* Adds global or parameter customization for mixins `allAbove()` and `allBelow()` 

# Global customization:
Will only works for the app's own stylesheets, won't retroactively affect imported Eufemia components.
```scss
@import '@dnb/eufemia/style/core/utilities.scss';

// allAbove() and allBelow() will subtract -10em from the size
$breakpoint-offset: -10em;

// the sizes 'medium' and 'large' for allAbove() and allBelow() are changed.
$breakpoints: map-merge(
  $breakpints,
  (
    'medium': 40em,
    'large': 90em,
  )
);
```

# Mixin parameter
```scss
@import '@dnb/eufemia/style/core/utilities.scss';

@include allAbove($size: small, $offset: 10em) {
  // adds 10em to $size
}
```

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/helpers.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/helpers.mdx
@@ -3,16 +3,6 @@ title: 'Helpers'
 description: 'A couple of helper functions used inside of components and extensions.'
 icon: 'helper_classes'
 order: 4
-showTabs: true
-tabs:
-  - title: Info
-    key: /uilib/helpers/info
-  - title: CSS Classes
-    key: /uilib/helpers/classes
-  - title: Functions
-    key: /uilib/helpers/functions
-  - title: Tools
-    key: /uilib/helpers/tools
 redirect_from:
   - /uilib/helper-classes
 ---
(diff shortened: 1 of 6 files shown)


tujoworker (commented):
Nice addition, Snorre! 

I think, it should be possible for the global offset to give different offsets per media query size. It would be at least useful. Lets say, on a small screen, a menu is not visible, then we don't want an offset.

snorrekim:
> Nice addition, Snorre!
> 
> I think, it should be possible for the global offset to give different offsets per media query size. It would be at least useful. Lets say, on a small screen, a menu is not visible, then we don't want an offset.

I think that would be covered by having access to the `$breakpoints` list. If you need to, say, offset everything except small by 10em, then you could change small to be 10em smaller. Or if you need a different offset for all of the sizes, you might as well edit all the `$breakpoints`. And you can edit the breakpoints based on their existing value if you want as well.

```scss
$breakpoints: map-merge(
  $breakpints,
  (
    'medium': map-get($breakpoints, medium) + 10em,
    'large': map-get($breakpoints, large) + 5em,
  )
);
```

tujoworker (approved):

PR #2774, opened 2023-10-19
fix(useMedia): enhance SSR support by setting state during component mount

tujoworker:
To avoid rehydration disturbance. The return values like `isSmall` will now always be false to start with. 
This should not be a breaking change, because on the client, it results as before.

I have not tested this hook IRL. Maybe someone can do that for me?


langz (approved):
Seems great to be able to provide a `initialValue` 👍

PR #2775, opened 2023-10-20
fix(forms): Support undefined in toggle and button fields

henit:
Before, passing no value defaulted to the `valueOff` option in `Field.Toggle`, leading to the same in `Field.Button` (false option being chosen). Now, no buttons is selected when value is undefined.

Screenshot 2023-10-20 at 14 38 13


tujoworker (approved):
Nice fix!

PR #2783, opened 2023-10-23
feat(Upload): singular texts when uploading single file

langz:
This PR adds singular texts/translations (file instead of files, etc) in the Upload component, when the `filesAmountLimit` is 1.

The changes can be seen and tested here: https://eufemia-git-feat-provide-singular-texts-in-uploa-6743ea-eufemia.vercel.app/uilib/components/upload/demos/#upload-single-filefixed-amount-of-files

From:
Screenshot 2023-10-23 at 16 01 53

To:
Screenshot 2023-10-23 at 16 01 19

### DIFF
--- a/packages/dnb-eufemia/src/components/upload/Upload.tsx
+++ b/packages/dnb-eufemia/src/components/upload/Upload.tsx
@@ -26,11 +26,19 @@ export { defaultProps }
 const Upload = (localProps: UploadAllProps) => {
   const context = React.useContext(Context)
 
+  const { buttonTextSingular, textSingular, ...translations } =
+    context.getTranslation(localProps).Upload
+
+  if (localProps?.filesAmountLimit === 1) {
+    translations.buttonText = buttonTextSingular
+    translations.text = textSingular
+  }
+
   const extendedProps = extendPropsWithContext(
     localProps,
     defaultProps,
     { skeleton: context?.skeleton },
-    context.getTranslation(localProps).Upload,
+    translations,
     context.Upload
   )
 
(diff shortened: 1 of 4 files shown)


packages/dnb-eufemia/src/components/upload/Upload.tsx
```diff
@@ -26,11 +26,22 @@ export { defaultProps }
...
+  const {
+    buttonTextSingleFileLimit,
+    textSingleFileLimit,
+    ...translations
+  } = context.getTranslation(localProps).Upload
+
+  if (localProps?.filesAmountLimit === 1) {
+    translations.buttonText = buttonTextSingleFileLimit
+    translations.text = textSingleFileLimit
+  }
```
langz, line 38: I'm not sure if the code above is the best(most correct way) of achieving what we want, but I found it to be pretty effective 🐎

tujoworker: Yes, nice! We don't have a better system of handling singular/plural translations.

packages/dnb-eufemia/src/shared/locales/en-GB.js
```diff
@@ -130,11 +130,14 @@ export default {
     Upload: {
       title: 'Upload documents',
       text: 'Drag & drop your files or choose files to upload.',
+      textSingleFileLimit:
```
tujoworker, line 133:
```suggestion
      textSingular:
```

Should probably be enough to understand the purpose.

tujoworker (approved):
Nice work! Except of my `Singular` comment, it looks good!

packages/dnb-eufemia/src/shared/locales/en-GB.js
langz: I'll change from `textSingleFileLimit` to `textSingular`, and from `buttonTextSingleFileLimit` to `buttonTextSingular` 👍

PR #2784, opened 2023-10-23
fix(NumberFormat): show em dash instead of NaN

tujoworker:
[More info](https://dnb-it.slack.com/archives/C014NMBQ5SQ/p1698044456753389).


langz (approved):
🎳

PR #2798, opened 2023-10-26
fix(Breadcrumb): Pass rest props to item when not interactive

langz:
I believe/hope this should fix the reported issue here: https://github.com/dnbexperience/eufemia/pull/2676/files#r1368576095

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/about-the-lib/releases/eufemia/v11-info.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/about-the-lib/releases/eufemia/v11-info.mdx
@@ -52,4 +52,10 @@ For more information on how to replace these, check out [these docs](/uilib/layo
 - Find `FormRow:` and replace with `formElement:`.
 - Find `import { includeValidProps } from '@dnb/eufemia/components/form-row/FormRowHelpers'` and replace with `import { pickFormElementProps } from '@dnb/eufemia/shared/helpers/filterValidProps'`.
 
+## Removal of passing down props to BreadcrumbItem's span
+
+We don't think this has been used for anything other than passing down `data-testid`'s for testing. We believe the potential side effects of passing down props to this span is greater than the advantages it gives for those who want to test this span using data-testid as their way of selecting the span.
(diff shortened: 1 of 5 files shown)


thaytharma (approved):
Thanks 😍

tujoworker (approved):
Nice! And nice tests as well! Sounds also that you not found any issues forwarding such props to the span 🎉

langz:
> Nice! And nice tests as well! Sounds also that you not found any issues forwarding such props to the span 🎉

I've not looked extensively at this, is there a greater risk of passing props to the span now, than before https://github.com/dnbexperience/eufemia/pull/2676/files#r1368576095?

Maybe we should "filter out" the "button props" to not getting drilled down to the `span`?

I can take a look tomorrow in our storybook and test the consequences of passing down all button props to the span, and see what happens 🙏

tujoworker:
We may add the filterProps addition #2802 – I have a commit ready locally and can push it after a rebase.

langz:
Perfect, thanks!
I’ll continue with this after todays morning meetings

langz:
Maybe we should add a comment, or add this PR to a list of things we probably can remove in the next breaking change/release?
Reason being, is because I'm not sure if this feature is really needed, or if it's just to be able to pass down `data-testid`'s.

I think it's fair that we now add it back to a v10 release for now, as we introduced sort of a "breaking change" by removing this in https://github.com/dnbexperience/eufemia/pull/2676/files#r1368576095 as part of a minor release. But in v11, maybe it's okay to remove it, and document it, as I think it's only used for passing down `data-testid`'s for now, which maybe is not a good enough reason to have it? 🤔

tujoworker:
Yes, I agree.

langz:
I've added a few commits
- https://github.com/dnbexperience/eufemia/pull/2798/commits/98f7d472a45460af4020c79ed5e75c322d2aa3be
- https://github.com/dnbexperience/eufemia/pull/2798/commits/0e119b365a5197adc4bdfc963a2fcab2e4ed3abd
- https://github.com/dnbexperience/eufemia/pull/2798/commits/8220d5dab333b806b91da901f661db9d788d1e50

And believe this should be good to go now 🙏

PR #2810, opened 2023-10-30
fix(NumberMask): as_number should default to locally given decimalSymbol

tujoworker:
Ensure that when setting `decimalSymbol={undefined}` the default behavior still works, which is e.g. using a coma or dot, based on the given locale.

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/Number/Number.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/Number/Number.tsx
@@ -46,7 +46,7 @@ function NumberComponent(props: Props) {
     percent,
     mask,
     thousandSeparator,
-    decimalSymbol = ',',
+    decimalSymbol,
     decimalLimit = 12,
     prefix,
     suffix,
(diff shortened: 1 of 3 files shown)


packages/dnb-eufemia/src/extensions/forms/Field/Number/Number.tsx
```diff
@@ -46,7 +46,7 @@ function NumberComponent(props: Props) {
     percent,
     mask,
     thousandSeparator,
-    decimalSymbol = ',',
+    decimalSymbol,
```
tujoworker, line 49: @henit This made the Number field always using a coma. But by default, it uses the correct separator based on the given locale.

henit: Where is that locale-based choice done then (when not set specifically through that prop)? In the number mask features of InputMasked?

henit (approved):

packages/dnb-eufemia/src/extensions/forms/Field/Number/Number.tsx
tujoworker: All good! We still can customize it. But its handled by default in the InputMasked component itself 👍

PR #2813, opened 2023-10-30
fix: BreadcrumbItem when using to and element props

skvisli:
Small fix to support using router links as breadcrumbs


tujoworker (approved):
Nice! Perfect 👌🚀

Hope its ok for you that I pushed some docs updates. You may have a look over them, just in case there are a spelling mistake.

Also, when you merge to the `main` branch, use **Squash** merge.

langz (approved):

PR #2816, opened 2023-10-30
feat(Theme): Add colors for ui, sbanken & eiendom

thaytharma:
Created and exported colors for ui, sbanken & eiendom to easily import these in code.
With this approach, users can simply import colors by importing `colors` from Eufemia. 

Example for import:  `import { colors } from '@dnb/eufemia/shared'` 
Example for usage (hex): `colors.ui.oceanGreen.hex`.
Example for usage (rgb): `colors.ui.oceanGreen.rgb`.

### DIFF
--- a/packages/dnb-eufemia/src/shared/colors.ts
+++ b/packages/dnb-eufemia/src/shared/colors.ts
@@ -1,0 +1,397 @@
+/**
+ * DNB UI / Sbanken / DNB Eiendom colors
+ */
+
+const ui = {
+  oceanGreen: createColorEntry(
+    'Ocean green',
+    '#00343E',
+    'rgb(0, 52, 62)',
+    'var(--color-ocean-green)'
+  ),
+  emeraldGreen: createColorEntry(
+    'Emerald green',
+    '#14555A',
+    'rgb(20, 85, 90)',
+    'var(--color-emerald-green)'
+  ),
+  seaGreen: createColorEntry(
+    'Sea green',
+    '#007272',
+    'rgb(0, 114, 114)',
+    'var(--color-sea-green)'
+  ),
+  mintGreen: createColorEntry(
+    'Mint green',
+    '#A5E1D2',
+    'rgb(165, 225, 210)',
+    'var(--color-mint-green)'
+  ),
+  summerGreen: createColorEntry(
+    'Summer green',
+    '#28B482',
+    'rgb(40, 180, 130)',
+    'var(--color-summer-green)'
+  ),
+  accentYellow: createColorEntry(
+    'Accent yellow',
+    '#FDBB31',
+    'rgb(253, 187, 49)',
+    'var(--color-accent-yellow)'
+  ),
+  indigo: createColorEntry(
+    'Indigo',
+    '#23195A',
+    'rgb(35, 25, 90)',
+    'var(--color-indigo)'
+  ),
+  violet: createColorEntry(
+    'Violet',
+    '#6E2382',
+    'rgb(110, 35, 130)',
+    'var(--color-violet)'
+  ),
+  skyBlue: createColorEntry(
+    'Sky blue',
+    '#4BBED2',
+    'rgb(75, 190, 210)',
+    'var(--color-sky-blue)'
+  ),
+  lavender: createColorEntry(
+    'Lavender',
+    '#F2F2F5',
+    'rgb(242, 242, 245)',
+    'var(--color-lavender)'
+  ),
+  sandYellow: createColorEntry(
+    'Sand yellow',
+    '#FBF6EC',
+    'rgb(251, 246, 236)',
+    'var(--color-sand-yellow)'
+  ),
+  pistachio: createColorEntry(
+    'Pistachio',
+    '#F2F4EC',
+    'rgb(242, 244, 236)',
+    'var(--color-pistachio)'
+  ),
+  seaGreen30: createColorEntry(
+    'Sea green 30%',
+    '#B3D5D5',
+    'rgb(179, 213, 213)',
+    'var(--color-sea-green-30)'
+  ),
+  mintGreen50: createColorEntry(
+    'Mint green 50%',
+    '#D2F0E9',
+    'rgb(210, 240, 233)',
+    'var(--color-mint-green-50)'
+  ),
+  mintGreen25: createColorEntry(
+    'Mint green 25%',
+    '#E9F8F4',
+    'rgb(233, 248, 244)',
+    'var(--color-mint-green-25)'
+  ),
+  mintGreen12: createColorEntry(
+    'Mint green 12%',
+    '#F4FBF9',
+    'rgb(244, 251, 249)',
+    'var(--color-mint-green-12)'
+  ),
+  accentYellow30: createColorEntry(
+    'Accent yellow 30%',
+    '#FEEBC1',
+    'rgb(254, 235, 193)',
+    'var(--color-accent-yellow-30)'
+  ),
+  signalOrange: createColorEntry(
+    'Signal orange',
+    '#FF5400',
+    'rgb(255, 84, 0)',
+    'var(--color-signal-orange)'
+  ),
+  fireRed: createColorEntry(
+    'Fire red',
+    '#DC2A2A',
+    'rgb(220, 42, 42)',
+    'var(--color-fire-red)'
+  ),
+  fireRed8: createColorEntry(
+    'Fire red 8%',
+    '#FDEEEE',
+    'rgb(253, 238, 238)',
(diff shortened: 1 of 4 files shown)


packages/dnb-eufemia/src/shared/colors.ts
```diff
@@ -0,0 +1,158 @@
...
+  export type EiendomColor = {
+    emeraldGreen50: ColorEntry;
+    emeraldGreen25: ColorEntry;
+    emeraldGreen10: ColorEntry;
+    mintGreen12: ColorEntry;
+  };
+  
+  export const ui: UiColor = {
+    oceanGreen: createColorEntry('Ocean green', '#00343E', '0 52 62', '--color-ocean-green'),
+    emeraldGreen: createColorEntry('Emerald green', '#14555A', '20 85 90', '--color-emerald-green'),
```
thaytharma, line 87:
```suggestion
  emeraldGreen50: {
    name: 'Emerald green 50%',
    hex: '#89aaac',
    rgb: '137 170 172',
    cssVariable: '--color-emerald-green-50',
  },```

Could also do this instead.

Comment (thaytharma) [reply packages/dnb-eufemia/src/shared/colors.ts:87]: Could also do this instead.

File: packages/dnb-eufemia/src/shared/colors.ts [hunk 2]
```diff
@@ -0,0 +1,463 @@
...
+
+type EiendomColor = {
+  emeraldGreen50: ColorEntry
+  emeraldGreen25: ColorEntry
+  emeraldGreen10: ColorEntry
+  mintGreen12: ColorEntry
+}
+
+const ui: UiColor = {
+  oceanGreen: createColorEntry(
```
thaytharma, line 91:
```suggestion
    oceanGreen: {
    name: 'Ocean green',
    hex: '#00343E',
    rgb: 'rgb(0, 52, 62)',
    cssVariable: 'var(--color-ocean-green)',
  },
```

Could also consider this instead.

packages/dnb-eufemia/src/shared/colors.ts
```diff
@@ -0,0 +1,463 @@
...
+function createColorEntry(
+  name: string,
+  hex: string,
+  rgb: string,
+  cssVariable: string
+): ColorEntry {
+  return { name, hex, rgb, cssVariable }
+}
+
+type UiColor = {
```
kimroen, line 21: You might have considered this already, but if we use a version of typescript that supports `satisfies`, we can use that to declare these colors instead. That way we can infer the keys afterwards, and skip needing to repeat all the keys like we do here.

I made an example of what that looks like [in this TS playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAwg9gGzgJwKIDtjJFAvFAbwCgop0BDAWwgC4oBnLAS3QHMSoBje+gNXORNyAIwS0GzNhwAWEAB51GgqaWSthiyewC+RIgHp9UAFIBXRlAAmETggEtWDOFACSVuOgDkwMhAiWoYGcWW1NrQNkoJkowMWpMcmAmDyJrWwFoADNTdE4kjy5kCESIeCQ0TGwACgBKOjKUDCwQAG49Tg8LUyY8Qg5SOE5i9ABxIoh0Ok4ikoaK5tqAGn6oCGpkcgRLMb9JwuLgUsRGypAlom0GRKZ6TKYIeigAJRsUSwAeJQdF2GP57AAfG0iB10BZ6MJyOgANYTXrEUikMDIOB3MQABVMyFi4mmByO5Sa1Rqy0RUGRqKYGKxOIAgghDsgKEkAG64maHOZEs4ki5XJK3e6PF4dZAfL5sH5c05A9qdHwQe7oSxwSjwjhrCAbLY7CYAVgADFMOQSTgteaRNdrtuN0AAmPXG-HS83LS70a6Ch7PV5iz5aKV-bmygxGABycAA7hFoHBhAArGw+UCQR4ZKLoTJaooBciPeSQPL+KCRpjAaRceVQWEgNPK36E05QPMxqAszamWggquUECoORgFA+fAEUjdH4QqGw9A-RUTFVq7RtMmkQxQAB6AH49KGoAB1aCcKHNhD0ZzmaAAAxTEFRl6imRL0EjUOTsjVLfoEDAAhKgXA3otpGEAIAgNB6DeUAAKpMHMvQ3qiUDdMuK5rluRCQQAypCMITHB+AIY+k64egKFkmh26QagSoLh05TwQBiFzsqqpkYiFFAA), but in short:

```ts
const ui = {
    oceanGreen: createColorEntry(),
    emeraldGreen: createColorEntry(),
} satisfies Record<string, ColorEntry>;
```

The type of `ui` is now as below, *and* you get type errors if you don't provide a `ColorEntry` as the value:

```ts
{
  oceanGreen: ColorEntry;
  emeraldGreen: ColorEntry;
}
```

thaytharma: That is a very good suggestion.
I've updated the code with these suggestions. 🙏🏼

rodynenko:
Why do we want to expose hex/rgb values for components? Are we considering dependency on static hex/rgb values somewhere and not on theme-based CSS variables only?

tujoworker (commented):
Nice idea! I also wonder the use case. And I have some thoughts on how to automate this, eventually. Because the single source of through principle is important in a DS.

But first, today we can import colors like so;

```ts
import properties from '@dnb/eufemia/style/themes/theme-ui/properties'

const seaGreenColor = properties['--color-sea-green']
```

But if `--color-sea-green` not exists, we just get `undefined` during a build. And the build does not break.

So I like the idea to be able to import such values explicit.

Now, there is work going on for renaming all of our colos. 
Therefore I think we should automate this. And we do already automate the `properties` JavaScript file. So we could enhance this one I suggest. And we already have the [color](https://www.npmjs.com/package/color) dependency in Eufemia, so we can convert values in build time.

At the end we may consider to import it from the existing place, just for consistency and easier documentation:

```ts
import { colors } from '@dnb/eufemia/style/themes/theme-ui/properties'

const seaGreenColor = colors.ui.seaGreen.hex
```

Let me know what do you think and also, where you need this feature.

thaytharma:
> Nice idea! I also wonder the use case. And I have some thoughts on how to automate this, eventually. Because the single source of through principle is important in a DS.
> 
> But first, today we can import colors like so;
> 
> ```ts
> import properties from '@dnb/eufemia/style/themes/theme-ui/properties'
> 
> const seaGreenColor = properties['--color-sea-green']
> ```
> 
> But if `--color-sea-green` not exists, we just get `undefined` during a build. And the build does not break.
> 
> So I like the idea to be able to import such values explicit.
> 
> Now, there is work going on for renaming all of our colos. Therefore I think we should automate this. And we do already automate the `properties` JavaScript file. So we could enhance this one I suggest. And we already have the [color](https://www.npmjs.com/package/color) dependency in Eufemia, so we can convert values in build time.
> 
> At the end we may consider to import it from the existing place, just for consistency and easier documentation:
> 
> ```ts
> import { colors } from '@dnb/eufemia/style/themes/theme-ui/properties'
> 
> const seaGreenColor = colors.ui.seaGreen.hex
> ```
> 
> Let me know what do you think and also, where you need this feature.



> ```ts
> import { colors } from '@dnb/eufemia/style/themes/theme-ui/properties'
> 
> const seaGreenColor = colors.ui.seaGreen.hex
> ```

I fully agree that we should automate this if possible. The end goal is to have the possible colors with type safety as you suggested here:

`import { colors } from '@dnb/eufemia/style/themes/theme-ui/properties'

const seaGreenColor = colors.ui.seaGreen.hex`

We have different places where we could use CSS variables for colors. We want to add support for the Sbanken theme in these components as well. The most important thing is to have type safety all the way. 

In this example, I could have mistyped one of the colors, e.g. `var(--color-mint-green-25)`.
`            <Area
              dataKey="value"
              fill="var(--color-mint-green-25)"
              fillOpacity={1}
              isAnimationActive={false}
              stroke="var(--color-success-green)"
              strokeWidth={2}
            />
            <CartesianGrid
              stroke="var(--color-emerald-green)"
              opacity={0.25}
              vertical={false}
            />`

thaytharma:
> Why do we want to expose hex/rgb values for components? Are we considering dependency on static hex/rgb values somewhere and not on theme-based CSS variables only?

Good question!
All the colors and values were purely based on the[ colors table](https://eufemia.dnb.no/quickguide-designer/colors/#dnb-color-table-for-web). This includes hex, css-variables, and rbg values.

thaytharma:
@tujoworker , I'll close this MR for now, we've integrated this `colors` approach in our shared libraries for now and will swap it out with Eufemia's version if/when it's available 😊

PR #2817, opened 2023-10-31
fix(Flex): export Flex correct

langz:
https://dnb-it.slack.com/archives/CMXABCHEY/p1698744253915779

Fixes the following error in TS, when trying to import and use Grid.X like `Grid.Item`:
`Property 'Item' does not exist on type 'typeof import("<my_path>/node_modules/@dnb/eufemia/components/grid/Grid")'.`


langz:
Fixed in https://github.com/dnbexperience/eufemia/pull/2819

PR #2823, opened 2023-11-01
chore(Storybook): update to latest

tujoworker:


langz (approved):
📖

tujoworker:
Looks like there is an deps lock file audit issue. It did work fine with Yarn v4 when I tested i locally. So this PR relies on #2825 in order to run properly.

langz:
> Looks like there is an deps lock file audit issue. It did work fine with Yarn v4 when I tested i locally. So this PR relies on #2825 in order to run properly.

Thanks for following-up on this, I saw the build failing after I had approved 🏎️ 🏁

PR #2826, opened 2023-11-02
chore(Section): ensure invalid style_type has still a valid fallback

tujoworker:
Whit this PR we do miss out of a valid fallback, when an invalid `style_type` was given (regarding the issue in the Portal Tools on the Sbanken brand theme).

```
<Section style_type="invalid">
  will be white now, instead of black
</Section>
```


langz (approved):
🏳️

PR #2841, opened 2023-11-02
docs(PaymentCard): sets card_number as required

langz:
I did a quick analysis, trying to figure out if `card_number` actually is optional(as the docs says) or required.
But it looks to me as `card_number` is actually required from the code's point of view:
- https://github.com/dnbexperience/eufemia/blob/main/packages/dnb-eufemia/src/extensions/payment-card/PaymentCard.js#L63
- https://github.com/dnbexperience/eufemia/blob/main/packages/dnb-eufemia/src/extensions/payment-card/PaymentCard.d.ts#L88


joakbjerk (approved):
Yeah it actually looks to be required judging by the code 🤔 💳

PR #2849, opened 2023-11-06
docs: fix font docs

langz:
When running `/quickguide-designer/fonts/#font-weights` [locally](http://localhost:8000/quickguide-designer/fonts/#font-weights), I get the following errors:
```
Warning: validateDOMNesting(...): <p> cannot appear as a descendant of <p>.
    at p
    at ElementInstance
```

What's happening is that `\n` somehow gets inserted when a prettier formats a `<p>`, leading to the content being wrapped in a `<p>`, so that we get `<p>`'s inside of `<p>`'s.
Here's a link to a related issue: https://github.com/storybookjs/storybook/issues/18921 

If this would only lead to the console error pasted above, it all would be fine, but in some scenarios when wrapping a `p` in a `p`, the visual result is actually different(so what's displayed in the docs is actually not what we want), compare the [fix](https://eufemia-git-docs-fix-font-docs-eufemia.vercel.app/quickguide-designer/fonts/#font-weights) vs. what's on [main](https://eufemia.dnb.no/quickguide-designer/fonts/#font-weights). Screenshot as reference:

Screenshot 2023-11-06 at 13 39 33

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/quickguide-designer/naming-conventions.mdx
+++ b/packages/dnb-design-system-portal/src/docs/quickguide-designer/naming-conventions.mdx
@@ -36,7 +36,7 @@ The same can apply to components such as cards, form rows, etc. Responsive break
 
 Eufemia for the web uses a simple naming system:
 
-<div className="typography-box">Color name + percentage</div>
+<TypographyBox>Color name + percentage</TypographyBox>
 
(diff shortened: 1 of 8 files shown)


tujoworker (approved):

langz (commented):
Looks great, thanks for purposing a better alternative @tujoworker 🥇

Issue #2850, opened 2023-11-06
Searching and selecting country in PhoneNumber, selects wrong country.


langz:
## 🐛 Bug Report

Searching and selecting country in PhoneNumber, selects wrong country.

Try it out here: https://eufemia.dnb.no/uilib/extensions/forms/feature-fields/PhoneNumber/demos/#empty

https://github.com/dnbexperience/eufemia/assets/1359205/787749b8-7e8b-47d7-8f18-ba3ba0121bde

## Eufemia Version

v10.11.0

PR #2852, opened 2023-11-06
fix(PhoneNumber): ensure correct selection of value

### LINKED ISSUE #2862: (PhoneNumber): Sort country codes by display language
## 🚀 Feature Proposal

As of today, the country codes in the autocomplete-dropdown are sorted by `iso` and not the `i18n` values used to display the country.

https://github.com/dnbexperience/eufemia/assets/25927156/36547b7a-b120-48ec-bd26-8b61d4306e46

## Motivation

Make it easier and more logical for the user to look for country codes when scrolling through the dropdown. Going from A to Ø(sterrike) and then suddenly H(viterussland) is a tad confusing. 

## Example

The user would scroll through the list and find the country codes sorted alphabetically by the display names shown to the user, instead of the `iso`-value

## Pitch

Seems like good UX.

langz:
Added a test for the issue https://github.com/dnbexperience/eufemia/issues/2850

Not 100% sure about the actual fix yet 🤔 
If someone else has a clue, please give it a go, and add a commit to this branch 🙇 

---

Update:

- This PR relies on PR #2863
- Fixes #2862 and #2850

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber/PhoneNumber.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber/PhoneNumber.tsx
@@ -20,6 +20,11 @@ export type Props = FieldHelpProps &
     width?: 'large' | 'stretch'
     onCountryCodeChange?: (value: string | undefined) => void
     onNumberChange?: (value: string | undefined) => void
+  } & {
+    /**
+     * For internal testing purposes
+     */
+    noAnimation?: boolean
   }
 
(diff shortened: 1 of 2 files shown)


tujoworker:
We probably should split the two commits in separate PRs. 
Also, added a sort/order to fix #2862.

What's changed? Now we process the whole list first when focusing instead of the first render.
This way, we skip processing it during SSR. Because its not visible anyway.

langz (commented):
Looks good to me 🥇

tujoworker (approved):

PR #2857, opened 2023-11-06
feat(HelpButton): Add inline display variant

Sundfjord:
This PR will add a new variant of the `HelpButton` component, that displays its contents in an inline element, as so:
image

The help content container will by default be appended to the HelpButton instance's parent element, but a `contentId` may also be specified to place the help content container in any existing DOM element.

In addition to the necessary changes in the `HelpButton` component, `FormLabel` now accepts a new `help` prop which will by default display what is sent with the `contents` prop in an inline help button. The changes to the component are mainly done to ensure correct positioning of both the help button itself and its contents regardless of label direction.

Test cases have been added for this new variant.


tujoworker:
Closing this in favor of #4280

PR #2865, opened 2023-11-08
fix(Autocomplete): ensure on_change is firered during dataset update

### LINKED ISSUE #2847: Autocomplete not firing on_change in async
When using Autocomplete in async mode, where the options returned by the async function **differ** between different calls (in our use case the string the user searches for is first filtered on the server - since the options are so many) the component will not trigger when there is only one option available. We have several users being tricked by this - thinking they gave changed the option but not being registered since on_change is not called.

In particular we notice this when there is only 1 option available:

1. The user types something in the autocomplete box
2. This triggers a async call to the `on_type` handler
3. Since the entry typed by the user is so specific only one option is available, so the `on_type` handler calls `updateData([{selectedKey: "entry1", content: "Entry 1"}])`
4. The user selects "Entry 1"
5. `on_change` is successfully called
6. A while later the user types in the autocomplete again, again something very specific
7. Again, only one option is available so `on_type` calls `updateData([{selectedKey: "Specific1", content. "Specific 1"}])`
8. The user clicks the (only available) entry "Specific 1"
9. **Critically** `on_change` is Not called (only `on_hide`) => mismatch between what the user has selected and registered.

The *same* problem occurs if the user at point 6, 7, 8 types something where the area multiple options available but clicks the first one.

It seems to me that the code is incorrectly detecting a "change" using a index based lookup (i.e. is it still the first item) instead of checking the key - a logic that works for static lists but not whit the async callback functionality (which is a critical part of the autocomplete functionality).

@tujoworker is this something  you could either consider a point release for or suggest some workarounds? (I have tried various ways to work around but nothing working consistently and thus creating problems for our users)

langz:
Added a test for the issue https://github.com/dnbexperience/eufemia/issues/2847

Not 100% sure about the actual fix yet 🤔
If someone else has a clue, please give it a go, and add a commit to this branch 🙇

---

- [CSB](https://codesandbox.io/s/hardcore-worker-x5nrv7?file=/src/App.tsx) with a reproduction.
- Will fix #2847

### DIFF
--- a/packages/dnb-eufemia/src/components/autocomplete/__tests__/Autocomplete.test.tsx
+++ b/packages/dnb-eufemia/src/components/autocomplete/__tests__/Autocomplete.test.tsx
@@ -14,13 +14,7 @@ import {
   mockImplementationForDirectionObserver,
   testDirectionObserver,
 } from '../../../fragments/drawer-list/__tests__/DrawerListTestMocks'
-import {
-  cleanup,
-  fireEvent,
-  render,
-  act,
-  waitFor,
-} from '@testing-library/react'
+import { fireEvent, render, act, waitFor } from '@testing-library/react'
 import {
   DrawerListData,
   DrawerListDataObject,
@@ -1200,6 +1194,10 @@ describe('Autocomplete component', () => {
       />
     )
 
+    expect(document.querySelector('input').value).toBe(
+      Array.from(mockData[2].content).join(' ')
+    )
+
     // 2. Then update the data
     rerender(
       <Autocomplete
@@ -1209,6 +1207,10 @@ describe('Autocomplete component', () => {
         data={newMockData}
         value={2}
       />
+    )
+
+    expect(document.querySelector('input').value).toBe(
+      Array.from(newMockData[2].content).join(' ')
     )
 
     // 3. And change the value again
@@ -1222,8 +1224,6 @@ describe('Autocomplete component', () => {
       />
     )
 
-    // It should not trigger the resetSelectionItem method
-    expect(on_change).toHaveBeenCalledTimes(0)
     expect(document.querySelector('input').value).toBe(
       newMockData[1].content
     )
@@ -1238,6 +1238,8 @@ describe('Autocomplete component', () => {
         value={null}
       />
     )
+
+    expect(document.querySelector('input').value).toBe('')
 
     fireEvent.change(document.querySelector('input'), {
       target: { value: 'cc' },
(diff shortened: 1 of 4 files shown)


packages/dnb-eufemia/src/components/autocomplete/__tests__/Autocomplete.test.tsx
```diff
@@ -1280,6 +1280,74 @@ describe('Autocomplete component', () => {
...
+    const on_change = jest.fn()
+    const on_type = jest.fn(onTypeHandler)
+
+    render(
+      <Autocomplete
+        {...mockProps}
+        on_change={on_change}
+        on_type={on_type}
+        data={mockDataA}
+        mode="async"
```
langz, line 1304: As of now, `on_change` is only called once in this test, but if we would have had `mode="sync"` instead of `mode="async"`, `on_change` is called multiple times.

To me, it's a bit strange, as I assumed that `mode` would only do "visual changes" when changing.
Based on what's said in the [docs](https://eufemia.dnb.no/uilib/components/autocomplete/properties/): _mode (optional) if set to async, it prevents showing the "no options" message during typing / filtering. Defaults to sync._

vron:
So are you then saying that I can solve my problem by setting mode="sync" - I hade (perhaps incorrectly) assumed that I had to set it to async when using an async/callback way to set the data rather then setting the data as a static array when rendering the component?

tujoworker:
> So are you then saying that I can solve my problem by setting mode="sync" 

No, you did it correctly. I work on a fix for it. @langz did just find (probably) another issue 🙏

packages/dnb-eufemia/src/components/autocomplete/__tests__/Autocomplete.test.tsx
tujoworker: While updating the data, and if the current item does not have the same key, we need to remove the selection, because it is something different.

So, while removing the selection, we also call `on_change` but with a undefined return value.

That's why we need to expect several more `on_change` calls.

My commit does ensure better consistency in that behavior.

tujoworker (approved):

langz (commented):
Thanks for the great effort @tujoworker 👏 
I did some testing in the following [CSB](https://codesandbox.io/s/eloquent-brown-cgdd8t), and the specific/reported issue seems to be fixed 🤗 

However, I did encounter a scenario which I think is a bit misleading for the user, as I as a user did think that my selected value would always be kept until I select a new one. But that's not the case in the following scenario:
- User types "A"
- User selects "AAA"
- User types "something" or "c"
- User presses outside of the autocomplete(on blur) and the selected value is still "AAA" as expected.
- User types "b"
- User presses outside of the autocomplete(on blur) and the selected value is now empty.

I feel like as a user, the these two actions(type/search and press outside) should result in the same result.

It might be that this is intentional, and that there's no better way from a technical point of view 🤔 

Here's a video of the scenario described:

https://github.com/dnbexperience/eufemia/assets/1359205/2cf08878-a13a-4c77-ba8b-f0d2065f8b65

PR #2869, opened 2023-11-09
fix(PhoneNumber): keep selected countryCode value on blur

tujoworker:
The change: When the user changes the countryCode value – without making a "selection" – and then the input blurs, it will show the current selected value. This way, it will not be possible to "clear" a country code or add a custom. 

But that also requires us to deliver all possible available codes. The error rate will be way lower for sure. Now, its more like a Dropdown, but with fee text search ability.

Here's [a demo](https://eufemia-git-feat-phone-number-keep-value-eufemia.vercel.app/uilib/extensions/forms/feature-fields/PhoneNumber/demos/).

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber/PhoneNumber.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber/PhoneNumber.tsx
@@ -73,16 +73,11 @@ function PhoneNumber(props: Props) {
       ? value.match(/^(\+[^ ]+)? ?(.*)$/)
       : [undefined, '', '']
 
-  const getCountryData = ({ filter = null } = {}) => {
-    const lang = sharedContext.locale?.split('-')[0]
-    return countries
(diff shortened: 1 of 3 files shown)


langz (approved):
I did some testing, and it works as expected 💯 

An additional thought, not 100% related to this fix/PR, is if this feature(that the user can't unselect the country code) will "limit" the usage of the component? As if this component would be included in a form with a lot of optional fields(non-required) then this field would always have a value 🤔

tujoworker:
Good point! That's actually anyway an issue. So I suggest we should only show an error when required, if the second field is empty. And maybe only return the code value if a phone number value exists?

langz:
> Good point! That's actually anyway an issue. So I suggest we should only show an error when required, if the second field is empty. And maybe only return the code value if a phone number value exists?

Agree agree 👍 
Wouldn't that be a "breaking change"? Maybe we wouldn't have to treat it as a breaking change in terms of semver, since these features are under the "beta"-flag 😎

PR #2870, opened 2023-11-09
feat(Autocomplete): add keepSelection property

tujoworker:
like `keep_value_and_selection` – but would not reset to the input value during input field blur.


tujoworker:
Rethinking this, I guess this makes no sense UX wise.

PR #2874, opened 2023-11-10
feat: add "use client" to components that use client side only React APIs (support for RSC)

tujoworker:
We have received several request to add a ["use client"](https://react.dev/reference/react/use-client) directive on top of each React component file that uses client side only React APIs.

The reason is because Next.js takes in use a beta version of React to make RSC work (which is not OK in my opinion – its a super aggressive move 😣).

To make a first move on this from our side, I think we should automate this. And one way to do so, is to use Babel.js – because we use it in our tool-chain.

There is already a very well written and tested plugin, written by Kyle Tsang I suggest to try out.

👉 [babel-plugin-transform-next-use-client](https://github.com/kyletsang/babel-plugin-transform-next-use-client)

This must be the first time we take in use a dependency with 0 GitHub stars. If it turns out to work as expected, we should all give it a star of course!

**Other notes:**

- I have tested it locally and also added a post build test.
- We skip rollup builds (UMD/ESM) bundles. But they are not uses in Next.js anyway.
- Time will tell of we need to [add more custom config](https://github.com/kyletsang/babel-plugin-transform-next-use-client#react-client-apis-in-custom-modules).

### DIFF
--- a/packages/dnb-eufemia/babel.config.js
+++ b/packages/dnb-eufemia/babel.config.js
@@ -77,6 +77,10 @@ const productionPlugins = [
   ],
 ]
 
+if (global.bundler !== 'rollup') {
+  productionPlugins.push('babel-plugin-transform-next-use-client')
+}
+
 if (typeof process.env.BABEL_ENV !== 'undefined') {
   productionPlugins.push([
     'babel-plugin-search-and-replace',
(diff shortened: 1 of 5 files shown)


langz (approved):
Great, let's give it a go 👍 
Better than adding these lines of code in all our components, etc.

PR #2882, opened 2023-11-10
feat(forms): improved state management and reacting to more changed props

henit:
There has been som limitations in the handling of changed external props, as well as the prioritization between different error sources. The different issues are closely linked together, so here are a summary of the changes:

- Errors from Provider and local errors in field components could overwrite each other incorrectly because they where put in the same state. They now have separate states in `useDataValue` and both are taken into account when returning an error to display in the component (as part of the return value of the hook)
- Depdendencies on external props (like `errorMessages`, `schema` and `validator`) in validator callbacks could lead to endless loops because they revalidated when something changed, creating a new error object and if the props was recreated (like if a field component did not memoize `errorMessages`), that would trigger re-validation giving a new error object and so on (endless loop). To avoid this problem, all state in both `useDataValue` and `Provider` are now put in `useRef` instead of `useState`, and a function check if validation leads to the same error that is already in the error state, and if so, it skips replacing it and re-rendering the component.
- Some props was not monitored for external changes, or not re-validated so that a new value kept the error state from the previous value.
- Custom hooks for mount effects (only run on mount) and update effects (not running on mount), with tests.
- A custom hook ensure that race conditions does not cause problems. An example was when a unit test ran, the validation triggered by the first mount (validating the initial value) and then re-validating after a change callback, but some times the mount validation process finished after the change validation process, setting the error state based on the initial value, while a changed value was in the state. This is avoided using the new `useProcessManager` hook in `useDataValue`.
- Rewritten validation code for better readability
- Various small improvements and bugfixes


packages/dnb-eufemia/src/extensions/forms/Field/OrganizationNumber/OrganizationNumber.tsx
```diff
@@ -9,25 +9,35 @@ export type Props = StringComponentProps & {
...
+      required: tr.organizationNumberErrorRequired,
+      pattern: tr.organizationNumberErrorPattern,
+      ...props.errorMessages,
+    }),
+    [tr, props.errorMessages]
+  )
+  const mask = useMemo(
+    () =>
+      omitMask
+        ? [/\d/, /\d/, /\d/, /\d/, /\d/, /\d/, /\d/, /\d/, /\d/]
```
tujoworker, line 27:
```suggestion
        ? Array(9).fill(/\d/)
```

😄

packages/dnb-eufemia/src/extensions/forms/Field/SelectCountry/SelectCountry.tsx
```diff
@@ -33,7 +38,7 @@ function SelectCountry(props: Props) {
           title={country.i18n[lang] ?? country.i18n.en}
         />
       )),
-    [sharedContext.locale]
+    [lang]
```
tujoworker, line 41: 👍

packages/dnb-eufemia/src/extensions/forms/hooks/useDataValue.ts
```diff
@@ -122,130 +131,187 @@ export default function useDataValue<
...
+  }, [
+    emptyValue,
+    required,
+    startProcess,
+    persistErrorState,
+    clearErrorState,
+  ])
+
+  useUpdateEffect(() => {
+    if (typeof schema !== 'object') {
```
tujoworker, line 277: "null" will pass (here it will yield false), but thats probably OK and covered in ajv.compile?

packages/dnb-eufemia/src/extensions/forms/hooks/useDataValue.ts
```diff
@@ -324,47 +389,47 @@ export default function useDataValue<
...
       iterateElementIndex,
-      value,
       continuousValidation,
       onChange,
       validateValue,
       dataContextHandlePathChange,
       setShowFieldBlockError,
       handleIterateElementChange,
       fromInput,
+      forceUpdate,
```
tujoworker, line 406: Should this be documented or is it for "internal" use only?

packages/dnb-eufemia/src/extensions/forms/hooks/__tests__/useDataValue.test.ts
tujoworker, line 1: This change is not needed technically. But I'm fine with it.

tujoworker (approved):
Well done 👏

packages/dnb-eufemia/src/extensions/forms/hooks/useDataValue.ts
henit: internal only. It is just to keep full control of how data is being updated, instead of just relying on it happening by itself through changes in `useState` values

packages/dnb-eufemia/src/extensions/forms/hooks/useDataValue.ts
henit: `undefined` doesn't work. I haven't tested `null`. I can check

henit: Null didn't fail in ajv, but since the schema was just checked to be truthy (which would ignore both undefined and null) a few other places, I changed it to be the same here to be more consistent.

PR #2887, opened 2023-11-13
chore: adds Card stories

langz:
Adds card stories, using fields from forms extension.

THOUGHTS:
Seems like my stories is not working as expected as of right now, as I'm doing:
```
<Card stack> 
   <>
      ...component1
      ...component2
   </>
</Card>
```
And must rather do the following for it to work:
```
<Card stack> 
   ...component1
   ...component2
</Card>
```

Should/could the first example with React fragment(`<></>`) work?

--------------------------------------------------------------------------------

Added one more commit where I duplicate it, instead of sharing it, to make it work/look as expected.

### DIFF
--- a/packages/dnb-eufemia/src/components/card/stories/Card.stories.tsx
+++ b/packages/dnb-eufemia/src/components/card/stories/Card.stories.tsx
@@ -13,317 +13,88 @@ export default {
   title: 'Eufemia/Components/Card',
 }
 
+const AllFormFields = () => {
+  return (
+    <>
+      <Field.NationalIdentityNumber
+        path="/ssn"
+        label="NationalIdentityNumber"
+      />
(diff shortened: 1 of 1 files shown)


tujoworker:
Since the point of the Card component and its possibilities is missing out when not delivering a stack of components, I suggest to get each component programmatically. 

See latest commit. I used your first commit and did not compare it, just in case, you made a change to the list of components.

langz:
> Since the point of the Card component and its possibilities is missing out when not delivering a stack of components, I suggest to get each component programmatically.
> 
> See latest commit. I used your first commit and did not compare it, just in case, you made a change to the list of components.

Thanks, looks good, I'm very happy with that suggestion 🙏

tujoworker (approved):

PR #2888, opened 2023-11-13
feat(Radio): Sbanken styling

Sundfjord:
Sbanken theming of the Radio component.

A fairly significant number of CSS variables has been introduced here. I've put them in a mixin in order to avoid listing all these variables in three different sass files (default, ui and sbanken theme files), but I am far from sure this is the best way to approach this, so please share your thoughts on this upon review.


packages/dnb-eufemia/src/components/radio/style/themes/radio-color-scheme-mixin.scss
```diff
@@ -0,0 +1,63 @@
+@mixin radioColorScheme($theme: 'ui') {
+  @if $theme == 'ui' {
```
tujoworker, line 2: We did not touch this kind of solution in today's meeting. But I would much prefer to have these vars in the theme files, instead of this logic. 

Because its a step further away from where you think it should live. And with that we gain an imperative way of distribute/handle code.

tujoworker (approved):
Very nice outcome! I just think, we should re-think the mixin approach. Let me know if you have any further arguments or thoughts I need to know about 😉

packages/dnb-eufemia/src/components/radio/style/themes/radio-color-scheme-mixin.scss
Sundfjord: I agree with the theme files being the correct place to place the variables. The primary reason for the mixin was to avoid the clutter the variables create in the main style file, but I'm happy to change it if you consider it better!

tujoworker: I'm fine to keep `@include radioColorScheme();` as is, but move the vars into the theme files.

PR #2894, opened 2023-11-13
docs(PhoneNumber): remove unsupported props

tujoworker:
This way we can simply list props that are not supported for various reasons:

```mdx
<DataValueReadwriteProperties
  type="string"
  omit={[
    'layout',
    'label',
    'labelDescription',
    'labelSecondary',
    'emptyValue',
  ]}
/>
```


langz (approved):
Nice nice 🥳 🎆

joakbjerk (approved):

PR #2900, opened 2023-11-14
fix(forms): Field block error handling

henit:
FieldBlock did not receive errors from initial renders of inner field components, only after change callbacks to the value. Fixed, added example in portal and unit test.


packages/dnb-eufemia/src/extensions/forms/hooks/useDataValue.ts
```diff
@@ -401,15 +406,17 @@ export default function useDataValue<
       forceUpdate()
     },
     [
-      id,
+      // id,
```
tujoworker, line 409: Should this be removed?

```suggestion
```

tujoworker (approved):
Good work 🫶

PR #2903, opened 2023-11-15
fix(forms): Improved message value replacements

henit:
Errors based on schema validation in the Provider did not get values injected when providing custom error messages. This is now handeled by adding validation values to the error object (when validated in the Provider) and injecting them in the target field component with their specific error messages. With test to ensure it works as intended.


tujoworker (approved):
Nice – and the test looks 👌

PR #2904, opened 2023-11-15
feat(Checkbox): Sbanken styling

Sundfjord:
Added Sbanken theming and introduced some CSS variables that controls the color scheme of the Checkbox component, similar to how it was done in #2888 .


tujoworker (approved):
Awesome work ✨

PR #2909, opened 2023-11-17
chore: refactor styles and theme packages factory

tujoworker:
This change will not make any change to the current outcome.

We would also see the results (when run `yarn build`), because this process is all about collection SCSS files and put the imports into other SCSS "package" files.


langz (approved):
🤝

tujoworker:
Yes, Expiry had an issue, which got fixed in PR #2914 and #2915 👍

PR #2911, opened 2023-11-17
fix(Selection): correctly link id with label

### LINKED ISSUE #2842: Incorrect use of <label for=FORM_ELEMENT> in Field.Selection component
## 🐛 Bug Report

```
Incorrect use of <label for=FORM_ELEMENT>
The label's for attribute doesn't match any element id. This might prevent the browser from correctly autofilling the form and accessibility tools from working correctly.
To fix this issue, make sure the label's for attribute references the correct id of a form field.
```

## To Reproduce

```
<Field.Selection
  variant="button"
  label="Label text"
>
  <Field.Option value="foo" title="Foo!" />
  <Field.Option value="bar" title="Baar!" />
</Field.Selection>
```

Here's a screenshot of the same error:
image

## Thoughts

I believe the problem is related to that `ToggleButton`'s have some kind of own label, and that we should rather pass the related label properties to the ToggleButtonGroupContext.Provider or so 🤔 
Instead of sending it to FieldBlock...

tujoworker:
Fixes #2842

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/Selection/Selection.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/Selection/Selection.tsx
@@ -162,6 +162,7 @@ function Selection(props: Props) {
             >
               {options.map((option, i) => (
                 <ToggleButton
+                  id={options.length > 0 ? id : undefined}
                   key={`option-${i}-${option.value}`}
                   text={option.title}
                   on_change={option.handleSelect}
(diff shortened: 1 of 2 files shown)


langz (approved):
Great 👍 
Gave it a quick run locally, and it works, seems like it fixes https://github.com/dnbexperience/eufemia/issues/2842 ?

tujoworker:
Yes, true 👍

PR #2913, opened 2023-11-17
fix(FieldBlock): fix support for stretch alignment

tujoworker:
To make a field stretch within the Card (PR #2907), we need an improvement in the FieldBlock.


langz (approved):
Great finding 🥇

PR #2916, opened 2023-11-17
chore: where -> were

langz:


tujoworker (approved):
So many 😌

langz:
> wow, so many 😌

Yeah, not so important, but I found one that I wanted to improve, and then I did a quick search looking for more occurrences 👍

PR #2917, opened 2023-11-17
feat(Breadcrumb): correct gap between items and fix overlapping focus ring

snorrekim:
Shrinked the focus effect so it doesn't overlap with the other text. Also fixed the left padding in collapsed mode. Both seem to be overlooked errors caused by the same change.

[Slack Thread](https://dnb-it.slack.com/archives/CMXABCHEY/p1699349835359029).


tujoworker (changes requested):
Nice work so far!

Actually, this solution is not good for the DNB styles, because the space between the items should be larger, and therefor, the button padding does not need to be changed.

snorrekim:
> Nice work so far!
> 
> Actually, this solution is not optimal for the DNB styles, because the space between the items should be larger than currently, and therefor, the button padding does not need to be changed.

I've fixed the distances

tujoworker (approved):
100% 👌

PR #2918, opened 2023-11-17
feat(forms): include styles by default

tujoworker:
This PR includes the forms extension styles in the components package. This means, that theses styles do not need to be imported separately.

These styles are very small in size (6 KB), and will not grow substantially either.

It also allows automated generation of theme files, when needed.


packages/dnb-design-system-portal/src/docs/uilib/extensions/forms.mdx
```diff
@@ -130,12 +128,23 @@ In this example, all state data, validation process and error handling are done
 
 ## Getting started
 
-You need to import the styles once:
+You import the components from with scopes, such as `Form` and `Field`:
```
joakbjerk, line 131:
```suggestion
You import the components from scopes, such as `Form` and `Field`:
```

joakbjerk (approved):
Looks good 🙏

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms.mdx
```diff
@@ -28,15 +28,13 @@ import FormDiagram2 from 'Docs/uilib/extensions/forms/form-diagram-2.png'
 Field components can be used directly as they are, for example `Field.Email`:
 
 ```jsx
-import '@dnb/eufemia/extensions/forms/style'
```
langz, line 31: Do we need to add to the release docs that users of Eufemia that have done this import already, should remove it?
Or can they still have the import in their code, without it leading to any problems?.

If we don't add it in our release docs, I think we should remember to add a comment about this in slack when releasing this new feature 🙏

langz (approved):

PR #2920, opened 2023-11-20
fix(PhoneNumber): return country code only when a number is given

tujoworker:
- With this change, we only return country code when a phone number exists (`onChange`). This effects also the `onFocus` and `onBlur` event. Else `emptyValue` is returned.
- [x] Needs this change from PR #2922 and #2923

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber/PhoneNumber.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber/PhoneNumber.tsx
@@ -86,22 +86,33 @@ function PhoneNumber(props: Props) {
   const langRef = React.useRef(lang)
 
   /**
-   * Update countryCode and phoneNumber when value from outside changes.
-   * Use memo to update refs in sync
+   * We do not process the whole country list at the first render.
+   * Only when the Autocomplete opens (focus).
+   * To achieve this, we use memo instead of effect to update refs in sync.
+   *
+   * We set or update the data list depending on if the countrycode changes or lang changes.
+   * We then update countryCode and phoneNumber when value changes.
    */
   useMemo(() => {
     const [countryCode, phoneNumber] = splitValue(props.value)
     phoneNumberRef.current = phoneNumber
 
     if (
-      (countryCode && countryCodeRef.current !== countryCode) ||
-      !countryCodeRef.current
+      !countryCodeRef.current ||
+      (countryCode && countryCode !== countryCodeRef.current)
     ) {
       countryCodeRef.current = countryCode || defaultCountryCode
 
       dataRef.current = getCountryData({
         lang,
         filter: countryCodeRef.current,
+      })
+    }
+
+    if (lang !== langRef.current) {
+      langRef.current = lang
+      dataRef.current = getCountryData({
+        lang,
       })
     }
   }, [props.value, lang])
@@ -117,19 +128,6 @@ function PhoneNumber(props: Props) {
       : emptyValue
     updateValue(newValue)
   }, [props.value, emptyValue, updateValue])
-
-  /**
-   * Update whole contry list when lang changes.
-   * Use memo to update refs in sync.
-   */
-  useMemo(() => {
-    if (lang !== langRef.current) {
-      langRef.current = lang
-      dataRef.current = getCountryData({
-        lang,
-      })
-    }
-  }, [lang])
 
   const handleCountryCodeChange = useCallback(
     ({ data }: { data: { selectedKey: string } }) => {
@@ -179,13 +177,13 @@ function PhoneNumber(props: Props) {
     ({ updateData }) => {
       if (dataRef.current.length < 10) {
         dataRef.current = getCountryData({
-          lang: sharedContext.locale?.split('-')[0],
+          lang,
         })
         updateData(dataRef.current)
       }
       handleFocus()
     },
-    [handleFocus, sharedContext.locale]
+    [handleFocus, lang]
   )
 
   return (
--- a/packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber/PhoneNumber.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber/PhoneNumber.tsx
@@ -124,6 +124,7 @@ function PhoneNumber(props: Props) {
    */
   useMemo(() => {
     if (lang !== langRef.current) {
+      langRef.current = lang
       dataRef.current = getCountryData({
         lang,
       })
--- a/packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber/PhoneNumber.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber/PhoneNumber.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo, useContext, useCallback } from 'react'
+import React, { useMemo, useContext, useCallback, useEffect } from 'react'
 import { Autocomplete, Flex } from '../../../../components'
 import { InputMaskedProps } from '../../../../components/InputMasked'
 import classnames from 'classnames'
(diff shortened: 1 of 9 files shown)


packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber/stories/PhoneNumber.stories.tsx
```diff
@@ -1,14 +1,28 @@
...
+  }, [])
   return (
-    <Form.Handler onSubmit={console.log}>
-      <Field.PhoneNumber required validateInitially path="/phoneNumber" />
-      <Form.SubmitButton top />
-    </Form.Handler>
+    <Field.PhoneNumber
+      required
+      value={state}
+      // validateInitially
```
henit, line 19: //?

packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber/PhoneNumber.tsx
```diff
@@ -74,58 +76,109 @@ function PhoneNumber(props: Props) {
...
+    const [countryCode, phoneNumber] = splitValue(props.value)
+    phoneNumberRef.current = phoneNumber
+
+    if (
+      (countryCode && countryCodeRef.current !== countryCode) ||
+      !countryCodeRef.current
+    ) {
+      countryCodeRef.current = countryCode || defaultCountryCode
+
+      if (lang === langRef.current) {
```
henit, line 103: Since `lang` is not in the memo dependencies array, won't this lead to the memo not being re-called if `lang` is changed?

packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber/PhoneNumber.tsx
```diff
@@ -74,58 +76,109 @@ function PhoneNumber(props: Props) {
...
+    [countryCodeRef, emptyValue, handleChange, onNumberChange]
   )
 
-  const onFocusHandler = ({ dataList, updateData }) => {
-    // because there can be more than one country with same cdc
-    if (dataList.length < 10) {
-      updateData(
-        getCountryData({ lang: sharedContext.locale?.split('-')[0] })
-      )
+  const onFocusHandler = ({ updateData }) => {
```
henit, line 176: Any good reason not to `useCallback` this one?

henit (changes requested):

packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber/stories/PhoneNumber.stories.tsx
tujoworker: Actually, Storybook in Eufemia is used as a place for drafting and does not need to be that height quality.

packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber/PhoneNumber.tsx
tujoworker: This is only used for the first render. When lang changes, there is a new check we. 
This is because we only process the first/needed country. But later we process all.

packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber/PhoneNumber.tsx
tujoworker: No, we can add it 👍

packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber/PhoneNumber.tsx
henit: Not just not being recalled, but actually having the old value even if it was changed before memo is recalled because `value` was changed.... Which practically means that if will never run true?

henit: ignoring the exhausive-deps rule is risky 😅

tujoworker: It looks like we can remove that change 👍

packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber/PhoneNumber.tsx
tujoworker: Done

packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber/PhoneNumber.tsx
tujoworker: Also added deps ✅

langz (approved):
Looks very good to me.
Should we add a few lines to the docs, describing this functionality, perhaps in the info or events section? 🙏

packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber/PhoneNumber.tsx
```diff
@@ -74,61 +75,117 @@ function PhoneNumber(props: Props) {
...
+      ? joinValue([countryCode, phoneNumber])
+      : emptyValue
+    updateValue(newValue)
+  }, [props.value, emptyValue, updateValue])
+
+  /**
+   * Update whole contry list when lang changes.
+   * Use memo to update refs in sync.
+   */
+  useMemo(() => {
```
henit, line 125: If the hook is not returning anything, I would suggest using `useEffect` instead of `useMemo`. Another reason for using effect instead, is that a useMemo is running in the middle of the render process, and if you change the `dataRef` then, it might have consequences (now or in the future) for logic that is part of the render process and depend on that ref, when the logic of the render running before this hook has one value in the ref, while after it is another one. A `useEffect` runs after the render, and triggers a re-render if needed, which is less likely to cause issues like the described cases.

packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber/PhoneNumber.tsx
```diff
@@ -74,61 +75,117 @@ function PhoneNumber(props: Props) {
...
+    if (lang !== langRef.current) {
```
henit, line 126: I see that `langRef` is set to the initial value when calling `useRef(lang)`, then used in this if and nowhere else. If I understand the flow correctly, the `dataRef` need to be updated every time lang changes. This will not happen correctly now if I understand the current logic. The initial value of the `lang` prop will be put in the ref at first render, then the ref is never updated (as far as I can see). So if the `lang` change to something other than what it was initially, it will trigger this change in `dataRef`. However, if it is changed back to the original value it was during the first render, `lang` will not be !== `langRef.current` (since the ref was never updated after initial rende), so then the `dataRef` would remain what it was from the previous lang....

henit: The new value recieved by the `useRef` call itself will not update the ref, hence doc: https://react.dev/reference/react/useRef#parameters

henit (changes requested):

packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber/PhoneNumber.tsx
tujoworker: yeah, but in this case, we want the change in sync. We have just refs here, and can't trigger a re-render via state update.

packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber/PhoneNumber.tsx
tujoworker: Yes, added a new commit to make it possible to switch locales multiple times. Thanks for finding this issue 🙏

packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber/PhoneNumber.tsx
tujoworker: Nice, we still could enhance the code readability with your feedback. Thanks 🙏

henit (approved):

PR #2922, opened 2023-11-21
fix(Autocomplete): enhance reactivity by value change from outside

tujoworker:


langz (approved):
Looks good, thanks for the demo/explanation at work today 🏢  💻

PR #2927, opened 2023-11-21
fix(FormLabel): don't apply margin-right when no content was given

tujoworker:


packages/dnb-eufemia/src/components/form-label/FormLabel.tsx
```diff
@@ -97,7 +100,9 @@ export default function FormLabel(localProps: FormLabelAllProps) {
       size && `dnb-h--${size}`,
       isInteractive && 'dnb-form-label--interactive',
       createSkeletonClass('font', skeleton, context),
-      createSpacingClasses(props),
+      createSpacingClasses(
+        content ? { right: 'small', ...props } : omitSpacingProps(props)
```
langz, line 104: Why do we add a right margin to most scenarios(I believe most scenarios has either text or children)?
Based on the commit message, I believed this PR would [don't apply margin-right when no content was given](https://github.com/dnbexperience/eufemia/pull/2927/files#top), but too me this change also adds margin right to most scenarios? 🤔  See all the updated screenshot/snapshot

tujoworker: I think, its still good to have a right-margin by default, men now we set it only when either text or children is set.

langz (approved):

PR #2930, opened 2023-11-24
feat(Button, Anchor): fix launch icon for _blank button links

snorrekim:
* The padding around the launch icon is the same as other icons

Button now adds the launch icon itself, instead of relying on anchor's implementation. It already does that with other icons, so we shouldn't rely on Anchor's implementation for `_blank` links.


packages/dnb-eufemia/src/components/button/Button.js
```diff
@@ -177,6 +178,13 @@ export default class Button extends React.PureComponent {
       : 'button'
     if (Element === Anchor) {
       attributes.omitClass = true
+      if (
+        props.target === '_blank' &&
+        !/^(mailto|tel|sms)/.test(props.href) &&
```
tujoworker, line 183: With that we suddenly have this logic twice in our code. In the Anchor, we have integration tests for it, but not in the button.

Would it possible to still inherit this logic as before?

snorrekim: I've now moved the logic to Anchor

langz (approved):
Looks good to me.
Should we consider using the `fix` commit decoration for this?
Based on the screenshot changes, it sort of seems like it was a bug that we've fixed 🤔 
Or maybe feat is just fine?

snorrekim:
> Looks good to me. Should we consider using the `fix` commit decoration for this? Based on the screenshot changes, it sort of seems like it was a bug that we've fixed 🤔 Or maybe feat is just fine?

As I understood it, `fix` would affect release version in unexpected ways, so I've mainly used `feat` to make sure it get's included. Though I don't know how it works.

PR #2940, opened 2023-11-29
chore: rename tabs horizontal aligned screenshots

langz:
I've seen that there's a failing screenshot test in [main branch](https://github.com/dnbexperience/eufemia/commits/main):

image

Link to commits with failed builds(might be outdated, as builds has been re-run):
- https://github.com/dnbexperience/eufemia/commit/a36ffa719eb529abd80c22ab79cfa43df79f9d0b

I've tried re-running a few builds. and it seems like the screenshot tests sometimes work, and sometimes not.
Not entirely sure what makes this specific test a bit "random".

Not sure if this will "fix the build", but just gave it an attempt, by renaming and generating new screenshots 🙏 

I don't think this fixes anything, but can be seen as a minor bug report that we could look into 🙏


tujoworker (approved):
I'm fine with the change, but it should not have any effect on the randomness.
I looked at the last failing e2e test on the main branch, and this is the cause:

image

Which seems strange. I assume there is a caching issue. So if this persists, we should check it out.

PR #2942, opened 2023-11-29
chore(Icon): removes tag recieve for hand_money

langz:
Just a suggestion, as `recieve` is "not a word".

But this is perhaps "breaking"? in a sense if `recieve` was used to get the hand_money icon previously.
I quickly searched through the monorepo, and could not find any occurrences of `recieve`.

Also, not 100% sure which commit decorator to use(fix, chore, feat)...


tujoworker:
We need to make this change in the Figma file and run an update.

tujoworker:
hmm, it looks like its already fixed. And maybe we will miss out the update. So I think its valid to make this change manually 👍 

Screenshot 2023-11-30 at 12 06 54

PR #2947, opened 2023-11-30
fix(DatePicker): renders mask_placeholder from locale w/o provider

langz:
`mask_placeholder` shows `dd.mm.yyyy` and not `dd/mm/yyyy` which is the expected result, when setting locale to be `en-GB`.

[Slack thread](https://dnb-it.slack.com/archives/CMXABCHEY/p1701329166313829)
[CSB reproducing the issue](https://codesandbox.io/p/sandbox/optimistic-bassi-gsn7hw)
Expected result in the [CSB](https://codesandbox.io/p/sandbox/optimistic-bassi-gsn7hw) would be that the `mask_placeholder` should be `dd/mm/yyyy` as seen when [wrapping the DatePicker in an Eufemia Provider](https://codesandbox.io/p/sandbox/vigorous-williamson-km4rc7?file=%2Fsrc%2FApp.tsx).

### DIFF
--- a/packages/dnb-eufemia/src/components/date-picker/DatePicker.js
+++ b/packages/dnb-eufemia/src/components/date-picker/DatePicker.js
@@ -471,7 +471,10 @@ export default class DatePicker extends React.PureComponent {
       this.props,
       DatePicker.defaultProps,
       { skeleton: this.context?.skeleton },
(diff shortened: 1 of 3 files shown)


joakbjerk (approved):
Nice nice! 🔥🙏

PR #2952, opened 2023-12-01
fix(GlobalStatus): remove z-index to enhance flexibility

tujoworker:
The z-index usage was introduces in [this commit](https://github.com/dnbexperience/eufemia/commit/1fb638b12df290533e6a89e7f8135a7e23c0797e) in 2019 (v4).

Today, I can't find any good reasons why to have it anymore. Surely, it "could" break something visually. But I think we still are pretty safe going forward without a z-index.

More info in this [Slack thread](https://dnb-it.slack.com/archives/CMXABCHEY/p1700653977331569).


langz (approved):
Looks fine to me, let's give it a go🦾

PR #2954, opened 2023-12-01
chore: fix wait on non-headless local visual tests

tujoworker:
The custom `headlessTimeout` did not work the way it was implemented. Also, moving the wait action upwards makes more sense.


joakbjerk (approved):
🔥

PR #2963, opened 2023-12-04
chore(ToggleButton): remove left-over to fix Sbanken styling issue

tujoworker:
To fix the issue on failing e2e tests on the main branch, like below;

tabs-for-sbanken-have-to-match-horizontally-aligned-tabs snap-diff

we should remove an left-over style that did influence the Sbanken styles with a higher specificity.

Screenshot 2023-12-04 at 20 58 59

Why did this only happen on the main branch? Because the CSS is only there completely bundled, which creates some times a different CSS specificity.

Here is a [test-run](https://github.com/dnbexperience/eufemia/actions/runs/7091720735/job/19301471826) with the fix. The branch name starts (`startsWith(github.ref, 'refs/heads/main')`) with `main`, therefore it will run like the main branch.


langz (approved):
Great, good job figuring this out🙇‍♂️

Issue #2970, opened 2023-12-05
bug(Field.Date): required doesn't have any required/error message


langz:
## 🐛 Bug Report

`Field.Date`'s `required` property doesn't result in any required/error message as in other Field.X components

https://github.com/dnbexperience/eufemia/assets/1359205/f28c59fe-c482-4278-8ef7-e03937533704

## To Reproduce

Try removing the data/input of the `Field.Date` component/field.

https://codesandbox.io/p/sandbox/hopeful-rui-8twp6l?file=%2Fsrc%2FApp.tsx%3A12%2C1

## Expected behavior

I would expect the `Field.Date` to also show some kind of required error message, that the field is required.

PR #3017, opened 2023-12-07
fix(Field.Date): define empty value to trigger required error

joakbjerk:
proposal to fix https://github.com/dnbexperience/eufemia/issues/2970

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/Date/__tests__/Date.test.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/Date/__tests__/Date.test.tsx
@@ -31,7 +31,7 @@ describe('Field.Date', () => {
     })
 
     await userEvent.keyboard(
-      '{Backspace}{Backspace}{Backspace}{Backspace}{Backspace}{Backspace}{Backspace}{Backspace}'
+      '{Backspace>8}'
     )
 
     expect(datepicker.classList).toContain(
--- a/packages/dnb-eufemia/src/extensions/forms/Field/Date/__tests__/Date.test.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/Date/__tests__/Date.test.tsx
@@ -1,10 +1,76 @@
 import React from 'react'
-import { render } from '@testing-library/react'
+import { act, render } from '@testing-library/react'
 import Date, { Props } from '..'
+import userEvent from '@testing-library/user-event'
+
+const props: Props = {}
 
 describe('Field.Date', () => {
   it('should render with props', () => {
-    const props: Props = {}
     render(<Date {...props} />)
+  })
+
+  it('should show required warning', async () => {
+    render(<Date {...props} value="2023-12-07" required />)
+
+    const datepicker = document.querySelector('.dnb-date-picker')
+    const inputs: Array<HTMLInputElement> = Array.from(
+      datepicker.querySelectorAll('.dnb-date-picker__input')
+    )
+
+    expect(datepicker.classList).not.toContain(
+      'dnb-date-picker__status--error'
+    )
+    expect(
+      datepicker.querySelector('.dnb-form-status__text')
+    ).not.toBeInTheDocument()
+
+    act(() => {
+      inputs[inputs.length - 1].focus()
+      inputs[inputs.length - 1].setSelectionRange(4, 4)
+    })
+
+    await userEvent.keyboard(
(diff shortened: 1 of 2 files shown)


joakbjerk:
> Thanks for the great proposal 🚀 I tried to reproduce the same bug as reported in #2970, with this PR/fix in the following [CSB](https://codesandbox.io/p/sandbox/eufemia-starter-forked-vcctqq?file=%2Fsrc%2FApp.tsx%3A1%2C55). In the CSB I experienced the same bug/issue reported, selecting a date and then clicking tilbakestill(then no required error is displayed). That makes me wonder, what does this PR/fix improve in terms of displaying required error? 🤔
> 
> Anyhow, I'm not actually sure if what I've reported in #2970, as a bug actually is a bug? Or should the Field.Date actually behave different than other Field-components, because a datepicker is different? If we would want to improve this, perhaps the actual change needed would be in the DatePicker component, if so, I guess that would be a bit bigger job than a minor fix in Field.Date.
> 
> Many thoughts, perhaps we should discuss this with @tujoworker and the team

Ahh yeah, it seems to only work if you remove the date by pressing backspace, so this fix is not yet ready 🙏


Screenshot 2023-12-07 at 13 54 15

langz:
> > Thanks for the great proposal 🚀 I tried to reproduce the same bug as reported in #2970, with this PR/fix in the following [CSB](https://codesandbox.io/p/sandbox/eufemia-starter-forked-vcctqq?file=%2Fsrc%2FApp.tsx%3A1%2C55). In the CSB I experienced the same bug/issue reported, selecting a date and then clicking tilbakestill(then no required error is displayed). That makes me wonder, what does this PR/fix improve in terms of displaying required error? 🤔
> > Anyhow, I'm not actually sure if what I've reported in #2970, as a bug actually is a bug? Or should the Field.Date actually behave different than other Field-components, because a datepicker is different? If we would want to improve this, perhaps the actual change needed would be in the DatePicker component, if so, I guess that would be a bit bigger job than a minor fix in Field.Date.
> > Many thoughts, perhaps we should discuss this with @tujoworker and the team
> 
> Ahh yeah, it seems to only work if you remove the date by pressing backspace, so this fix is not yet ready 🙏
> 
> Screenshot 2023-12-07 at 13 54 15

But that could also be the "correct way" or "only way" of handling displaying the required message for this component, as it could be that it doesn't make sense to display it when pressing "tilbakestill"-button and that potentially could be a heavy change in DatePicker 🤔

joakbjerk:
> > > Thanks for the great proposal 🚀 I tried to reproduce the same bug as reported in #2970, with this PR/fix in the following [CSB](https://codesandbox.io/p/sandbox/eufemia-starter-forked-vcctqq?file=%2Fsrc%2FApp.tsx%3A1%2C55). In the CSB I experienced the same bug/issue reported, selecting a date and then clicking tilbakestill(then no required error is displayed). That makes me wonder, what does this PR/fix improve in terms of displaying required error? 🤔
> > > Anyhow, I'm not actually sure if what I've reported in #2970, as a bug actually is a bug? Or should the Field.Date actually behave different than other Field-components, because a datepicker is different? If we would want to improve this, perhaps the actual change needed would be in the DatePicker component, if so, I guess that would be a bit bigger job than a minor fix in Field.Date.
> > > Many thoughts, perhaps we should discuss this with @tujoworker and the team
> > 
> > 
> > Ahh yeah, it seems to only work if you remove the date by pressing backspace, so this fix is not yet ready 🙏
> > Screenshot 2023-12-07 at 13 54 15
> 
> But that could also be the "correct way" or "only way" of handling displaying the required message for this component, as it could be that it doesn't make sense to display it when pressing "tilbakestill"-button and that potentially could be a heavy change in DatePicker 🤔

Yeah we probably should discuss this in more detail

packages/dnb-eufemia/src/extensions/forms/Field/Date/__tests__/Date.test.tsx
```diff
@@ -1,10 +1,76 @@
...
+      datepicker.querySelector('.dnb-form-status__text')
+    ).not.toBeInTheDocument()
+
+    act(() => {
+      inputs[inputs.length - 1].focus()
+      inputs[inputs.length - 1].setSelectionRange(4, 4)
+    })
+
+    await userEvent.keyboard(
+      '{Backspace}{Backspace}{Backspace}{Backspace}{Backspace}{Backspace}{Backspace}{Backspace}'
```
langz, line 34:
```suggestion
      '{Backspace>8}'
```

langz (approved):

PR #3034, opened 2023-12-08
feat(PhoneNumber): add prioritized sort option

tujoworker:
Like in PR #3023 and since they share so many things, we move the shared code into the SelectCountry component.

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/feature-fields/PhoneNumber/info.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/feature-fields/PhoneNumber/info.mdx
@@ -21,7 +21,7 @@ Also, the `value` property should be a string with the country code, separated f
 
 The component returns the `emptyValue` when no number is set, which defaults to `undefined`.
 
-It supports the HTML [autofill](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete) attribute (`tel-country-code` and `tel-national`) if no value is given.
+It uses the HTML [autocomplete](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete) attribute (`tel-country-code` and `tel-national`) in their respective fields(country code and phone number) to provide automated assistance in filling out form field values, as well as guidance to the browser as to the type of information expected in the field..
 
 ### Default country code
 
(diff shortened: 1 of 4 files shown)


packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/feature-fields/PhoneNumber/info.mdx
```diff
@@ -21,6 +21,8 @@ Also, the `value` property should be a string with the country code, separated f
 
 The component returns the `emptyValue` when no number is set, which defaults to `undefined`.
 
+It supports the HTML [autofill](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete) attribute (`tel-country-code` and `tel-national`) if no value is given.
```
langz, line 24:
```suggestion
It uses the HTML [autocomplete](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete) attribute (`tel-country-code` and `tel-national`) in their respective fields(country code and phone number) to provide automated assistance in filling out form field values, as well as guidance to the browser as to the type of information expected in the field..
```

langz:
Just a thought, haven't looked in to it yet, but if/when using `omitCountryCodeField` could this lead to us wanting to provide a different value to autocomplete than  autoComplete="tel-national"? 🤔 

An other observation, not related directly to this PR, but to `omitCountryCodeField`, and the question above, is that when we do `omitCountryCodeField` we only support norwegian phone numbers of 8 digits 🤔

langz (approved):
Looks to work really well 👏 
Tested it on Chrome on mobile and desktop, and it worked as expected.
I was not able to make it work properly in safari on mobile and desktop, but that could be that they don't support it or that there's something off with my own browser settings regarding this 👍

joakbjerk:
Just some super nitpicky things, that don't need to be a part of this PR necessarily.

When searching for countries with `æ` `ø` or `å` in them, it does not seem to register/highlight or take those letters into account.
Notice how the `Ø` in Grønland is not bolder
Screenshot 2023-12-11 at 08 41 07

Same the `Æ` with Færøyene
Screenshot 2023-12-11 at 08 41 07

And then after typing `Fær`, Færøyene dissapears as an suggestion, even though it should match
Screenshot 2023-12-11 at 08 41 25

And shouldn't Grønland and Færøyene also show up in t
Screenshot 2023-12-11 at 08 42 36
he dropdown countries is set to `Nordic`?

joakbjerk (approved):
🔥

tujoworker:
@joakbjerk Nice findings! That is most certenly an "Autocomplete" bug. Can you write a quick bug report on this with links to your findings?

PR #3053, opened 2023-12-11
feat(Section): add dropShadow support

tujoworker:


joakbjerk (approved):
Should we maybe have visual tests for this too?

tujoworker:
Ideally we should test each property isolated in all states. 
But we actually do have this change in a visual test. The snapshot did fail and I updated it 👍

PR #3060, opened 2023-12-12
fix(Section): ensure cornerRadius does work together with dropShadow

tujoworker:


langz (approved):
Looks good, did not test it very much though 😓

PR #3070, opened 2023-12-13
fix: correctly compile CSS functional :has()

tujoworker:
In order to let our build setup correctly compile CSS functional :has() – we need to update CSS related deps, as it by now resulted in invalid code:

Screenshot 2023-12-13 at 09 00 24


langz (approved):
🚀

tujoworker:
In `v1.63.0` of sass, the maintainers have removed the possibility to compile sass in "non" Node.js environments by adding a check like this `self.fs` == null -> throw .. is only supported on Node.js.
 
But jest is a Node.js environment, its just that we run jest with jsdom, witch causes sass to think, its not an Node.js environment.

For now, we deal with that later.

Issue #3039, opened 2023-12-10
bug(Field.Selection): error does not work when variant is radio


langz:
## 🐛 Bug Report

In the light of the following two CSBs:
- [All Field.Components and variants with error and `without a value`](https://codesandbox.io/p/sandbox/eufemia-starter-forked-xgf9ww?file=%2Fsrc%2FApp.tsx%3A104%2C31)
- [All Field.Components and variants with error and `with a value`](https://codesandbox.io/p/sandbox/eufemia-starter-forked-ckv7qm?file=%2Fsrc%2FApp.tsx%3A120%2C31)

It's observed that the following code does not result in a red error border and no error message:
```
      <Field.Selection
        path="/selection3"
        label="Selection radio"
        error={new FormError('This is what is wrong...')}
        variant="radio"
      >
        <Field.Option value="foo" title="Foo!" />
        <Field.Option value="bar" title="Baar!" />
      </Field.Selection>
```

Results in:
<img width="339" alt="image" src="https://github.com/dnbexperience/eufemia/assets/1359205/b8f8677d-7a5e-404e-b915-77e442a1f9e7">

PR #3077, opened 2023-12-13
fix(Field.Selection): add error state and message

joakbjerk:
Proposal to fix https://github.com/dnbexperience/eufemia/issues/3039.

Not sure if wrapping the `Radio.Group` inside a `FieldBlock` or if the current approach is the best way to solve it. Please discuss 🙏

Before:
Screenshot 2023-12-13 at 14 59 12

After:
Screenshot 2023-12-13 at 14 58 56


langz:
Nice nice, to me, this approach looks good 👍 

But we should probably check in the existing forms code, if there's other examples of `Radio.Group`, `ToggleButton.Group` or similar components that's wrapped in a `FieldBlock`. If it exists, I guess we should do so 🤔

langz (commented):
Nice nice, to me, this approach looks good 👍 

But we should probably check in the existing forms code, if there's other examples of `Radio.Group`, `ToggleButton.Group` or similar components that's wrapped in a `FieldBlock`. If it exists, I guess we should do so 🤔

tujoworker:
I have taken a look on this. And found that we need a slightly larger restructure. 

I may commit these changes if that is ok?

tujoworker (approved):

PR #3081, opened 2023-12-13
`Field.PhoneNumber` renders an validation error initially when it is required and only country code is provided

tellnes:
The `Field.PhoneNumber` component renders an validation error initially when it is required and only country code is provided. Git bisect tells me this was introduced in 5b1a9b05341e93f6bb7e90c3d8c6d9c5accc47e1.

Here is a small example to reproduce it, but see the attached commit for full details.
```jsx
<PhoneNumber required value='+47' />
```

Work around:
```tsx
function DemoPhoneNumberField() {
  const countryCode = '+47'
  const phoneNumber = ''

  const value = (countryCode === '+47' && !phoneNumber) ? undefined : (countryCode + ' ' + phoneNumber)

  return <PhoneNumber required value={value} />
}
```


tujoworker:
Nice! Thank you for the fantastic PR report 💯 

I'v added a commit with a solution that is at least one approach to solve this. 
Let me know if you would have solved it differently.

The overall goal is to always handle the cc value as optional.

langz:
Looks good, do we have to think about this fix if/when used with `omitCountryCodeField`?

tujoworker:
We have good test coverage on `omitCountryCodeField` – and I tested it manually as well 👍 

Had a talk with @henit – where we have concluded a different approach would be better, especially because of schema validation.

To solve it I added `toOutput`, `fromOutput` and `validateRequired`.

packages/dnb-eufemia/src/extensions/forms/hooks/useDataValue.ts
```diff
@@ -103,7 +109,7 @@ export default function useDataValue<
   const externalValue = useMemo(() => {
     if (props.value !== undefined) {
       // Value-prop sent directly to the field has highest priority, overriding any surrounding source
-      return props.value
+      return fromOutput(props.value)
```
henit, line 112: Good method to implement, but maybe a bit of a confusing name? How about `fromExternal`?

packages/dnb-eufemia/src/extensions/forms/hooks/useDataValue.ts
```diff
@@ -454,12 +470,13 @@ export default function useDataValue<
         handleIterateElementChange?.(iterateValuePath, newValue)
       }
     },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
henit, line 473: Why is this necessary? It is a bit risky as it stops IDE/tests from alerting about missing deps in future changes, which might lead to incorrect data handling.

henit (commented):
I like the extended flexibility!

packages/dnb-eufemia/src/extensions/forms/hooks/useDataValue.ts
tujoworker: Yes! I agree! And what about `toOutput` vs `toExternal`?

packages/dnb-eufemia/src/extensions/forms/hooks/useDataValue.ts
tujoworker: Because when the transformers where declared, we got an infinity look, as they could not have a useCallback.

But I solved this witha useRef in this PR #3109

tujoworker:
The `useDataValue` changes/additions will be med in this PR #3109 instead. When merged, we will rebase this branch/PR.

tujoworker (approved):

PR #3085, opened 2023-12-14
chore(Number): document percentage and add test

tujoworker:


packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/base-fields/Number/Examples.tsx
```diff
@@ -204,6 +204,19 @@ export const ValidateMinimum = () => {
   )
 }
 
+export const Percentage = () => {
+  return (
+    <ComponentBox>
+      <Field.Number
```
langz, line 210: I think we should add the 
`percent` property in this example👌

langz (approved):
Great great👌
I added one comment about the example/demo🙏

Issue #3099, opened 2023-12-15
feat(Field.Expiry): Display number keyboard in input field


langz:
## 🚀 Feature Proposal

Would it be nice to show the number input in keyboard on a mobile phone for the Field.Expiry?

[As of today](https://eufemia.dnb.no/uilib/extensions/forms/feature-fields/Expiry/), when opening/focusing the input field, it shows the "regular keyboard", and then I as a user will have to change to numbers in the keyboard:
![Expiry  Eufemia](https://github.com/dnbexperience/eufemia/assets/1359205/b3b3a67f-9aea-49a1-8cc1-110b55d37d6d)

[Here's an example from Posten where I guess they use type="tel"](https://id.posten.no/login/4yc4uqbaqhwp5z2c0wsq9kre0qqin5ymfng6w86ihokmvl689n)(If link doesn't work, go to [posten](https://www.posten.no/) and press "Min side" in the top right corner), on how they have a input field, displaying the number keyboard:
![Postenid](https://github.com/dnbexperience/eufemia/assets/1359205/2ca57f3a-2704-49b0-9bdb-76448b64b22f)

[Here's an example from Material UI](https://mui.com/material-ui/react-text-field/) where we at least get the numbers(1,2,3,4,5,6,7,8,9,0) as part of the keyboard:
![React Text Field component - Material UI](https://github.com/dnbexperience/eufemia/assets/1359205/2d6cd714-c5d7-49f3-b6fc-42329acbe015)

langz:
The following is also an annoying observation, which I think will be fixed when this issue is fixed. 
That’s that I have to change/set the keyboard type twice to input the full value in this component: 

https://github.com/dnbexperience/eufemia/assets/1359205/5ff0dd89-d74a-494a-967c-ca454a2b8fec

joakbjerk:
> The following is also an annoying observation, which I think will be fixed when this issue is fixed. That’s that I have to change/set the keyboard type twice to input the full value in this component:
> RPReplay_Final1702676445.mp4
> 
> First when entering the input, the keyboard has only lettere, so I have to change keyboard to get numbers when entering month, and then I have to do the same again when entering year…

Yeah this seems to be working now on my android device at least, on [this pr](https://github.com/dnbexperience/eufemia/pull/3104) tested in the portal review. The backspace/delete functionality seems iffy on my android device though, but that is a fix for a later pr I think.

PR #3104, opened 2023-12-17
chore(Field.Expiry): make number keyboard default for mobile

joakbjerk:
Adds this feature proposal https://github.com/dnbexperience/eufemia/issues/3099

Tested and works on my android device, so need someone to test if it actually works on iOS 🙏


langz (approved):
I just gave this a quick test, and it works great on my iPhone in Safari🦏

langz:
We could consider adding a test for this functionality as well.
Not sure how easy/hard that is 🤔

Issue #3098, opened 2023-12-15
feat(Field.BankAccountNumber): Display number keyboard in input field


langz:
## 🚀 Feature Proposal

Would it be nice to show the number input in keyboard on a mobile phone for the Field.BankAccountNumber?

[As of today](https://eufemia.dnb.no/uilib/extensions/forms/feature-fields/Currency/), when opening/focusing the input field, it shows the "regular keyboard", and then I as a user will have to change to numbers in the keyboard:
![BankAccountNumber  Eufemia](https://github.com/dnbexperience/eufemia/assets/1359205/063ab82b-4756-4906-b2c2-b48bd6c6b7c9)

[Here's an example from Sparebanken where I guess they use type="tel"](https://www.sparebank1.no/nb/bank/privat/daglig-bruk/konto-og-betaling/utland.html):
![Betaling til og fra utlandet  SpareBank 1](https://github.com/dnbexperience/eufemia/assets/1359205/1073691f-e5ec-40c7-a212-6e5dd8ffc425)

[Here's an example from Material UI](https://mui.com/material-ui/react-text-field/) where we at least get the numbers(1,2,3,4,5,6,7,8,9,0) as part of the keyboard:
![React Text Field component - Material UI](https://github.com/dnbexperience/eufemia/assets/1359205/2d6cd714-c5d7-49f3-b6fc-42329acbe015)

PR #3105, opened 2023-12-17
chore(Field.BankAccountNumber): make numeric keyboard default for mobile

joakbjerk:
Proposal to implement this feature https://github.com/dnbexperience/eufemia/issues/3098

Had to add an `inputMode` prop to the `String` component to make it work, but we should probably discuss if that is the best way to add the `inputMode` prop 🙏


tujoworker:
This PR #3108 should fix it, so closing this one.

Issue #3097, opened 2023-12-15
feat(Field.Currency): Display number keyboard in input field


langz:
## 🚀 Feature Proposal

Would it be nice to show the number input in keyboard on a mobile phone for the Field.Currency?

[As of today](https://eufemia.dnb.no/uilib/extensions/forms/feature-fields/Currency/), when opening/focusing the input field, it shows the "regular keyboard", and then I as a user will have to change to numbers in the keyboard:
![Currency  Eufemia](https://github.com/dnbexperience/eufemia/assets/1359205/2bfe116d-e31a-4321-8093-7a6ff51b8f96)

[Here's an example from DNB where I guess they use type="tel"](https://www.dnb.no/lan/kalkulator/boliglan):
![Boliglånskalkulator  Beregn månedlig lånekostnad - DNB](https://github.com/dnbexperience/eufemia/assets/1359205/78f376a7-5a4b-42c7-a4b7-4ece0461751f)

[Here's an example from Material UI](https://mui.com/material-ui/react-text-field/) where we at least get the numbers(1,2,3,4,5,6,7,8,9,0) as part of the keyboard:
![React Text Field component - Material UI](https://github.com/dnbexperience/eufemia/assets/1359205/2d6cd714-c5d7-49f3-b6fc-42329acbe015)

joakbjerk:
Solved by this pr: https://github.com/dnbexperience/eufemia/pull/3108

PR #3106, opened 2023-12-17
chore(Field.Number): make numeric keyboard default on mobile

joakbjerk:
Implementation of this feature proposal https://github.com/dnbexperience/eufemia/issues/3093, this should also solve https://github.com/dnbexperience/eufemia/issues/3097

Tested on android in preview portal, and it works. Need confirmation for iOS 🙏


tujoworker:
The change should be made in input masked. 

We already have a logic and tests for it. But we may need to enhance it and also add pattern="\d*" (probably).

tujoworker:
This PR #3108 should fix it, so closing this one.

PR #3109, opened 2023-12-18
feat(useDataValue): add toOutput, fromExternal and validateRequired

tujoworker:
For data transformation, add these hooks:

- `toOutput`
- `fromExternal`

For custom control about validating the required prop:

- `validateRequired`

Also, adding docs about the `useDataValue` hook.

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/hooks/useDataValue.ts
+++ b/packages/dnb-eufemia/src/extensions/forms/hooks/useDataValue.ts
@@ -23,6 +23,7 @@ interface ReturnAdditional<Value> {
   id: string
   value: Value
   error: Error | FormError | undefined
+  hasError: boolean
   setHasFocus: (hasFocus: boolean, valueOverride?: unknown) => void
   handleFocus: () => void
   handleBlur: () => void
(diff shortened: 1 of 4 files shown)


langz (approved):

packages/dnb-eufemia/src/extensions/forms/hooks/useDataValue.ts
```diff
@@ -41,26 +42,41 @@ export default function useDataValue<
...
 
+  const transformersBase = {
+    toInput,
+    fromInput,
+    toEvent,
+    fromExternal,
+    validateRequired,
+  }
+  const transformers = useRef(transformersBase)
+  transformers.current = transformersBase
```
henit, line 78: This will cause the object in the ref with all the transformers to be new on every render, even if the transformer functions are unchanged. Not a problem unless a dependency array depend on them, but if that is done, it could have unforeseen consequences. May be ok to ignore, just pointing out.

packages/dnb-eufemia/src/extensions/forms/hooks/useDataValue.ts
```diff
@@ -41,26 +42,41 @@ export default function useDataValue<
...
     schema,
-    errorMessages,
     validateInitially,
     validateUnchanged,
     continuousValidation,
     toInput = (value: Value) => value,
     fromInput = (value: Value) => value,
+    toEvent = (value: Value) => value,
+    fromExternal = (value: Value) => value,
+    validateRequired = (value: Value, { emptyValue, required }) =>
```
henit, line 59: Not a big problem, but I like consistent patterns, and since we have other externally controllable validators (props) that return an error or undefined to indicate result of the validation, maybe it is better to do the same here? Then throw if an error is returned in `validateValue`...

henit (approved):

packages/dnb-eufemia/src/extensions/forms/hooks/useDataValue.ts
tujoworker: Yes. I have made this change. Let me know if there is anything we should change.

packages/dnb-eufemia/src/extensions/forms/hooks/useDataValue.ts
tujoworker: Yes. Its so its possible to update these functions. But I removed it as of now. Not sure if this is ever used.

PR #3125, opened 2023-12-20
fix(Field.String): validated empty value when required

tellnes:
If the form is submitted without filling out any any fields, I expect errors messages to be displayed and form submission to be prevented if the form is invalid.

I've not tested if this affects other validation rules, but at least it fails when using the `required` property.

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/String/__tests__/String.test.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/String/__tests__/String.test.tsx
@@ -328,6 +328,20 @@ describe('Field.String', () => {
         })
         expect(screen.queryByRole('alert')).not.toBeInTheDocument()
       })
+
+      it('should show error for initially empty value when required and validateInitially is set', async () => {
+        render(<Field.String value="" required validateInitially />)
+        expect(screen.getByRole('alert')).toBeInTheDocument()
+      })
+
+      it('should show error for initially empty value when required and blur event', async () => {
+        render(<Field.String value="" required />)
+        const input = document.querySelector('input')
+        act(() => {
+          input.blur()
+        })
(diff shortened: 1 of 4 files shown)


langz:
To assist with this problem, I've created a few CSBs looking at how `validateInitially` with `required` and different variations of `value` works:
- [validateInitially with required and value](https://codesandbox.io/p/sandbox/eufemia-starter-forked-mmrrg2?file=%2Fsrc%2FApp.tsx)
- [validateInitially with required and empty string value `""`](https://codesandbox.io/p/sandbox/eufemia-starter-forked-7rgqv4?file=%2Fsrc%2FApp.tsx)
- [validateInitially with required and null value](https://codesandbox.io/p/sandbox/eufemia-starter-forked-sd7d9z?file=%2Fsrc%2FApp.tsx%3A209%2C48)
- [validateInitially with required and undefined value](https://codesandbox.io/p/sandbox/eufemia-starter-forked-yk3x7r?file=%2Fsrc%2FApp.tsx%3A137%2C27)
- [validateInitially with required and without value](https://codesandbox.io/p/sandbox/eufemia-starter-forked-mx8x29?file=%2Fsrc%2FApp.tsx%3A147%2C21)

tujoworker:
Thank you for test great tests 🙏 👏 

The initial approach is, that it compares `emptyValue` to the given value. But because `emptyValue` is undefined by default, it did not invoke the error. But I agree, that an empty string should probably be handled as "invalid". 

So thereof the solution I committed. 

When to show an error: We should preferably (always) wait for the user to make a change or wait for the submit action, because, the user should not be distracted when just tabbing from field to field because of "navigation" reasons.

But we can omit this behavior with the `validateUnchanged` prop. So I added it to one of the tests.

langz (approved):
🚀

Issue #3035, opened 2023-12-08
docs(Field.SelectCountry): document the event return types


langz:
## 🐛 Bug Report

Add docs about the return values/objects in the events.
As of now it's not described at all. An example to follow for documenting the return object could be as done in [Autocomplete](https://eufemia.dnb.no/uilib/components/autocomplete/events/).

Maybe a few of the examples in demos should/could be updated to showcase the events better 🤔 

[This is a link to latest preview from main](https://eufemia-dnb-design-system-portal.vercel.app/uilib/extensions/forms/feature-fields/SelectCountry/events/)

PR #3129, opened 2023-12-20
chore(Field.SelectCountry): document event handlers argument type

joakbjerk:
Enhancement proposal for https://github.com/dnbexperience/eufemia/issues/3035

- [x] Update examples
- [x] Rewrite argument documentation to include both arguments and not just object


langz (approved):
Looks good 😄 
Could we add an example or so, to see how to access/use this?
In [all our examples](https://eufemia-git-chore-select-country-document-return-type-eufemia.vercel.app/uilib/extensions/forms/feature-fields/SelectCountry/demos/#empty) that demos the return type of the `onChange` event only returns the `iso`(NO, SE, FI, DK, etc...):

image

How could I as a developer, access the object you have described? 🙏

tujoworker:
Yes, you can make an example in `Examples.tsx` and import it `events.mdx` (like it demos.mdx).

PR #3131, opened 2023-12-20
chore(InputMasked): fixes failing unit test

langz:
There's a failing unit test in main, https://github.com/dnbexperience/eufemia/actions/runs/7273881239/job/19818731303

This PR tries to "fix it", though I've not validated that the "fix" actually is correct.
I've not read or tried to understand the semantics of the test, so don't actually know what the correct expectation of the test is.

### DIFF
--- a/packages/dnb-eufemia/src/components/input-masked/__tests__/InputMasked.test.tsx
+++ b/packages/dnb-eufemia/src/components/input-masked/__tests__/InputMasked.test.tsx
@@ -1991,7 +1991,7 @@ describe('inputmode', () => {
 
     expect(inputElement).toHaveAttribute('type', 'number')
     expect(inputElement).toHaveAttribute('placeholder', '1 234,5 kr')
-    expect(inputElement).toHaveAttribute('value', '1234,5')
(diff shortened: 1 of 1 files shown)


packages/dnb-eufemia/src/components/input-masked/__tests__/InputMasked.test.tsx
```diff
@@ -1991,7 +1991,7 @@ describe('inputmode', () => {
 
     expect(inputElement).toHaveAttribute('type', 'number')
     expect(inputElement).toHaveAttribute('placeholder', '1 234,5 kr')
-    expect(inputElement).toHaveAttribute('value', '1234,5')
+    expect(inputElement).toHaveAttribute('value', null)
```
langz, line 1994:
```suggestion
    expect(inputElement).not.toHaveAttribute('value')
```

tujoworker:
Yes, because of PR race conditions as we do not require branches to be rebased before merge.
Its true, that we don't expect a value attribute. Its fixed already, so I close this pr. 

Thanks for creating it 💯

PR #3135, opened 2023-12-20
fix(forms): update data context with initially given field value

### LINKED ISSUE #3079: bug(Forms): Setting value to all fields results in empty onSubmit event in Form
## 🐛 Bug Report

When setting value of `Fields` through their `value` property, the wrapped `Form` component's `onSubmit` event does not include the values in the return value/object.

## To Reproduce

Here's a [CSB](https://codesandbox.io/p/sandbox/eufemia-starter-forked-j5l7ct?file=%2Fsrc%2FApp.tsx%3A34%2C1), try to press "Send" and check the console to see the onSubmit event log what it returns. It returns null or undefined. 

## Expected behavior

Wouldn't it be expected that the `onSubmit` event would return all values from all fields?

I would expect that the onSubmit event would return the same as if I would have set the Field values through the Form's data property, like in this [CSB](https://codesandbox.io/p/sandbox/eufemia-starter-forked-hkkfk6?file=%2Fsrc%2FApp.tsx%3A32%2C18).

tujoworker:
Fixes #3079

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/DataContext/Context.ts
+++ b/packages/dnb-eufemia/src/extensions/forms/DataContext/Context.ts
@@ -12,6 +12,7 @@ export interface ContextState {
   /** Will set autoComplete="on" on each nested Field.String and Field.Number */
   autoComplete?: boolean
   handlePathChange: (path: string, value: any) => void
+  updateDataValue: (path: string, value: any) => void
   handleSubmit: (props?: HandleSubmitProps) => any
   // Error status
   showAllErrors: boolean
@@ -28,6 +29,7 @@ export interface ContextState {
 export const defaultContextState: ContextState = {
   data: undefined,
   handlePathChange: () => null,
+  updateDataValue: () => null,
   handleSubmit: () => null,
   showAllErrors: false,
   setShowAllErrors: () => null,
--- a/packages/dnb-eufemia/src/extensions/forms/DataContext/Provider/Provider.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/DataContext/Provider/Provider.tsx
@@ -108,6 +108,7 @@ export default function Provider<Data extends JsonObject>({
     // eslint-disable-next-line react-hooks/exhaustive-deps -- Avoid triggering code that should only run initially
   }, [])
   const internalDataRef = useRef<Partial<Data>>(initialData)
+  const dataCacheRef = useRef<Partial<Data>>(data)
   // - Validator
   const ajvSchemaValidatorRef = useRef<ValidateFunction>()
 
@@ -163,12 +164,11 @@ export default function Provider<Data extends JsonObject>({
     []
   )
 
-  const handlePathChange = useCallback(
+  const updateDataValue = useCallback(
     (path, value) => {
       if (!path) {
         return
       }
-      onPathChange?.(path, value)
 
       // Update the data even if it contains errors. Submit/SubmitRequest will be called accordingly
       const newData = structuredClone(
@@ -179,11 +179,10 @@ export default function Provider<Data extends JsonObject>({
             internalDataRef.current ??
               (path.match(isArrayJsonPointer) ? [] : {})
       )
+
       if (path !== '/') {
         pointer.set(newData as Data, path, value)
       }
-
-      onChange?.(newData as Data)
 
       if (sessionStorageId && typeof window !== 'undefined') {
         window.sessionStorage?.setItem(
@@ -194,11 +193,30 @@ export default function Provider<Data extends JsonObject>({
 
       internalDataRef.current = newData
 
-      validateData()
-      showAllErrorsRef.current = false
       forceUpdate()
+
+      return newData
     },
-    [onChange, onPathChange, validateData, sessionStorageId]
+    [sessionStorageId]
+  )
+
+  const handlePathChange = useCallback(
+    (path, value) => {
+      if (!path) {
+        return
+      }
+
+      onPathChange?.(path, value)
+
+      const newData = updateDataValue(path, value)
+
+      onChange?.(newData as Data)
+
+      showAllErrorsRef.current = false
+
+      validateData()
+    },
+    [onPathChange, updateDataValue, onChange, validateData]
   )
 
   // Mounted fields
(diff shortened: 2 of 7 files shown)


packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/create-component.mdx
```diff
@@ -107,26 +108,59 @@ const {
...
 
-- `setHasFocus` accepts a boolean as value. When called, it will update the internal logic - for event handling and validation. Will re-render the React Hook and its outer component.
+- `required` if true, it will call `validateRequired` for validation.
+- `schema` or `pattern` for JSON schema validation powered by [ajv](https://ajv.js.org/).
+- `validator` your custom validation function.
+- `validateRequired` does allow you to provide a custom logic for how the `required` prop should validate.
+- `onBlurValidator` your custom validation function.
+- `validateInitially` in order to show an error without a change and blur event. Should only be used for testing purposes.
+- `validateUnchanged` in order to validate without a change and blur event. Should only be used for testing purposes.
+- `continuousValidation` in order to validate without without a focus event beforehand. Should only be used for testing purposes.
```
langz, line 141: Is it correct that all of the props above(`validateInitially`, `validateUnchanged`, and `continuousValidation`) is only for testing purposes? Can it never be used in a real life scenario? 🤔

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/create-component.mdx
```diff
@@ -107,26 +108,59 @@ const {
...
 
-#### Validation
+**Error**
 
-- `validateRequired` does allow you to provide a custom logic for how the `required` prop should validate.
+- `error` error object like `FormError` that includes the string to display or an object with the key `validationRule`. More info down below.
```
langz, line 145:
```suggestion
- `error` object like `FormError` that includes the string to display or an object with the key `validationRule`. More info down below.
```

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/create-component.mdx
```diff
@@ -155,7 +189,7 @@ const { error, hasError } = useDataValue({
 
 During validation, the different APIs do have a prioritization order and will stop processing further when they match:
 
-1. `require` prop
+1. `required` prop
```
langz, line 192: 👍

langz (approved):

langz:
I tried testing the [reprod](https://codesandbox.io/p/sandbox/eufemia-starter-forked-j5l7ct?file=%2Fsrc%2FApp.tsx%3A34%2C1) you posted, but I think it was using the wrong eufemia source.
So I created a [reprod](https://codesandbox.io/p/sandbox/eufemia-starter-forked-kzpdh4?file=%2Fsrc%2FApp.tsx%3A34%2C1), and it seems to work great.

langz (changes requested):
I tried testing the [reprod](https://codesandbox.io/p/sandbox/eufemia-starter-forked-j5l7ct?file=%2Fsrc%2FApp.tsx%3A34%2C1) you posted, but I wouldn't say that it works as intended? 🤔 

The `onSubmit` event still returns `undefined`, as reported in https://github.com/dnbexperience/eufemia/issues/3079

I've also come up with one other example/variant that we perhaps should support, and that's when some data is set through `value` in the Component's prop, and other values are set through Form.Handlers `data` prop.
Let's say in the [following CSB](https://codesandbox.io/p/sandbox/eufemia-starter-forked-r55zs7?file=%2Fsrc%2FApp.tsx%3A37%2C1).
Should we merge these together and return in the onSubmit event?


And to further complicate things, if we set `a` to value `x` through data prop of `Form.Handler`, and in the `Field` with path `/a` we set `value="y"` and press `onSubmit`, will we return `{a: "x"}` or `{a:"y"}`? Who overwrites who
 🤔

tujoworker:
Very good feedback and very true.

I think the value "near" the component should win. Also, in some cases it may make sense to make "one" component controlled (with a local state), and then it's even more logical that the value should be prioritized, I would say. 

I will make a change in the PR regarding that 👍

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/create-component.mdx
tujoworker: Yeah, we may find a better formulation 🤔

tujoworker:
I'v made change. Can you check if it works as expected? Also, we need to get in the PR build version in [this CSB](https://codesandbox.io/p/sandbox/eufemia-starter-forked-j5l7ct).

langz:
> I'v made change. Can you check if it works as expected? Also, we need to get in the PR build version in [this CSB](https://codesandbox.io/p/sandbox/eufemia-starter-forked-j5l7ct).

Great work, thanks.

I've updated the following two CSBs, using the build from this PR, now:
- https://codesandbox.io/p/sandbox/eufemia-starter-forked-r55zs7 (example where we use both data and value)
- https://codesandbox.io/p/sandbox/eufemia-starter-forked-j5l7ct (example where we only use value)

tujoworker:
Nice findings! Thank you.

I added some tests with `StrictMode` and they did fail the same way as in the CSB.
Also, I [added fix](https://github.com/dnbexperience/eufemia/pull/3135/files#diff-286fe1bc4c98747f348b6f9ffba318ab1ceac84fe7054969fbc381b3a00acc77R293) so it works in `StrictMode` too.

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/create-component.mdx
```diff
@@ -138,7 +138,7 @@ All properties are optional and can be used as needed. These properties can be p
 - `onBlurValidator` your custom validation function. It will run on a `handleBlur()` call. Use it over `validator` for validations with side-effects.
 - `validateInitially` in order to show an error without a change and blur event. Used for rare cases.
 - `validateUnchanged` in order to validate without a change and blur event. Used for rare cases.
-- `continuousValidation` in order to validate without a focus event beforehand. Used for rare cases.
+- `continuousValidation` in order to validate without without a focus event beforehand. Used for rare cases.
```
joakbjerk, line 141: Not sure if you actually need to fix this typo, as it seems to be the same typo as fixed [here](https://github.com/dnbexperience/eufemia/pull/3158/commits/5c8f89d88aff843bd398a949eb35f8fea02f136a) 🤔

```suggestion
- `continuousValidation` in order to validate without a focus event beforehand. Used for rare cases.
```

joakbjerk (approved):
Nice! Seems to be working as intended now 😄

PR #3139, opened 2023-12-21
fix(Anchor): dnb-anchor--no-hover and dnb-anchor--contrast

langz:
Fixes the following code example, which would result in having a hover style:
```
<Anchor
    href="/uilib/components/anchor"
    className="dnb-anchor--no-hover dnb-anchor--contrast"
  >
    No hover
  </Anchor>
```
Looking like:
image

After the fix, it looks like this when hovering:
image

This PR also adds a few screenshot tests for `dnb-anchor--contrast` variations, together with other helper classes.


packages/dnb-eufemia/src/components/anchor/__tests__/Anchor.screenshot.test.ts
```diff
@@ -107,6 +107,71 @@ describe.each(['ui', 'sbanken'])('Anchor for %s', (themeName) => {
     expect(screenshot).toMatchImageSnapshot()
   })
 
+  it('have to match the anchor-contrast no-underline "default" state', async () => {
```
tujoworker, line 110: We should probably move all of the contrast tests in its own describe block.

tujoworker (approved):
Nice work. Thank you 💯

PR #3140, opened 2023-12-21
feat(Field.Number): add step control functionality

Sundfjord:
Adds a new prop `showStepControls` to the Number component. When `true`, this prop enables control buttons that adhere to a `step` property that is also added, as well as taking min/max values into account.

This variant of the component is Sbanken-ready, and a number of tests and screenshot tests has been implemented in relation to it.


tujoworker:
Very nice work so far! 

I gave it a first try with VoiceOver. And so far, we need to improve the experience there. But will have to come back to it in January. 

1. We need `aria-hidden` on the buttons.
2. We need probably some more aria attributes to make the "stepper" work as a stepper, when VoiceOver tells on what keyboard keys should be used. When I used them right now, it sets an invalid value of e.g. -10/+10.

Also, we need to test it on NVDA. I will come back to you once I tested it there.

tujoworker:
It works now:

- VoiceOver macOS: very good.
- VoiceOver iOS: very good.
- NVDA with Chrome/Firefox on Windows: ok, only first step is announced, needs live-update enhancement we should add to InputMasked.

tujoworker (approved):
Very nice work 🥳 Thank you 💯

PR #3141, opened 2023-12-21
fix(Section): ensure overlay is behind content

tujoworker:
- We do that by adding a needed negative z-index.
- Move both the outline and the shadow in the same pseudo element for better flexibility.


langz (approved):
🥇

PR #3148, opened 2023-12-28
chore(Field.Currency): add input mode tests

joakbjerk:


tujoworker (approved):
Regarding the support of "minus". Do we set inputMode in this component? Because we should not. The InputMasked should handle this.

joakbjerk:
> Regarding the support of "minus". Do we set inputMode in this component? Because we should not. The InputMasked should handle this.

This component is using the `Field.Number` under the hood, so the `inputMode` is determined by the `InputMasked` used in `Field.Number` 🙏 So we don't have to set the `inputMode` directly on any of the components that uses `Field.Number` since the `InputMasked` takes care of that 👍

So this PR is just taking care of the "test contract" for `Field.Currency` `inputMode` to so to say

tujoworker (approved):

PR #3151, opened 2023-12-28
chore(Field.Number): add input mode test

joakbjerk:


packages/dnb-eufemia/src/extensions/forms/Field/Number/__tests__/Number.test.tsx
```diff
@@ -115,6 +115,14 @@ describe('Field.Number', () => {
...
       expect(element.className).toContain('dnb-input__align--right')
     })
+
+    it('should have decimal input mode', () => {
+      render(<Field.Number />)
+
+      const input = document.querySelector('.dnb-input__input')
+
+      expect(input).toHaveAttribute('inputmode', 'decimal')
+    })
```
tujoworker, line 125: Also here you could use toHaveAttribute. 

But can you enter a minus in this field? Then we should not add it and just relay on what InputMasked is doing.

joakbjerk: It relies on the `InputMasked` to determine `inputmode`, and the `Field.Number` allows negative numbers, and I can't really see if there is a way to opt out of that, to dissallow negative numbers for `Field.Number`. So in a sense this test is kind of redundant, since we are already testing this in `InputMasked.test.tsx` 🤔 Should we perhaps just close this PR?

tujoworker: We could leave just the test. I see both pros and cons of having it. The "pro" is again the "contract" thinking, kind of what we expect this component to deliver.

tujoworker (approved):

PR #3160, opened 2024-01-02
fix(InputMasked): correct cursor position when navigating using keyboard

tellnes:
If you are tabbing into an `InputMasked` field which is partially filled and then uses your keyboard to un-select it by clicking your right arrow key, your input is ignored.

**Steps to reproduce:**

1. Partially fill an InputMasked field
2. Tab out of the field (Click <kbd>Tab</kbd>)
3. Tab back to the field (Click <kbd>Shift</kbd>+<kbd>Tab</kbd>)
4. Unselect by moving the cursor to the end of the text (Click <kbd>Right Arrow</kbd>)
5. Trying to continue filling out the value is ignored

**Work around:**

Set the `show_guide` property to `false`

### DIFF
--- a/packages/dnb-eufemia/src/components/input-masked/InputMaskedHooks.js
+++ b/packages/dnb-eufemia/src/components/input-masked/InputMaskedHooks.js
@@ -417,7 +417,10 @@ const useCallEvent = ({ setLocalValue }) => {
     const cleanedValue =
       numberValue === 0 && String(num).charAt(0) !== '0' ? '' : num
 
-    if (name === 'on_change' && numberValue === 0) {
+    if (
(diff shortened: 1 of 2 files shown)


tujoworker:
I'v added a possible solution. This should be tested extensively. Also, possibly we should dive deeper into the origin of this issue, as the helper function `correctCaretPosition` is only a fix.

Here is a [CSB](https://codesandbox.io/p/sandbox/eufemia-input-masked-selection-ntttdc) reprod.

tujoworker (approved):
I think we should give this a chance as is. 
The fix will first kick in once it needs to.

joakbjerk (approved):
🥧

PR #3163, opened 2024-01-03
feat(Card): responsive breakout on small screens

tujoworker:
On small screens, the Card should breakout. To be able to do that, we use the Section component.


joakbjerk (approved):
Looks good 🙏

kimroen:
Hey Tobias! It looks like before this change, the card component had padding on small screens. With this, it doesn't.

Is this accurate, and if so - intentional?

I made this example to demonstrate a different issue, but you can see this by changing the Eufemia version from 1.15.1 to 1.16.0: https://codesandbox.io/p/sandbox/laughing-bash-f2d5zm

PR #3171, opened 2024-01-04
chore(DatePicker): add onBlur event

joakbjerk:
Proposal to add onBlur event prop for the `DatePicker`

This could be solve by either hooking the onBlur event to the `fieldset` wrapping the inputs, or adding it to the existing internal `DatePickerInput` `onBlur` handler. 

I went with the latter approach, as adding it to the `fieldset` would requrie adding a check for if the `InputButton` is being clicked to abort the onBlur event, preventing the onBlur from firing every time the user clicks the `InputButton`

### DIFF
--- a/packages/dnb-eufemia/src/components/date-picker/__tests__/DatePicker.test.tsx
+++ b/packages/dnb-eufemia/src/components/date-picker/__tests__/DatePicker.test.tsx
@@ -1498,7 +1498,7 @@ describe('DatePicker component', () => {
     expect(thirdDateButton.children[2]).toHaveTextContent('24')
   })
 
-  it.only('should fire blur event when input loses focus', async () => {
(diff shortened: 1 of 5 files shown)


packages/dnb-eufemia/src/components/date-picker/__tests__/DatePicker.test.tsx
```diff
@@ -1497,6 +1497,41 @@ describe('DatePicker component', () => {
     )
     expect(thirdDateButton.children[2]).toHaveTextContent('24')
   })
+
+  it.only('should fire blur event when input loses focus', async () => {
```
tujoworker, line 1501:
```suggestion
  it('should fire blur event when input loses focus', async () => {
```

tujoworker (approved):
Nice 🚀 (we need to remove "only")

PR #3175, opened 2024-01-05
fix(Field.Number): replace `rightAligned` prop with `align`

Sundfjord:
Removed Number and Currency component support for the `rightAligned` prop and replaced it with a more flexible  `align` prop that accepts one of `left`, (default) `center` and `right`.


tujoworker (approved):
Looks good 🍡  thank you 💯

PR #2799, opened 2023-10-26
feat(DatePicker): Convert to functional components with typescript and hooks

joakbjerk:
- [x] `DatePicker`
- [x] `DatePickerAddon`
- [x] `DatePickerCalc`
- [x] `DatePickerCalendar`
- [x] `DatePickerContext`
- [x] `DatePickerFooter`
- [x] `DatePickerInput`
- [x] `DatePickerProvider`
- [x] `DatePickerRange`
- [x] Fix tests
- [x] Fix types
- [x] Fix `onClickOutside` functionality
- [x] Make tests for `onClickOutside`
- [x] Fix `DatePickerProvider` `getDerivedStateFromProps` logic
- [x] Fix `DatePickerInput` mask and value logic
- [x] Fix `DatePickerInput` paste functionality
- [x] General Bugfixes
- [x] Move provider date logic to hook
- [x] Move provider view logic to hook
- [x] Move date correction logic closer to trigger
- [x] Move `lastEventCallCache` to hook
- [x] Fix `callOnChangeHandler` callback functionality
- [x] Fix Input onChange handling
- [x] Wrap necessary functions or values in `useCallback` or `useMemo` hooks
- [x] Remove `locale` prop, and depricate (Solved by only importing the two relevant functions used by the `Locale` object)
- [x] Replace all `any` types with more accurate ones
- [x] Fix Range Keyboard functionality
- [x] Clean up mess 🙏
- [x] fix type errors
- [x] Fix shortcuts
- [x] Fix cancel button
- [x] Stop cancel button from resetting date
- [x] Fix latest discrepancies
- [x] Figure out the `DatePickerEvent` situation
- [x] Rename `ReturnObject`-types to give better context
- [x] Fix Input date removal when max or min is active
- [x] Add tests range shortcuts
- [x] Add tests for `link` selection
- [x] Add tests for range pickers cancel button behaviour
- [x] Add tests to make sure shortcut callback receives all dates as parameters
- [x] Add tests for `endDate` backspace removal when in `range`
- [x] Update month handling
- [x] Add more tests for range view functionality
- [x] Fix clicking same date for end and start date while in range mode, and add test
- [x] Contemplate life 🤔
- [x] Fix range selection of start and end date in same month behaviour
- [x] Fix date selection with arrow keys and add test

PR #3176, opened 2024-01-05
fix(DatePicker): fix bug where minDate set to today is disabled

joakbjerk:
`minDate` would be disabled was set to be `new Date()`, since the `isAfter` function that does the comparison to determine of the date should disabled in the calendar, also compares time. The calendar `date` param passed down to the `isDisabledCalc` has time set to `00:00:00` meaning that `date` is after `minDate` if its set to `new Date()` since `minDate` can have a time of `16:57:13`

Before:
Screenshot 2024-01-05 at 09 41 01

After:
Screenshot 2024-01-05 at 09 40 44

### DIFF
--- a/packages/dnb-eufemia/src/components/date-picker/DatePickerCalc.js
+++ b/packages/dnb-eufemia/src/components/date-picker/DatePickerCalc.js
@@ -170,12 +170,19 @@ const isWithinSelectionCalc = (date, startDate, endDate) => {
 // date is before minDate or after maxDate
 const isDisabledCalc = (date, minDate, maxDate) => {
   // isBefore and isAfter return false if comparison date is undefined, which is useful here in case minDate and maxDate aren't supplied
+  // resetting the the `minDate` and `maxDate` hours for comparison, as the `date` param comes with a time of 00:00:00
   return (
-    (minDate && isBefore(date, minDate)) ||
-    (maxDate && isAfter(date, maxDate))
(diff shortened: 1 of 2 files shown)


packages/dnb-eufemia/src/components/date-picker/DatePickerCalc.js
```diff
@@ -170,13 +170,21 @@ const isWithinSelectionCalc = (date, startDate, endDate) => {
 // date is before minDate or after maxDate
 const isDisabledCalc = (date, minDate, maxDate) => {
   // isBefore and isAfter return false if comparison date is undefined, which is useful here in case minDate and maxDate aren't supplied
+  // resetting the the `minDate` and `maxDate` hours for comparison, as the `date` param comes with a time of 00:00:00
   return (
-    (minDate && isBefore(date, minDate)) ||
-    (maxDate && isAfter(date, maxDate))
+    (minDate && isBefore(date, resetHours(minDate))) ||
```
tujoworker, line 175: DateFNS has a function called `startOfDay` – can you check if that does the same as your reset function? Then `startOfDay` should be used.

tujoworker (approved):
Nice 👍 

Also, its probably a good time to rebase #2799

joakbjerk:
> Nice 👍
> 
> Also, its probably a good time to rebase #2799

True 😄

PR #3178, opened 2024-01-05
fix(DatePicker): add return object for onBlur

joakbjerk:
Adds a return object for the onBlur event, to make it possible to extract the date for the event object.

### DIFF
--- a/packages/dnb-eufemia/src/components/date-picker/DatePicker.d.ts
+++ b/packages/dnb-eufemia/src/components/date-picker/DatePicker.d.ts
@@ -260,11 +260,6 @@ export interface DatePickerProps
   ) => void;
 }
 
-type TestType = {
-  target: any;
-  value: any;
-};
-
 export default class DatePicker extends React.Component<
   DatePickerProps,
   any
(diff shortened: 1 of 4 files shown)


packages/dnb-eufemia/src/components/date-picker/DatePicker.d.ts
```diff
@@ -247,8 +255,11 @@ export interface DatePickerProps
   /**
    * Will be called once the input loses focus.
    */
-  onBlur?: (event: React.FocusEventHandler<HTMLInputElement>) => void;
+  onBlur?: (
+    event: DatePickerEvent<React.FocusEvent<HTMLInputElement>>
+  ) => void;
 }
+
 export default class DatePicker extends React.Component<
```
tujoworker, line 263: Should probably not be there 😉

tujoworker (approved):
Nice catch! Thanks for the tests too 🙏

packages/dnb-eufemia/src/components/date-picker/DatePicker.d.ts
tujoworker: Ahh. You saw it already 👍

joakbjerk: Ahh, thought you meant the space, silly Github 😄

PR #3180, opened 2024-01-05
fix(Autocomplete): only set aria-controls attribute when expanded

tujoworker:
Because the corresponding HTML element does not exists in the DOM when not opened.


joakbjerk (approved):
👨🏻‍🔧

PR #3183, opened 2024-01-06
chore(Field.Number): center align value when showStepControls is used

tujoworker:
Use `chore` as we have not released this feature yet.


joakbjerk (approved):
⁑

PR #3185, opened 2024-01-07
chore(useDataValue): refactor

tujoworker:
Move `trim` and `capitalize` feature out of `useDataValue`, because they are a Field.String feature. To be able to make that move, we need an additional transformer simply called `transformValue`, because it is meant to just do that.

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/create-component/useDataValue/info.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/create-component/useDataValue/info.mdx
@@ -173,25 +173,12 @@ The transformers are hooks to transform the value on different stages.
 
 They should return a transformed value: `(value) => value`
 
-- `toInput` transforms the value before it gets returned by the hook:
+- `toInput` transforms the value before it gets returned as the `value`:
 
-```ts
-const { value } = useDataValue(props)
-```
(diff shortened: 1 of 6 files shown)


packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/create-component/useDataValue/info.mdx
```diff
@@ -173,25 +173,12 @@ The transformers are hooks to transform the value on different stages.
...
+- `fromInput` transforms the value given by `handleChange` before it is used in the further process flow. Use it to destruct the value form the original event object.
 
-- `fromInput` transforms the value given by `handleChange` before it is used in the further process flow.
-
-```ts
-handleChange(value)
-```
-
-- `toEvent` transforms the internal value before it gets returned by even callbacks such as `onChange`, `onFocus` and `onBlur`.
+- `toEvent` transforms the internal value before it gets returned by even callbacks such as `onChange`, `onFocus` and `onBlur`. The second parameter returns the event type: `onChange`, `onFocus`, `onBlur` or `onBlurValidator`.
```
joakbjerk, line 180:
```suggestion
- `toEvent` transforms the internal value before it gets returned by event callbacks such as `onChange`, `onFocus` and `onBlur`. The second parameter returns the event type: `onChange`, `onFocus`, `onBlur` or `onBlurValidator`.
```

joakbjerk (approved):
Great refactor, makes more sense 👍

PR #3187, opened 2024-01-08
fix(FieldBlock): ensure extra spacing of labels when on small screens

### LINKED ISSUE #3102: bug(Forms): Different amount of spacing between Field components.
To me, when browsing the docs on iPhone in Safari, it seems like there’s different spacing between the Field components in a card demo/example here https://eufemia.dnb.no/uilib/layout/#flex-usage-in-forms

IMG_4967

To me, it seems like the spacing is different between the fields e-post and landskode, compared to spacing between all other fields.

snorrekim:
Fixes #3102


tujoworker (approved):
Very good to reuse what already exists 💯 ❤️

PR #3189, opened 2024-01-08
chore(Field.Date): wrap in FieldBlock

tujoworker:
Relies on #3188 and will be rebased once it is in main: ✅


joakbjerk (commented):
Should we perhaps set a default width so the component is not so wide out of the box?
Screenshot 2024-01-08 at 14 39 21

and the error styling seems to not be applied anymore
Screenshot 2024-01-08 at 14 40 52

tujoworker:
Yes! Nice you check the actual outcome 🥇 – I've added two visual tests now ✅ 
Also, we do not handle width, as we are better with the given width. Also for future compatibility for date-range support.

joakbjerk (approved):
Great, looks perfect now! 🔥

Issue #3191, opened 2024-01-08
bug(DatePicker): set correct cursor position when pressing ArrawRight 


tujoworker:
## Issue

Once selecting the input and pressing the RightArrow key, the cursor is gets set on position 1 (between the two first charts).

Video:

https://github.com/dnbexperience/eufemia/assets/1501870/c70d5755-d796-4b5d-9a45-6a414a551372 

## Next step

I assume that this works as expected once we have:

1. Get in this PR #2799
2. And use once we use `MultiInputMask`(#2736)

Because the [Expiry field](https://eufemia.dnb.no/uilib/extensions/forms/feature-fields/Expiry/) does not have this issue.

But I still wanted to report this issue.

langz:
This issue was fixed or made redundant in https://github.com/dnbexperience/eufemia/pull/6658

PR #3192, opened 2024-01-08
fix(MultiInputMask): enhance handling of right, left and backspace key usage

tujoworker:
This PR will change the behavior when a user users the keyboard keys on a `MultiInputMask` input. When a selection is made, it should not navigate to the next field.

Its related to this issue #3191 – even `MultiInputMask` did not have that issue – it now enhance its appearance even more.

Before it jumps to the next input field:
https://github.com/dnbexperience/eufemia/assets/1501870/049d8ead-2197-4f6c-a14d-e1b026f154a5

After it does stay on the same input field:
https://github.com/dnbexperience/eufemia/assets/1501870/af8d0030-6c84-43d5-b6a8-43430cc930c0

### DIFF
--- a/packages/dnb-eufemia/src/components/input-masked/__tests__/MultiInputMask.test.tsx
+++ b/packages/dnb-eufemia/src/components/input-masked/__tests__/MultiInputMask.test.tsx
@@ -4,7 +4,7 @@
  */
 
 import React from 'react'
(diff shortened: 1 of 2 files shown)


joakbjerk (approved):
Nice! 🙏

PR #3198, opened 2024-01-09
fix(Flex.Container): show line divider even when heading is present

### LINKED ISSUE #3078: bug(Forms): using Form.SubHeading inside a card with stack doesn't display the stack border
## 🐛 Bug Report

Using `Form.SubHeading` inside a card with stack doesn't display the stack border between the elements in the card.

image

I'm not 100% sure what we want the expected result to be here, but I find it strange that the "stack border" doesn't display in this scenario.

There's a lot of demos/examples in the docs that's written like this, and the "stack border" is not displayed. So it would be nice to know if it should be like this, or not 🙏 

## To Reproduce

Take a look in the following [CSB](https://codesandbox.io/p/sandbox/eufemia-starter-forked-jwp3gq?file=%2Fsrc%2FApp.tsx%3A27%2C21).
It includes a few different demos/examples, using Form and Card stack.

tujoworker:
Fixes #3078


joakbjerk (approved):
Looks good, thank you! 🔥

PR #3207, opened 2024-01-10
fix(Textarea): ensure correct height based on rows for Firefox browser

tujoworker:
Before:
Screenshot 2024-01-10 at 15 08 15

After:
Screenshot 2024-01-10 at 15 08 28


joakbjerk (approved):
🎆

PR #3209, opened 2024-01-10
chore(forms): remove labelDescription and labelSecondary

tujoworker:


joakbjerk (approved):
🙏

tujoworker:
@kimroen Thank you for following up this potential issue. 

We have already added a better version to the original textarea, this time with screen reader support. 

PR: https://github.com/dnbexperience/eufemia/pull/3210

But you use it on single line inputs?

kimroen:
Thanks! It sounds like the fields documentation might need updating then, and mentioning it in the release notes would be appreciated 👏

We do indeed, for short fields like KID-number for instance.

tujoworker:
Yeah, regarding the release notes – I strongly agree and I'm sorry for the confusion. The `labelDescription` and `labelSecondary` was never a part of Eufeia. But projects used it anyways. That's not how it should be. But I think, at least the `labelDescription` we should get back in.
 
Regarding the counter; UX decided, they don't want a counter for normal fields. Also, I just got feedback, that it should up and down. But let me see what we can do about the counter for inputs.

tujoworker:
We add back in [labelDescription](https://github.com/dnbexperience/eufemia/pull/3251) for v10.19 🙌

kimroen:
Thanks for following up here!

> Regarding the counter; UX decided, they don't want a counter for normal fields.

Thanks, I'll take this back to our UX designers to see what they think we should do instead for these fields.

> Also, I just got feedback, that it should up and down.

I don't think I understand what this means 🙈

tujoworker:
@kimroen we have the same props as in v10.17, except labelSecondary. While characterCounter accepts now a number (max) or the props for TextCounter.

kimroen:
Thanks! I tried this now, and it seems that `characterCounter` still has no effect on single-line fields (but it's a valid prop again now at least). Is that correct and intentional? If so, it seems I misunderstood what you were saying here.

I took this back to UX, and they had been part of discussions about this and did not have the impression that the counter would only be for multiline input fields.

Separately, the docs for `Fields.String` still says that `characterCounter` accepts a boolean, but the types seem to suggest that it only accepts a number or an object.

kimroen:
> Separately, the docs for `Fields.String` still says that `characterCounter` accepts a boolean, but the types seem to suggest that it only accepts a number or an object.

Looks like the type is correct, but perhaps supporting boolean and falling back to the supplied `maxLength` would be a little more backwards compatible if it used to support boolean too?



I like this new counter for multiline fields though, don't get me wrong! We'll go back to the drawing board here for how to design the KID number field, and just need to keep this in mind when upgrading Eufemia for existing applications.

tujoworker:
Are you able to use the TextCounter for your use case?

PR #3210, opened 2024-01-10
feat(Textarea): add characterCounter

tujoworker:
- [x] Implement screen reader support (relies on #3217)
- [x] Align styles
- [x] Align texts in counter
- [x] Add tests
- [x] Add visual test

A couple of visual snapshots needed to get update because a new margin was added, to align elements below the checkbox.


joakbjerk (approved):
𝍡

PR #3213, opened 2024-01-11
fix(NumberFormat): setting selectall to false will no longer select all

joakbjerk:


tujoworker (approved):
Nice – and sweet test!

PR #3214, opened 2024-01-11
fix selectall logic

mallsoft:


tujoworker:
Than you so much for your contribution 💯 right before you, @joakbjerk made the same fix #3213 with tests. So we need to close this PR. But I hope that not holds you back from a new contribution. 

Also, I can add you to the contributors team, so you can push directly to the repo. Is that something you will consider?

mallsoft:
💯  
Hehe yeah i saw someone had made a PR right after

Issue #3028, opened 2023-12-08
bug(Field.Selection): value does not display which value is selected when field is disabled and variant is button


langz:
## 🐛 Bug Report

In the light of the following two CSBs:
- [All Field.Components and variants with disabled and `without a value`](https://codesandbox.io/p/sandbox/eufemia-starter-forked-gn3cg7?file=%2Fsrc%2FApp.tsx%3A59%2C8-59%2C23)
- [All Field.Components and variants with disabled and `with a value`](https://codesandbox.io/p/sandbox/eufemia-starter-forked-6wry6m?file=%2Fsrc%2FApp.tsx)

It's observed that the following code does not display which value that's selected/used:
```
<Field.Selection
        value="bar"
        path="/selection2"
        label="Selection button"
        disabled
        variant="button"
      >
        <Field.Option value="foo" title="Foo!" />
        <Field.Option value="bar" title="Baar!" />
</Field.Selection>
```

Results in:
<img width="266" alt="image" src="https://github.com/dnbexperience/eufemia/assets/1359205/bf626f96-d98b-4b03-b2da-db30dc38f441">

## To Reproduce

[CSB with a reproduction](https://codesandbox.io/p/sandbox/eufemia-starter-forked-tfj5wg?file=%2Fsrc%2FApp.tsx%3A17%2C5)

Issue #3027, opened 2023-12-08
bug(Field.Boolean): value does not display which value is selected when field is disabled and variant is buttons


langz:
## 🐛 Bug Report

In the light of the following two CSBs:
- [All Field.Components and variants with disabled and `without a value`](https://codesandbox.io/p/sandbox/eufemia-starter-forked-gn3cg7?file=%2Fsrc%2FApp.tsx%3A59%2C8-59%2C23)
- [All Field.Components and variants with disabled and `with a value`](https://codesandbox.io/p/sandbox/eufemia-starter-forked-6wry6m?file=%2Fsrc%2FApp.tsx)

It's observed that the following code does not display which value that's selected/used:
```
      <Field.Boolean
        value={true}
        path="/boolean3"
        label="Boolean buttons"
        variant="buttons"
        disabled
      />
```

Results in:
<img width="192" alt="image" src="https://github.com/dnbexperience/eufemia/assets/1359205/3c7d6977-358c-4cf0-abda-66027b4ecf1c">

## To Reproduce

[CSB with a reproduction](https://codesandbox.io/p/sandbox/eufemia-starter-forked-25r825?file=%2Fsrc%2FApp.tsx%3A11%2C26)

PR #3224, opened 2024-01-15
fix(ToggleButton): add checked disabled style

joakbjerk:
proposal to solve https://github.com/dnbexperience/eufemia/issues/3028 and https://github.com/dnbexperience/eufemia/issues/3027

Not sure if its the best way though.

before:
Screenshot 2024-01-15 at 10 12 42

after:
Screenshot 2024-01-15 at 10 12 22


tujoworker (approved):
Looks good! Did you check the Sbanken styles?

tujoworker:
Also, do we have a visual test on the disabled state? Then we may add a "checked" example state in there as well (same test).

joakbjerk:
I'll add the same to Sbanken and add visual tests 🙏

tujoworker:
I added a commit. We actually had a visual test for the disabled state. So we just need to make one of them as "selected".

PR #3225, opened 2024-01-15
fix(Field): show error state without error object if parent FieldBlock has error

snorrekim:
Fixes #2958

The `hasError` prop is used to tell "regular" input components that they have an error


tujoworker (approved):
Perfect! Looks good. 

Would be great to have tests these cases as well 🚀

PR #3235, opened 2024-01-17
fix(Input): should not clear input value with escape key

tujoworker:
When an input is used inside a Dialog/Modal/Drawer, the escape key should only close the modal. Key clear button is still available for keyboard users by tabbing to the clear button.


joakbjerk (approved):
🆑

PR #3240, opened 2024-01-18
chore(DatePicker): document correct_invalid_date prop, and make sure functionality fires when min_date or max_date is set

joakbjerk:
Several developers have noted that the user is still able to manually input invalid dates in the `DatePickerInput` even though a `max_date` or `min_date` has been defined. This is a proposal to solve that.

It turns out that we already had functionality and tests to correct this with the `correct_invalid_date` prop, but no documentation for the prop itself in the portal. I have added the missing documentation.

I also made it so that the correction functionality triggers when either the `min_date` or `max_date` props are set, to reduce the amount of component configuration developers have to do, since a lot of people where not aware of this prop. 

- [x] Remove newly added implementation.
- [x] Document `correct_invalid_date prop`.
- [x] Make `correct_invalid_date prop` opt out when `min_date` or `max_date` is given.
- [x] Make it work when only `max_date` is given


tujoworker:
Cam you add a test for all three props? 🙏

joakbjerk:
> Cam you add a test for all three props? 🙏

Woops, sure 😄

tujoworker (approved):
Nice ⭐

PR #3243, opened 2024-01-18
fix(Field): Refactor Field and Value to use consistent namespace and alias

snorrekim:
We can't really use `Field.String` inside another Field component, because that would lead to a cyclical dependency. So instead I've used the alias `StringField` where needed. I've also made a clearer distinction between `Value.String` and `Field.String` as that was causing a lot of confusion.


tujoworker (approved):
This is a solid improvement – thank you 💯

tujoworker:
> maybe we could rename all the field component **files** with the suffix "Field"? `StringField.ts`.

I agree, that this makes sense as well.

PR #3244, opened 2024-01-18
feat(Flex): add `line-framed` to divider

tujoworker:
Relies on PR #3242
Also, I think, a white background for [this example](https://eufemia-git-feat-flex-container-bookend-line-eufemia.vercel.app/uilib/layout/flex/container/#framed-line-dividers), would be better. Therefor, we may get in this PR #3245 for that.

image

I think, `line-framed` describes the use-case good. But maybe something different is better?

- outlined
- grooved
- bordered
- enclosed


joakbjerk (approved):
Nice! 🔥

Issue #3199, opened 2024-01-09
Dropdown is not showing the currently selected value


vron:
## 🐛 Bug Report

Then using value to set the initial value the Dropdown is not showing the correct item

## To Reproduce

Goto the link below and try the buttons - I woudl have expected the dropdown to show the correct value..

https://codesandbox.io/p/sandbox/eufemia-starter-forked-f8k2dg?file=%2Fsrc%2FApp.tsx%3A45%2C14

## Eufemia Version

Latest

PR #3247, opened 2024-01-19
fix(DrawerList): update original data when data prop changes

joakbjerk:
closes https://github.com/dnbexperience/eufemia/issues/3199

Issue caused by the the function updating selected items using the `original_data` to find and set the selected item. This causes a issue since `original_data` does not update on `props.data` change.


tujoworker:
We need a test for this that will verify the change and fail without. This way, a later rewrite of the component, can be verified easily.

PR #3248, opened 2024-01-19
chore(useData): enhance flexibility

tujoworker:
When the initial value was given in the Form.Handler, it never got set in useData. Now we have several test cases for this too. This change needed some more rewrite of the internal structure. Also, we move from state to ref in the useData hook.


joakbjerk (approved):
Nice! 🏆

PR #3250, opened 2024-01-20
feat(TextCounter): add new fragment used in Textarea

tujoworker:
This is to compliment PR #3210 and make it possible to easier use it in other cases as well.


joakbjerk (approved):
Great addition! 🙏

PR #2762, opened 2023-10-18
fix(Breadcrumb): fix rehydration disturbance

tujoworker:
Ensure to not update the state before the first component mount.

PR #3254, opened 2024-01-21
fix(Breadcrumb): fix rehydration disturbance

tujoworker:
This PR #2762 was originally merged into #2671 – and was with that never merged into main.


joakbjerk (approved):
🏖️

PR #3257, opened 2024-01-22
fix(forms): rename `contents` to `content`

tujoworker:
This PR should align the term we already use before in other components.

Also, internally used only, we rename `contentsWidth` to `contentWidth`.


joakbjerk (approved):
🥳

PR #3262, opened 2024-01-22
feat(NumberUtils.format): Only return object if returnAria: true

kimroen:
The function `format` only returns an object if the option `returnAria` was passed in as `true`, but TypeScript doesn’t know this.

With this change, we’re making an overload for the function that only returns the object in the right situation. This way, the caller doesn’t need to check the returned value before being able to use it as a string (or number).

The `formatReturnType` type is now unused, but given it's been exported I opted to not remove it as that might be considered a breaking change.

You can consider this more of a suggestion than a request though, I'm not sure what your policy or testing setup is like for type changes :)


tujoworker (approved):
Thank you! Very nice 👌

PR #3272, opened 2024-01-24
fix(Flex.Container): enhance children parsing

tujoworker:


packages/dnb-design-system-portal/src/docs/uilib/layout/flex/container/info.mdx
```diff
@@ -49,6 +49,8 @@ render(
 )
 ```
 
+During render, the items within the "Wrapper" container are wrapped with the same properties. This ensures that all the items have a same appearance.
```
joakbjerk, line 52:
```suggestion
During render, the items within the "Wrapper" container are wrapped with the same properties. This ensures that all the items have the same appearance.
```

joakbjerk (approved):
Looks good 🙏

PR #3293, opened 2024-01-31
HeightAnimation maintenance and improvements

tujoworker:
This is a collection PR where each commit represents a fix that will make its way to the release notes. This is because the changes in HeightAnimation do effect all these components.

The reason for the changes is to avoid flickering and an overall more stable height animation detection by cloning the element, instead of manipulating the styles on the target element.


packages/dnb-eufemia/src/components/height-animation/useHeightAnimation.tsx
```diff
@@ -54,147 +59,206 @@ export function useHeightAnimation(
...
+       * In certain cases, the targetRef.current is outdated, so we need to set it again.
+       * E.g. in Tabs.js, when we switch tabs, the targetRef.current is not updated.
+       */
+      instRef.current.setElement(targetRef.current)
+
+      /**
+       * Ensure we don't have height, while we get the "toHeight" again.
        * We may move this inside of the HeightAnimationInstance class later,
-       * but the "GlobalStatus" is currently relaying on "getUnknownHeight" inside of adjustTo
+       * but the "GlobalStatus" is currently relaying on "getUnknownHeight" inside of adjustTo.
```
joakbjerk, line 251:
```suggestion
       * but the "GlobalStatus" is currently relying on "getUnknownHeight" inside of adjustTo.
```

packages/dnb-eufemia/src/extensions/forms/FieldBlock/style/dnb-field-block.scss
```diff
@@ -80,6 +80,9 @@ fieldset.dnb-forms-field-block {
 
   &__status {
     grid-area: status;
+    .dnb-form-status__shell {
+      margin-top: 0.5rem; // use padding, so we can animate the height
```
joakbjerk, line 84: should this be `padding-top` instead of `margin-top`, or am I misreading the comment?

joakbjerk (approved):
Nice! Good to see all that `css` being shaved off 😄

PR #3295, opened 2024-02-01
feat(Slider): Add marker functionality

Sundfjord:
This PR adds ability to "mark" a given point in a Slider through a new `marker`prop, which accepts a `value` prop determining where to set the marker, as well as a `text` prop which, if given, will display the given text in a tooltip.

image


tujoworker:
We need to test with different font sizes and screen readers. We may look at it next Tuesday in the meeting. 👍

tujoworker (approved):
Very nice work ⭐

PR #3296, opened 2024-02-02
fix(DatePicker): make return object have corrected value

joakbjerk:
Up until now, the return object for the different `DatePicker` events did not included the corrected date value of `correct_invalid_date` was true.


tujoworker (approved):
Nice 👌

PR #3306, opened 2024-02-07
feat(FieldBlock): enhance support for composition fields

tujoworker:
This PR adds docs, examples and tests on how to create composition fields. 

- We add easy handling of the hasError prop we get from useDataValue.

- We also lists combined status messages with a list, instead of a pipe.

- If a status message appears to be the same, we now show just one of them.

- We also add `status` and `warning` to be supported by FieldBlock.

- Enhance handling of aria attributes.

We also needed to change how MultiInputMask has handled aria labels, as axe don't allows invisible aria-labelledby and aria-describedby combined. So we used there aria-label instead.


packages/dnb-eufemia/src/components/form-status/stories/FormStatus.stories.tsx
```diff
@@ -158,7 +158,7 @@ export const ToggleAnimation = () => {
   const toggleStatus = () => {
     setStatus((s) => (!s ? 'You have to fill in this field' : null))
   }
-  const [showError, setShowError] = useState(false)
+  const [showError, showFieldError] = useState(false)
```
joakbjerk, line 161:
```suggestion
  const [showError, setShowError] = useState(false)
```

packages/dnb-eufemia/src/shared/component-helper.js
```diff
@@ -685,8 +685,15 @@ export function combineLabelledBy(...params) {
 export function combineDescribedBy(...params) {
   return combineAriaBy('aria-describedby', params)
 }
+export function combineDetails(...params) {
+  return combineAriaBy('aria-details', params)
+}
 function combineAriaBy(type, params) {
   params = params.map((cur) => {
+    // console.log('cur', cur)
```
joakbjerk, line 693:
```suggestion
  
```

joakbjerk (approved):
Looks good to me, and from testing in the portal 🔥🥇 Just a few nit picky things 🙏

PR #3312, opened 2024-02-12
feat(InfoCard): add optional dropShadow-prop

thaytharma:
Added optional `dropShadow`-prop to `InfoCard`-component.

Skjermbilde 2024-02-12 kl  16 24 17

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/components/info-card/Examples.tsx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/components/info-card/Examples.tsx
@@ -104,6 +104,16 @@ export const InfoCardCentered = () => (
   </ComponentBox>
 )
 
+export const InfoCardWithoutDropShadow = () => (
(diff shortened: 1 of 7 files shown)


packages/dnb-design-system-portal/src/docs/uilib/components/info-card/Examples.tsx
```diff
@@ -104,6 +104,16 @@ export const InfoCardCentered = () => (
   </ComponentBox>
 )
 
+export const InfoCardWithoutDropShadow = () => (
+  <ComponentBox data-visual-test="info-card-drop-shadow">
```
tujoworker, line 108:
```suggestion
  <ComponentBox>
```
If we need a visual test on this, we need to add it on the test file as well.
But I think, we probably don't need a visual test for this case as its a minor visual detail.

tujoworker (approved):
Very nice work! Thank you ⭐

PR #3317, opened 2024-02-15
feat(NumberFormat): align currency display of "name" to be displayed without Norwegian

tujoworker:
To adapt the formatting of currency to what the "Norske Språkrådet" recommends, we remove "Norwegian" (norske) when currency display is set to name.


joakbjerk (approved):
Nice! 👑

PR #3318, opened 2024-02-15
chore(Slider): refactor to not include the marker by default (slider extension)

tujoworker:
This way, the JavaScript code will not be exposed in app bundles, if they don't use the marker.


joakbjerk (approved):
✅

PR #3322, opened 2024-02-16
chore: add DocTable for easy documenting properties inside components directory

tujoworker:
This PR introduces a new way of how we may/can place and write our component properties. The idea is that we move it from Markdown Tables to TypeScript, but inside the component directory itself.

This opens up for more flexibility and much better Git diffs as we never will change the whole table, just because the max lengths has changed of one description.

For now, only Input, InputMasked and Textarea are moved. 
Later we also may move the events properties.

Questions:

- What about the naming of things?


joakbjerk (approved):
Nice 👍 Looks more flexible. And I think `DocTable` works fine. It's either that or `PropertyTable` I guess 🤔

PR #3324, opened 2024-02-16
chore(Portal): properly simulate active and height changes in screenshots

snorrekim:
* simulating `active` no longer clicks the element
* screenshots can recalculate height after simulating


tujoworker (approved):
Very nice fix! Looks good to me ⭐

PR #3338, opened 2024-02-26
fix(FormLabel): align width to fit content instead of being stretched 100% on clickable labels

tujoworker:
The visual test does also verify that new appearance. Visually it should not make any difference. But the "clickable" area will be more logically. 

Before
Screenshot 2024-02-26 at 09 55 40

After
Screenshot 2024-02-26 at 09 55 50


joakbjerk (approved):
Looks good 🔥

PR #3344, opened 2024-02-26
fix(ProgressIndicator): align label spacing when size is `small` and use span instead of divs to support inline elements

tujoworker:
After:
Screenshot 2024-02-27 at 08 27 20

Before:
Screenshot 2024-02-27 at 08 27 40

### DIFF
--- a/packages/dnb-eufemia/src/components/progress-indicator/ProgressIndicator.js
+++ b/packages/dnb-eufemia/src/components/progress-indicator/ProgressIndicator.js
@@ -124,9 +124,9 @@ export default class ProgressIndicator extends React.PureComponent {
       class: _className,
       children,
       title,
-      progress: _progress, //eslint-disable-line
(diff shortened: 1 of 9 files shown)


packages/dnb-eufemia/src/components/progress-indicator/ProgressIndicator.js
```diff
@@ -179,11 +180,11 @@ export default class ProgressIndicator extends React.PureComponent {
           />
         )}
         {indicatorLabel && (
-          <div className="dnb-progress-indicator__label">
-            <p className="dnb-p">{indicatorLabel}</p>
-          </div>
+          <span className="dnb-progress-indicator__label">
+            <label className="dnb-p">{indicatorLabel}</label>
```
joakbjerk, line 184: is the label correctly connected to the progress indicator now?

joakbjerk (approved):
Looks good 👍

PR #3348, opened 2024-02-27
fix(Form.useData): add support for StrictMode

tujoworker:


joakbjerk (approved):
Nice, great job 😄 works perfectly when testing and comparing between versions in codesandbox 🙏

PR #3349, opened 2024-02-28
chore(useLocale): add localisation hook

joakbjerk:
Conversion of locale files from `JS` to `TS` will happen in a different PR, as this caused a error in the type testing, that is a little convoluted and time consuming to fix.

- [x] Create hook
- [x] Connect to `SharedContext`
- [x] Add tests
- [x] Remove locale parameter in function call
- [x] Add possibility to retrieve component specific translations
- [x] Create dynamic return type for better developer experience


tujoworker (approved):
Nice work 💪

PR #3351, opened 2024-02-28
feat(FormStatus): remove hover color, so it does not change its color anymore

tujoworker:
This is based on a UX decision from the web maintainer group. It also effects radio buttons, toggle buttons, switch and checkbox.


joakbjerk (approved):
Great! Looks much better and less "noisy" now that the highlighting is only given to the actual interactive element 🔥

Just one issue. It seems like all inputs inside a `FieldBlock` receives a focus border when a `label` is hovered over. Tested [here](https://eufemia-git-feat-form-label-hover-eufemia.vercel.app/uilib/extensions/forms/create-component/FieldBlock/)

https://github.com/dnbexperience/eufemia/assets/25927156/f6690cb8-c414-4b14-80e2-178402d44efb

tujoworker:
Nice findings. Corrected ✅

PR #3359, opened 2024-03-01
feat(Button): refactor button css

snorrekim:
* Padding is always used on the "outside" of text and icon
* Margin is always used between text and icon
* Css based on button height and icon size is now calculated using variables


packages/dnb-eufemia/src/components/global-status/__tests__/__image_snapshots__/globalstatus-for-ui-have-to-match-the-close-button-in-focus-state.snap.png
tujoworker, line 1: It appears that the left padding on the button has been altered. Was this change intentional?

tujoworker (approved):
Really nice work ⭐ 

Left one comment you may check out before merge.

packages/dnb-eufemia/src/components/global-status/__tests__/__image_snapshots__/globalstatus-for-ui-have-to-match-the-close-button-in-focus-state.snap.png
snorrekim: Yes, that's intentional, the tertiary focus effect was too wide compared to the design

PR #3360, opened 2024-03-01
chore(Docs): remove deprecated pxtoem

joakbjerk:
the previous owner of pxtoem might have sold the domain, as the link as of today takes you to an online casino 🎰 which has nothing to do with px nor em.


tujoworker (approved):
👌

PR #3365, opened 2024-03-04
chore(CI): reposition screenshot wrapper to whole number coordinate

snorrekim:
If the screenshot element (the wrapper we create) has an x coordinate that isn't a whole number, the screenshot will add an extra pixel to the bottom. This causes inconsistent behaviour based on changes in the previous elements. If we simply add that decimal distance to the `top` of the wrapper, we will get more consistent results.


tujoworker (approved):
Nice finding and solution 👏

PR #3371, opened 2024-03-06
feat: show error in console when Eufemia CSS and JS versions do not match

tujoworker:
It will show an error like this:

Screenshot 2024-03-06 at 20 58 40


joakbjerk (approved):
Great idea 💡

PR #3372, opened 2024-03-06
fix(Table): fix type error when using array of elements

snorrekim:
`children` prop should rarely, if ever, be checked using typescript typing. It has limited effect, and is more likely to cause issues than not.


tujoworker (approved):
👍

PR #3374, opened 2024-03-07
chore(DatePicker): make correct_invalid_date opt-in

joakbjerk:
Remove the automatic correction of invalid dates when `min_date` or `max_date` is defined.

This change also reflects the expected behavior in the documentation.


packages/dnb-eufemia/src/components/date-picker/__tests__/DatePicker.test.tsx
```diff
@@ -593,265 +593,6 @@ describe('DatePicker component', () => {
     expect((yearElem as HTMLInputElement).value).toBe(year)
   })
 
-  it('has to auto-correct invalid min/max dates', async () => {
```
tujoworker, line 596: Why should we remove the tests?

joakbjerk: Woops, deleted by mistake, well add it back 🙏 Only the tests related to the automatic date correction should be removed.

tujoworker (approved):

PR #3375, opened 2024-03-07
fix(Tag): delete and backspace key press for removable tag

snorrekim:
* Fixed Delete and Backspace not triggering `onClick` for `removable` variant
* Fixed mistakes in documentation and updated examples
* Handling of deprecated behaviour is separated into its own function


snorrekim:
Tobias suggestion: use FieldBlock label, not heading for remove/add example

tujoworker (approved):
Nice work ⭐

PR #3377, opened 2024-03-08
feat(Anchor): Sbanken style based on icon

snorrekim:
* Sbanken Anchor styling now depends on wether it has an icon or not (instead of on the class `.dnb-anchor--inline`)
* Anchors without an icon will use the "inline" style, high contrast underline, and no extra padding
* Anchors with icon will use the green underline and add extra padding
* class `.dnb-anchor--inline` can still be used to force styling
* documentation now focuses on normal use cases: link in test, link with icon

### Refactoring:
These changes are also planned for dnb theme in a seperate PR.
* Sbanken icon is now `inline`, instead of `inline-block`
* Sbanken icon no longer wraps to new line alone, it includes the closest word
* Sbanken icon scales with font size


tujoworker (approved):
Perfectissimo 🪄

PR #3378, opened 2024-03-12
feat(forms): implement support for async form submission and autosave feature

### LINKED ISSUE #3333: feat(forms): async handling of onSubmit and validator / onBlurValidator
When `onSubmit` of the `Form.Handler` uses an async function, the form should:

- show an indicator on the form bottom
- disable form fields

When `validator` or `onBlurValidator` of a `Field` uses an async function, the form should:

- show an indicator on the field (label would be a good place a11y wise, UX RFC will be created)
- show an indicator on the form bottom
- disable form fields (when blur)

`validator` should not disable the current field, while `onBlurValidator` should disable the related field.

tujoworker:
This PR adds a couple of new event names for the `Form.Handler`; 

- Such as `onSubmitAsync` and `onChangeAsync` will lets you make "async" operations very easily. It will handle the form state for you (disable all input fields) and show an indicator. 
- When an "async" validator or onBlurValidator are used, it will behave similar as described above (showing an indicator on the corresponding field label).
- When an "async" field `onChange` event handler is used, it will behave similar as described above (showing an indicator on the corresponding field label).

**Feedback wanted:**

- Should we use the suffix `Async` on the onSubmit and onChange events for the `Form.Handler`? 
- Technically, we can check if the given function is async and let the form behave after that.
- We do that with the field `onChange` and the `validator` and `onBlurValidator`, because it would be too messy to have so many extra properties.

**Update**: We are going for only using `async` function as the on/off switch for the async form behavior.

**Additional**

- This is a multi feature PR with several features separated in each commit.
- closes #3333

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/extended-features/Form/SubmitIndicator/info.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/extended-features/Form/SubmitIndicator/info.mdx
@@ -16,4 +16,4 @@ render(<Form.SubmitIndicator state="pending" />)
 
 ### Integration
 
-The indicator is implemented by default in the [SubmitButton](/uilib/extensions/forms/extended-features/Form/SubmitButton/) and [FieldBlock](/uilib/extensions/forms/create-component/FieldBlock/) label. It will be displayed when the `onSubmit` event is triggered or when `enableAsyncBehavior` is set to true on the [Form.Handler](/uilib/extensions/forms/extended-features/Form/Handler/) component. You find some examples down below.
+The indicator is implemented by default in the [SubmitButton](/uilib/extensions/forms/extended-features/Form/SubmitButton/) and [FieldBlock](/uilib/extensions/forms/create-component/FieldBlock/) label. It will be displayed when the `onSubmit` event handler on the [Form.Handler](/uilib/extensions/forms/extended-features/Form/Handler/) component is an async function. You can find some examples down below.
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/extended-features/StepsLayout/StepsLayout/Examples.tsx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/extended-features/StepsLayout/StepsLayout/Examples.tsx
@@ -1,6 +1,14 @@
 import ComponentBox from '../../../../../../../shared/tags/ComponentBox'
 import { Card, P } from '@dnb/eufemia/src'
-import { StepsLayout, Form } from '@dnb/eufemia/src/extensions/forms'
+import {
+  StepsLayout,
+  Form,
+  Field,
+  Value,
+} from '@dnb/eufemia/src/extensions/forms'
+import { debounceAsync } from '@dnb/eufemia/src/shared/helpers/debounce'
+import React from 'react'
+import { createRequest } from '../../Form/SubmitIndicator/Examples'
 
 export const Default = () => {
   return (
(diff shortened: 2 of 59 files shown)


joakbjerk:
Just a comment to the feedback you wanted 😄 

If we do have the ability to check if the function is async or not(and if it does not complicate the code too much), then I think we should go for that approach, instead extra async suffix. I agree with the sentiment that it would be messy with so many extra props, and boilerplate to make it work.

Also doing a check instead of adding extra props lends itself to a better developer experience too 🙏

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/extended-features/Form/SubmitIndicator/info.mdx
```diff
@@ -16,4 +16,4 @@ render(<Form.SubmitIndicator state="pending" />)
 
 ### Integration
 
-The indicator is implemented by default in the [SubmitButton](/uilib/extensions/forms/extended-features/Form/SubmitButton/) and [FieldBlock](/uilib/extensions/forms/create-component/FieldBlock/) label. It will be displayed when the `onSubmit` event is triggered or when `enableAsyncBehavior` is set to true on the [Form.Handler](/uilib/extensions/forms/extended-features/Form/Handler/) component. You find some examples down below.
+The indicator is implemented by default in the [SubmitButton](/uilib/extensions/forms/extended-features/Form/SubmitButton/) and [FieldBlock](/uilib/extensions/forms/create-component/FieldBlock/) label. It will be displayed when the `onSubmitAsync` event is used on the [Form.Handler](/uilib/extensions/forms/extended-features/Form/Handler/) component. You find some examples down below.
```
joakbjerk, line 19:
```suggestion
The indicator is implemented by default in the [SubmitButton](/uilib/extensions/forms/extended-features/Form/SubmitButton/) and [FieldBlock](/uilib/extensions/forms/create-component/FieldBlock/) label. It will be displayed when the `onSubmitAsync` event is used on the [Form.Handler](/uilib/extensions/forms/extended-features/Form/Handler/) component. You can find some examples down below.
```

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/extended-features/StepsLayout/StepsLayout/Examples.tsx
```diff
@@ -1,82 +1,174 @@
...
+              <StepsLayout
+                onStepChangeAsync={onStepChangeAsync}
+                variant="drawer"
+              >
+                <StepsLayout.Step title="Step 1">
+                  <Card stack>
+                    <Field.String
+                      label="Required field with async validator"
+                      validator={validator1}
+                      path="/field1"
```
joakbjerk, line 141:
```suggestion
                      path="/field1"
                      required
```

packages/dnb-eufemia/src/extensions/forms/Form/SubmitIndicator/SubmitIndicatorDocs.ts
```diff
@@ -2,12 +2,12 @@ import { PropertiesTableProps } from '../../../../shared/types'
 
 export const SubmitIndicatorProperties: PropertiesTableProps = {
   state: {
-    doc: 'Provide `pending` to make it visible.',
+    doc: 'Provide `pending` to make the dots visible and `success` when to show the checkmark icon.',
```
joakbjerk, line 5:
```suggestion
    doc: 'Provide `pending` to make the dots visible and `success` to show the checkmark icon.',
```

joakbjerk (approved):
Looks good to me from testing in the portal 🔥

PR #3381, opened 2024-03-15
feat(useFieldProps): rename useDataValue to useFieldProps

tujoworker:
For a more semantic name I propose to rename this hook. It is made with backwards compatibility.


joakbjerk (approved):
Like the new name, makes more sense 👍

PR #3383, opened 2024-03-15
feat(useFieldProps): add support for data-attributes and replace `ariaAttributes` with `htmlAttributes`

joakbjerk:


tujoworker:
We need at least one test in each of the three files: String, Number and useFieldProps 🚀

tujoworker (approved):
Good addition ⭐

PR #3384, opened 2024-03-15
feat(Tabs): add `breakout` prop for Tabs

joakbjerk:
This does not feel like the best way to solve this, so this is just a quick proposal.

Some developers have [requested](https://dnb-it.slack.com/archives/CMXABCHEY/p1710405329367549) the possiblity to make the `Tabs` bottom border to stay within the bounds of its parent element.

Screenshot 2024-03-14 at 09 03 41

Is the current way the bottom-border behaves the intended behavior? Or could this thing easily be solved by just changing the 100vw and -100vw inside the ::content to 100% and 0?

dependent on https://github.com/dnbexperience/eufemia/pull/3065

- [x] rename `noBleedBorder` to `breakout`
- [x] Implement same logic in #3065 
- [x] Use TabsDocs.ts from #3326


tujoworker:
Yes, its very similar to what we did with the Section and the usage of the `breakout` property. I suggest to use the same name. Its enabled by default, but can be disabled by using `breakout={false}`.

tujoworker:
Maybe we should get this PR #3065 in first 🤦‍♂️

joakbjerk:
> Maybe we should get this PR #3065 in first 🤦‍♂️

Yeah that's probably wise 😄

joakbjerk:
> Maybe we should get this PR #3065 in first 🤦‍♂️

So I have added to breakout prop to that branch, with t[his commit](https://github.com/dnbexperience/eufemia/pull/3065/commits/676a76933e4b7c757ca95ecdf8a50b4d517faf1d). Please correct me if I misunderstood something @tujoworker 😄

tujoworker (approved):

PR #3397, opened 2024-03-20
feat(Table): add support for scope="row" with odd/even background color

tujoworker:
How it should look ([example preview](https://eufemia-git-feat-table-row-scope-eufemia.vercel.app/uilib/components/table/#responsive-table-in-a-card)):

Screenshot 2024-03-20 at 19 05 04

table-have-to-match-table-with-one-td snap


joakbjerk (approved):

packages/dnb-design-system-portal/src/docs/uilib/components/table/Examples.tsx
```diff
@@ -1304,3 +1320,117 @@ export const VariantCombinations = () => {
...
+            {isLarge ? (
+              <Tr>
+                <Td>{content.title}</Td>
+                <Td>{content.description}</Td>
+                <Td>{content.status}</Td>
+                <Td>{content.deadline}</Td>
+              </Tr>
+            ) : (
+              <>
+                <Tr>
```
yervand-stepanyan, line 1365: @tujoworker I think we are missing one prop here.

On large screens, every Table item has the same color and it is clear that this is a new item.
On Small screens, the color is not the same according to the [Example](https://eufemia-git-feat-table-row-scope-eufemia.vercel.app/uilib/components/table/#responsive-table-in-a-card:~:text=Responsive%20table%20in%20a%20Card). 
See the Large screen option, Mobile screen and the desired solution.

SInce I am the person who works on this sketch, I could fix this by adding `variant="odd"` to the first `<Tr>` `(<Tr variant="odd">)`.

Can you please, look into this before merging the code?


Screenshot 2024-03-21 at 13 12 04
Screenshot 2024-03-21 at 13 12 53
Screenshot 2024-03-21 at 13 16 48

tujoworker: Thank you 🙏

packages/dnb-design-system-portal/src/docs/uilib/components/table/Examples.tsx
```diff
@@ -1304,3 +1320,117 @@ export const VariantCombinations = () => {
...
+                  <Td>{content.status}</Td>
+                </Tr>
+                <Tr>
+                  <Th scope="row">{header.deadline}</Th>
+                  <Td>{content.deadline}</Td>
+                </Tr>
+              </>
+            )}
+
+            <Tr>
```
yervand-stepanyan, line 1384: @tujoworker Probably, we need the`variant="odd"` here also. (`<Tr variant="odd">`).

I was testing that when the number of rows before this button block row is **odd** then this row becomes **gray**. 

WIth the `variant="odd"` prop added, we ensure that the button actions row is always **white** regardless of the number of the rows above.

Maybe this can be clarified with a designer(Tonje?), because I am pretty sure she wants to have this buttons row **white** to have this row as a separator among the Table items. 

Screenshot 2024-03-21 at 19 56 38
Screenshot 2024-03-21 at 20 09 21

PR #3406, opened 2024-03-22
feat(Accordion): add additional Sbanken styling variables

Sundfjord:
Some properties of the Accordion component are often subject to overriding styles in the consumer apps. To facilitate easier styling of these properties, some new CSS variables (that can be overridden in the consuming app) has been added.
These are:
- Vertical icon alignment
- Content padding
- Content margin

Default styles are not changed, although two image snapshots had to be updated because the removal of a `margin-top` property ensured that Accordion.Content spacing properties worked as intended.


tujoworker (approved):
Nice addition, looks good 🙌

PR #3407, opened 2024-03-24
fix(Autocomplete): use current value to filter changed data

### LINKED ISSUE #3355: Autocomplete: Replacing the data inside `on_type` temporarily disables the filter/highlight
## 🐛 Bug Report
We have a use case where we pass the data in from outside the `Autocomplete` component, and then inside `on_type` (after a debounce), we update that data with new items.

When this happens, the filter and highlighting for the current query disappears for the first render, showing the full unfiltered list, with no highlights.

## To Reproduce

Here is [a reduced codesandbox that demonstrates this behavior](https://codesandbox.io/p/sandbox/confident-vaughan-ty4tz4?file=%2Fsrc%2FApp.tsx%3A23%2C20).

The sandbox shows the issue by replacing the list with a new list once you stop typing in the input field. This only happens once, so you need to refresh to show the bug again.

In the sandbox:

1. Click in the input field, observe that the list contains "first list"
2. Type "second"
3. Observe that the list has changed

Expected:
- The new list is filtered, showing only the matching item ("second list")
- The word "second" is highlighted in the list

Actual:
- The new list is not filtered, showing all items
- The search term is not highlighted

## Expected behavior

When the data is updated while the user has typed a query, the filter and highlighting should still apply right away.

## Additional info

- This behavior does not manifest when StrictMode is enabled
- If we remove the call to `hideIndicator()`, the bug does not happen
- The bug is not there if we keep the data internal and update it with `updateData()`

## Eufemia Version

10.21.0

tujoworker:
Fixes #3355

There is a big potential downside to our ability to avoid potential performance issues when going forward with the implementation as is.

Because the data prop changes, we need to check if it has changed and update the search index. But not all devs will memorise the given data – and if that is not the case, each re-render will set a new search index, regardless if it has changed or not.

At the moment (late Sunday evening) I can't come up with a smart solution of mitigate that potential downside.

Any suggestions?

And thank you again for the fantastic reproduction and issue description: https://codesandbox.io/p/sandbox/confident-vaughan-ty4tz4?file=%2Fsrc%2FApp.tsx

### DIFF
--- a/packages/dnb-eufemia/src/components/autocomplete/__tests__/Autocomplete.test.tsx
+++ b/packages/dnb-eufemia/src/components/autocomplete/__tests__/Autocomplete.test.tsx
@@ -2252,6 +2252,80 @@ describe('Autocomplete component', () => {
     ).toBe('c value')
   })
 
+  it('will filter items with current value after data prop has changed', async () => {
+    const mockDataA = ['first', 'foo']
+    const mockDataB = ['second', 'bar', 'baz']
+
+    const WithState = () => {
+      const [data, setData] = React.useState(mockDataA)
+
+      const onTypeHandler = ({ debounce, ...args }) => {
+        debounce(() => {
+          args.showIndicator()
+          setTimeout(() => {
+            args.hideIndicator()
+            setData(mockDataB)
+          }, 1)
+        }, 1)
+      }
+
+      return (
+        <Autocomplete
+          {...mockProps}
+          mode="async"
+          data={data}
+          on_type={onTypeHandler}
+          show_submit_button
+        />
+      )
+    }
+
+    render(<WithState />)
+
+    const options = () =>
+      document.querySelectorAll(
+        'li.dnb-drawer-list__option:not(.dnb-autocomplete__show-all)'
+      )
+
+    toggle()
+
+    expect(options()).toHaveLength(2)
+
+    fireEvent.change(document.querySelector('.dnb-input__input'), {
+      target: { value: 'first' },
+    })
+
+    expect(options()).toHaveLength(1)
+    expect(options()[0]).toHaveTextContent('first')
+    expect(options()[0].innerHTML).toBe(
+      '<span class="dnb-drawer-list__option__inner"><span class="dnb-drawer-list__option__item"><span><span class="dnb-drawer-list__option__item--highlight">first</span></span></span></span>'
+    )
+
+    fireEvent.change(document.querySelector('.dnb-input__input'), {
+      target: { value: 'second' },
+    })
+
+    await waitFor(() => {
+      expect(options()).toHaveLength(1)
+      expect(options()[0]).toHaveTextContent('second')
+      expect(options()[0].innerHTML).toBe(
+        '<span class="dnb-drawer-list__option__inner"><span class="dnb-drawer-list__option__item"><span><span class="dnb-drawer-list__option__item--highlight">second</span></span></span></span>'
+      )
+    })
+
+    fireEvent.change(document.querySelector('.dnb-input__input'), {
+      target: { value: '' },
+    })
+
+    expect(options()).toHaveLength(3)
+    expect(options()[0]).toHaveTextContent('second')
+    expect(options()[0].innerHTML).toBe(
+      '<span class="dnb-drawer-list__option__inner"><span class="dnb-drawer-list__option__item">second</span></span>'
+    )
+    expect(options()[1]).toHaveTextContent('bar')
+    expect(options()[2]).toHaveTextContent('baz')
+  })
+
   it('has correct selected value', () => {
     render(<Autocomplete {...props} data={mockData} />)
     expect(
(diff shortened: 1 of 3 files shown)


tujoworker:
Maybe we should at lest mention in in the docs?

joakbjerk:
Hmm, I can't really think of a better solution right now either. 

Would checking if the content is the same, instead of comparing if the two arrays are not the same (which would think are always going to result in false)`prevProps.every((item, index) => item === props.data[index])` bring any benefits at all on that all in the componentDidUpdate function?

joakbjerk (approved):

tujoworker:
Yeah, a comparison by itself will have an impact on performance. That's why React don't shallow compare, but we devs need to be aware of our data flow and re-render issues. 

Another option would be to force devs to use the current approach by updating the data via the internal updateData handler (during typing). I'm not 100% what the arguments against it are. 

cc @kimroen 

Regarding the PR; I think the docs should be enough for now.

tujoworker:
I will remove the irrelevant docs updates to its own PR #3413.

packages/dnb-eufemia/src/components/autocomplete/__tests__/Autocomplete.test.tsx
```diff
@@ -2252,6 +2252,64 @@ describe('Autocomplete component', () => {
     ).toBe('c value')
   })
 
+  it('will filter items with current value after data prop has changed', async () => {
```
kimroen, line 2255: It looks like this test passes without the changes made in this PR, so I just wanted to check that it's testing what you intended.

If intentional, would it be valuable to also have one that breaks without the fix?

kimroen:
Thanks for this great PR @tujoworker, I appreciate the responsiveness 👍 

> Yeah, a comparison by itself will have an impact on performance. That's why React don't shallow compare, but we devs need to be aware of our data flow and re-render issues.

Indeed! Comboboxes are a pretty complicated topic and I'm not an expert here, but from what I've seen from (recent) headless implementations of comboboxes, they handle this specific problem by using children to define the list items, and then these children register themselves to a combobox-specific context with an id when they mount.

More about this topic below, but I realize rewriting Autocomplete isn't necessarily in scope here, so just allowing the user to update the data from outside like we do in this PR would help.

If our data prop didn't support so many permutations then we could perhaps compare the values of each item with the previous one as @joakbjerk  suggests (just comparing the items themselves doesn't work because they can be object references or react fragments in addition to plain text items), but perhaps allowing the user to pass a compare function could help there.

At that point we could also consider allowing a dependency array, and when we get there we might realize that we might as well have the user use `useMemo` and get all of that for free 😄 

> Another option would be to force devs to use the current approach by updating the data via the internal updateData handler (during typing). I'm not 100% what the arguments against it are.

At the moment we allow a weird combination of controlled and uncontrolled for this component, which I at least personally find confusing. The data in our use case is a combination of statically known items, items from an initial data fetch, and items that depend on a network call made during typing. We could combine these inside the on type callback and update the internal data using `updateData`, but what do we do about the other data sources? As long as the autocomplete doesn't own the source of truth, it is just simpler to keep this data handling external to the component.

---

I thought this description of the composition pattern from Ryan Florence was a good one ([Advanced Element Composition in React](https://ryanflorence.dev/p/advanced-element-composition-in-react)). We might see that Eufemia mostly lives in the "Option 1" area, but newer concepts like Eufemia Forms and the layout components are leaning closer to the later options, so I'm sure you're aware of these 😊 

Anyway, we can see this in a few implementations. All of these has some kind of store in a parent context that the child components interact with:

- [CMDK](https://cmdk.paco.me/) ([source](https://github.com/pacocoursey/cmdk/blob/main/cmdk/src/index.tsx))
- [Headless UI combobox](https://headlessui.com/react/combobox)
- [Reach UI combobox](https://reach.tech/combobox/)
- [React ARIA ComboBox](https://react-spectrum.adobe.com/react-aria/ComboBox.html)

What I like about these is that the user can just render the items, and the component handles the change. No memoization necessary.

If we look at other options, we can also find that for instance [downshift](https://github.com/downshift-js/downshift) handles these things with a fairly complicated redux-like internal state that the user is allowed to tap into to change the internal state whenever anything happens. Powerful, but maybe _too_ powerful? 😅  Not quite sure how they keep track of changes as the data changes externally though.

packages/dnb-eufemia/src/components/autocomplete/__tests__/Autocomplete.test.tsx
tujoworker: ohh wow. Thank you for checking this out. Normally I check the negative outcome. 

The example code missed the `showIndicator` which was actually the only thing the test needed to fail – without the fix.

kimroen: Awesome, looks great! 👍 🎉

tujoworker:
Very good feeback @kimroen – thank you.

I agree, the Autocomplete is long overdue. The plan is to rewrite it to Hooks, so its possible to use the search with the formatting inline with a normal search.

Ideally, all of Eufemia components should not keep any state internally (controlled). But we have a history with AEM where Eufemia had huge focus on Web Components support, so having uncontrolled components would have been way more effective to work with. Today, we have still a lot of legacy code.

Another viewpoint is, that Eufemia is not directly comparable to open source libraries, as DNB UX wants a certain direction of whats possible to achieve with these components.
But sometimes we see designers braking the borders, and suddenly, we need more flexibility anyway 😄

Thank you for all contributions 🙏

PR #3416, opened 2024-03-26
feat(forms): add max width to Card when used in forms

tujoworker:
As we can see in the visual snapshot change, this was a welcome addition to large screens.


joakbjerk (approved):
Looks much better 😄

PR #3420, opened 2024-03-26
fix(InputMasked): ensure decimal value update does format correctly 

kimroen:
Adds a testcase showing a problem that happens with decimal handling and values being updated from outside the `InputMasked` component. This can be seen [in this codesandbox](https://codesandbox.io/p/sandbox/shy-grass-tdynr7?file=%2Fsrc%2FApp.tsx%3A25%2C13) - entering "20.00" in the top textbox results in "2000." in the InputMasked field.

@tujoworker then swooped in and fixed the issue by skipping the custom handling of decimals for the local copy of the value if the passed-in value has a decimal separator.

---

Adding the async call to `userEvent.type` seems to break some other unrelated tests - not quite sure what's up with that.

### DIFF
--- a/packages/dnb-eufemia/src/components/input-masked/InputMaskedUtils.js
+++ b/packages/dnb-eufemia/src/components/input-masked/InputMaskedUtils.js
@@ -117,9 +117,9 @@ export const correctNumberValue = ({
   if (localValue !== null) {
     const localNumberValue = localValue.replace(/[^\d,.-]/g, '')
     const numberValue = value.replace(/[^\d,.-]/g, '')
-    const hasDecimal = numberValue.includes(decimalSymbol)
+    const valueHasDecimal = numberValue.includes(decimalSymbol)
 
-    if (!hasDecimal) {
+    if (!valueHasDecimal) {
       const endsWithDecimal = localNumberValue.endsWith(decimalSymbol)
       const endsWithZeroAndDecimal = localNumberValue.endsWith(
         `${decimalSymbol}0`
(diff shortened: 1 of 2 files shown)


tujoworker:
Nice! Thank you so much. I added a fix and test cases for the whole utils file.

kimroen:
> Nice! Thank you so much. I added a fix and test cases for the whole utils file. I hope this was OK for you? 🙏

By all means, great idea! Looks like the unrelated cursor test still fails when this new test is added though—really strange.

tujoworker:
> Looks like the unrelated cursor test still fails when this new test is added though—really strange.

Agree 👍  I moved the new tests over to a new PR as it is not directly related (except for the `correctNumberValue`).

packages/dnb-eufemia/src/components/input-masked/InputMaskedUtils.js
```diff
@@ -117,24 +117,27 @@ export const correctNumberValue = ({
...
-    ) {
-      /**
-       * When the users has 20,02, then hits "backspace",
-       * the returned {numberValue} in the onChange event would then be "20",
-       * but we want it to be 20,0
-       */
-      value = `${value}${decimalSymbol}0`
+    const hasDecimal = numberValue.includes(decimalSymbol)
+
+    if (!hasDecimal) {
```
kimroen, line 122: I understand what this does now after some more reading, but even with this knowledge the name `hasDecimal` isn't entirely clear on what it's referencing.

We're checking specifically if the `value` has a decimal symbol, and if it does then we shouldn't try to update the `value` based on the decimal symbols in the `localNumberValue`.

The other values in this area like `endsWithDecimal` are about the local value, which makes this potentially more confusing.

How about `valueHasDecimal` instead to make this clearer?

pbthorste:
Hi. I was trying out the unit tests and found one case that might be worth looking into.

Adding this to the test cases:
```js
await userEvent.type(nativeInput, '1,')
expect(nativeInput).toHaveValue('1,00')
expect(inputMasked).toHaveValue('1,00')
```
Seems to give a test failure:
```js
```

tujoworker (approved):

PR #3422, opened 2024-03-26
fix(Card): ensure "spacing" children do stretch by default

tujoworker:
**With the change:**

Screenshot 2024-03-26 at 23 05 34

   
   
**Before:**

Screenshot 2024-03-26 at 23 05 18


tujoworker:
I added a test visual test that fails without the fix.

packages/dnb-design-system-portal/src/docs/uilib/components/card/Examples.tsx
```diff
@@ -33,36 +33,49 @@ export const Default = () => {
...
-            </tbody>
-          </Table>
-        </Table.ScrollView>
-      </Card>
+      {() => {
+        const MyTable = () => (
+          <Table.ScrollView>
+            <Table border outline size="medium">
+              <thead>
+                <Tr noWrap>
```
yervand-stepanyan, line 41: Just a discovery regarding the `noWrap` for the `<Tr ...>`.

If the header or `Table` row texts are long, the Table is overflowing from the `Card` and even no horizontal scroll is possible when the content is out of the screen.

As much as I tested the problem is caused by the `Card` Component and maybe we should have a look to the `Card` in the future.

PS. adding `overflow-x: auto;` to the `Card` fixes the horizontal scroll, but need to be tested more for other use-cases.

PPS. without `noWrap` on `<Tr>` texts are wrapping and not causing any overflow from the `Card`.

If the meaning of a `Card` is to have a fixed `width` regardless of the nested content, then maybe we should consider an edge-case when the `Card` is used as a `Wrapper` for a `Table`(can be that some other components also may have problems when wrapped by a `Card`).


Screenshot 2024-03-27 at 09 01 21
Screenshot 2024-03-27 at 09 05 00
Screenshot 2024-03-27 at 09 05 42
Screenshot 2024-03-27 at 09 26 39

yervand-stepanyan (approved):

yervand-stepanyan (approved):

packages/dnb-design-system-portal/src/docs/uilib/components/card/Examples.tsx
tujoworker: It turned out that the "fix" in this PR did also fix your mentioned findings. So thank you twice for the feedback and the good collaboration ❤️

PR #3426, opened 2024-04-02
feat(forms): refactor API from StepsLayout to Steps.Layout

tujoworker:
With 100% backwards compatibility.

One question:

- Should we rename `Steps.Step` to `Steps.Case`?


AadneRo:
I added some docs changes as a seperate commit. Just so the texts reflects the api changes made 


Screenshot 2024-04-02 at 14 54 37

AadneRo:
`Steps.Step` is a bit awkward, but it describes the use case better than Case in my opinion. If anyone else has a strong opinion I'm fine with either

AadneRo:
There are still a few references to StepsLayout, but since it's only the api we're changing I think it's fine. I've updated the doc text references I found https://github.com/dnbexperience/eufemia/pull/3426#issuecomment-2031973720 
So please double check I didn't mess up before merging.

Apart from that it looks good!

snorrekim:
I think `Steps.Step` is better than `Steps.Case`. "Case" might even be a bad name, it doesn't describe that it's a sequence. "Case" in coding generally mean that you only use one of the "case"s.

I think the word `Wizard´ could be used. Thats basically what this is. Do we have other Wizard types? This could be structured like this:
```
<Wizard type="step">
  <Wizard.Step>
    ...
    <Wizard.NextButton/>
  </Wizard.Step>

  <Wizard.Step>
    ...
    <Wizard.PreviousButton/>
    <Wizard.NextButton/>
  </Wizard.Step>
</Wizard>
```

snorrekim:
### Side-note: NextButton, PreviousButton and Buttons should be one automatic component

First, I think `StepsLayout.Buttons` might be unneeded or confusing, unlike `NextButton` and `PreviousButton` it implements it's own `<Form.ButtonRow>`. But you still have to manually use `<Form.ButtonRow>` in the first and last steps. So why change that behaviour? It should just add both buttons and be named `BothButtons`. 

But I think we should go one step further and just merge all three component into one that automatically knows what buttons to add. With a `variant="next" | "previous" | "both"` if you need to override?

Like this:

```
  <Wizard variant="drawer">

    <Wizard.Step>
      ...
      <Form.ButtonRow>
        <Wizard.Buttons/> // automatically only shows "next" button
      </Form.ButtonRow>
    </Wizard.Step>

    <Wizard.Step>
      ...
      <Form.ButtonRow>
        <Wizard.Buttons/> // automatically shows "next" and "previous" button
      </Form.ButtonRow>
    </Wizard.Step>

    <Wizard.Step>
      ...
      <Form.ButtonRow>
        <Wizard.Buttons/>
        <Form.SubmitButton/> // automatically only shows "previous" button
      </Form.ButtonRow>
    </Wizard.Step>

  </Wizard>
```

If the user really needs more control of each button, they can still do that:
```
</Form.ButtonRow>
  <Wizard.Buttons variant="previous" disabled/>
  <MyCustomWrapperComponent>
    <Wizard.Buttons variant="next"/> 
  </MyCustomWrapperComponent>
</Form.ButtonRow>
```

PR #3428, opened 2024-04-03
feat(forms): add `transformIn` and `transformOut` to fields for value transformation

tujoworker:
This new transformer API is a declarative addition that can be used on every field, regardless if the field internals already use one of the other transformers, such as `toInput` and `fromInput`.

These two transformers are meant to be used form the field consumer, and not internally in fields.

```tsx
const transformIn = (value) => {
  return value?.toUpperCase()
}
const transformOut = (value) => {
  return value?.toLowerCase()
}
render(
  <Form.Handler onChange={console.log}>
    <Field.String
       path="/myField"
       transformIn={transformIn}
       transformOut={transformOut}
    />
  </Form.Handler>
)
```


AadneRo (approved):
LGTM 🎉

PR #3430, opened 2024-04-03
fix(Table): ensure tables with one td align the th width

tujoworker:
Screenshot 2024-04-03 at 15 05 29


yervand-stepanyan (approved):
Looks good!

PR #3432, opened 2024-04-04
chore: remove unused useEventEmitter

tujoworker:
Removing EventEmitter is still left, as it is used in Tabs. But when #3065 lands, it should use `useSharedState` and we can then remove the EventEmitter entirely.


packages/dnb-eufemia/src/fragments/drawer-list/DrawerList.d.ts
```diff
@@ -107,8 +191,18 @@ export type DrawerListOptionsProps = {
 export type DrawerListItemProps = {
   children: React.ReactNode;
   selected: boolean;
+  /**
+   * Define a preselected data entry (index) or key inside an array item. Can be a string or integer.
```
AadneRo, line 195:
```suggestion
   * Define a preselected data entry (index) or key inside an array item. Can be a string or number.
```
This shouldn't be that confusing for a developer, but nice to be consistent

packages/dnb-eufemia/src/fragments/drawer-list/DrawerList.d.ts
```diff
@@ -107,8 +191,18 @@ export type DrawerListOptionsProps = {
...
+   */
   value: string;
-  on_click: ({ value }: { value: string }) => void;
+  on_click: ({
+    value
+  }: {
+    /**
+     * Define a preselected data entry (index) or key inside an array item. Can be a string or integer.
```
AadneRo, line 202: https://github.com/dnbexperience/eufemia/pull/3432/files#r1551662084

AadneRo (approved):
Some language nitpicks

PR #3433, opened 2024-04-04
feat(forms): refactor API from StepsLayout to Wizard.Container

tujoworker:
More [info here](https://github.com/dnbexperience/eufemia/pull/3426#issuecomment-2034150647).

Related PR #3426 which we need to remove from the release notes when released.


snorrekim (approved):
Approved.

Though I think `<Wizard.Layout>` should be just ´<Wizard>` to make clear that it's the container component for all the `<Wizard.XXX>` components.

PR #3438, opened 2024-04-04
fix(NumberFormat): ensure support for screen reader speak text under mouse

tujoworker:
This PR fixes an issue where screen readers could not read the content when reading with "read text under mouse" mode.

VoiceOver reads `aria-label` fine. But NVDA does only read "visible" text with its "read text under mouse" mode. Therefore we additionally disable the `aria-hidden` while hovering. 

From my tests so far, this approach works very well and I can't come up with downsides.

**VoiceOver on mac**

Screenshot 2024-04-04 at 21 38 50

**NVDA on Windows**

Screenshot 2024-04-04 at 21 39 28


AadneRo:
I tried this preview with a SR and compared it to the current, but I'm not able to spot any change in behaviour. Could you verify this fix @snorrekim

tujoworker:
Jepp, no diff. for VoiceOver. Reading the content with "Speak text under mouse" did work on VoiceOver before as well.

Only NVDA had this issue from my testing, regardless what browser I test it with (even we should only use Safari and Firefox, because they have official VoiceOver support).

- Before: https://eufemia.dnb.no/uilib/components/number-format/
- After: https://eufemia-git-fix-number-format-hover-sr-support-eufemia.vercel.app/uilib/components/number-format/

snorrekim (approved):
Looks good to me

PR #3441, opened 2024-04-08
docs: document all colors

snorrekim:
* Color tables are now generated based on `properties.js` file
* Added tertiary colors for Sbanken
* Verified and updated all DNB and Sbanken colors with Figma (Eiendom is still not verified)
* Added screenshots of color tables


tujoworker (approved):
Very good work – thank you ⭐

PR #3448, opened 2024-04-09
feat(FormLabel): enlarge bounding area of clickable labels

tujoworker:


snorrekim (commented):
I assume this is a way to ensure at least 44px click area? 

I think the area should be directly linked to the input field size. A default input field is only `32px`, so the label shouldn't be larger.

tujoworker:
The solution is too "heavy" in what we get in return. Also, the need of an larger bounding area for a label is not that important, compared to other interactive elements, as it is a part of other components anyway, most of the time.

PR #3450, opened 2024-04-09
fix(Card): ensure 100% width does only apply when ScrollView is used

tujoworker:
Screenshot 2024-04-09 at 14 38 47


joakbjerk (approved):
📜

PR #3452, opened 2024-04-10
chore(Input): ensure border hover color has emerald-green

tujoworker:
This PR fixes an issue cause in this #3440 PR. Its not released, so we use `chore` as the git tag.

Screenshot 2024-04-10 at 08 08 46


joakbjerk (approved):
❇️

PR #3459, opened 2024-04-12
feat(Typography): Replace inline font with font files (Sbanken)

Sundfjord:
The relevant Sbanken font files have been added and taken into use.
In addition, a simple license document has been added, but feedback from someone in the organisation with experience with this kind of licensing endeavour is desired.

Using woff format for all font-face definitions for consistency.


tujoworker:
Really nice work ⭐ I suggest we use `woff2`, like we use for the DNB/UI. These files are way smaller and should yield the same quality. We also use `woff` as a fallback. 

> The WOFF 2.0 Web Font compression format offers a 30% average gain over WOFF 1.0.

I added these files with additional commit.

Sundfjord:
Seems sensible! But the font seems to be marginally different from the woff ones, enough to break some screenshot tests (at least locally for me), but that might not be a problem?

tujoworker (approved):
Looks good to me now ✨

Thank you for the good work.

snorrekim (commented):
I don't know how fonts-files are imported. Is it part of core?

tujoworker:
> If it doesn't happen automatically, like in css, we should maybe update documentation.

It does happen likewise automatically. Bundlers to handle ti as a part of asset management.

PR #3460, opened 2024-04-12
chore(docs) Add comment in all legacy component 

AadneRo:
add a text in all legacy class components letting all new developers know this is a legacy component


tujoworker:
Nice 👌 


Checkbox, ProgrssIndicator, Tabs and all DatePicker components we probably should omit, as they are in works and it will create a conflict due to the renaming of the file extension.

AadneRo:
> Nice 👌
> 
> Checkbox, ProgrssIndicator, Tabs and all DatePicker components we probably should omit, as they are in works and it will create a conflict due to the renaming of the file extension.

Done ✅

tujoworker (approved):
Nice improvement ⭐

PR #3463, opened 2024-04-16
fix(forms): ensure proper `Iterate` data context and `filterData` support

tujoworker:


joakbjerk (approved):
Nice work, and the new examples looks much better 😄

The `ArrayPushButton` demos does not seem to be working properly though.

https://github.com/dnbexperience/eufemia/assets/25927156/c787536c-01f1-406b-a781-81478aca9da7

The current portal version has the same issue, so it's not introduced with this pr though 🙏

PR #3464, opened 2024-04-16
chore: enhance docs and examples

tujoworker:


packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/quick-start.mdx
```diff
@@ -10,6 +10,8 @@ import { Field } from '@dnb/eufemia/extensions/forms'
 render(<Field.Email />)
 ```
 
+> **NB:** Only the email field will be a part of your application bundle. Unused code will be tree-shaken away.
```
joakbjerk, line 13:
```suggestion
> **NB:** In the above example, only the email field will be a part of your application bundle. Unused code will be tree-shaken away.
```

joakbjerk: Don't have to use the exact wording in the suggestion, but I think we just need to clarify a little, so people don't think that only the email field will be part of the bundle, regardless of what they do 😄

joakbjerk (approved):
Looks good 😄

PR #3470, opened 2024-04-18
fix(Anchor): icon position when line-height increased by another element

snorrekim:
* No longer using css unit `lh`, it does not match line-height if a larger element has increased it
* Shrink icon height to 50% before using transform to scale it back up. This way, browser line height is calculated with the tiny 50% icon.
* `Breadcrumbs` fix for "unavoidable" `0.5px` line-height when inline element is flex item


tujoworker (approved):
Looks good. Great work ⭐

PR #3481, opened 2024-04-22
fix(forms): change Wizard default scroll behaviour to be active

tujoworker:


joakbjerk (commented):
Hmm, the code looks correct, but the `omitScrollManagement`does not seem to work when testing it in the preview portal, I still get scrolled to the top when the step changes.

And also clicking the related code-editor-block also takes me to the top when `omitScrollManagement` is falsy 😄

https://github.com/dnbexperience/eufemia/assets/25927156/5504c2fc-f60d-41c2-8803-4ccf882c7ba4

joakbjerk:
So after doing some testing, it seems like the the `omitScrollManagement` is ignored when navigating to the `Summary` `Step` 🤔

joakbjerk (approved):

joakbjerk:
> Thanks for testing it out. We missed `preventScroll: true` in the focus. The focus handling made it to scroll.
> 
> Can you test it again? 🙏

Works much better now!

PR #3487, opened 2024-04-23
feat(ProgressIndicator): add `countdown` type, center label, and customization options

snorrekim:
* Added new type `countdown` that moves in the reverse direction of `circular`
* Added new label position `inside` for `circular` and `countdown`
* Can send in custom size in the `size` prop
* Can customize colors with `customColors` prop
* Can customize line thickness with `customLineWidth` prop


tujoworker (approved):
Looks good 👌 very nice work 👏

PR #3493, opened 2024-04-25
fix(Flex.Container): fix issue where `space` object was ignored in nested `Flex.Containers`

joakbjerk:
Proposal to fix issue where the space prop was overwritten on `Flex.Container ``children` nested inside another `Flex.Container`.

Not sure if this is the best method to fix this issue 🤔

With this composition: 
Screenshot 2024-04-25 at 10 45 34

Before:
Screenshot 2024-04-25 at 10 45 11

After:

Screenshot 2024-04-25 at 10 44 46


tujoworker:
We should check if this #3564 does the same and if so, add these tests in there instead.

PR #3494, opened 2024-04-25
feat(forms): add `Value.Composition` component

tujoworker:


joakbjerk (approved):
Looks good 😁

Just asking because I don't know the correct answer, but should the `<dt />` for the combinded `First Name` `Last Name` be empty?
Screenshot 2024-04-25 at 13 16 03

PR #3500, opened 2024-04-26
chore: refactor to internal field boundary

tujoworker:
This way we can re-use the field boundary logic in other parts.


joakbjerk (approved):
Smart 🧐

PR #3505, opened 2024-04-29
fix(Autocomplete): fix issue where `React.Element` would not render when used as `selected_value` 

joakbjerk:
- [x] Find a better solution for Iterate/Array/demo portal error


packages/dnb-eufemia/src/components/autocomplete/__tests__/Autocomplete.test.tsx
```diff
@@ -1461,6 +1461,53 @@ describe('Autocomplete component', () => {
     ).toBe('')
   })
 
+  it('should render "select_value" when set to React.Element', async () => {
```
tujoworker, line 1464:
```suggestion
  it('should render "selected_value" when set to React.Element', async () => {
```

tujoworker (approved):
Nice fix 👍

PR #3517, opened 2024-05-03
feat(forms): add Wizard transition animation

tujoworker:


joakbjerk (approved):
Looks very smooth! 🔥 less sudden and easier to notice step change

PR #3522, opened 2024-05-06
fix(Eufemia): enhance CSS version check for dnb-core-style

snorrekim:
* Also check the `.dnb-core-style` element if version is not found on `body`
* warning instead of error if version is `unknown`
* Explain that `unknown` might be caused by lazy loading or a version older than "10.25.0"
* Add a .js import for theme basis and update documentation for css importing
* Fix .js import of eiendom theme


tujoworker (approved):
Nice improvement 🙏

PR #3526, opened 2024-05-06
fix(Context) revert to less strict local types

AadneRo:
Making this more strict caused a breaking change for some applications


tujoworker (approved):
🚀

PR #3528, opened 2024-05-07
feat(forms): add `Field.Name` and `Value.Name`

tujoworker:
Includes PR #3529 and will be rebased


joakbjerk (approved):
Looks good 👍

Should the `Value.Name` component also do the same case transformation as the input field?

Screenshot 2024-05-07 at 08 04 27
Screenshot 2024-05-07 at 08 03 19

And should the name field show up in the leftside menu?

Screenshot 2024-05-07 at 08 03 31

PR #3534, opened 2024-05-07
fix(Drawer): prevent `auto-focus` from triggering in inital mount when `openState` is `false`

joakbjerk:


idamje:
I have an example where openState is always false, but the auto-focus still occurs. If I remove openState completely, the auto-focus does not occur.

PR #3556, opened 2024-05-14
feat(forms): add `required` prop to Wizard.Step

tujoworker:


packages/dnb-eufemia/src/extensions/forms/Wizard/Step/Step.tsx
```diff
@@ -49,6 +56,9 @@ function Step(props: Props) {
     return null
   }
 
+  const fieldProps =
```
joakbjerk, line 59: Could this be added directly to the required attribute for the `FieldProps` wrapper below, or are we planning to add more props to the `fieldProps` variable later?

joakbjerk (approved):
Looks good 👍

packages/dnb-eufemia/src/extensions/forms/Wizard/Step/Step.tsx
tujoworker: yeah, the change is there that we add more.

PR #3563, opened 2024-05-15
chore(Flex): add innerSpace as a valid spacing prop when using `pickSpacingProps`

tujoworker:
Use existing list of valid spacing props to pick among other props.


joakbjerk (approved):
⭐️

PR #3572, opened 2024-05-16
feat(Visibility): deprecate `pathValue` in favour of `visibleWhen`

tujoworker:


packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Visibility/demos.mdx
```diff
@@ -12,6 +12,19 @@ import * as Examples from './Examples'
 
 ### Matching value
 
+`visibleWhen` is pretty powerful. You can use it to show/hide based on the value of a `path`. You can also give it a `withValue` function that gives you the current value, so you can assert it and return a boolean based on that.
```
joakbjerk, line 15:
```suggestion
`visibleWhen` is pretty powerful. You can use it to display based on the value of a `path`. You can also give it a `withValue` function that gives you the current value, so you can assert it and return a boolean based on that.
```

joakbjerk (approved):
Looks good! Much better and more concise syntax 🔥

PR #3576, opened 2024-05-20
feat(forms): add path based `filterData` support

tujoworker:
Based on #3574

PR #3577, opened 2024-05-20
feat(forms): change `filterData` method to return object based parameters `{ path, value, data, props, internal }`

tujoworker:
Based on #3576


joakbjerk (approved):

tujoworker:
I will have a dialog with the devs using this feature, as this is technically a breaking change.

PR #3584, opened 2024-05-22
docs: add more details about media query breakpoint ranges

tujoworker:
To make it more clear when media query breakpoints do break, we can add info in pixels, as it is more accurate.

Here is [the preview](https://eufemia-git-docs-breakpoint-clarification-eufemia.vercel.app/uilib/layout/media-queries/#breakpoint-ranges).


joakbjerk (approved):
🔥

PR #3585, opened 2024-05-22
fix(forms): set focus back on existing element after async submission

tujoworker:
It can be tested in [this example](https://eufemia-git-fix-set-focus-after-submit-eufemia.vercel.app/uilib/extensions/forms/Form/demos/#in-combination-with-a-submitbutton) vs the [existing one](https://eufemia.dnb.no/uilib/extensions/forms/Form/demos/#in-combination-with-a-submitbutton).


joakbjerk (approved):
Works on my machine 🙏

PR #3588, opened 2024-05-22
fix(Breakpoint): ensure correct breakpoint ranges when using useMedia, useMediaQuery or MediaQuery

tujoworker:
In this PR we ensure that break-point ranges do not overlap.


joakbjerk (approved):
Looks good 👍

PR #3590, opened 2024-05-22
feat(Table): add `Th` (header) width attribute for tables with `fixed` layout 

tujoworker:


snorrekim:
Turns out the `width` attribute is deprecated and it only supports percentage and `px` in chrome at the moment, not `rem` or `em`. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/th#width

We might have to use `style={{width: "25%"}}`

Also, we can avoid the small width change on expand if we leave 1 column to be "free". Since there is a hidden column for the expand icon, the total widths adds up to more than 100%, and we can't predict the percentage. But if one column doesn't have a width, it will fill the rest. Basically, any total width above 100% seems to be handled like a non-fixed table, so it changes when expanding.

Though we should probably change the example to be more useful. Most real tables will have at least one column that can be flexible.

snorrekim (changes requested):
Use css width `style` instead, as that supports more units and isn't deprecated.

PR #3593, opened 2024-05-23
chore: add tests for when field path is not defined

tujoworker:


packages/dnb-eufemia/src/extensions/forms/hooks/__tests__/useFieldProps.test.tsx
```diff
@@ -83,6 +83,42 @@ describe('useFieldProps', () => {
     })
   })
 
+  it('should throw then "path" without slash was given', () => {
```
joakbjerk, line 86:
```suggestion
  it('should throw error when "path" without slash is given', () => {
```

packages/dnb-eufemia/src/extensions/forms/hooks/useFieldProps.ts
```diff
@@ -153,28 +153,18 @@ export default function useFieldProps<
...
     path: iteratePath,
     handleChange: handleChangeIterateContext,
   } = iterateElementContext ?? {}
   const { setFieldError } = fieldBoundaryContext ?? {}
 
-  if (path && path.substring(0, 1) !== '/') {
-    throw new Error(
-      'Invalid path. Data value path JSON Pointers must be from root (starting with a /).'
-    )
+  if (path && path[0] !== '/') {
```
joakbjerk, line 163:
```suggestion
  if (path && path.startsWith('/')) {
```

Not a necessary change, but maybe we could replace this with `startsWith `to make it more readable?

not that `path[0] !== '/'` is unreadable when we now that path is a `string `though!

joakbjerk (approved):
Looks good!

PR #3607, opened 2024-05-27
fix(forms): add support for `trueText` and `falseText` to Boolean checkbox variant

tujoworker:


joakbjerk (approved):
works as intended! 🔥

PR #3610, opened 2024-05-27
fix(ArraySelection): make explicit type of value can be undefined

pdhenriques:
If the project is using "exactOptionalPropertyTypes": true in tsconfig, then this component will not work unless you explicitly say it can be undefined in the type.

I'm not sure how to test this change, since changing the tsconfig flag will break everything else. So I didn't include a test.


tujoworker (approved):
Nice 🚀 Thank you. Does that mean there may be a number of components where this affects the same issue?

pdhenriques:
Probably yes. When I tried to set the flag in my local eufemia fork, and ran the test, I got a mountain of errors.

PR #3612, opened 2024-05-27
feat(Checkbox, Radio): fix border width for Sbanken brand

snorrekim:
* made some border widths 1.5px for dnb and 1px for sbanken


tujoworker (approved):
Nice work, thank you ⭐

PR #3619, opened 2024-05-28
feat(Blocks): add ChildrenWithAge block

tujoworker:


joakbjerk:
Should we hide the "antall barn" card of "nei" is selected?
Screenshot 2024-05-29 at 09 13 24

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/blocks/info.mdx
```diff
@@ -0,0 +1,70 @@
+---
+showTabs: true
+---
+
+## Description
+
+Blocks are a collection of reusable fields and values. They can also be 100% cstomized and nested in each other. This makes it easy to reuse the same blocks in different contexts.
```
joakbjerk, line 7:
```suggestion
Blocks are a collection of reusable fields and values. They can also be 100% customized and nested in each other. This makes it easy to reuse the same blocks in different contexts.
```

joakbjerk (approved):

PR #3630, opened 2024-05-30
fix(TableContainer): take full width when inside `Card`

joakbjerk:
before:
Screenshot 2024-05-30 at 10 34 19

after:
Screenshot 2024-05-30 at 10 34 05


tujoworker (approved):
Nice ⭐

PR #3639, opened 2024-05-31
feat(forms): add `active` prop to Wizard.Step

tujoworker:
Is there any better prop name than `active`?


joakbjerk:
> Is there any better prop name than `active`?

From reading to comment, the only other suggestions I can come up with would be show/display 🤔

PR #3652, opened 2024-06-05
feat(forms): add edit and view containers to `Form.Section`

tujoworker:
Very similar to:

- https://eufemia.dnb.no/uilib/extensions/forms/Iterate/EditContainer/
- https://eufemia.dnb.no/uilib/extensions/forms/Iterate/ViewContainer/


packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Section.mdx
```diff
@@ -3,6 +3,11 @@ title: 'Section'
 description: '`Form.Section` lets you compose blocks of fields and values to be reused in different contexts.'
 hideInMenu: true
 showTabs: true
+tabs:
+  - title: Info
+    key: '/info'
+  - title: Demos
+    key: '/demos'
```
langz, line 10: We must remember to add properties as a tab, so that we can see the props(packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Section/EditContainer/properties.mdx)

tujoworker: True. I'll add it.

langz: The context of my question is that I looked at the example here: https://eufemia-git-feat-form-section-edit-view-containers-eufemia.vercel.app/uilib/extensions/forms/Form/Section/EditContainer/#demos

And I saw the prop containerMode used `<Form.Section path="/nestedPath" required containerMode="edit">`, but I don't know where I can read about that prop.

langz:
One experience testing the EditContainer, is that I don't feel like it works 100% as intended/described, but I believe perhaps the problem is that the example is opened by default.

In our docs it says, `The "Cancel" button resets any changes made to the item content, restoring it to its original state.`.

But in the following example https://eufemia-git-feat-form-section-edit-view-containers-eufemia.vercel.app/uilib/extensions/forms/Form/Section/EditContainer/#demos and video, it doesn't reset the "etternavn"-field(it still shows "a"):


https://github.com/dnbexperience/eufemia/assets/1359205/529f7d2b-ccb7-4770-92b6-657955014228

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Section/ViewContainer/info.mdx
```diff
@@ -0,0 +1,32 @@
...
+    </Form.Section.ViewContainer>
+  </Form.Section>,
+)
+```
+
+## Accessibility
+
+The `ViewContainer` component has an `aria-label` attribute, which is set to the `title` prop value. It uses a section element to wrap the content, which helps users with screen readers to get the needed announcement.
+
+When the item (view and edit) container gets removed, the active element focus will be set on the previews item.
```
langz, line 32:
```suggestion
When the item (view and edit) container gets removed, the active element focus will be set on the previous item.
```

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Section/info.mdx
```diff
@@ -235,6 +236,50 @@ export function MyForm() {
 }
 ```
 
+## Edit and View container
+
+The `Form.Seciton` supports an [EditContainer](/uilib/extensions/forms/Form/Section/EditContainer/) and [ViewContainer](/uilib/extensions/forms/Form/Section/ViewContainer/) container. The edit container should be used for data input with fields. While the view container is used to display the data in a read-only way.
```
langz, line 241:
```suggestion
The `Form.Section` supports an [EditContainer](/uilib/extensions/forms/Form/Section/EditContainer/) and [ViewContainer](/uilib/extensions/forms/Form/Section/ViewContainer/) container. The edit container should be used for data input with fields. While the view container is used to display the data in a read-only way.
```

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Section/info.mdx
```diff
@@ -235,6 +236,50 @@ export function MyForm() {
...
+
+render(
+  <Form.Handler>
+    <MySection />
+    <Form.SubmitButton />
+  </Form.Handler>,
+)
+```
+
+Note: The reason `Done` and not `Save` is used in the toolbar button tet is because validation is done in fields. If we decouple the data entered in a section, the `Form.Handler` will not be able to validate the data on submit. This can also lead to missing data because the user can press the submit button before `Save` is pressed.
```
langz, line 281:
```suggestion
Note: The reason `Done` and not `Save` is used in the toolbar button is because validation is done in fields. If we decouple the data entered in a section, the `Form.Handler` will not be able to validate the data on submit. This can also lead to missing data because the user can press the submit button before `Save` is pressed.
```

langz:
I'm a bit confused about the edit and view container of `Section` and `Iterate`, and how they are "intertwined" with each other.

There are multiple files where I would assume we care about the edit and view container of Iterate, and not Section, but where we've in this PR changed to care about Section in an "Iterate"-file.
For instance in the file `Iterate`/EditContainer/__tests__/EditAndViewContainer.test.tsx, we now query for `.dnb-forms-section-block` and not `.dnb-forms-iterate-block` which I would expect, since the test file is in `Iterate`/EditContainer/.

https://github.com/dnbexperience/eufemia/pull/3652/files#diff-5de40ef30a8ee79350a42cf7087760651b7aeadb35fff5c9503c1227114a58b8

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Section.mdx
langz: Thanks :)

langz (approved):

tujoworker:
> I'm a bit confused about the edit and view container of `Section` and `Iterate`, and how they are "intertwined" with each other, or if they are just separate stuff.

Yes, only the styles are shared at the moment. But I guess we should look into sharing also the React code from  `Seciton.EditContainer` with -> `Iterate.EditContainer`.

tujoworker:
> One experience testing the EditContainer, is that I don't feel like it works 100% as intended/described, but I believe perhaps the problem is that the example is opened by default.

Very nice finding 💯 This has been fixed 👍

langz:
> > One experience testing the EditContainer, is that I don't feel like it works 100% as intended/described, but I believe perhaps the problem is that the example is opened by default.
> 
> 
> 
> Very nice finding 💯 This has been fixed 👍 and tests where added.
> 
> 

Great work @tujoworker, thanks⭐️

PR #3665, opened 2024-06-05
docs(Tabs): improve property docs

langz:
I feel that the [properties docs of tabs](https://eufemia.dnb.no/uilib/components/tabs/properties/#key) is not very readable because of how the `data` property's `type` is formatted.

This is just a suggestion, but instead of changing the values of the type property(as I've done in the PR here) we should probably change the styling or so, to accomplish the same. Perhaps the `PropertiesTable` is the correct place to change? 🤔 

From:

Screenshot 2024-06-05 at 13 54 05

To:

Screenshot 2024-06-05 at 13 53 58

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/components/tabs/properties.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/components/tabs/properties.mdx
@@ -3,11 +3,18 @@ showTabs: true
 ---
 
 import PropertiesTable from 'dnb-design-system-portal/src/shared/parts/PropertiesTable'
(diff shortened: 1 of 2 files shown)


joakbjerk (approved):
I find the new proposal more readable 👍 Not a big fan of vertical scrolling when reading, so mitigating that would be nice 😄

tujoworker (approved):
Nice improvement 🚀

tujoworker:
Actually, looking at this closer, we should use the same approach as done with [ProgressIndicator](https://eufemia.dnb.no/uilib/components/progress-indicator/properties/#data-object-customcolors), where a value as object gets a new table.

PR #3666, opened 2024-06-06
feat(Table): added handle to collapse all accordion rows

snorrekim:
* added prop `collapseAllHandleRef` that exposes a ref handle that can be called to collapse all expanded rows in a table
* refactor `useTableAccordion` into a component  `<TableAccordion>` as that makes more sense in this case


tujoworker (approved):
Nice work. Thank you 💯

PR #3611, opened 2024-05-27
feat(Logo): Refactor Logo to ts

AadneRo:
- Refactor the Logo component to typescript
- Add deprecated tag to wrongly cased components and unused components
- Add typed docs

PR #3673, opened 2024-06-06
chore(Logo): fix size prop in logo

AadneRo:
The size prop was broken in earlier refactor to ts (#3611). This fix makes it backwards compatible.

### DIFF
--- a/packages/dnb-eufemia/src/components/logo/Logo.tsx
+++ b/packages/dnb-eufemia/src/components/logo/Logo.tsx
@@ -112,7 +112,7 @@ function Logo(localProps: LogoProps) {
   }, [brand, variant])
 
   /** @deprecated Can remove this in v11 */
-  const height = !isNaN(parseFloat(size)) ? size : heightProp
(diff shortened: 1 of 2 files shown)


packages/dnb-eufemia/src/components/logo/Logo.tsx
```diff
@@ -111,6 +111,9 @@ function Logo(localProps: LogoProps) {
     return 'dnb'
   }, [brand, variant])
 
+  /** @deprecated Can remove this in v11 */
+  const height = !isNaN(parseFloat(size)) ? size : heightProp
```
tujoworker, line 115:
```suggestion
  const height = heightProp ?? size
```
I think, this would be more "readable"  `const height = parseFloat(size)>0 ? size : heightProp`, but the code suggestion above should be fine as well, I think. Because, if height is defined, we want to prioritize it. And if size is defined, it acts as a fallback for height. And if nothing is defined, its undefined, and will not be set.

tujoworker (approved):

PR #3676, opened 2024-06-06
chore: improve home button on small devices

langz:
I guess the code/styling in this PR could be made better.
The reason for me going with the current approach(hiding/showing two different home buttons based on screen size) is because of the existing approach(as of today) does not work, because a "icon button with hidden text(display:none) still styles/acts as it is text there, as the node is still rendered".
Resulting in the "strange icon" we see today in the "From example", and this:

Screenshot 2024-06-06 at 15 01 36

From:
Screenshot 2024-06-06 at 14 37 11

To:

image

### DIFF
--- a/packages/dnb-design-system-portal/src/shared/menu/StickyMenuBar.module.scss
+++ b/packages/dnb-design-system-portal/src/shared/menu/StickyMenuBar.module.scss
@@ -1,3 +1,4 @@
+@use 'sass:math';
 @import '@dnb/eufemia/src/style/core/utilities.scss';
 
 :global {
@@ -44,12 +45,6 @@
     #github-button {
       display: inline-flex;
     }
-    #toggle-main-menu {
-      display: inline-flex;
-    }
-    #toggle-main-menu-small-screen {
-      display: none;
-    }
 
     /*
     God for a mobile menu instead
@@ -67,8 +62,10 @@
       #toggle-main-menu {
         display: none;
       }
+    }
+    @include allAbove(small, math.div(1, 16)) {
       #toggle-main-menu-small-screen {
-        display: inline-flex;
+        display: none;
       }
     }
   }
--- a/packages/dnb-design-system-portal/src/shared/menu/StickyMenuBar.module.scss
+++ b/packages/dnb-design-system-portal/src/shared/menu/StickyMenuBar.module.scss
@@ -44,7 +44,12 @@
     #github-button {
       display: inline-flex;
     }
-
+    #toggle-main-menu {
+      display: inline-flex;
+    }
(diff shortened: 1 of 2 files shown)


AadneRo:
This is only a fix for the portal, right? Is this something we should look at in the Button component as well?

langz:
> This is only a fix for the portal, right? Is this something we should look at in the Button component as well?

Yeah, that's correct, only in the portal. 

I'm not sure if this is possible to easily fix in the button itself? @tujoworker 
As of now, for this scenario the button relies heavily on hasText/has-text which is based on javascript if the button has text property or text as children. Not sure if it would be easy to make this work together with if the text is hidden in CSS(like display: none).

tujoworker:
@snorrekim has a task assigned to him and will look into the "button" issue itself. 

I think he should take a look on this PR and decide what we should do about it.

langz:
> @snorrekim has a task assigned to him and will look into the "button" issue itself.
> 
> I think he should take a look on this PR and decide what we should do about it.

That's fine by me, I'll leave the PR hanging here then 🙏

tujoworker:
Let's talk about this in tomorrow's meeting.

snorrekim:
I think this is the correct way to do it. The other option is to change the props based on screen size in javascript. I don't think we should try and make changes to the `<Button>` component to support something more. An icon button is very different from a text button (the size, spacing, and icon-size are all different), and trying to make it all work probably would be worth the effort.

I pushed a commit that makes the size correct and with some refactoring. Mainly reduced the customization so we only define `display: none`. Also added some invers media queries mixins so we don't get that 1px overlap (even adding an offset pixel didn't work if the screen was zoomed in)

langz:
> I think this is the correct way to do it. The other option is to change the props based on screen size in javascript. I don't think we should try and make changes to the `<Button>` component to support something more. An icon button is very different from a text button (the size, spacing, and icon-size are all different), and trying to make it all work probably would be worth the effort.
> 
> 
> 
> I pushed a commit that makes the size correct and with some refactoring. Mainly reduced the customization so we are less likely to have issues down the line. Also added some invers media queries mixins so we don't get that 1px overlap (even adding an offset pixel didn't work if the screen was zoomed in)
> 
> 
> 
> * Set size `default` for icon button
> 
> * Reduces code repetition for home button.
> 
> * Made sass mixins `notAbove`, `notBelow`, and `notBetween` that are inverses of the regular media queries. They allows us to avoid having to define an offset to prevent overlaps or gaps between queries.

I agree. I really like the imrpovements you've made here. Let's move the notMixins out from this PR, and merge this PR🫡

snorrekim:
I've removed the new mixins. I'll start working on a separate PR for them

PR #3680, opened 2024-06-07
fix(forms): solve issue when schema is used in Wizard

tujoworker:
> Cannot access 'errorMessagesRef' before initialization


langz (approved):
🙌

langz:
I don't think this is necessary related to the actual changes in this PR.
But I did some quick testing of the Wizard, and I've experienced that when changing the step that only the content of the step is changed, and not the "header.

Take a look:

https://github.com/dnbexperience/eufemia/assets/1359205/8f0ffffa-08d5-43c0-a15d-d48db074211d



Should I report this as a issue? 🤔

tujoworker:
> Should I report this as a issue? 🤔

Yes, that wold be nice 👍

langz:
> > Should I report this as a issue? 🤔
> 
> Yes, that wold be nice 👍

Issue https://github.com/dnbexperience/eufemia/issues/3685 is now created ✅

Issue #3687, opened 2024-06-07
bug(TextCounter): exceeded message when variant="up" is not correct


langz:
## 🐛 Bug Report

When showing message as exceeded using variant="up", like so: `<TextCounter variant="up" text="test" max={2} />`, the message displayed says `4 characters over the limit of 2.`, which is not correct. The correct message should be  `2 characters over the limit of 2.`

## To Reproduce

CSB: https://codesandbox.io/p/sandbox/strange-gauss-nh8d8z?file=%2Fsrc%2FApp.tsx%3A3%2C1

You can also experience it by changing the demo/example here https://eufemia-git-main-eufemia.vercel.app/uilib/components/fragments/text-counter/demos/

## Expected behavior

The correct message that should be displayed is  `2 characters over the limit of 2.`.
An other option could be to have two different "exceed messages", one for each variant. Thought I'm not sure how the exceed message for variant up should be like.

PR #3688, opened 2024-06-07
fix(TextCounter): exceeded message when variant up

langz:
Haven't actually made the fix yet, just added tests.

fixes https://github.com/dnbexperience/eufemia/issues/3687


tujoworker (approved):
Nice catch 🤩

PR #3693, opened 2024-06-11
docs(forms): add translations table

tujoworker:


langz (approved):
Looks great, it worked when I tested it now. 

A thought I have, should we list the translations under the properties tab? Maybe make a new tab for translations?
Or maybe a new tab for translations is too much. 

Anyways, just a few thoughts.

PR #3589, opened 2024-05-22
feat(FormStatus, Section): change sbanken info color to purple

snorrekim:
* Fix colors for sbanken `<FormStatus>` icons
* Sbanken `info` `<FormStatus>` now have the same color at `marketing`
* Update other uses of the "info" status style
* `<Breadcrumb>` now use custom background color in sbanken theme

PR #3695, opened 2024-06-11
fix(FormStatus, Section): reverts #3589

langz:
This reverts commit 5356eb8acbfb8fa62626e99fcc4e4b051a89f47f, and reverts #3589


langz:
Closed this, since I don't believe we have to revert this.

PR #3697, opened 2024-06-12
fix(StepIndicator): avoid re-assigning functions that can cause titles to be not in sync

### LINKED ISSUE #3685: bug(Wizard): changing step using the step indicator does not update/reflect in the step indicator
## 🐛 Bug Report

When changing the step using the step indicator, only the content of the step is changed, and not the step indicator("header"/part) where I clicked, it still says step 2 out of 2, and that step 2 is active, even though I pressed step 1.

Take a look:

https://github.com/dnbexperience/eufemia/assets/1359205/cef8bb37-f41d-44d6-8825-813fd0c7352a

## To Reproduce

https://eufemia.dnb.no/uilib/extensions/forms/Wizard/#intro-example

tujoworker:
Fixes #3685

Here is the same example as mentioned in the issue: https://eufemia-git-fix-wizard-step-title-eufemia.vercel.app/uilib/extensions/forms/Wizard/#intro-example

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Wizard/Container/__tests__/WizardContainer.test.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Wizard/Container/__tests__/WizardContainer.test.tsx
@@ -653,9 +653,9 @@ describe('Wizard.Container', () => {
     expect(wizardList()).not.toBeInTheDocument()
 
     rerender(
-      <Wizard.Container sidebarId="drawer-please">
+      <Wizard.Container variant="drawer">
         <Wizard.Step title="Step 1">
-          <output>Step 1</output>
+          <output>ensure re-render</output>
         </Wizard.Step>
 
         <Wizard.Step title="Step 2">
@@ -664,6 +664,7 @@ describe('Wizard.Container', () => {
       </Wizard.Container>
     )
 
+    expect(output()).toHaveTextContent('ensure re-render')
     expect(stepTrigger()).toBeInTheDocument()
     expect(wizardList()).not.toBeInTheDocument()
   })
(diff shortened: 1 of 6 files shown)


packages/dnb-eufemia/src/extensions/forms/Wizard/Container/__tests__/WizardContainer.test.tsx
```diff
@@ -653,7 +653,7 @@ describe('Wizard.Container', () => {
     expect(wizardList()).not.toBeInTheDocument()
 
     rerender(
-      <Wizard.Container sidebarId="drawer-please">
+      <Wizard.Container variant="drawer">
```
langz, line 656: Not sure if this is related to the fix?🤔

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Wizard/Container/Examples.tsx
```diff
@@ -213,7 +213,10 @@ export const AsyncWizardContainer = () => {
 
           return (
             <Form.Handler onSubmit={onSubmit}>
-              <Wizard.Container onStepChange={onStepChange}>
+              <Wizard.Container
+                onStepChange={onStepChange}
+                variant="drawer"
```
langz, line 218: Not sure if this is related to the fix?🤔

langz (approved):
Looks good, I have not been able to retest this yet, but can do so when I arrive at the office🚎

packages/dnb-eufemia/src/extensions/forms/Wizard/Container/__tests__/WizardContainer.test.tsx
tujoworker: It does, because it should. It was done so to ensure it re-renders, because that was the intention I think with the changed prop. But I made it more clear with a new commit 👍

PR #3699, opened 2024-06-12
chore(Icon): adds test for when icon provided is FC using hooks

langz:
- [x] Added a test case


langz:
For fun, I added this test to

tujoworker (approved):
Whoop 🙌

PR #3710, opened 2024-06-14
chore: replaceHydrationFunction only when testing

langz:
Just for build purposes.


packages/dnb-design-system-portal/gatsby-browser.tsx
```diff
@@ -22,9 +22,11 @@ function setIsTest(location) {
 export const replaceHydrateFunction = () => {
   // Added to solve the following errors, which prevented us from running screenshot tests
   // https://github.com/gatsbyjs/gatsby/discussions/36232
-  return (element: React.ReactElement, container: HTMLElement) => {
-    const root = ReactDOM.createRoot(container)
-    root.render(element)
+  if (globalThis.IS_TEST) {
```
tujoworker, line 25: We can use:

```js
const params = new URLSearchParams(document.location.search);
const themeName = params.get("eufemia-theme"); 
```

PR #3716, opened 2024-06-17
chore: replaceHydrationFunction only when testing

langz:


tujoworker (approved):
I did not check all the result extensively. But a quick look seems to show that we actually can go for this approach, right?

Thank you for the effort Anders 🌟

langz:
I've removed the commits and changes we discussed during todays meeting.
Will just test the preview a few times before I think it's okay to merge it.

PR #3721, opened 2024-06-18
fix(toCapitalized): remove lookbehind regular expression in order to support iOS Safari <v16 

tujoworker:
The test does fail when using the regexp version:

Screenshot 2024-06-18 at 22 08 08

Reprod: https://codesandbox.io/p/sandbox/eufemia-js-ios16-issue-gz77wm?file=%2Fsrc%2FApp.js%3A7%2C5


joakbjerk (approved):

tujoworker:
It does work when the fix is applied:

Screenshot 2024-06-18 at 22 48 36

PR #3726, opened 2024-06-19
feat(Accordion): allow multiple accordions inside a group to be expanded at the same time

joakbjerk:
- [x] Add expansion behavior prop instead

### DIFF
--- a/packages/dnb-eufemia/src/components/accordion/Accordion.tsx
+++ b/packages/dnb-eufemia/src/components/accordion/Accordion.tsx
@@ -228,13 +228,17 @@ function Accordion({
     if (context?.expanded_id && context.expanded_id === props.id) {
       setExpanded(true)
     }
-  }, [context.flush_remembered_state, context.expanded_id])
+  }, [
+    context.flush_remembered_state,
+    context.expanded_id,
+    props.expanded,
+    props.id,
+    store,
+  ])
 
   // Add callback for closing all accordions inside a group if collapseAllHandleRef is defined
   if (context?.collapseAllHandleRef && !hasAddedCallbackRef.current) {
-    context?.collapseAccordionCallbacks?.current.push(() =>
-      changeOpened(false)
-    )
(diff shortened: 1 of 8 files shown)


packages/dnb-eufemia/src/components/accordion/Accordion.tsx
```diff
@@ -228,13 +228,17 @@ function Accordion({
     if (context?.expanded_id && context.expanded_id === props.id) {
       setExpanded(true)
     }
-  }, [context.flush_remembered_state, context.expanded_id])
+  }, [
+    context.flush_remembered_state,
+    context.expanded_id,
+    props.expanded,
+    props.id,
```
AadneRo, line 235:
```suggestion
id,
```

Is there a reason we don't use the deconstructed props like this?

joakbjerk: Nope 😄 We should propably clean that up in a seperate pr, since there seem to be a lot of those occurances

packages/dnb-eufemia/src/components/accordion/__tests__/Accordion.test.tsx
```diff
@@ -465,6 +466,58 @@ describe('Accordion container component', () => {
         .getAttribute('style')
     ).toBe('transition-duration: 1ms; min-height: 6rem;')
   })
+
+  it('should allow all accordions inside a group to be expanded at the same time', async () => {
+    render(
+      <Accordion.Group allowAllExpanded>
```
tujoworker, line 472: I think the prop name is misleading. What if we rather add a `expandBehavior` prop? That defaults to "single" or something. But can be changed to "multi".

joakbjerk: Good idea, that seems more readable 👍

tujoworker (approved):
Nice 👍

PR #3727, opened 2024-06-20
feat(Switch): refactor to typescript

AadneRo:


packages/dnb-eufemia/src/components/switch/Switch.tsx
```diff
@@ -0,0 +1,256 @@
...
+      if (innerRef.current) {
+        innerRef.current.focus()
+      }
+    },
+    [callOnChange, innerRef, isChecked, onChangeEnd, readOnly]
+  )
+
+  const onKeyDownHandler = useCallback(
+    (event) => {
+      switch (keycode(event)) {
```
joakbjerk, line 125: Should probably be replaced with `event.key` or `event.code`, so we can get rid of the `keycode` dependency 🤔 and then change the case to `case 'Enter' to make it work

packages/dnb-eufemia/src/components/switch/SwitchDocs.ts
```diff
@@ -0,0 +1,87 @@
+import { PropertiesTableProps } from '../../shared/types'
+
+export const SwitchProperties: PropertiesTableProps = {
+  checked: {
+    doc: 'determine whether the switch is checked or not. The default will be `false`.',
```
joakbjerk, line 5: as Tobias reminded me, the start of the doc texts should be capitalized from now on 😄 

I used AI to quickly add the capitalized changes when I did similar work on the `Accordion` docs

joakbjerk (approved):
Looks good to me 👍

But we should probably add the noted changes before merge

packages/dnb-eufemia/src/components/switch/types.ts
```diff
@@ -0,0 +1,112 @@
...
+    | ((elem: HTMLInputElement) => void)
+} & Omit<
+  React.HTMLProps<HTMLElement>,
+  'ref' | 'size' | 'onChange' | 'innerRef'
+> &
+  SpacingProps &
+  DepricatedSwitchProps
+
+// deprecated, can be removed in v11
+type DepricatedSwitchProps = {
```
langz, line 95:
```suggestion
type DeprecatedSwitchProps = {
```

packages/dnb-eufemia/src/components/switch/SwitchDocs.ts
```diff
@@ -0,0 +1,87 @@
...
+    type: ['default', 'medium', 'large'],
+    status: 'optional',
+  },
+  status: {
+    doc: 'Text with a status message. The style defaults to an error message. You can use `true` to only get the status color, without a message.',
+    type: 'ReactNode',
+    status: 'optional',
+  },
+  statusState: {
+    doc: 'Defines the state of the status. Currently, there are two statuses `[error, info]`. Defaults to `error`.',
```
langz, line 40: Note to self, I think we should update/change this in many places in our docs, as the type says `type: ['error', 'warn', 'info', 'success', 'marketing']`, so it got to be more that only two statuses 🤔 

Nothing to improve in this PR, I'll look into it after this PR is merged.

langz (approved):

joakbjerk:
I think this one is ready for merge now 🤔

tujoworker (approved):
Looks good 👍

PR #3728, opened 2024-06-21
feat(Card): add support for nesting a Section with breakout until edge of card

tujoworker:
We will also need this the the inline help button.

Screenshot 2024-06-21 at 12 03 54


joakbjerk (approved):
Nice work ⭐️

PR #3732, opened 2024-06-25
feat(HeightAnimation): add property `compensateForGap`

tujoworker:
This feature is to make it possible to use `HeightAnimation` as a child inside a container using CSS row-gap, without the "jump" when the HeightAnimation gets display: none when hidden.

### DIFF
--- a/packages/dnb-eufemia/src/components/height-animation/HeightAnimationDocs.ts
+++ b/packages/dnb-eufemia/src/components/height-animation/HeightAnimationDocs.ts
@@ -1,0 +1,77 @@
+import { PropertiesTableProps } from '../../shared/types'
+
+export const HeightAnimationProperties: PropertiesTableProps = {
+  open: {
+    doc: 'Set to `true` on second re-render when the view should animate from 0px to auto. Defaults to `true`.',
+    type: 'boolean',
+    status: 'optional',
+  },
+  animate: {
+    doc: 'Set to `false` to omit the animation. Defaults to `true`.',
+    type: 'boolean',
+    status: 'optional',
+  },
+  keepInDOM: {
+    doc: 'Set to `true` ensure the nested children content will be kept in the DOM. Defaults to `false`.',
+    type: 'boolean',
+    status: 'optional',
+  },
+  compensateForGap: {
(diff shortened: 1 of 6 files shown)


packages/dnb-eufemia/src/components/height-animation/__tests__/HeightAnimation.test.tsx
```diff
@@ -160,6 +164,37 @@ describe('HeightAnimation', () => {
...
+        </HeightAnimation>
+      )
+
+      const inner = document.querySelector(
+        '.dnb-height-animation > .compensateForGap'
+      )
+      expect(inner).toBeInTheDocument()
+    })
+
+    it('should marginTop when compensateForGap is given', () => {
```
langz, line 181:
```suggestion
    it('should set marginTop when compensateForGap is given', () => {
```

packages/dnb-eufemia/src/components/height-animation/useHeightAnimation.tsx
```diff
@@ -18,6 +18,12 @@ export type useHeightAnimationOptions = {
    */
   animate?: boolean
 
+  /**
+   * To compensate for CSS gap between the rows so animation not jumps during the animation.
```
langz, line 22:
```suggestion
   * To compensate for CSS gap between the rows so animation does not jump during the animation.
```

langz (approved):

packages/dnb-eufemia/src/components/height-animation/HeightAnimationDocs.ts
```diff
@@ -0,0 +1,77 @@
...
+    type: 'boolean',
+    status: 'optional',
+  },
+  keepInDOM: {
+    doc: 'Set to `true` ensure the nested children content will be kept in the DOM. Defaults to `false`.',
+    type: 'boolean',
+    status: 'optional',
+  },
+  compensateForGap: {
+    doc: 'To compensate for CSS gap between the rows so animation not jumps during the animation. Provide a CSS unit or `auto`. Defaults to `null`.',
```
langz, line 20:
```suggestion
    doc: 'To compensate for CSS gap between the rows, so animation does not jump during the animation. Provide a CSS unit or `auto`. Defaults to `null`.',
```

PR #3746, opened 2024-06-28
fix(PaymentCard): set sbanken MC code to P103

haigol:
The previous sourceProductCode was wrong for Sbanken-cards. This one should be correct.


langz (approved):
Thanks, looks great :) 

I rebased this branch with main branch, as there was some conflict with changes done in https://github.com/dnbexperience/eufemia/pull/3747

PR #3747, opened 2024-06-30
fix(PaymentCard): change SBanken to Sbanken

thaytharma:
- changed SBanken to Sbanken
 - added proper brand name for Sbanken: 'Sbanken – et konsept fra DNB'


langz (approved):
Thanks, looks great. 
I'll take a closer look at it tomorrow. I see there's a failing visual test(atleast when I approved the PR), it could be random, or that we just have to update the test. 

🫡

PR #3752, opened 2024-07-02
feat(Table): add navigation mode

langz:
Navigation Table looks like this:
Screenshot 2024-07-02 at 16 11 12


langz:
@dsa015, we have started implementing what we believe you want/need. 
It would be great if you can set aside a few minutes to test it and provide feedback, so that we know if we are moving in the correct direction 🫡 

Demos/docs can be seen here:
https://eufemia-git-feat-table-navigation-mode-eufemia.vercel.app/uilib/components/table/#table-with-navigation

Press the «Code»-button to be able to explore and edit the code and example.

dsa015:
@langz this is great! This addresses our current design perfectly! 👍

joakbjerk (approved):
Nice work! 🔥 Looks good to me ⭐️

PR #3758, opened 2024-07-05
Chore/adds disallow leading zeros test

langz:


packages/dnb-eufemia/src/extensions/forms/Field/Number/__tests__/Number.test.tsx
```diff
@@ -265,6 +265,23 @@ describe('Field.Number', () => {
 
       expect(input).toHaveValue('365')
     })
+
+    it('should not allow leading zeroes when `disallowLeadingZeroes` is true', async () => {
+      render(<Field.Number value={123} />)
+
+      const input = document.querySelector('input')
+
+      await fireEvent.keyDown(document.querySelector('input'), {
```
langz, line 274: This is a false positive, doesn't matter if we provide disallowLeadingZeroes prop for this test or not.
So we'll have to re-write this, to work.

tujoworker (approved):

tujoworker:
Thank you for the test 🫶:

PR #3762, opened 2024-07-08
feat(Forms.Wizard): separate translations for buttons

langz:
This PR splits the `Step` translation, into the following separate translations:

- `WizardEditButton`
- `WizardPreviousButton`
- `WizardNextButton`

This PR also adds docs for the translation props of:

- `Wizard`
- `Wizard.PreviousButton`
- `Wizard.EditButton`
- `Wizard.Step.EditButton`

I've still kept the `Step` translation with the `summaryTitle`, as it's documented in so many places so there's a high chance changing that might break something for our users.


joakbjerk (approved):
Looks good to me 🧙‍♂️

PR #3767, opened 2024-07-09
chore(DrawerList): make for a more automated internal state managment of selection (WIP)

joakbjerk:
## WIP

- [x] Automate it
- [ ] Make data change trigger `onChange`
- [x] Fix tests
- [x] Fix visual tests
- [ ] Add Tests
- [x] Add Docs


langz (commented):
As a user of this component in the given scenario described in the PR description, wouldnt this only fix the problem visually? I would still have to manage my own state for the value, right?

joakbjerk:
> As a user of this component in the given scenario described in the PR description, wouldnt this only fix the problem visually? I would still have to manage my own state for the value, right?

Yeah, we need to verify this 🙏

PR #3771, opened 2024-07-10
feat(Forms): add validation for dnr and fnr in `Field.NationalIdentityNumber`

langz:
Motivation from https://dnb-it.slack.com/archives/CMXABCHEY/p1720521252444219

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/feature-fields/NationalIdentityNumber/Examples.tsx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/feature-fields/NationalIdentityNumber/Examples.tsx
@@ -112,6 +112,28 @@ export const ValidationRequired = () => {
   )
 }
 
+export const ValidationFnr = () => {
+  return (
+    <ComponentBox>
+      <Field.NationalIdentityNumber
+        value="29020112345"
+        validateInitially
+      />
+    </ComponentBox>
+  )
+}
+
+export const ValidationDnr = () => {
+  return (
+    <ComponentBox>
+      <Field.NationalIdentityNumber
+        value="69020112345"
+        validateInitially
+      />
+    </ComponentBox>
+  )
+}
+
 export const ValidationFunction = () => {
   return (
     <ComponentBox>
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/feature-fields/NationalIdentityNumber/demos.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/feature-fields/NationalIdentityNumber/demos.mdx
@@ -44,13 +44,17 @@ import * as Examples from './Examples'
 
 ### Validation - Norwegian national identity numbers
 
-It validates [norwegian national identity numbers(fnr)](https://www.skatteetaten.no/en/person/national-registry/identitetsnummer/fodselsnummer/) using the [fnrvalidator](https://github.com/navikt/fnrvalidator).
+It validates [Norwegian national identity numbers(fnr)](https://www.skatteetaten.no/en/person/national-registry/identitetsnummer/fodselsnummer/) using the [fnrvalidator](https://github.com/navikt/fnrvalidator).
+
+Below is an example of the error message displayed when there's an invalid Norwegian national identity number(fnr):
 
 <Examples.ValidationFnr />
 
 ### Validation - D numbers
 
 It validates [D numbers](https://www.skatteetaten.no/en/person/national-registry/identitetsnummer/d-nummer/) using the [fnrvalidator](https://github.com/navikt/fnrvalidator).
+
+Below is an example of the error message displayed when there's an invalid D number:
 
 <Examples.ValidationDnr />
 
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/feature-fields/NationalIdentityNumber/demos.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/feature-fields/NationalIdentityNumber/demos.mdx
@@ -42,8 +42,20 @@ import * as Examples from './Examples'
 
 <Examples.ValidationRequired />
 
+### Validation - Norwegian national identity numbers
+
+It validates [norwegian national identity numbers(fnr)](https://www.skatteetaten.no/en/person/national-registry/identitetsnummer/fodselsnummer/) using the [fnrvalidator](https://github.com/navikt/fnrvalidator).
+
+<Examples.ValidationFnr />
+
+### Validation - D numbers
+
+It validates [D numbers](https://www.skatteetaten.no/en/person/national-registry/identitetsnummer/d-nummer/) using the [fnrvalidator](https://github.com/navikt/fnrvalidator).
+
+<Examples.ValidationDnr />
+
 ### Validation function
 
-You can provide your own validation function or may use the one [from NAV](https://github.com/navikt/fnrvalidator).
+You can provide your own validation function.
 
 <Examples.ValidationFunction />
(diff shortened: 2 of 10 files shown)


joakbjerk:
Seems to be working nicely though 😄

Would there be any potential issues with having this as a dependency, instead of doing our own implementation?

I can't really think of any concrete examples right now, so asking out of curiosity 😄

langz:
> Seems to be working nicely though 😄
> 
> 
> 
> Would there be any potential issues with having this as a dependency, instead of doing our own implementation?
> 
> 
> 
> I can't really think of any concrete examples right now, so asking out of curiosity 😄

Yeah, was thinking the same. 
Would like to discuss this with @tujoworker as well.

tujoworker:
We have an example on how to implement the same: https://eufemia.dnb.no/uilib/extensions/forms/feature-fields/NationalIdentityNumber/demos/#validation-function


We need to have a discussion about whether we should implement it by default or not. 

We could also make it easy to add, so it still is not a part of the bundle by default, but where you import the validator and feed it to the component.

langz:
> We have an example on how to implement the same: https://eufemia.dnb.no/uilib/extensions/forms/feature-fields/NationalIdentityNumber/demos/#validation-function
> 
> We need to have a discussion about whether we should implement it by default or not.
> 
> We could also make it easy to add, so it still is not a part of the app bundle by default, but where you import the validator and feed it to the component.
> 
> But most importantly, we should check the size of the added validator so we know the addition.

[Unpacked Size 7.69 kB](https://www.npmjs.com/package/@navikt/fnrvalidator)

langz:
@tujoworker & @joakbjerk, I've updated the PR now by adding validation support for d number as well as fnr 🤞

joakbjerk:
Seems like both examples are validating d-number?

Screenshot 2024-08-28 at 15 08 06

langz:
> Seems like both examples are validating d-number?
> 
> 
> 
> Screenshot 2024-08-28 at 15 08 06
> 
> 
> 
> 

Yes, you as a user can't decide if you want one or the other validation(of course you could add your own validation using the validator prop). It validates both fnr and dnr. 
And since you've entered what could be a dnr(starting with a value between 4-7) then it says that it's not a dnumber. 

The examples are just meant for displaying the two different error messages, one for fnr and one for dnr.

joakbjerk:
> > Seems like both examples are validating d-number?
> > Screenshot 2024-08-28 at 15 08 06
> 
> Yes, as you've entered something that can only be a d number, not a fnr.
> 
> You as a user can't decide if you want one or the other validation(of course you could add your own validation using the validator prop). It validates both fnr and dnr.
> 
> In the [original examples](https://eufemia-git-feat-add-fnr-validation-on-blur-to-n-632efb-eufemia.vercel.app/uilib/extensions/forms/feature-fields/NationalIdentityNumber/demos/#validation---norwegian-national-identity-numbers), it's different values used than what you've entered. And since you've entered what could only be a dnr(starting with a value between 4-7) then it says that it's not a d number.
> 
> The examples are just meant for displaying the two different error messages, one for fnr and one for dnr.

Ahh good to know 🙏 

This might just be me misunderstanding again, so bear with me 😝 but copy pasting the numbers from `validDnum` array from the tests, into the input in the demo example, still yields the ´Invalid d-number´ error message. Is that because it's a d-number and not a fnr, or is there something else I'm misunderstanding here? 😄 other than that looks good to me 👍

https://github.com/user-attachments/assets/42507b7a-0e0c-4782-bd7e-13a8195ca13b

joakbjerk (approved):

langz:
> > > Seems like both examples are validating d-number?
> 
> > > Screenshot 2024-08-28 at 15 08 06
> 
> > 
> 
> > Yes, as you've entered something that can only be a d number, not a fnr.
> 
> > 
> 
> > You as a user can't decide if you want one or the other validation(of course you could add your own validation using the validator prop). It validates both fnr and dnr.
> 
> > 
> 
> > In the [original examples](https://eufemia-git-feat-add-fnr-validation-on-blur-to-n-632efb-eufemia.vercel.app/uilib/extensions/forms/feature-fields/NationalIdentityNumber/demos/#validation---norwegian-national-identity-numbers), it's different values used than what you've entered. And since you've entered what could only be a dnr(starting with a value between 4-7) then it says that it's not a d number.
> 
> > 
> 
> > The examples are just meant for displaying the two different error messages, one for fnr and one for dnr.
> 
> 
> 
> Ahh good to know 🙏 
> 
> 
> 
> This might just be me misunderstanding again, so bear with me 😝 but copy pasting the numbers from `validDnum` array from the tests, into the input in the demo example, still yields the ´Invalid d-number´ error message. Is that because it's a d-number and not a fnr, or is there something else I'm misunderstanding here? 😄 other than that looks good to me 👍
> 
> 
> 
> https://github.com/user-attachments/assets/42507b7a-0e0c-4782-bd7e-13a8195ca13b
> 
> 
> 
> 
> 
> 

Definitely something fishy going on here, I'll look into it🫡

langz:
> > > > Seems like both examples are validating d-number?
> > 
> > 
> > > > Screenshot 2024-08-28 at 15 08 06
> > 
> > 
> > > 
> > 
> > 
> > > Yes, as you've entered something that can only be a d number, not a fnr.
> > 
> > 
> > > 
> > 
> > 
> > > You as a user can't decide if you want one or the other validation(of course you could add your own validation using the validator prop). It validates both fnr and dnr.
> > 
> > 
> > > 
> > 
> > 
> > > In the [original examples](https://eufemia-git-feat-add-fnr-validation-on-blur-to-n-632efb-eufemia.vercel.app/uilib/extensions/forms/feature-fields/NationalIdentityNumber/demos/#validation---norwegian-national-identity-numbers), it's different values used than what you've entered. And since you've entered what could only be a dnr(starting with a value between 4-7) then it says that it's not a d number.
> > 
> > 
> > > 
> > 
> > 
> > > The examples are just meant for displaying the two different error messages, one for fnr and one for dnr.
> > 
> > 
> > Ahh good to know 🙏
> > This might just be me misunderstanding again, so bear with me 😝 but copy pasting the numbers from `validDnum` array from the tests, into the input in the demo example, still yields the ´Invalid d-number´ error message. Is that because it's a d-number and not a fnr, or is there something else I'm misunderstanding here? 😄 other than that looks good to me 👍
> > 
> >   
> >     
> >     
> > 
> >     Screen.Recording.2024-08-28.at.17.16.37.mov
> 
> Definitely something fishy going on here, I'll look into it🫡 Thanks for finding this 👏

@joakbjerk, thanks again 🙇 
So the `d number` you tried in that scenario was actually not a correct d number.
My mistake for providing it, since I had made my own d numbers based on fnrs, but that clearly did not work.
The tests did not fail for this value, since the tests did actually not run correctly 😱 
I've fixed them now, and double-checked, see the following commit https://github.com/dnbexperience/eufemia/pull/3771/commits/d27d096f73f09499ac08d8de75dee966efe70810
Reason why it wasn't validating the tests correctly is something about how the `validator`  prop in String is implemented, so have to do some kind of different [assertion](https://github.com/dnbexperience/eufemia/pull/3771/commits/d27d096f73f09499ac08d8de75dee966efe70810#diff-79d33bdc66b768e3946ad33182de3b439e3834f7c2eff8c60a9244efcd83795cR9), which I've copied from the Field.String tests(which is the underlying component of NationalIdentityNumber).

I've also used the following [library](https://github.com/navikt/nav-faker?tab=readme-ov-file) to now generate correct d numbers as testdata.

tujoworker (approved):
Very nice work 🚀

langz:
Some additional thoughts about implementing this:
- What's the difference betweeen pattern and validator, as a user/application. Will I know/carr about the difference?
- As a user, if I want to extend the validator, by adding a check for if fnr is older than 18 years, how can I do that, while still keeping the existing fnr and dnr validations?

langz:
Note to self, will the [validate](https://eufemia.dnb.no/uilib/extensions/forms/feature-fields/NationalIdentityNumber/properties/) prop still work now?
Will it deactivate the dnr and fnr validations?

PR #3778, opened 2024-07-16
feat(Img): add support for loading-prop

thaytharma:
Add support for `loading`-prop in `Img`. Mainly to be able to use native '`lazy`'-loading.


tujoworker (approved):
Thank you 🥇

PR #3779, opened 2024-07-16
chore(Portal): fix Github button size

thaytharma:
The  Github button size was slightly larger than the "more options"-menu button.

Before:
Skjermbilde 2024-07-16 kl  23 00 39

After:
Skjermbilde 2024-07-16 kl  23 00 24


langz (approved):
❤️

PR #3780, opened 2024-07-16
(fix) Can now click on arrow when text is highlighted to expand/collapse

### LINKED ISSUE #3750: Table: Accordion button does not work(expand or collapse) when pressed, after selecting some other text from the table row
https://github.com/dnbexperience/eufemia/assets/1359205/89def15f-a610-493a-9267-74287d7b00c4

tariknh:
Fixes #3750

This removes  the previous: `user-select: none; // prevent selection on double-click` . 

If by selection they meant text highlight, then this previous fix i removed only solved double clicking on the arrow only.

### DIFF
--- a/packages/dnb-eufemia/src/components/table/style/table-accordion.scss
+++ b/packages/dnb-eufemia/src/components/table/style/table-accordion.scss
@@ -57,7 +57,6 @@
     &.dnb-table__td {
       padding: 0;
     }
-    user-select: none; // prevent selection on double-click
   }
 
   &__tr--clickable &__button {


tujoworker:
Thank you for this PR 🫶

Looking at it, we should keep the `user-select: none;` because it serves a different issue. 
To fix the overall downside reported #3750, we should fix this issue on the JavaScript side and where the "prevent click" check happens. Its located in the [`onClickTr`](https://github.com/dnbexperience/eufemia/blob/702118a966858e73f3e7bae77ab5e026f083fd54/packages/dnb-eufemia/src/components/table/TableClickableHead.tsx#L92) handler inside `TableClickableHead.tsx` file.

tujoworker:
Thank you for this PR 🫶

Looking at it, we should keep the `user-select: none;` because it serves a different issue. 
To fix the overall downside reported #3750, we should fix this issue on the JavaScript side and where the "prevent click" check happens. Its located in the [`onClickTr`](https://github.com/dnbexperience/eufemia/blob/702118a966858e73f3e7bae77ab5e026f083fd54/packages/dnb-eufemia/src/components/table/TableClickableHead.tsx#L92) handler inside `TableClickableHead.tsx` file.

PR #3782, opened 2024-07-22
fix(forms): add support for horizontal layout in SummaryList and adjust spacing and alignment

### LINKED ISSUE #3741: Value.SummaryList horizontal layout is broken
## 🐛 Bug Report

Horizontal layout on Value.SummaryList does not make new lines for each Value, and spacing seems to be off.
(And for some reason `<strong>` does not seem to have an effect in the reprod...)

Also happens with `grid` layout:
Also when not having a `label`, the Value does not take all of the row space.
This results in the next Value label appearing on the same row as the first Value

I noticed that there are no examples for horizontal layout in the docs or storybook.

Tested in Edge and Firefox
image

## To Reproduce

https://[codesandbox.io/p/sandbox/kind-hertz-cqhp39?file=%2Fsrc%2FApp.tsx%3A1%2C14](https://codesandbox.io/p/sandbox/kind-hertz-cqhp39?file=%2Fsrc%2FApp.tsx%3A1%2C14)

## Expected 

Something like this:
image

## Eufemia Version

**Browser JS:** 10.38.0

**Browser CSS:** "10.38.0"

tujoworker:
fixes #3741


langz (approved):
Looks good to me :) 
I see there's quite a few changes(in width and height px) to screenshot tests, especially in the Upload component. Are these changes expected? Could these changes be breaking?

tujoworker:
> I see there's quite a few changes(in width and height px) to screenshot tests, especially in the Upload component. Are these changes expected? Could these changes be breaking?

Yes they are expected. I don't can imagine it would break other places as it's just a small visual adjustment when it comes to the changed details.

PR #3794, opened 2024-08-06
docs(Portal): add accordion to side menu

snorrekim:
## Accordion
* Can make a menu item into an accordion with the doc option `accordion: true` (default is `false`)
* Accordions automatically open and close depending on which page you navigate to.
* A "+/-" button indicates the accordions and allows for manual opening and closing.
* Fallback auto-scroll _after_ expanding if the item is no longer in view. Because the default auto-scroll can't handle multiple accordions opening and closing at once.

## Menu ordering
* Added ability to use negative `order` to add to the bottom (`-1` is the the first of them, followed by `-2` etc.).
* docs without order will be placed in the middle.
* fix: docs with `order` number now at the top instead of the bottom (except negative order)
* fix: docs with same title and level no longer ignored
* fix: docs with dot `.` in the file name no longer ignored

## New doc structure
* "Releases and versions" can now be expanded to see each release page
* "Pagination" and "Fragments" children now hidden behind accordion
* "Form", "Wizard", "Iterate", "Value", and "Create your own component" can now be expanded to see sub-components
* Changed "All feature fields" to accordion "More feature fields..."

## Other
* "Getting started" table of contents filled out to include all headings
* "Releases and versions" now uses `<ListSummaryFromEdges>` component
* `<ListSummaryFromEdges>` 
  * component now shows description in `returnListItems` mode
  *  fix: `<ListSummaryFromEdges>` made an `<ul>`for each `<li>` in `returnListItems` mode. Now there is only one `<ul>` for the entire list.
* Side menu is now a hierarchy of nested `<ul>` instead of one flat list.

## Not-implemented ideas for the future
* Auto-scroll
  * instead of scrolling the item to the top it should be smart enough to include context. Scroll to the parent item, or until the item is centered if the list is too long.
  * Should animate the item when finished scrolling (a blink effect maybe?), so the user can easily see where they are
  * It should completely control scrolling by stopping all other scrolling at the start of expansion so we don't get the occasional double scroll.
* The accordion button interferes with the sbanken styling, this should be improved
* The "Usage" sub-menu should probably also implement some accordions
* "Base fields" and "Feature fields" could also be accordions, then showing all the children would take less space. today we hide a few, and that's not optimal.


snorrekim:
Discrussion:
* non cursive description in UL edges. Can we markdown kursive in description?
* "All feature fielst" should exist as an exåpandable "more... +/-"

tujoworker (approved):
Fantastic improvement 💯 Thank you so much for the hard work 🫶

PR #3801, opened 2024-08-08
fix(Button): Allow `react-router` Link as `element`

jonasws:
This commit updates the type signature of `Button` to allow passing a
`react-router`-style `Link` in the `element` props.


langz (approved):
Looks good, thanks for the great contribution❤️

This fix will be included in release https://github.com/dnbexperience/eufemia/pull/3800 which is hopefully very soon🤞

PR #3804, opened 2024-08-09
feat(Form): continuousValidation did not update error state correctly

snorrekim:
`FieldBoundaryContext`'s `hasVisibleError` was not updated if attempting to `showError()` and there is none. If there was an error before, then any component that used `hasVisibleError` (only `Iterate.EditContainer` and `Form.Section.EditContainer` at the moment) would still think there was an error visible. 

`showError()` is run at any change whenever `continuousValidation` is used. Some components always use `continuousValidation` like `ArraySelection`.

* `hasVisibleError` is now set to `false` if there is no error when attempting to "showError"
* `showError` renamed to `revealError` to make it more clear that it does not actually mean there is an error.


tujoworker (approved):
Fantastic 👏

PR #3809, opened 2024-08-12
feat(forms): add `Form.clearData` method to remove existing/shared data forms with an id

### LINKED ISSUE #3803: Forms: Provide a way of resetting or scoping form data stored by id?
## 🐛 Bug Report

When a form has been defined with an `id`, the contents of the form persists even when react re-mounts the application. This sounds like it might be the intention, so I don't necessarily want to call it a bug, but wanted to raise it since it causes problems for us.

We have a form with an id that we render in a test. When we run multiple tests, the state of the form persists from the first test to the next.

## To Reproduce

Make a test file with this content (sorry, I couldn't make testing-library run properly in codesandbox):

```tsx
import "@testing-library/jest-dom/extend-expect";
import React from "react";
import { render, screen } from "@testing-library/react";
import user from "@testing-library/user-event";
import { Form, Field } from "@dnb/eufemia/extensions/forms";

function MyForm() {
  return (
    <Form.Handler
      id="my-id"
      defaultData={{
        firstName: "",
      }}
    >
      <Field.String path="/firstName" label="First name" />
    </Form.Handler>
  );
}

test("Filling out the form", async () => {
  render(<MyForm />);

  const nameField = screen.getByRole("textbox", { name: /first name/i });
  await userEvent.type(nameField, "Sally");

  expect(nameField).toHaveValue("Sally");
});

test("Filling out the form again", async () => {
  render(<MyForm />);

  const nameField = screen.getByRole("textbox", { name: /first name/i });
  await userEvent.type(nameField, "John");

  expect(nameField).toHaveValue("John");
});
```

The second test fails, because the value of the field is "SallyJohn", and not "John".

## Expected behavior

The form does not retain the entered data between tests. Either by some way of resetting the state between tests, or by providing a way of scoping it to a context (perhaps the `EufemiaProvider`).

## Eufemia Version

**Node.js:** 10.43.0

tujoworker:
Closes #3803

- [x] Figure out the [best API/name](https://github.com/dnbexperience/eufemia/issues/3803#issuecomment-2283661946)
- [x] Add docs

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/DataContext/Provider/Provider.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/DataContext/Provider/Provider.tsx
@@ -587,6 +587,11 @@ export default function Provider<Data extends JsonObject>(
       sharedData.data !== internalDataRef.current
     ) {
       cacheRef.current.shared = sharedData.data
+      if (sharedData.data.resetForm) {
+        const reset = {} as Data
+        sharedData.set(reset)
+        return reset
+      }
       return {
         ...internalDataRef.current,
         ...sharedData.data,
--- a/packages/dnb-eufemia/src/extensions/forms/DataContext/Provider/Provider.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/DataContext/Provider/Provider.tsx
@@ -587,10 +587,10 @@ export default function Provider<Data extends JsonObject>(
       sharedData.data !== internalDataRef.current
     ) {
       cacheRef.current.shared = sharedData.data
-      if (sharedData.data.resetForm) {
-        const reset = {} as Data
-        sharedData.set(reset)
-        return reset
+      if (sharedData.data.clearForm) {
+        const clear = {} as Data
+        sharedData.set(clear)
+        return clear
       }
       return {
         ...internalDataRef.current,
--- a/packages/dnb-eufemia/src/extensions/forms/Form/data-context/__tests__/clearData.test.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Form/data-context/__tests__/clearData.test.tsx
@@ -1,0 +1,45 @@
+import React from 'react'
+import { act, render } from '@testing-library/react'
+import { Field, Form } from '../../..'
+
+describe('Form.clearData', () => {
+  it('should reset a form with an id', () => {
+    render(
+      <Form.Handler id="unique-id" data={{ myString: 'my string' }}>
+        <Field.String path="/myString" />
+      </Form.Handler>
+    )
+
+    expect(document.querySelector('input')).toHaveValue('my string')
+
+    act(() => Form.clearData('unique-id'))
+
+    expect(document.querySelector('input')).toHaveValue('')
+  })
+
+  describe('can be used in beforeEach', () => {
+    beforeEach(() => {
+      Form.clearData('unique-id')
+    })
+
+    it('first set the value', () => {
+      render(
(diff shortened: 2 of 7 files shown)


packages/dnb-eufemia/src/extensions/forms/Tools/__tests__/formReset.test.tsx
```diff
@@ -0,0 +1,45 @@
...
+      render(
+        <Form.Handler id="unique-id" data={{ myString: 'my string' }}>
+          <Field.String path="/myString" />
+        </Form.Handler>
+      )
+
+      expect(document.querySelector('input')).toHaveValue('my string')
+    })
+
+    it('second, it should not be present anymore', () => {
```
langz, line 35: Does this actually "prove that" it works?
I'm pretty sure this test would also work(go green) even without the:

```
  beforeEach(() => {
      Tools.formReset('unique-id')
    })
```

langz (approved):

packages/dnb-eufemia/src/extensions/forms/Tools/__tests__/formReset.test.tsx
tujoworker: nope, it will not. Because, when an `id` is defined, it automatically syncs the data to a shared store and because the ID is the same, it will pick it up again, and fail.

langz: Thanks, I learned something new today as well 🙏

packages/dnb-eufemia/src/extensions/forms/Form/data-context/__tests__/clearData.test.tsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react'
+import { act, render } from '@testing-library/react'
+import { Field, Form } from '../../..'
+
+describe('Form.clearData', () => {
+  it('should reset a form with an id', () => {
```
langz, line 6:
```suggestion
  it('should clear a form with an id', () => {
```

langz (approved):
Looks great :) 

I see that there's a few references to `reset` left, but that might not be in scope of this PR to change/rename, or perhaps not even necessary 🤔

packages/dnb-eufemia/src/extensions/forms/DataContext/Provider/Provider.tsx
```diff
@@ -587,6 +587,11 @@ export default function Provider<Data extends JsonObject>(
       sharedData.data !== internalDataRef.current
     ) {
       cacheRef.current.shared = sharedData.data
+      if (sharedData.data.resetForm) {
+        const reset = {} as Data
+        sharedData.set(reset)
+        return reset
+      }
```
tujoworker, line 594:
```suggestion
      if (sharedData.data.clearForm) {
        const clear = {} as Data
        sharedData.set(clear)
        return clear
      }
```

packages/dnb-eufemia/src/extensions/forms/Form/data-context/clearData.ts
```diff
@@ -0,0 +1,9 @@
+import { createSharedState } from '../../../../shared/helpers/useSharedState'
+
+export default function clearData(id: string) {
+  const sharedAttachments = createSharedState(id + '-attachments')
+  sharedAttachments.set({})
+
+  const sharedData = createSharedState(id)
+  sharedData.update({ resetForm: true })
```
tujoworker, line 8:
```suggestion
  sharedData.update({ clearForm: true })
```

PR #3819, opened 2024-08-13
feat(Forms): seperate translations for Boolean and Toggle

langz:
Could potentially be breaking if there's someone who manually sets BooleanField for the Field.Toggle.


tujoworker:
I'm thinking about a good reasons to separate these. Will they not always be the same? Probably for logical reasons?

langz:
The reason I can see, is if users would want to have different translation values for these two different components. 

And also from a user perspective, it's more logical to change translation X for component X, and translation Y for component Y. Rather than changing translation X and both component X and Y changes.

tujoworker (approved):

PR #3829, opened 2024-08-15
fix(forms): add `Form.Isolation` path support when used inside `Form.Section`

tujoworker:


packages/dnb-eufemia/src/extensions/forms/Form/Isolation/IsolationDocs.ts
```diff
@@ -31,7 +31,7 @@ export const IsolationProperties: PropertiesTableProps = {
 
 export const IsolationEvents: PropertiesTableProps = {
   onCommit: {
-    doc: 'Will be called on a nested form context commit – if validation has passed.',
+    doc: 'Will be called on a nested form context commit – if validation has passed. The first parameter is the committed data object. The second parameter is on object containing a method to clear the internal data `{ clearData }`.',
```
langz, line 34:
```suggestion
    doc: 'Will be called on a nested form context commit – if validation has passed. The first parameter is the committed data object. The second parameter is an object containing a method to clear the internal data `{ clearData }`.',
```

langz (approved):
Looks good to me, thanks for the demo 🍿

andlbrei:
There is something weird happening, and I'm not sure exactly how.
The isolated value ends up on the root of the committed object.

https://github.com/user-attachments/assets/809bbb31-e9a2-4427-b3e4-ba286193f40c

Also, might be unrelated to this MR and not necessarily a problem, but I would have thought that the `onCommit` `data` parameter would only contain the object that is in the scope of the `Form.Isolation` component.
And perhaps the same for `onPathChange`.
At least that's how I think about this component. It is isolated from the rest of the form, doesn't know or care what the parent paths are, and the data will be committed to the place it belongs to in the hierarchy, which the parent has control over (or somewhere else if it is overriden via props).
This is a bit taxing to think about, so I might be completely off base, missing some cruical reasons as to why it can't be, or shouldn't be, like this 😅 

```
<Form.Section path="/mySection">
  <Field.String path="/name" />
  <Form.Isolation onCommit={(data) => console.log(data)}> // Will log { address: "Some user input" } and inserted/merged into { mySection } resulting in { mySection: { name: "Some name", address: "Some user input" }}
    <Field.String path="/address" />
  </Form.Isolation>
</Form.Section>
```

tujoworker:
yes, very good points 💯 Both of your concerns are implemented now. Thank you @andlbrei 🥇 
Happy if you can verify it and eventually approve the PR ✅

andlbrei:
@tujoworker If the user presses commit without doing any changes, then the default value won't be commited.

Other than that it seems good 👌

tujoworker:
Ohh yes. The latest commit fixes this issue, including a test.

andlbrei (approved):
👌

PR #3830, opened 2024-08-15
feat(Field): adds typing for event callbacks

snorrekim:
* Added the type `FieldPropsWithExtraValue` if we need to type events with two arguments.
* The type `FieldProps` will now type events like `onChange`, `onBlur` and `onFocus` with only one argument.
* Added the type `FieldPropsGeneric` that makes no assumptions on whether events have one or two parameters
* `Field.PhoneNumber` and `Field.SelectCountry` now types their second event arguments.

Other fixes:
* fixed `Field.Expiry` missing type for empty value causing useless event type
* fixed `Iterate.Array` having types for non existing props

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/hooks/useFieldProps.ts
+++ b/packages/dnb-eufemia/src/extensions/forms/hooks/useFieldProps.ts
@@ -13,7 +13,7 @@ import { errorChanged } from '../utils'
 import { ajvErrorsToOneFormError } from '../utils/ajv'
 import {
   FormError,
-  FieldPropsBase,
+  FieldPropsGeneric,
   AdditionalEventArgs,
   SubmitState,
   EventReturnWithStateObjectAndSuccess,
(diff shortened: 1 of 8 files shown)


packages/dnb-eufemia/src/extensions/forms/hooks/useFieldProps.ts
```diff
@@ -111,10 +108,10 @@ export default function useFieldProps<
     transformAdditionalArgs = (additionalArgs: AdditionalEventArgs) =>
       additionalArgs,
     fromExternal = (value: Value) => value,
-    validateRequired = (value: Value, { emptyValue, required, error }) => {
+    validateRequired = (value, { emptyValue, required, error }) => {
       const res =
         required &&
-        (value === emptyValue ||
+        ((value as any) === emptyValue ||
```
tujoworker, line 114: Do we need any here?

tujoworker (approved):
Looks good to me 👏

packages/dnb-eufemia/src/extensions/forms/hooks/useFieldProps.ts
langz: Not sure if we "need it" or if we could/should make the types "comparable".

This is the TS error I get when removing the as any here:

`This comparison appears to be unintentional because the types 'Value' and 'EmptyValue' have no overlap.ts(2367)
`

image

langz (approved):

packages/dnb-eufemia/src/extensions/forms/hooks/useFieldProps.ts
snorrekim: It's an annoying error, because the === comparison is also technically a type check in it self. Though it technically exposes the fact that our types doesn't state a relationship between the generic types `Value` and `EmptyValue`, when there definitely is one. Most places where we use `Value` we actually mean `Value | EmptyValue`.

I think we'll just keep the `any` for now, because the. `===` comparison is definitely safe.

PR #3841, opened 2024-08-19
chore(PostalCodeAndCity): fix iterate over array example

langz:
Fixes the following error:
`ReferenceError: Iterate is not defined`

image


tujoworker (approved):

tujoworker:
I think we may add `Iterate` to the default imported things, like Form and Field. What do you think?

langz:
Seems like a great idea. It's used in quite many examples/demos.

PR #3845, opened 2024-08-20
chore: default import Iterate in ComponentBox

langz:
As suggested [here](https://github.com/dnbexperience/eufemia/pull/3841#issuecomment-2297024013)


tujoworker (approved):
Nice 👏

PR #3847, opened 2024-08-20
fix(forms): make `Form.Isolation`s data flow "from outside" stricter

### LINKED ISSUE #3844: Form.Isolation defaultData + useState = problem
## 🐛 Bug Report

When a `useState` `set` function is called the `data` from `Form.Handler` seems to override `defaultData` set on `Form.Isolation`.

https://github.com/user-attachments/assets/40a15928-3730-41f6-881a-3aa7848fb937

## To Reproduce

https://codesandbox.io/p/sandbox/form-isolation-defaultdata-usestate-r2slx6

Have a `Form.Handler` with `data` and a child `Form.Isolation` with some other `defaultData` and some `set` function which is used when a `Field` changes it's value.
When the component re-renders the value from `Form.Handler data` will override the inner `Form.Isolation defaultData`.

## Expected behavior
I expect values from `Form.Handler` to not have any impact in `Form.Isolation` when `Form.Isolation defaultData` i set.

## Eufemia Version

10.44.0

tujoworker:
Fixes #3844

The test fails with the old implementation 👍

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Isolation/info.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Isolation/info.mdx
@@ -12,10 +12,10 @@ It's a provider that lets you provide a `schema` or `data` very similar to what 
 ### Good to know
 
 - It needs to be used inside of a `Form.Handler` component.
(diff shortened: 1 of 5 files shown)


andlbrei (approved):
Looks good 👍

langz (approved):
🥳

PR #3849, opened 2024-08-20
fix(forms): enhance `Field.Name` validation rule

tujoworker:


langz (approved):
Looks fine to me 👍 

I don't 100% understand the validation rule, or why we want/need validation rules for names. Does it exist a rule for how a name could look like? Is it the government who decides this? Is it different from country to country? Or perhaps its DNB who wants these rules for names? 🤔

tujoworker:
We had a very weak check. The new regex is still not very strict, but as you can see in the tests, it takes care of obvious wrongly written names.
I don't think there is the perfect validation for names. But I think this brings us a long way. And yes, the change is made from feedback by the Trident team. That said, they used a much stricter regex.

langz:
> We had a very weak check. The new regex is still not very strict, but as you can see in the tests, it takes care of obvious wrongly written names.
> 
> I don't think there is the perfect validation for names. But I think this brings us a long way. And yes, the change is made from feedback by the Trident team. That said, they used a much stricter regex. 

Thanks for the explanation🫡

PR #3858, opened 2024-08-21
feat(forms): add `inheritLabel` prop to value components

tujoworker:


joakbjerk (approved):
Looks good! ☀️

tujoworker:
As discussed in the meeting today, we rather introduce the `inheritLabel` property for the same functionality.

PR #3859, opened 2024-08-21
feat(Upload): define max file size for file type

langz:
Screenshot of a "complex variant" of this feature:
Screenshot 2024-08-22 at 20 40 33

PR #3860, opened 2024-08-21
chore: bump typescript to 5.5.4

langz:


joakbjerk:
This functionality seems to be only supported in newer versions of safari, is that something we should take into consideration, or is there some kind of polyfill we could use to mitigate this? 🤔

joakbjerk:
> > This functionality seems to be only supported in newer versions of safari, is that something we should take into consideration, or is there some kind of polyfill we could use to mitigate this? 🤔
> 
> I think/hope that babel would just take care of this, but could be that something else is needed, and I also think that I have to bump the version of node to support this as well. I'll postpone this for a while, and not use Object.groupBy() in #3859 as of now

Ahh okay. The functionality itself looks great though, so it would be nice if we could use it 😕

tujoworker:
We postpone the upgrade for now.

PR #3863, opened 2024-08-22
feat(Button): rewise documented `title` prop with aria-label documentation

snorrekim:
* `aria-label` is automatically set to the same value as `title` if the button has no content
* documentation updated to make it clear that `title` is not good enough for accessibility


snorrekim:
Looking at things. Shouldn't `tooltip` actually be used instead of either `title` and `aria-label`? It uses `aria-describedby` which I believe has good support, it creates a much more readable hover text, and screenreaders are not at risk of repeating the text.

packages/dnb-design-system-portal/src/docs/uilib/components/button/properties.mdx
```diff
@@ -8,7 +8,7 @@ showTabs: true
...
 | `text` or `children`                    | _(optional)_ the content of the button can be a string or a React Element.                                                                                                                                                                                             |
-| `title`                                 | _(optional)_ title of the button. Optional, but should always be included because of accessibility.                                                                                                                                                                    |
+| `aria-label` or `title`                 | _(optional)_ required if there is no text in the button. If `text` and `children` are undefined, setting the `title` prop will automatically set `aria-label` with the same value.                                                                                     |
```
langz, line 11:
```suggestion
| `ariaLabel` or `title`                 | _(optional)_ required if there is no text in the button. If `text` and `children` are undefined, setting the `title` prop will automatically set `aria-label` with the same value.                                                                                     |
```

Should the propname be ariaLabel(which I assume is the prop name for the React component, or aria-label which is the HTML attribute).

langz (approved):

packages/dnb-eufemia/src/components/button/Button.js
```diff
@@ -279,7 +282,7 @@ export default class Button extends React.PureComponent {
 Button.propTypes = {
   text: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
   type: PropTypes.string,
-  title: PropTypes.node,
+  title: PropTypes.string,
```
langz, line 285: When making this change, will it still be possible to provide a translation from like react-intl, like `<FormattedMessage id="greeting" / >` ?
I believe some apps use the title prop this way.

packages/dnb-design-system-portal/src/docs/uilib/components/button/properties.mdx
snorrekim: For some reason `aria-` attributes still use hyphens in React. It's mentioned in their documentation:

> [Note that all aria-* HTML attributes are fully supported in JSX. Whereas most DOM properties and attributes in React are camelCased, these attributes should be hyphen-cased (also known as kebab-case, lisp-case, etc) as they are in plain HTML](https://legacy.reactjs.org/docs/accessibility.html#wai-aria)

packages/dnb-eufemia/src/components/button/Button.js
snorrekim: I don't think it can actually do that today. The "string" becomes `[object Object]`. This is the pure html prop `title` and it only works with strings. In their documentation Format.JS they een say

> [While our [components](https://formatjs.io/docs/react-intl/components) provide a seamless integration with React, the imperative API are recommended (sometimes required) in several use cases:

Setting text attributes such as title, aria-label and the like where a React component cannot be used (e.g )](https://formatjs.io/docs/react-intl/api/#why-imperative-api)

PR #3869, opened 2024-08-26
docs(Selection): value should be string when variant is radio or button

langz:
Motivation:
The [value property of Field.Selection](https://eufemia.dnb.no/uilib/extensions/forms/base-fields/Selection/properties/) states the type is string and number, however when using variant radio or button with a value of type number it breaks.

Reason it breaks, is because the value property of [RadioButton](https://eufemia.dnb.no/uilib/components/radio/properties/) and [ToggleButton](https://eufemia.dnb.no/uilib/components/toggle-button/properties/) states it only supports string. 

Any better ways to fix/improve on this, rather than just adding a description to the docs? 🤔

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/base-fields/Selection/properties.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/base-fields/Selection/properties.mdx
@@ -14,7 +14,4 @@ import { SelectionProperties } from '@dnb/eufemia/src/extensions/forms/Field/Sel
 
 ### General props
 
-<PropertiesTable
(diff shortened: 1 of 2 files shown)


joakbjerk:
I think we maybe could do some TypeScript magic here, but can't come up with a good example on the go right now 😅

langz:
> I think we maybe could do some TypeScript magic here, but can't come up with a good example on the go right now 😅

Yeah, I agree, but if I remember correct. The last time I looked into this, the TypeScript Magic results in code that's very difficult to read. So decided against it.

tujoworker (approved):
Nice 👏

PR #3870, opened 2024-08-26
chore(CI): remove portal build cache

tujoworker:
It seems to me that the `Build portal` step is not slower by removing the cache. So we probably should do that so we avoid the CSS file import issues, like [here](https://github.com/dnbexperience/eufemia/actions/runs/10559105350/job/29249868805).


langz (approved):
Looks good to me 😃 
Does this mean that we should also remove the key in our project/repo  settings?

tujoworker:
We use `CACHE_VERSION` for dependency caching. I think it's still worth to have it there 👍

PR #3874, opened 2024-08-27
fix(Forms): enhance `Field.Email` validation pattern

tujoworker:
See the tests for more details about what we now allow. In short, its not possible to have the perfect validation. Here is an [article about the challenge](https://www.abstractapi.com/guides/email-validation/email-address-pattern-validation). 
The Regex gets way larger when we try to validate IPv4 and especially IPv6. But I think we don't need a perfect validation for these, as these are exceptions anyways.


langz (approved):
Great comments in the code, and tests 🚀

PR #3895, opened 2024-09-04
fix(Forms): provide `connectWithPath` in the change validator and onBlurValidator

tujoworker:
This feature allows a validator for a field to connect to another data path. Changes to the data path value will be reflected in a rerender of the validator. But only if the validator is actually in use. This can be controlled with field properties, such as `validateUnchanged` etc.

**Good to know:** Why do validators not rerender on every data change? Because field A should not emit the validator of field B. It can be used for various cases, such as backend requests.

But with the new feature called `connectWithPath` you as a dev get the possibility to control this in a fine granted way.

Quick examples:

```tsx
import { Form, Field } from '@dnb/eufemia/extensions/forms'

const validator = (value, { connectWithPath }) => {
  const { getValue } = connectWithPath('/myReference')
  const amount = getValue()
  if (amount >= value) {
    return new Error(`The amount should be greater than ${amount}`)
  }
}

render(
  <Form.Handler>
    <Field.Number path="/myReference" defaultValue={2} />

    <Field.Number
      path="/withValidator"
      defaultValue={2}
      validator={validator} // NB: You may use "onBlurValidator" depending on use case.

      //
    />
  </Form.Handler>,
)
```

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/error-messages/info.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/error-messages/info.mdx
@@ -24,6 +24,20 @@ const validator = (value) => {
 render(<Field.PhoneNumber validator={validator} />)
 ```
 
+## Reuse existing error messages in a validator function
+
+You can reuse existing error messages in a validator function. The types of error messages available depend on the field type.
+
+For example, you can reuse the `required` error message in a validator function:
+
+```tsx
+const validator = (value, { errorMessages }) => {
+  // Your validation logic
+  return new Error(errorMessages.required)
+}
+render(<Field.String validator={validator} />)
+```
+
 ### FormError object
 
 You can use the JavaScript `Error` object to display a custom error message:
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/feature-fields/NationalIdentityNumber/Examples.tsx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/feature-fields/NationalIdentityNumber/Examples.tsx
@@ -141,7 +141,7 @@ export const ValidationFunction = () => {
         const fnr = (value: string) =>
           value.length >= 11 ? { status: 'valid' } : { status: 'invalid' }
 
-        const validator = (value, errorMessages) => {
+        const validator = (value, { errorMessages }) => {
           const result = fnr(value)
           return result.status === 'invalid'
             ? new Error(errorMessages.pattern)
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/getting-started.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/getting-started.mdx
@@ -355,25 +355,25 @@ const schema = {
 The `onBlurValidator` and `validator` properties accepts a function that takes the current value of the field as an argument and returns an error message if the value is invalid:
 
 ```tsx
-const validator = (value) => {
+const onChangeValidator = (value) => {
   const isInvalid = new RegExp('Your RegExp').test(value)
   if (isInvalid) {
     return new Error('Invalid value message')
   }
 }
-render(<Field.PhoneNumber validator={validator} />)
+render(<Field.PhoneNumber validator={onChangeValidator} />)
 ```
 
 You can find more info about error messages in the [Error messages](/uilib/extensions/forms/Form/error-messages/) docs.
 
 ##### Connect with another field
 
-You can also use the `connectWithPath` function to connect the validator to another field. This allows you to rerender the validator function once the value of the connected field changes:
+You can also use the `connectWithPath` function to connect the validator to another field. This allows you to rerun the validator function once the value of the connected field changes:
 
 ```tsx
 import { Form, Field } from '@dnb/eufemia/extensions/forms'
 
-const validator = (value, { connectWithPath }) => {
+const onChangeValidator = (value, { connectWithPath }) => {
   const { getValue } = connectWithPath('/myReference')
   const amount = getValue()
   if (amount >= value) {
@@ -388,21 +388,19 @@ render(
     <Field.Number
       path="/withValidator"
       defaultValue={2}
-      validator={validator} // NB: You may use "onBlurValidator" depending on use case.
-
-      //
+      validator={onChangeValidator} // NB: You may use "onBlurValidator" depending on use case.
     />
   </Form.Handler>,
 )
 ```
 
-By default, the validator function will only run when the "/withValidator" field is changed (and blurred). When the error message is shown, it will update the error message with the new value of the "/myReference" field.
+By default, the validator function will only run when the "/withValidator" field is changed. When the error message is shown, it will update the message with the new value of the "/myReference" field.
 
-You can also change this behaviour by using the following props:
+You can also change this behavior by using the following props:
 
-- `validateInitially` Will run the validator initially.
-- `continuousValidation` Will run the validator on every change, including when the connected field changes.
-- `validateUnchanged` Is a combination of `validateInitially` and `continuousValidation`.
+- `validateInitially` will run the validation initially.
+- `continuousValidation` will run the validation on every change, including when the connected field changes.
+- `validateUnchanged` will validate without any changes made by the user, including when the connected field changes.
 
 ##### Async validation
 
@@ -411,7 +409,7 @@ Async validation is also supported. The validator function can return a promise 
 In this example we use `onBlurValidator` to only validate the field when the user leaves the field:
 
 ```tsx
-const validator = async (value) => {
+const onChangeValidator = async (value) => {
   try {
     const isInvalid = await makeRequest(value)
     if (isInvalid) {
@@ -421,7 +419,7 @@ const validator = async (value) => {
     return error
   }
 }
-render(<Field.PhoneNumber onBlurValidator={validator} />)
+render(<Field.PhoneNumber onBlurValidator={onChangeValidator} />)
 ```
 
 ##### Async validator with debounce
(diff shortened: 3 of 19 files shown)


packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/getting-started.mdx
```diff
@@ -350,9 +350,9 @@ const schema = {
 <Field.PhoneNumber schema={schema} />
 ```
 
-#### validator
+#### onBlurValidator and validator
 
-The `validator` (including `onBlurValidator`) property is a function that takes the current value of the field as an argument and returns an error message if the value is invalid:
+The `onBlurValidator` and `validator` accepts a function that takes the current value of the field as an argument and returns an error message if the value is invalid:
```
langz, line 355:
```suggestion
The `onBlurValidator` and `validator` properties accepts a function that takes the current value of the field as an argument and returns an error message if the value is invalid:
```

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/getting-started.mdx
```diff
@@ -366,6 +366,44 @@ render(<Field.PhoneNumber validator={validator} />)
...
+}
+
+render(
+  <Form.Handler>
+    <Field.Number path="/myRefference" defaultValue={2} />
+
+    <Field.Number
+      path="/withValidator"
+      defaultValue={2}
+      validator={validator} // NB: You may use "onBlurValidator"
```
langz, line 391:
```suggestion
      validator={validator} // NB: You may use "onBlurValidator" depending on your use case.
```

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/getting-started.mdx
```diff
@@ -366,6 +366,44 @@ render(<Field.PhoneNumber validator={validator} />)
...
+  const { getValue } = connectWithPath('/myReference')
+  const amount = getValue()
+  if (amount >= value) {
+    return new Error(`The amount should be greater than ${amount}`)
+  }
+}
+
+render(
+  <Form.Handler>
+    <Field.Number path="/myRefference" defaultValue={2} />
```
langz, line 386:
```suggestion
    <Field.Number path="/myReference" defaultValue={2} />
```

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/getting-started.mdx
```diff
@@ -366,6 +366,44 @@ render(<Field.PhoneNumber validator={validator} />)
...
+      defaultValue={2}
+      validator={validator} // NB: You may use "onBlurValidator" depending on use case.
+
+      //
+    />
+  </Form.Handler>,
+)
+```
+
+By default, the validator function will only run when the "/withValidator" field is changed (and blurred). When the error message is shown, it will update the error message with the new value of the "/myRefference" field.
```
langz, line 399:
```suggestion
By default, the validator function will only run when the "/withValidator" field is changed (and blurred). When the error message is shown, it will update the error message with the new value of the "/myReference" field.
```

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/getting-started.mdx
```diff
@@ -366,6 +366,44 @@ render(<Field.PhoneNumber validator={validator} />)
...
+)
+```
+
+By default, the validator function will only run when the "/withValidator" field is changed (and blurred). When the error message is shown, it will update the error message with the new value of the "/myReference" field.
+
+You can also change this behaviour by using the following props:
+
+- `validateInitially` Will run the validator initially
+- `continuousValidation` Will run the validator on every change, including when the connected field changes
+- `validateUnchanged` Is a combination of validateInitially and continuousValidation
```
langz, line 405:
```suggestion
- `validateUnchanged` Is a combination of `validateInitially` and `continuousValidation`.
```

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/getting-started.mdx
```diff
@@ -366,6 +366,44 @@ render(<Field.PhoneNumber validator={validator} />)
...
+  </Form.Handler>,
+)
+```
+
+By default, the validator function will only run when the "/withValidator" field is changed (and blurred). When the error message is shown, it will update the error message with the new value of the "/myReference" field.
+
+You can also change this behaviour by using the following props:
+
+- `validateInitially` Will run the validator initially
+- `continuousValidation` Will run the validator on every change, including when the connected field changes
```
langz, line 404:
```suggestion
- `continuousValidation` Will run the validator on every change, including when the connected field changes.
```

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/getting-started.mdx
```diff
@@ -366,6 +366,44 @@ render(<Field.PhoneNumber validator={validator} />)
...
+    />
+  </Form.Handler>,
+)
+```
+
+By default, the validator function will only run when the "/withValidator" field is changed (and blurred). When the error message is shown, it will update the error message with the new value of the "/myReference" field.
+
+You can also change this behaviour by using the following props:
+
+- `validateInitially` Will run the validator initially
```
langz, line 403:
```suggestion
- `validateInitially` Will run the validator initially.
```

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/error-messages/info.mdx
```diff
@@ -24,6 +24,20 @@ const validator = (value) => {
 render(<Field.PhoneNumber validator={validator} />)
 ```
 
+## Reuse existing error messages in a validator function
+
+You can reuse existing error messages in a validator function. What kind of error messages are available depends on the field type.
```
langz, line 29:
```suggestion
You can reuse existing error messages in a validator function. The types of error messages available depend on the field type.
```

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/getting-started.mdx
```diff
@@ -366,6 +366,44 @@ render(<Field.PhoneNumber validator={validator} />)
 
 You can find more info about error messages in the [Error messages](/uilib/extensions/forms/Form/error-messages/) docs.
 
+##### Connect with another field
+
+You can also use the `connectWithPath` function to connect the validator to another field. This allows you to rerender the validator function once the value of the connected field changes:
```
langz, line 371:
```suggestion
You can also use the `connectWithPath` function to connect the validator to another field. This allows you to render the validation function when the value of the connected field changes:
```

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/feature-fields/NationalIdentityNumber/Examples.tsx
```diff
@@ -141,7 +141,7 @@ export const ValidationFunction = () => {
         const fnr = (value: string) =>
           value.length >= 11 ? { status: 'valid' } : { status: 'invalid' }
 
-        const validator = (value, errorMessages) => {
+        const validator = (value, { errorMessages }) => {
```
langz, line 144: I just tried out this [example](https://eufemia-git-fix-refresh-validator-on-rerender-eufemia.vercel.app/uilib/extensions/forms/feature-fields/NationalIdentityNumber/#validation-function), but was not able to see the error messages errorFnr and errorDnr in `errorMessages`:

image

Is it just because it's not rebased with latest main? Or is there only some error messages that's included, and not all?

langz (approved):
🍦

joakbjerk (approved):

packages/dnb-eufemia/src/extensions/forms/hooks/useFieldProps.ts
```diff
@@ -507,24 +600,18 @@ export default function useFieldProps<Value, EmptyValue, Props>(
...
 
-    const opts = {
-      ...contextErrorMessages,
-      ...errorMessagesRef.current,
-    }
-
     const tmpValue = valueRef.current
-
-    // Run async regardless to support Promise based validators
-    const result = await validatorRef.current(valueRef.current, opts)
```
kimroen, line 518: I'm not sure what our stance is on breaking changes for extensions like `forms` now that it is "out of beta", but I think this change to the second argument for the validator function counts like one so I just wanted to bring it up just in case.

`validator` used to take two arguments: the current value and an object with error messages.

Now it instead takes the current value and then an object with extra tools, one of which is the error messages object.

 Of course, Typescript should help the users notice this so it might not matter too much.

kimroen: Fair! So undocumented is okay to break then - is that the rule of thumb?

We're using those error messages in our project, but I'm not sure how it was discovered 😅 Maybe just looking at the types?

To be clear—I don't mind that much, it's probably fine 😊

tujoworker: > is that the rule of thumb

Yes. But I'll take another look if I can find something. Else we need to use a third argument. But I like the objects better. Because then you can use the connector without a dummy var in the middle.

kimroen: > Else we need to use a third argument. But I like the objects better. Because then you can use the connector without a dummy var in the middle.

Completely agree! I also prefer an object as the second argument here.

If we cannot detect or accommodate the existing and new behavior at the same time (and mark the old as deprecated), then I vote for breaking it 😅

kimroen: We could include the translation keys in the object and mark them as deprecated like in this example:

CleanShot 2024-09-04 at 21 05 47@2x

However, since the second argument is marked as `| undefined`, people could have code that assumes the error messages are missing if it isn't there, and that wouldn't be true any more. I have no idea what the use case for that would be though.

tujoworker: Yes, we could do that. I'll take a look. Bu its not that easy, as they are dynamic. Its the same `errorMessages` you send in to the field. We have used that internally. I took another look and could not find any docs about it. Did search with this `alidator=\{\(.+,` and this `alidator = \(.+,` regex.

PR #3896, opened 2024-09-04
fix(Breakpoint): ensure matching breakpoint ranges in mixins with docs & hooks

langz:
### The problem/challenge
The [docs](https://eufemia.dnb.no/uilib/layout/media-queries/#breakpoint-ranges) says that:
- `isSmall` should correspond to `allBelow(small)`
- `isMedium` should correspond to `allBetween(small, medium)`
- `isLarge` should correspond to `allAbove(medium)`

This is not how it works today, as the mixins doesn’t match the hooks. 

That the mixins should match the hooks, is stated in both [the table](https://eufemia.dnb.no/uilib/layout/media-queries/#breakpoint-ranges), and  [the image/visualization ](https://eufemia.dnb.no/static/breakpoint-ranges-fae719ed85bf0a5499e92e733e9ad01e.svg): 
image

### [CSB demonstrating how it works today](https://codesandbox.io/p/sandbox/suspicious-bell-w49j63). Please test it out by changing window size in the [full view](https://w49j63.csb.app/) .
Observations: 
- At 960px `isMedium` is `true`, but `allBetween(small, medium)` is `false`. Based on the docs , at 960px, they both should be `true`.
- At 960px `isLarge` is `false`(which is correct), but `allAbove(medium)` is `true`. Based on the docs, at 960px they, both should be `false`.

Screenshot from the CSB above, at 960px:
image

### [CSB demonstrating how it works after this fix](https://codesandbox.io/p/sandbox/sleepy-monad-clcx58?workspaceId=9eb653cb-4406-4b8f-bc96-714927cc7fcd). Please test it out by changing window size in the [full view](https://clcx58.csb.app/).

Screenshot from the CSB above, at 960px:
image

Screenshot from the CSB above, at 961px:
image

### Discussion

Does this fix seem appropriate, something we want? Or should we rather just add new mixins that match the hook? Like isSmall, isMedium, isLarge?

If we are going with the fix, is it just a fix? Or a breaking change?

### General thoughts

I like the functionality of the mixins `allBelow`, `allAbove`, and `allBetween`. But when they actual kick in(is valid) both before this PR and in this PR is to me difficult to understand. When small is `40em`(640px) then to me it doesn't make sense that `allBelow(small) `is valid at 640px, as 640px is not below 640px. 639px is below 640px. Same goes for the other mixins. But I guess fixing or improving upon this is something else than this PR.


tujoworker:
The offset screenshot tests should be aligned. Probably get a higher offset, so it matches the outcome from before.

tujoworker (approved):
Else it looks good 👏

langz:
I should take a second look at this screenshot test.

[...snapshots__/media-have-to-match-media-offset-for-below-ismedium-offset.snap.png](https://github.com/dnbexperience/eufemia/pull/3896/files#diff-7e17543a65d5830a29a493a20cc9a843ab05aeaab6e4575e13b5c95aeaae51dd)

PR #3898, opened 2024-09-05
fix(LabelDescription): nothing was returned

langz:
Motivation from [this Slack thread](https://dnb-it.slack.com/archives/CMXABCHEY/p1725367722056839).

Importing/using `<Field.SelectCountry />` and/or `<Field.PhoneNumber />` from `import { Field } from '@dnb/eufemia/extensions/forms'` with eufemia 10.46.0 and react 17.0.2 results in the following error: `LabelDescription(...): Nothing was returned from render. This usually means a return statement is missing. Or, to render nothing, return null.`, see [CSB](https://codesandbox.io/p/sandbox/new-cookies-82klmm?workspaceId=9eb653cb-4406-4b8f-bc96-714927cc7fcd)

Here's the [same CSB with the fix in this PR](https://codesandbox.io/p/sandbox/elegant-http-2kc6y5?workspaceId=9eb653cb-4406-4b8f-bc96-714927cc7fcd).

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/FieldBlock/FieldBlock.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/FieldBlock/FieldBlock.tsx
@@ -519,7 +519,10 @@ function CombineMessages({
 
 function LabelDescription({ labelDescription, children }) {
   if (!labelDescription) {
-    return children
+    if (children) {
+      return children
+    }
+    return null
   }
   return <div className="dnb-forms-field-block__label">{children}</div>
 }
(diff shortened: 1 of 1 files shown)


packages/dnb-eufemia/src/extensions/forms/FieldBlock/FieldBlock.tsx
```diff
@@ -519,7 +519,10 @@ function CombineMessages({
 
 function LabelDescription({ labelDescription, children }) {
   if (!labelDescription) {
-    return children
+    if (children) {
+      return children
+    }
+    return null
```
tujoworker, line 525:
```suggestion
    return children ?? null
```

tujoworker (approved):
I would prefer the nullish check here. This way it is still possible to actually return `0` or `false` or even an empty string `''` – but not with the if check.

Issue #3894, opened 2024-09-04
Country list is not complete per ISO 3166-1 alpha-2


andlbrei:
## 🐛 Bug Report

This isn't really a bug I guess, but I think the country list should be based on a standard list of countries, and not a phone country code list.
Our team requires compliance with https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2 and cannot use Eufemia for the country list unless it is updated.
One issue with this approach is that we need to do some research to figure out which phone country code a country has for the countries that are in ISO 3166 alpha-2 but not in the phone country code list. Most likely these countries are autonomous regions of other countries.

I will make an MR with a ISO 3166-1 alpha-2 compliant list of countries.

PR #3899, opened 2024-09-05
feat(countries): Added missing countries from ISO 3166 #3894

andlbrei:
The reasoning for adding these countries is that they are required for compliance reasons for application in some of DNBs solutions.

The country list in Eufemia serves two purposes in letting users select countries and letting users select the dialing code for a country.
The country codes missing from ISO 3166 are mostly for external territories of other states.
The dialing code for some of them are certain, and are contained in the wikipedia page the country list was based on earlier.
For others the dialing code is not certain, and most likely they do not have a dialing code. These countries are mostly research bases, or otherwise not permanently inhabited.

The likelyhood of any of these countries actually being used is probably very small.

In the event that someone should want to use a number from one of the uncertain dialing codes, they will most likely identify that the dialing code is wrong. As far as I can tell each of these countries belong to a group of islands or a territory that they share a dialing code with, and we can probably assume that the users of these dialing codes knows which other country to look for in the list.

Added the following countries:
- AX	Åland Islands - Shared dialing code with Finland
- BV	Bouvet Island - Most likely no dialing code, uses the dialing code of Norway
- TF	        French Southern Territories - Most likely no dialing code, uses the dialing code of Mayoette and Reunion
- HM	Heard Island and McDonald Islands - Same dialing code as Australian External Territories
- MQ	Martinique
- NF	Norfolk Island - Same dialing code as Australian External Territories
- GS	South Georgia and the South Sandwich Islands - Same dialing code as Falkland Islands
- UM	United States Minor Outlying Islands - Most likely no dialing code, uses the dialing code of USA
- BQ	Bonaire, Sint Eustatius and Saba

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber/__tests__/PhoneNumber.test.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber/__tests__/PhoneNumber.test.tsx
@@ -289,7 +289,7 @@ describe('Field.PhoneNumber', () => {
       keyCode: 13,
     })
 
-    expect(europe).toHaveLength(51)
+    expect(europe).toHaveLength(52)
   })
 
   it('should return correct value onChange event', async () => {
--- a/packages/dnb-eufemia/src/extensions/forms/constants/countries.ts
+++ b/packages/dnb-eufemia/src/extensions/forms/constants/countries.ts
@@ -1,5 +1,8 @@
 /**
- * Source: Wikipedia's official list of phone number country codes
+ * Sources:
+ * - https://en.wikipedia.org/wiki/List_of_country_calling_codes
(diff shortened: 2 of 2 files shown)


tujoworker (approved):
Nice 🤩 Looks good to me 🫶

tujoworker:
The `PhoneNumber.test.tsx` test needs to be adjusted with the amount for countries it finds with the filter functionality.

andlbrei:
In relation, I notice that the list contains 253 countries, while ISO 3166 consists of 249 official countries.
XK (Kosovo) is one code that is needed for compliance, which is a so called "User-assigned" code in ISO 3166, and is contained in Eufemia. This makes 250 countries, which is what my department needs for compliance.

I have not checked which extra countries are contained, other than Kosovo.

PR #3901, opened 2024-09-05
docs(useTheme): updates docs to match TS type definitions

langz:
Motivation from this [Slack thread](https://dnb-it.slack.com/archives/CMXABCHEY/p1725533657841919).

The follow code:

```
  const { isSbanken } = useTheme()
  console.log(isSbanken)
```

Results in this error message:

`Property 'isSbanken' does not exist on type 'UseThemeReturn'.typescript(2339)`

[CSB displaying the error.](https://codesandbox.io/p/sandbox/silent-star-67jztx)

This is because of the type definition introduced in this PR https://github.com/dnbexperience/eufemia/pull/3861/files

This PR types just the `object`, and not `object | null`(which it was), since all fields in JavaScript (and in TypeScript) can have the value null or undefined? So by just typing the `object`, it could also be `null` instead of the `object`?

Here's a [CSB with the fix](https://codesandbox.io/p/sandbox/cocky-rumple-5hdk2m?workspaceId=9eb653cb-4406-4b8f-bc96-714927cc7fcd).

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/usage/customisation/theming/theme/info.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/usage/customisation/theming/theme/info.mdx
@@ -12,7 +12,7 @@ The Theme component is a helper component that lets you create nested theming so
 import { Theme, useTheme } from '@dnb/eufemia/shared'
 
 const Component = () => {
-  const { name } = useTheme()
+  const themeName = useTheme()?.name
   return 'My Component'
 }
 
@@ -58,7 +58,8 @@ In order to change or map CSS properties, you can make use of the `propMapping` 
 import { Theme, useTheme } from '@dnb/eufemia/shared'
 
 const Component = () => {
-  const { name, propMapping } = useTheme()
+  const theme = useTheme()
+  const { name, propMapping } = theme || {}
   return 'My Component'
 }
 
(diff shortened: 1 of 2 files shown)


tujoworker:
The challenge is that we, as of now, actually return `null` when the context not exit. Therefore `const { isSbanken } = useTheme()` should result in a runtime error. When providing a fallback `useTheme() || {}` everything should be fine.

Other option is to always return an object.

langz:
Aha, I see, yeah, I'm not exactly sure what's the best thing to do here. This PR is just an attempt at fixing it, and highlighting that it's not working 100% in TS as of today

tujoworker:
We can either update the docs with `useTheme() || {}` or check again with @ahmetacer5 if an empty object would be ok?

ahmetacer5:
> We can either update the docs with `useTheme() || {}` or check again with @ahmetacer5 if an empty object would be ok?

This change is going to revert the previous change. I think we should keep the previous change.

Other solution is to make useTheme hook throw error when condition https://github.com/dnbexperience/eufemia/pull/3901/files#diff-48cd736dcf2c35fc9dd697f9820b65d849cf2e718ba1d73ee0472f52cf9cfeb2R19 is false.

ahmetacer5:
> > We can either update the docs with `useTheme() || {}` or check again with @ahmetacer5 if an empty object would be ok?
> 
> This change is going to revert the previous change. I think we should keep the previous change.
> 
> Other solution is to make useTheme hook throw error when condition https://github.com/dnbexperience/eufemia/pull/3901/files#diff-48cd736dcf2c35fc9dd697f9820b65d849cf2e718ba1d73ee0472f52cf9cfeb2R19 is false. This will alert developer that ThemeProvider should be provided during the developement.

PR #3902, opened 2024-09-05
docs: add component translations table to properties docs

langz:


packages/dnb-design-system-portal/src/docs/uilib/components/global-error/properties.mdx
```diff
@@ -16,3 +18,7 @@ showTabs: true
 | `skeleton` | _(optional)_ applies a loading skeleton to all included elements.                                                                |
 | `locale`   | _(optional)_ set a [supported locale](/uilib/usage/customisation/localization/) if needed.                                       |
 | `children` | _(optional)_ to display additional related content, like useful links etc.                                                       |
+
+## Translations
+
+<TranslationsTable localeKey="GlobalError" />
```
langz, line 24: Any ideas on how I can improve upon this one?
The GlobalError translation is a bit different structured than other translations, with nested objects:
```
    GlobalError: {
      404: {
        title: "We can't find the page you're looking for …",
        text: 'Are you sure you have entered the correct address? Or have we messed with the links?',
      },
      500: {
        title: 'Sorry, a technical error happened!',
        text: 'The service is not working properly at the moment. Try again later.',
      },
      code: 'Error code:',
      help: 'Here are some links that might help:',
    },
```

This now looks like this:
Screenshot 2024-09-05 at 21 13 26

Can also be viewed in the deploy preview.

tujoworker: Try something like that:

```tsx
<TranslationsTable
  localeKey={[
    'GlobalError.404',
    'GlobalError.500',
    'GlobalError.code',
    'GlobalError.help',
  ]}
/>
```

Its not that nice, and needs custom support when they change. But we could make some magic inside the `TranslationsTable`, and check for nested object and destruct them. I think we already do it for one lever, do we?

tujoworker (approved):
Nice 🤩

packages/dnb-design-system-portal/src/docs/uilib/components/global-error/properties.mdx
langz: Thanks for the suggestion, that resulted in the same result.

I quickly tried to do some magic inside TranslationsTable, not the best code, but it does the job.

langz: Now looking like:

image

PR #3911, opened 2024-09-09
fix(Forms): ensure label supports HTML formatting

tujoworker:
Here is a [demo](https://eufemia-git-fix-field-block-label-html-eufemia.vercel.app/uilib/extensions/forms/Form/Handler/#complex-async-autosave-example) of this feature (without HTML).

And here is a quick screen recording where the label has HTML inside:

https://github.com/user-attachments/assets/56fdc553-b78c-4b88-9de3-fb696c710c68


langz (approved):
🏁

PR #3912, opened 2024-09-09
chore(Forms): add demos to showcase exclusive to Field.Currency and Field.Number

tujoworker:
The motivation is that these two fields are very often used, and I saw a dev using the `onChange` event for throwing an error, instead of using the `valdiator` prop.


langz (approved):
Looks good to me, but shouldn't the app/user of this code rather use the maximum property? Or is that an other use case 🤔

PR #3917, opened 2024-09-10
fix(Forms): rename "Hviterussland" to "Belarus"

boyum:
Ever since May 2022, the Norwegian Ministry of Foreign Affairs has decided to use the name "Belarus" for Belarus [(source)](https://www.regjeringen.no/no/aktuelt/hviterussland-blir-til-belarus/id2916338/). Unless DNB has made a executive decision to keep the name "Hviterussland", I say we should change it.


tujoworker (approved):
Thank you so much 👏

langz (approved):

PR #3921, opened 2024-09-11
feat(Forms): show optional label when a field uses `required={false}` and add `labelSuffix` prop to each field

tujoworker:
**Quick example:**

```tsx
<Form.Handler required>
    <Field.Email path="/email" required={false} />
</Form.Handler>
```

```tsx

<Field.Email path="/email" labelSuffix="(recommended)" />

```

👉 **PostalCodeAndCity:**
Screenshot 2024-09-11 at 07 03 25

👉 **PhoneNumber:**
Screenshot 2024-09-11 at 07 04 50

👉 **Screenshot from the new demo in the Form.Handler:**

Screenshot 2024-09-11 at 06 59 25


packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/getting-started.mdx
```diff
@@ -329,7 +329,25 @@ The `required` property is a boolean that indicates whether the field is require
...
+```tsx
+<Form.Handler required>
+  <Field.PhoneNumber />
+  <Field.String />
+</Form.Handler>
+```
+
+#### optional
+
+The `optional` property adds "(optional)" to the field label and ensures the field is not required, even if all fields in the form are made required by inheriting the required property from a component like [Form.Handler](/uilib/extensions/forms/Form/Handler/).
```
langz, line 343:
```suggestion
The `optional` property adds "(optional)" to the field label and ensures the field is not required. Even if all fields in the form are made required by inheriting the required property from a component, like [Form.Handler](/uilib/extensions/forms/Form/Handler/), the optional field will still not be required. 
```

packages/dnb-eufemia/src/extensions/forms/FieldBlock/FieldBlock.tsx
```diff
@@ -62,6 +62,8 @@ export type Props = Pick<
   asFieldset?: boolean
   /** use `true` to make the label only readable by screen readers. */
   labelSrOnly?: boolean
+  /** use `true` to add an optional label to the field. */
+  optional?: boolean
```
langz, line 66: To me, the optional property does not match my initial thoughts when reading/seeing it. 
Isn't all fields that are not required optional?

In my head, these two components would be the same:

<Field />
<Field optional />

The opposite would be
<Field required />

tujoworker: Yes, so maybe rename it to `optionalLabel={true}` / `optionalLabel="(custom)"`?

joakbjerk: So going by the naming, this label is just optional for you to add if you want to, and it's not really tied to the field being optional or not?

tujoworker: Yes, then a dev would still need to back out of the required state with `<Field.String required={false} optionalLabel />`.

I fear that devs would "forget" to add "optionalLabel", because when its just a "visual" change, how do you know that you should add it?

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Handler/demos.mdx
```diff
@@ -6,6 +6,14 @@ import * as Examples from './Examples'
 
 ## Demos
 
+### Required and Optional Fields
+
+To make all fields required, set the `required` prop on the `Form.Handler` component.
+
+For fields that should remain optional, use `required={false}` prop on the specific field, which will add "(optional)" to the label.
```
langz, line 13:
```suggestion
For fields that should remain optional, use `required={false}` prop on the specific field. When doing so, it will append "(optional)" to the optional field's label (`labelSuffix`).
```

packages/dnb-eufemia/src/extensions/forms/hooks/DataValueDocs.ts
```diff
@@ -42,10 +42,15 @@ export const dataValueProperties: PropertiesTableProps = {
     status: 'optional',
   },
   required: {
-    doc: 'When set `true`, the field will give an error if the value cannot be empty.',
+    doc: 'When set `true`, the field will give an error if the value fails the required validation. When set to `false`, the field will not be required, but will add a "(optional)" suffix to the label.',
     type: 'boolean',
     status: 'optional',
   },
+  labelSuffix: {
+    doc: 'Will add an additional text to the label, like "(optional)" or "(recommended)".',
```
langz, line 50: Should we mention something about why they should use the labelSuffix property, instead of just using the label?
You had a great point in terms of it not being included in a value that uses inheritLabel?
Not sure if this should be here in the doc property, or if we should make an example/demo which highlights/demos this.

packages/dnb-eufemia/src/extensions/forms/constants/locales/en-GB.ts
```diff
@@ -9,6 +9,7 @@ export default {
       errorSummary: 'Please correct the following errors:',
       errorRequired: 'This field is required.',
       errorPattern: 'The value is invalid.',
+      optionalLabel: '(optional)',
```
langz, line 12:
```suggestion
      optionalLabelSuffix: '(optional)',
```

packages/dnb-eufemia/src/extensions/forms/constants/locales/nb-NO.ts
```diff
@@ -9,6 +9,7 @@ export default {
       errorSummary: 'Feil som må rettes:',
       errorRequired: 'Dette feltet må fylles ut.',
       errorPattern: 'Verdien er ugyldig.',
+      optionalLabel: '(valgfritt)',
```
langz, line 12:
```suggestion
      optionalLabelSuffix: '(valgfritt)',
```

langz (commented):
If some of our users already have done the following in their existing code today:
`<Field.Email label="e-post (valgfritt)" required={false} />`
Which would result in:
image



Then the same code after this feature will result in:
`<Field.Email label="e-post (valgfritt)" required={false} />`
image

Is this "breaking"? Or at least it could be worth making a post about in Eufemia Web channel or so.

packages/dnb-eufemia/src/extensions/forms/hooks/DataValueDocs.ts
tujoworker: yeah, sure. Another reason is that UX may potentially change the visual style of this suffix at one point of time. Which is also a good reasons in my book.

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Handler/demos.mdx
```diff
@@ -6,6 +6,14 @@ import * as Examples from './Examples'
 
 ## Demos
 
+### Required and Optional Fields
+
+To make all fields required, set the `required` prop on the `Form.Handler` component.
+
+For fields that should remain optional, use `required={false}` prop on the specific field. When doing so, it will append "(optional)" to the optional field's label (`labelSuffix`).
```
langz, line 13:
```suggestion
For fields that should remain optional, use `required={false}` prop on the specific field. When doing so, it will append "(optional)" to the optional field's label(`labelSuffix`).
```

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/getting-started.mdx
```diff
@@ -329,7 +329,24 @@ The `required` property is a boolean that indicates whether the field is require
...
+**Note:** You can use the `required` property on the [Form.Handler](/uilib/extensions/forms/Form/Handler/) or [Wizard.Step](/uilib/extensions/forms/Wizard/Step/) components ([example](/uilib/extensions/forms/Form/Handler/demos/#required-and-optional-fields)). Additionally, the [Form.Section](/uilib/extensions/forms/Form/Section/) component as well as the [Form.FieldProps](/uilib/extensions/forms/Form/FieldProps/) provider has a `required` property, which will make all nested fields within that section required.
+
+```tsx
+<Form.Handler required>
+  <Field.PhoneNumber />
+  <Field.String />
+</Form.Handler>
+```
+
+When you need to opt-out of the required field validation, you can use the `required={false}` property. This will also add a "(optional)" suffix to the field label (`labelSuffix`).
```
langz, line 341:
```suggestion
When you need to opt-out of the required field validation, you can use the `required={false}` property. This will also add a "(optional)" suffix to the field label(`labelSuffix`).
```

langz:
Shouldn't the following example result in having the label suffix (recommended) and not (valgfritt)?

image

langz:
> Shouldn't the following example result in having the label suffix (recommended) and not (valgfritt)?
> 
> image

Gave it a retest now, and now I could not reproduce it, thanks for improving 💯

tujoworker:
> Is this "breaking"?

The latest commit includes a check and test for this situation and will ensure we don't add the same text twice.

packages/dnb-eufemia/src/extensions/forms/hooks/DataValueDocs.ts
```diff
@@ -42,10 +42,15 @@ export const dataValueProperties: PropertiesTableProps = {
     status: 'optional',
   },
   required: {
-    doc: 'When set `true`, the field will give an error if the value cannot be empty.',
+    doc: 'When set `true`, the field will give an error if the value fails the required validation. When set to `false`, the field will not be required, but will add a "(optional)" suffix to the label.',
     type: 'boolean',
     status: 'optional',
   },
+  labelSuffix: {
+    doc: 'Will add an additional text to the label, like "(optional)". When using `inheritLabel`, the suffix will not be inherited. NB: The visual appearance of the `labelSuffix` may change in the future.',
```
langz, line 50:
```suggestion
    doc: 'Will append an additional text to the label, like "(optional)". When using `inheritLabel`, the suffix will not be inherited. NB: The visual appearance of the `labelSuffix` may change in the future.',
```

packages/dnb-eufemia/src/extensions/forms/types.ts
```diff
@@ -245,6 +245,10 @@ export type FieldBlockProps = {
    * Main label text for the field
    */
   label?: React.ReactNode
+  /**
+   * Will add an additional text to the label, like "(optional)" or "(recommended)"
```
langz, line 249:
```suggestion
   * Will append an additional text to the label, like "(optional)" or "(recommended)"
```

langz:
> > Is this "breaking"?
> 
> The latest commit includes a check and test for this situation and will ensure we don't add the same text twice.

Great, thanks, that sounds smart 👍 
But could it lead to problems? 🤔 

Shouldn't I be able to see the labelSuffix provided in the example below?
`<Value.String label="penger" labelSuffix="er" showEmpty />`
image

packages/dnb-eufemia/src/extensions/forms/hooks/DataValueDocs.ts
```diff
@@ -42,10 +42,15 @@ export const dataValueProperties: PropertiesTableProps = {
     status: 'optional',
   },
   required: {
-    doc: 'When set `true`, the field will give an error if the value cannot be empty.',
+    doc: 'When set `true`, the field will give an error if the value fails the required validation. When set to `false`, the field will not be required, but will add a "(optional)" suffix to the label.',
```
langz, line 45:
```suggestion
    doc: 'When set to `true`, the field will give an error if the value fails the required validation. When set to `false`, the field will not be required, but will add a "(optional)" suffix to the label.',
```

langz (approved):
Looks great(recommended) :) 

One [comment](https://github.com/dnbexperience/eufemia/pull/3921#issuecomment-2346030583) worth considering before merging 🙏

tujoworker:
> Shouldn't I be able to have/see the labelSuffix er provided in the example below, where label is penger?

If we only look for the translation string `optionalLabelSuffix` we lower the possibility of problems very much.

PR #3924, opened 2024-09-12
feat(Wizard): add `preventNavigation` callback function to `onStepChange`

joakbjerk:
- [x] Add example to `Wizard.Container`
- [x] Update docs
- [x] Update tests
- [x] Fix types

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Wizard/Container/info.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Wizard/Container/info.mdx
@@ -140,7 +140,7 @@ const MyForm = () => {
 }
 ```
 
-You can also prevent the user from from navigating to the next or previous step, by using the `preventNavigation` callback function found as the third parameter, in the `onStepChange` event.
+You can also prevent the user from navigating to the next or previous step, by using the `preventNavigation` callback function found as the third parameter, in the `onStepChange` event.
 
 ```tsx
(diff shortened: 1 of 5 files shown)


packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Wizard/Container/info.mdx
```diff
@@ -140,6 +140,41 @@ const MyForm = () => {
 }
 ```
 
+You can also prevent the user from from navigating to the next or previous step, by using the `preventNavigation` callback function found as the third parameter, in the `onStepChange` event.
```
langz, line 143:
```suggestion
You can also prevent the user from navigating to the next or previous step, by using the `preventNavigation` callback function found as the third parameter, in the `onStepChange` event.
```

langz (approved):
Other than the failing unit test, it looks good to me  🎆 🥳 
Maybe @tujoworker also have the time to quickly look through this?

In general I like providing interactive examples instead of just "textual code" as done for this feature in https://eufemia-git-feat-wizard-preventnavigation-eufemia.vercel.app/uilib/extensions/forms/Wizard/Container/#controlling-the-wizard-steps

When it's interactive, I think the users will better understand the use case of the example/feature, it's easier for a reviewer to test if it actually works, you'll write the example/demo as actual code and not just "textual code" which will find potential errors in the example, such as random human coding errors, typesript errors, etc.

Just give it a consideration 🙏

tujoworker (approved):
Nice. Added another test to verify that the validation is run before the preventNavigation

PR #3927, opened 2024-09-12
chore(Forms): enhance the not provided label warning

tujoworker:
Today, we have a warning for when an async validator or onChange event is used, but no label is given. Because we want then show the indicator. But maybe this is not that useful? Then we can consider to remove it.

This PR enhances at least the message given to the user.


langz (approved):
Now I understand ❤️

PR #3929, opened 2024-09-12
fix(HeightAnimation): enhance adjusting the height to 0

tujoworker:
When animating/adjusting the height for some value to 0, it did not. Also animating/adjusting from 0 to a a certain value did not work properly.

From value to 0, first the issue and then with the fix:

https://github.com/user-attachments/assets/316d7288-2a2d-4792-90e1-3d41313602a0

https://github.com/user-attachments/assets/93dc3198-0ffd-4eb3-b5de-b7e6f0d6ac76

From 0 to value, first the issue and then with the fix:

https://github.com/user-attachments/assets/ddd21697-60a8-4555-af5d-a1d3a8b6ebc0

https://github.com/user-attachments/assets/cd97dbca-8a1d-46d1-8fd8-16f7cea85b5a

### DIFF
--- a/packages/dnb-eufemia/src/components/height-animation/HeightAnimationInstance.ts
+++ b/packages/dnb-eufemia/src/components/height-animation/HeightAnimationInstance.ts
@@ -168,7 +168,7 @@ export default class HeightAnimation {
     this.callAnimationEnd()
   }
   getHeight() {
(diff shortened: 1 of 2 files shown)


langz (approved):
Great🫡
Do you want to make a bugfix release for this fix alone, or just wait to release this fix until the next minor release?

PR #3931, opened 2024-09-13
fix(PushContainer): Add missing variant type

andlbrei:
`PushContainer` can use the prop `variant` to change between outline or no outline surrounding the contents.
The prop has not been exposed in the prop type, and using it will result in a typescript error even though it functionally works.


tujoworker (approved):
Nice! Thank you 🫶

langz (approved):

joakbjerk (approved):

PR #3933, opened 2024-09-13
feat(Forms): add `preventChangeValidator` to each field

tujoworker:
Should we even call it "Validator"? It acts very similar and has the same API. But it does also a little more, as it actually prevents the "new value" to be a part of the data set. Something which is not possible with our other validators. But on the other side, it also shows an error. So it "validates" the field. Any thoughts?

Quick example:

```tsx
<Field.SelectCountry
  label="Land du er statsborger i"
  required
  preventChangeValidator={(value, { connectWithPath }) => {
    const { getValue } = connectWithPath('/countries')
    if ((getValue() || []).includes(value)) {
      return new Error(
        'You can not have the same country twice',
      )
    }
  }}
/>
```

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/hooks/DataValueDocs.ts
+++ b/packages/dnb-eufemia/src/extensions/forms/hooks/DataValueDocs.ts
@@ -86,7 +86,7 @@ export const dataValueProperties: PropertiesTableProps = {
     type: 'function',
     status: 'optional',
   },
-  preventInputValidator: {
+  preventChangeValidator: {
     doc: 'Custom validator function (synchronous) that is triggered when the user makes a change to a field (e.g. making a change to a dropdown). When the function returns an error, the field will not store the new value in the data context or call the onChange event. The given error will be shown.  The first parameter is the value, and the second parameter returns an object containing { errorMessages, connectWithPath, validators }.',
     type: 'function',
     status: 'optional',
--- a/packages/dnb-eufemia/src/extensions/forms/hooks/DataValueDocs.ts
+++ b/packages/dnb-eufemia/src/extensions/forms/hooks/DataValueDocs.ts
@@ -87,7 +87,7 @@ export const dataValueProperties: PropertiesTableProps = {
     status: 'optional',
   },
   preventChangeValidator: {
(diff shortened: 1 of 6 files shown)


packages/dnb-eufemia/src/extensions/forms/hooks/DataValueDocs.ts
```diff
@@ -82,7 +82,12 @@ export const dataValueProperties: PropertiesTableProps = {
...
-    doc: 'Custom validator function that is triggered when the user leaves a field (e.g., blurring a text input or closing a dropdown). The function can be either asynchronous or synchronous. The first parameter is the value, and the second parameter returns an object containing { errorMessages, connectWithPath, validators }.',
+    doc: 'Custom validator function that is triggered when the user leaves a field (e.g. blurring a text input or making a change to a dropdown). The function can be either asynchronous or synchronous. The first parameter is the value, and the second parameter returns an object containing { errorMessages, connectWithPath, validators }.',
+    type: 'function',
+    status: 'optional',
+  },
+  preventChangeValidator: {
+    doc: 'Custom validator function (synchronous) that is triggered when the user makes a change to a field (e.g. making a change to a dropdown). When the function returns an error, the field will not store the new value in the data context or call the onChange event. The given error will be shown.  The first parameter is the value, and the second parameter returns an object containing { errorMessages, connectWithPath, validators }.',
```
langz, line 90:
```suggestion
    doc: 'Custom validator function (synchronous) that is triggered when the user makes a change to a field (e.g. making a change to a dropdown). When the function returns an error, the field will not store the new value in the data context or call the onChange event. The given error will be shown. The first parameter is the value, and the second parameter returns an object containing { errorMessages, connectWithPath, validators }.',
```

langz (commented):
Yeah, I'm not 100% sure about the naming of the prop yet. Will try to look further into it on monday, perhaps you can explain the motivation for the feature, and we can discuss possible prop names🤞

langz:
I think we should think about this PR:
- What we want to solve
- How we want to solve it

Here are two scenarios I quickly came across when testing the [newly updated example](https://eufemia-git-feat-prevent-input-validator-eufemia.vercel.app/uilib/extensions/forms/Iterate/Array/demos/#initially-open):


Not being able to submit after removing the duplicate item #1:
https://github.com/user-attachments/assets/8cc315fe-e743-4d56-bf10-b02fbfc4e5f2

Not being able to submit after removing the duplicate item #2:
https://github.com/user-attachments/assets/9f5b3b51-33d5-46f3-a410-f573f737d9d9

tujoworker:
Thank you for the feedback 🫶

I agree, we should not add this validator, because it creates a couple of UX issues you demonstrate, but rather use the `validator` we have on the Iterate.Array.

This PR #3953 adds a duplication check.

PR #3936, opened 2024-09-14
feat(Forms): add `{nextItemNo}` support to Iterate.PushButton

tujoworker:
Quick example:

```tsx
<Form.Handler>
  <Iterate.Array path="/myList">
    <Field.String itemPath="/" />
  </Iterate.Array>

  <Iterate.PushButton
    path="/myList"
    pushValue="push value"
    text="Add no. {nextItemNo}"
  />
</Form.Handler>,
```

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Iterate/PushButton/info.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Iterate/PushButton/info.mdx
@@ -29,7 +29,7 @@ In order to create new items you can also use the [Iterate.PushContainer](/uilib
 
 ## Show the next item number in the button
 
-You can use the `{nextItemNr}` variable in the `text` or `children` property to display the next item number.
+You can use the `{nextItemNo}` variable in the `text` or `children` property to display the next item number.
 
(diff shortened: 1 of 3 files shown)


langz (approved):
Looks good, it probably makes sense that the propname is itemNr even though it displays the itemNr for the next item? 
Is that how the itemNr works elsewhere as well?

Just don't want out consumers to be confused about an itemNr here and one there, and why it does not display the same value...

tujoworker:
Agree, I had nextItemNr before committing as well. 
It can never be prevItemNr. 
So maybe it will be more declarative to name it nextItemNr.

Just a small detail; Should we name it "No" instead of "Nr"? "No" is actually the more correct term in English. So we can start with the new one and change the others incrementally.

langz:
> Agree, I had nextItemNr before committing as well. 
> It can never be prevItemNr. 
> So maybe it will be more declarative to name it nextItemNr.
> 
> Just a small detail; Should we name it "No" instead of "Nr"? "No" is actually the more correct term in English. So we can start with the new one and change the others incrementally. 
> 

Sounds smart, I agree 🫡

langz (approved):

PR #3939, opened 2024-09-15
fix(Forms): enhance container boundary error handling

tujoworker:
This PR adds the "visible error" logic to the field boundary. 

Before it just inherited it from the global data context, which could lead in worst case to error disturbance, similar to what [happened here](https://github.com/dnbexperience/eufemia/pull/3938).


langz (approved):
🫡

PR #3946, opened 2024-09-16
chore: split out ChildrenWithAge into its own docs

langz:
TODO before merging: 

- [x] Check if there's any links that's now outdated(as I've made own Pages for info, demos, properties). 
- [x] Display the translations for this block? Not sure if that's easy, as the translations is sort of "on the side" in the file `ChildrenWithAgeTranslations.ts`, and it’s structured a bit differently than other translations.


tujoworker (approved):
Nice work 💯

Originally, I thought we may make these pages as simple as we can. But I think having the tabs makes these docs so much easier to consume 🫶

langz:
> Nice work 💯
> 
> Originally, I thought we may make these pages as simple as we can. But I think having the tabs makes these docs so much easier to consume 🫶

Yeah, I agree. I also like having the same file structure for stuff when developing. I know that I find demos/examples in the demos.mdx for the given component, etc.

PR #3948, opened 2024-09-16
feat(Forms): add `inheritVisibility` to each `Value.*` component

tujoworker:
This PR adds a new property called `inheritVisibility`. When it is used, it will look for a field with the same path. If it exists, it will sync its visibility to the mounting state of the field.

A following up PR will make the same, but for the submit data.

- [x] We should test this on the ChildrenWithAge block


joakbjerk (approved):
Looks good to me, testing it locally 🙏

Is the idea that developers can filter out invisible data on submit, by using the `isVisible` object property?

langz (approved):
Looks good 👍 

My only thought is that up until this point in [Value](https://eufemia-git-feat-remove-unmounted-fields-data-eufemia.vercel.app/uilib/extensions/forms/Value/String/properties/) we only use the word inherit in the following prop `inheritLabel`, which sort of inherits the value of a prop(label) in the [Field](https://eufemia-git-feat-remove-unmounted-fields-data-eufemia.vercel.app/uilib/extensions/forms/base-fields/String/properties/).

But for this new prop `inheritVisibility`, there's no prop named `visibility` or so in The [Field](https://eufemia-git-feat-remove-unmounted-fields-data-eufemia.vercel.app/uilib/extensions/forms/base-fields/String/properties/).

Not saying the naming `inheritVisibility` is wrong, but based on the existing pattern, it's not the same(doesn't have to be).




In the [example provided](https://eufemia-git-feat-remove-unmounted-fields-data-eufemia.vercel.app/uilib/extensions/forms/Value/demos/#inherit-visibility-from-fields), we use  `<Form.Visibility />`, is that the only way of setting visibility for a given field?

If so, then I think we would be fine by just adding some info to the docs of inheritVisibility, that it only works when wrapped in a <Form.Visibility /> or so...

tujoworker:
> is that the only way of setting visibility for a given field?

No, it supports every possible way. As long as the field "unnmounts". But I recommend to use Visibility, because its declarative, and not imperative like most other ways are. Also, you then have nice options such as `keepInDOM` (if needed) and `animate` 🪄

So yes, we should mention it, but probably not give an explicit example.

I like your reflection about the name vs functionality origin. 
But I think, it will be hard to find any other prop name or another metodologi.

On the other side, the word "inherit" describes clearly what's happening. We could add `inheritFieldVisibility` – but then we would probably also rename `inheritLabel` to `inheritFieldLabel`. And still, it does not explain that the linkage happens through the path.

Other names could be:

- `linkVisibility`
- `syncVisibility`
- `deriveVisibility`
- `inheritPathVisibility`

But for simplicity, I think `inheritVisibility` still stands out.

Any thoughts?

langz:
> > is that the only way of setting visibility for a given field?
> 
> No, it supports every possible way. As long as the field "unnmounts". But I recommend to use Visibility, because its declarative, and not imperative like most other ways are. Also, you then have nice options such as `keepInDOM` (if needed) and `animate` 🪄
> 
> So yes, we should mention it, but probably not give an explicit example.
> 
> I like your reflection about the name vs functionality origin. But I think, it will be hard to find any other prop name or another metodologi.
> 
> On the other side, the word "inherit" describes clearly what's happening. We could add `inheritFieldVisibility` – but then we would probably also rename `inheritLabel` to `inheritFieldLabel`. And still, it does not explain that the linkage happens through the path.
> 
> Other names could be:
> 
> * `linkVisibility`
> * `syncVisibility`
> * `deriveVisibility`
> * `inheritPathVisibility`
> 
> But for simplicity, I think `inheritVisibility` still stands out.
> 
> Any thoughts?

Thanks for your detailed thoughts.
I agree with out 🤝 
Seems like we could help the user by describing what "visibility" actually is, or what affects that 👍

packages/dnb-eufemia/src/extensions/forms/Value/ValueDocs.ts
```diff
@@ -21,6 +21,11 @@ export const ValueProperties: PropertiesTableProps = {
     type: 'boolean',
     status: 'optional',
   },
+  inheritVisibility: {
+    doc: 'Use `true` to inherit the visibility from a field with the same path. You find more info in the [Value section](/uilib/extensions/forms/Value/#inherit-visibility-from-fields-based-on-path).',
```
langz, line 25:
```suggestion
    doc: 'Use `true` to inherit the visibility from a field with the same path. You can find more info in the [Value section](/uilib/extensions/forms/Value/#inherit-visibility-from-fields-based-on-path).',
```

langz:
Thanks, looks very good 🙇

PR #3959, opened 2024-09-18
fix(Forms): ensure `defaultValue` works in React.StrictMode

tujoworker:
This can be experienced e.g. in the ChildrenWithAge block rendered in React.StrictMode. Then non of the default values are selected.

Screenshot 2024-09-18 at 22 02 08


langz (approved):
Great, nice find.
I've experienced it myself, but somehow never thought about why it happened 😂

PR #3963, opened 2024-09-19
chore(Form.Section): remove code duplication

tujoworker:
Switching the mode when there is a submit error happens on two other places already. We don't need this third one.


langz (approved):
🍾

PR #3964, opened 2024-09-19
feat(Forms): add `limit` prop to Iterate.Array

tujoworker:


langz:
I'm just trying out the limit property, by adding `limit={1}`, to a few of our Iterate.Array examples:

https://eufemia-git-feat-iterate-limit-eufemia.vercel.app/uilib/extensions/forms/Iterate/Array/demos/#conditions-using-visibility with `limit={1}`:



https://github.com/user-attachments/assets/15e96f54-d435-4c4b-9243-f3a2053c67c3

packages/dnb-eufemia/src/extensions/forms/Iterate/Array/ArrayDocs.ts
```diff
@@ -12,6 +12,11 @@ export const ArrayProperties: PropertiesTableProps = {
     type: 'string',
     status: 'optional',
   },
+  limit: {
+    doc: 'Limit the number of items to iterate over. Defaults to `undefined`.',
```
langz, line 16: Just me thinking here, "to iterate over" probably makes sense, but could it be "limit the number of visible items"? Or that's probably self-explanatory, since the Iterate.Array is only used to display items? 🤔 

Reason for me thinking about this, is because I don't really feel like the "limit"-property is "limiting" the users sort of. You could think that limit only makes it so that the user could enter a limit of items, like in the following [example](https://eufemia-git-feat-iterate-limit-eufemia.vercel.app/uilib/extensions/forms/Iterate/Array/demos/#with-datacontext-and-addremove-buttons) using `limit={1}` but here you can enter as many as you want, it's only that it's one that's displayed/visible sort of.


https://github.com/user-attachments/assets/a0498ed3-1d01-4aa1-ac12-69819747eaa1




Just rambling thoughts....

tujoworker: Yes, I see. 

It's actually possible letting the push button use/inherit the same limit. We then should show probably a message.

langz: I'm not sure if we have to do this though, and if this is the way to go. This was just my initial thought when first testing.

tujoworker: Glad if you can tests this out 🤩

langz: I'll do it before lunch time 🥫 🥘

langz: Good good, thanks for improving upon this 🚀 

Two thoughts:

- Warning message changes color to be an error message(before it animates out) when removing the one and only item:



https://github.com/user-attachments/assets/5b28e254-6a18-484b-aafd-c61d3612e836

- When setting `limit={1}` to the existing [example](https://eufemia-git-feat-iterate-limit-eufemia.vercel.app/uilib/extensions/forms/Iterate/Array/demos/#with-datacontext-and-addremove-buttons) with two items, the warning message sort of feels a bit strange, as I know that I already have two items in the list, but the message says "Du har nådd grensen på: 1", which is sort of true, hehe.
It's not like we should display an warning message if there's more items in the iterate.array then the limit? 🤔 

https://github.com/user-attachments/assets/85345528-8f73-48bf-ae09-335b67d1868c

langz (approved):

packages/dnb-eufemia/src/extensions/forms/Iterate/Array/ArrayDocs.ts
tujoworker: Yes, the yellow to read is not nice. But it should be solved in the `FormStatus`. It should basically "keep" the state when animating out. I don't think we should fix it in this PR.
The reason is that we remove the warning, which is used to both show and hide but also to change the color. And when its not true anymore, we close it with animation, but the state was then already set to "error", because this is the default state.

PR #3969, opened 2024-09-20
fix(Forms): deprecate and rename `{itemNr}` in favor of `{itemNo}`

tujoworker:


langz (approved):
Sort of breaking, isn't it? But we'll let it slide? Probably not many people using `itemNr` yet?

tujoworker:
Its backwards compatible by using the regex. So should be fine.

PR #3975, opened 2024-09-20
fix(Forms): add support for `transformIn` and `transformOut` to Field.SelectCountry

tujoworker:
Based on PR #3974

Here is a [demo](https://eufemia-git-fix-select-country-transform-eufemia.vercel.app/uilib/extensions/forms/feature-fields/SelectCountry/#transformin-and-transformout).


joakbjerk (approved):
Looks really good to me, from testing on the demo page 🙏

PR #3976, opened 2024-09-20
feat(OrganizationNumber): adds validation

langz:


packages/dnb-eufemia/src/extensions/forms/Field/OrganizationNumber/OrganizationNumber.tsx
```diff
@@ -26,15 +29,54 @@ function OrganizationNumber(props: Props) {
...
+    pattern: props.pattern ?? (validate ? validationPattern : undefined),
+    label: props.label ?? label,
     errorMessages,
     mask,
     width: props.width ?? 'medium',
     inputMode: 'numeric',
+    validator: validate
+      ? props.validator || organizationNumberValidator
+      : undefined,
+    // exportValidators: { organizationNumberValidator },
```
langz, line 79: By adding this line, which I want(need to do), I somehow get the following error message: `Maximum call stack size exceeded error`. Not sure if the error is in my code, or the `exportValidators`.

It happens in the unit tests, and in the examples when I "paste in" a full value(org number), not when typing one digit at the time.

tujoworker: I'll take a look at it.

tujoworker: I think we should use the `onBlurValidator`, not that this fixes the issue. In fact, it did execute the same issue just by writing 123 and bluring the field.

What do you think?

langz: Yeah, I'm fine with that🫡🤝

tujoworker: Yes. Because we don't need to validate on each keystroke. We show the error first on blur anyway.

The tests needs to be aligned then. It needs:

```tsx
fireEvent.blur(document.querySelector('input'))
```

tujoworker: Created a separate PR #3977 that fixes the maximum call stack error issue.

tujoworker (approved):
Very good to get this in 💙

PR #3977, opened 2024-09-20
fix(Forms): make `exportValidators` work with the same validator as given in `validator` or `onBlurValidator`

tujoworker:


langz (approved):
As long as tests for existing components that use this go green, then I'm happy🤞

PR #3979, opened 2024-09-21
feat(Forms): add `filterCountries` prop to Field.SelectCountry to filter out certain countries

tujoworker:


langz (approved):
Looks great🫡
A failing test that we should look into. 

An example/demo would also be Nice🤓

Could this be needed in PhoneNumber as well? We use countries there too🤔

PR #3984, opened 2024-09-23
fix(Field.Toggle): removes invisible label when label not provided

langz:
Fixes https://jira.tech.dnb.no/browse/DCE-12453

Before this PR:
image

After this PR:
https://github.com/dnbexperience/eufemia/pull/3984/files#diff-d4c280bed02d42be9b1a172ed18eea80d2cb23e0a33ad9dd27b88847815d5177

image


packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/base-fields/Toggle/demos.mdx
```diff
@@ -66,6 +66,12 @@ import * as Examples from './Examples'
 
 <Examples.VariantButtonsWithHelp />
 
+#### Buttons without label
```
langz, line 69:
```suggestion
```

joakbjerk (approved):

tujoworker (approved):
I would rather move the logic into the `FieldBlock` itself. Because other fields may suffer from the same. I pushed a commit with tests. 

I could not find any `isFragment` or `fragmentHasChildren`, so added for now to the FieldBlock.

langz:
> I would rather move the logic into the `FieldBlock` itself. Because other fields may suffer from the same. I pushed a commit with tests.
> 
> I could not find any `isFragment` or `fragmentHasChildren`, so added for now to the FieldBlock.

Great, thanks.
I see that the existing fix does not work anymore now, I'll look into it.
Now it looks like this(as it did before the fix):
image

langz:
> > I would rather move the logic into the `FieldBlock` itself. Because other fields may suffer from the same. I pushed a commit with tests.
> > I could not find any `isFragment` or `fragmentHasChildren`, so added for now to the FieldBlock.
> 
> Great, thanks. I see that the existing fix does not work anymore now, I'll look into it. Now it looks like this(as it did before the fix): image

I've added a fix for this now, in the following commit https://github.com/dnbexperience/eufemia/pull/3984/commits/7fa80f515409f51cc562afb17835867c6d66f9bc

Happy with how the fix looks now 🙏

langz:
I also had to update screenshot tests in this commit  https://github.com/dnbexperience/eufemia/pull/3984/commits/68370f6b4f2edc2c3abd7980ea33aea63b47140e

Looks good to me:
image

PR #3985, opened 2024-09-23
feat(SASS mixins): media queries mixins now uses 0.1px instead of 1px to avoid overlap

snorrekim:
The `1px` overlap-avoidance between media queries created a  `0.999...px` gap between mixin media queries. That is visible if user zooms in. This PR changes that to `0.1px` so you would have to zoom in more than 1000% in order to get a gap.

### Example: browser with 200% zoom and 640.5px with 
1. `'small'` breakpoint `40em = 640px
3. `allBelow('small')` `0px to 640px` is narrower than `640.5px`
4. `allAbove('small')` `641px to infinity` is wider than `640.5px`
5. Changing `allAbove('small')` to `640.1px to infinity` includes `640.5px`

### Changes:
* changed overlap-avoidance from `1px` to `0.1px` in both SASS mixins and react hooks.
* updated documentation for "SASS and Mixins" and "Media Queries and Breakpints" to make it clearer that we do not include the minimum breakpoint in our intervals.

### other fixes:
* fixed Portal Home button breakpoint off by 1 px
* fixed a bug where you could not use any other unit than `em` for `allAbove`  and `allBetween` mixin.
* fixed a bug where `$breakpoint-offset`. was not used in `allBetween` mixin.
* fixed a bug with parsing `'and'` in array sizes for `MediaQueryUtils.ts` 
* added a comment that `handheld` is a [deprecated](https://developer.mozilla.org/en-US/docs/Web/CSS/@media#sect2) media-type in 'MediaQueryUtils.ts' (should we remove it?)

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/helpers/sass.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/helpers/sass.mdx
@@ -78,7 +78,7 @@ You can import Eufemia _mixins_ directly into your SCSS styles:
 
 Use the `allAbove`, `allBelow` and `allBetween` mixins to add media queries to your css.
 
-To avoid overlapping media queries we add `0.00625` to the minimum width. This is adds `0.1px` to any `em` size. If you send in a different unit than `em` you will probably need to make adjustments if you want to avoid an overlap (`0.00625px` is too small to avoid an overlap)
+To prevent overlapping media queries, `0.00625em` gets added to the minimum width. This results in an increment of approximately `0.1px` when using `em` units. If you're using a unit other than `em`, you may need to adjust this value accordingly, as `0.00625px` is typically too small to effectively prevent overlaps.
 
 | mixin                   | actual interval (em)   | actual interval (px) |
 | ----------------------- | ---------------------- | -------------------- |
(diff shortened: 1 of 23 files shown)


packages/dnb-design-system-portal/src/docs/uilib/helpers/sass.mdx
```diff
@@ -78,16 +78,27 @@ You can import Eufemia _mixins_ directly into your SCSS styles:
 
 Use the `allAbove`, `allBelow` and `allBetween` mixins to add media queries to your css.
 
+To avoid overlapping media queries we add `0.00625` to the minimum width. This is adds `0.1px` to any `em` size. If you send in a different unit than `em` you will probably need to make adjustments if you want to avoid an overlap (`0.00625px` is too small to avoid an overlap)
```
tujoworker, line 81:
```suggestion
To prevent overlapping media queries, `0.00625em` gets added to the minimum width. This results in an increment of approximately `0.1px` when using `em` units. If you're using a unit other than `em`, you may need to adjust this value accordingly, as `0.00625px` is typically too small to effectively prevent overlaps.
```

packages/dnb-eufemia/src/shared/MediaQueryUtils.ts
```diff
@@ -27,7 +27,7 @@ export type MediaQueryCondition =
       minWidth?: number | string | MediaQuerySizes
       maxWidth?: number | string | MediaQuerySizes
       orientation?: string
-      handheld?: boolean
+      handheld?: boolean // DEPRECATED: this is no longer an accepted media-type. Should we remove it?
```
tujoworker, line 30: Yes, I think so too.

tujoworker (approved):
Very good work. Thank you so much 🤩

(Left some comments here and there. )

PR #3990, opened 2024-09-24
docs: prop -> property, and props -> properties

langz:
It's of course correct to use the word `prop` and `props`.
It's just that we mostly use the words `property` and `properties`, hence why I opened this PR


tujoworker (approved):
wow 🫣

PR #4000, opened 2024-09-25
feat(Forms): implement TypeScript version of `json-pointer`

tujoworker:
I believe this is the right step for now because the dependency we’re using is written in CJS and doesn't treeshake unused parts. Since we rely on this pointer in the default Eufemia Context, reducing its footprint is my main motivation.

By using the native `foreach` method and the export approach in this PR, we can ship smaller bundle (maybe 20%). In the future, we may consider submitting a PR to the original dependency to address this issue directly.

PR no. 4000

### DIFF
--- a/packages/dnb-eufemia/package.json
+++ b/packages/dnb-eufemia/package.json
@@ -113,7 +113,6 @@
     "classnames": "2.3.1",
     "core-js": "3.26.1",
     "date-fns": "2.25.0",
-    "json-pointer": "0.6.2",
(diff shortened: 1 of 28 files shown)


langz (approved):
Wow, big job 👏 
4000 🚀 

Should we consider using https://www.npmjs.com/package/@types/json-pointer as well or wouldn't that make sense?

tujoworker:
Ohh, I was not aware of this! Nice findings 👍 We need to take a look at this 💯

PR #4001, opened 2024-09-25
fix(Forms): `Form.useData` should not throw error when used in Wizard

langz:
As reported here https://dnb-it.slack.com/archives/C07JEJVSHJR/p1727258016481019

TODO:
- [x] The actual fix...


tujoworker (approved):
Nice findings and thank you for the test 🫶

langz:
> Nice findings and thank you for the test 🫶

Thanks to you for the fix(the important part) 🚀

PR #4005, opened 2024-09-26
feat(Forms): add Value.Provider to propagate value props down to Value.* components (e.g. `inheritVisibility`)

tujoworker:
- [x] We may add this instead of #4003 or maybe both. Will have to discuss this in today's meeting.


langz (approved):
Looks good to me, will this also work with `inheritLabel` implemented in https://github.com/dnbexperience/eufemia/pull/4009 ?

tujoworker:
Yes, should work with all value props.

PR #4008, opened 2024-09-26
docs(Forms): improves docs for Value.Number & Value.Currency

langz:
Can be demoed here:
- [Currency](https://eufemia-git-chore-improve-currency-docs-eufemia.vercel.app/uilib/extensions/forms/Value/Currency/properties/)
- [Number](https://eufemia-git-chore-improve-currency-docs-eufemia.vercel.app/uilib/extensions/forms/Value/Number/properties/)

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Value/Number/NumberDocs.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Value/Number/NumberDocs.tsx
@@ -1,6 +1,8 @@
 import { PropertiesTableProps } from '../../../../shared/types'
+import { NumberFormatPropertiesCamelCase } from '../../../../components/number-format/NumberFormatDocs'
 
 export const NumberProperties: PropertiesTableProps = {
+  ...NumberFormatPropertiesCamelCase,
   minimum: {
     doc: 'Defines the minimum value of the rendered number. Defaults to `Number.MIN_SAFE_INTEGER`.',
     type: 'number',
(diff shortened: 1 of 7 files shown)


packages/dnb-eufemia/src/extensions/forms/Value/Number/NumberDocs.tsx
```diff
@@ -1,6 +1,8 @@
 import { PropertiesTableProps } from '../../../../shared/types'
+import { NumberFormatPropertiesCamelCase } from '../../../../components/number-format/NumberFormatDocs'
 
 export const NumberProperties: PropertiesTableProps = {
+  ...NumberFormatPropertiesCamelCase,
```
tujoworker, line 5: Here we need to hand pick what we think should be supported. I think we only should show the "currency" relevant props. And not show e.g. `nin`.

tujoworker (approved):
Nice work 🙏

I think we need to adjust what prop we show for currency.

packages/dnb-eufemia/src/extensions/forms/Value/Number/NumberDocs.tsx
langz: _What we think should be supported?_

Isn't all `Value.Number` properties supported in `Value.Currency`? Especially from a [TS perspective](https://github.com/dnbexperience/eufemia/blob/aac721b63daadb9cf8dbf39baf5fbe7d3717eb1f/packages/dnb-eufemia/src/extensions/forms/Value/Currency/Currency.tsx#L13).

If we want to change the supported properties, shouldn't we change the TS as well as in the property docs?
Or can/should these be different?

langz: Do we have any suggestions on which subset of props from its available ts props(NumberValueProps), that we want to display?

langz: I'll open a new PR with my own suggestions, and then we can take it further from there 👍

PR #4011, opened 2024-09-26
fix(Forms): removes cards max-width in Wizard (StepIndicator)

langz:
I don't think we should merge this right away.

I don't 100% understand why we would want to set a max width for a card used in eufemia forms? 
As introduced [here](https://github.com/dnbexperience/eufemia/pull/3416/).

I believe it probably exists multiple scenarios where you would want the card to take the full available width?
At least, it's been reported as a wish [here](https://dnb-it.slack.com/archives/CMXABCHEY/p1727171072702699).
How can we cater for doing so? Is it best to implement a prop, or just to remove the max-width(as done in this PR for now)? 🤔 

Internal discussion here: https://dnb-it.slack.com/archives/C04NL73S2Q5/p1727351567888759?thread_ts=1727246953.089389&cid=C04NL73S2Q5


tujoworker (approved):
I think we should give this a try. Looks like having a max-width gives not a better experience than stretched cards.

PR #4023, opened 2024-09-27
fix(Dropdown): align tertiary dropdown triangle icon with button

joakbjerk:
Proposal on how to solve the alignment issue

Before:
image-2024-09-26-10-09-57-403

After:
Screenshot 2024-09-30 at 12 32 34

Screenshot 2024-09-30 at 12 32 27

### DIFF
--- a/packages/dnb-eufemia/src/components/autocomplete/__tests__/__snapshots__/Autocomplete.test.tsx.snap
+++ b/packages/dnb-eufemia/src/components/autocomplete/__tests__/__snapshots__/Autocomplete.test.tsx.snap
@@ -2142,12 +2142,6 @@ button .dnb-form-status__text {
 .dnb-dropdown__row {
   display: inline-flex;
 }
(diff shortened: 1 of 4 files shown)


langz (approved):
Looks good to me 👍 
A shame that we can't  access dropdown-padding-horizontal, is there some other way of making use of this variable? 🤔

langz:
Does it make sense to have a screenshot test for these scenarios, to see if the different variations of arrows align up correctly?

joakbjerk:
> Does it make sense to have a screenshot test for these scenarios, to see if the different variations of arrows align up correctly?

Sounds like that could be handy 🤔

joakbjerk:
> Looks good to me 👍 A shame that we can't access dropdown-padding-horizontal, is there some other way of making use of this variable? 🤔

The only way I can think of is making it global through `:root` but, then it would loose the dropdown scope

PR #3987, opened 2024-09-24
feat(Forms): add support for `defaultValue` (and `value`) for fields used in `Iterate.Array`

tujoworker:
Fixes #3882

PR #4025, opened 2024-09-30
fix(Forms): keep field state during a Wizard step change when used inside Iterate.Array (which used defualtValue as the data source)

tujoworker:


tujoworker:
Jepp. The easier implementation had drawbacks and resulted in this bug. So better to make the shared solution, which all fields use, better. Which happens here. Also, most of the code change here will be used in PR #3987

Issue #3840, opened 2024-08-19
bug(Field.Selection): keyboard navigation in variant radio


langz:
## 🐛 Bug Report

The keyboard navigation in [Field.Selection in variant radio](https://eufemia.dnb.no/uilib/extensions/forms/base-fields/Selection/demos/#radio-buttons) does not work as expected(compared to [Eufemia's Radio component](https://eufemia.dnb.no/uilib/components/radio/#radio-group)).
In [Field.Selection with variant radio](http://localhost:8000/uilib/extensions/forms/base-fields/Selection/demos/#radio-buttons), I'm not able to:
- Tab into an option and select it using space/enter key
- Navigate between options using the arrow keys

I'm able to do all of the above in the [Radio component](https://eufemia.dnb.no/uilib/components/radio/#radio-group).

## To Reproduce

Visit the [demos of Field.Selection](https://eufemia.dnb.no/uilib/extensions/forms/base-fields/Selection/demos/#radio-buttons) and try to navigate using the keyboard.

## Expected behavior

It should work as in the [Radio component](https://eufemia.dnb.no/uilib/components/radio/#radio-group).

PR #4037, opened 2024-10-01
fix(Field.Selection): prevent `radio` re-render onClick to enable keyboard navigation

joakbjerk:
closes [this issue](https://github.com/dnbexperience/eufemia/issues/3840)

Before:

https://github.com/user-attachments/assets/4629c16f-111e-47a5-85d2-3e6ed66fb003

After:

https://github.com/user-attachments/assets/f7419a58-a0ea-4430-9eb9-461c7298968e

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/Selection/Selection.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/Selection/Selection.tsx
@@ -342,7 +342,7 @@ function renderRadioItems({
     return (
       <Component
         id={optionsCount === 1 ? id : undefined}
-        key={`option-${i}-${value}`}
+        key={`option-${i}-${id}`}
         label={variant === 'radio' ? label : undefined}
         text={variant === 'button' ? label : undefined}
         value={String(value ?? '')}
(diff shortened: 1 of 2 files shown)


packages/dnb-eufemia/src/extensions/forms/Field/Selection/Selection.tsx
```diff
@@ -342,7 +342,7 @@ function renderRadioItems({
     return (
       <Component
         id={optionsCount === 1 ? id : undefined}
-        key={`option-${i}-${value}`}
+        key={`option-${i}-${id}`}
```
langz, line 345: Looks good to me 👍
Is it safe to use `id` instead of `value` for all different Selection `variants`?
I guess so, but could be worth a thought 🤔

langz (approved):
🧗‍♂️

packages/dnb-eufemia/src/extensions/forms/Field/Selection/Selection.tsx
joakbjerk: I'll test 🙏

joakbjerk: Seems to be working nicely for both `radio` and `toggle` variants 👍

tujoworker (approved):
Nice! Can you add a test too?

PR #4038, opened 2024-10-01
feat(Forms): add `isValid` to Form.Visibility for showing content based on the validation of a field

tujoworker:
As of now, the Visibility does only react on data content, but not on validation.
This PR is an attempt to add support for showing something based on the validation result by another field:

```tsx
<Form.Visibility
  visibleWhen={{
    path: '/foo',
    isValid: true,
  }}
>
  <Value.Name.First path="/foo" />
</Form.Visibility>
```

Here is [a demo](https://eufemia-git-feat-visibility-validate-eufemia.vercel.app/uilib/extensions/forms/Form/Visibility/#show-children-when-field-has-no-errors-validation).

## Why first act on the blur event?

To avoid hiding/showing its content during e.g. typing, it will wait until the field has blurred.

E.g. the name field does validate negatively when a space is added after a name. During typing are we going then from false to true to false several times.

Screenshot 2024-10-01 at 10 32 22

This blur behavior could for sure be omitted with a property such as `continuously: true` (a little bit like `continuousValidation`).

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Visibility/info.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Visibility/info.mdx
@@ -47,7 +47,7 @@ render(
 
 ### Validation driven visibility
 
-You can conditionally display children based on field validation by using the `visibleWhen` property with `hasValidated: true`:
+You can conditionally display children based on field validation by using the `visibleWhen` property with `isValid: true`:
 
 ```tsx
 import { Form, Field } from '@dnb/eufemia/extensions/forms'
@@ -58,7 +58,7 @@ render(
     <Form.Visibility
       visibleWhen={{
         path: '/myField',
-        hasValidated: true,
+        isValid: true,
       }}
     >
       show me when the validation succeeds
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Visibility/info.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Visibility/info.mdx
@@ -4,19 +4,70 @@ showTabs: true
 
 ## Description
 
-The `Form.Visibility` component makes it possible to show or hide components on the screen based on the state of data. It can either be fed with the values directly via properties, or it can read data from a surrounding [Form.Handler](/uilib/extensions/forms/Form/Handler) and show or hide components based on the data it points to.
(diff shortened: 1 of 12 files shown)


langz:
I made a CSB where I try to use this feature with `Field.OrganizationNumber` https://codesandbox.io/p/sandbox/friendly-architecture-6prw47

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Visibility/info.mdx
```diff
@@ -4,20 +4,71 @@ showTabs: true
 
 ## Description
 
-The `Form.Visibility` component makes it possible to show or hide components on the screen based on the state of data. It can either be fed with the values directly via properties, or it can read data from a surrounding [Form.Handler](/uilib/extensions/forms/Form/Handler) and show or hide components based on the data it points to.
+The `Form.Visibility` component allows you to conditionally show or hide components based on the state of data or field validation. You can either provide the values directly via properties or let it read data from a surrounding [Form.Handler](/uilib/extensions/forms/Form/Handler). This enables dynamic visibility control based on the paths it points to.
+
+### Data driven visibility
+
+There are several properties you can use to control visibility, such as `pathDefined`, `pathTruthy`, `pathTrue` etc.
```
langz, line 11:
```suggestion
There are several [properties](/uilib/extensions/forms/Form/Visibility/properties/) you can use to control visibility, such as `pathDefined`, `pathTruthy`, `pathTrue` etc.
```

packages/dnb-eufemia/src/extensions/forms/Form/Visibility/Visibility.tsx
```diff
@@ -16,12 +16,23 @@ import VisibilityContext from './VisibilityContext'
...
     }
   | {
       itemPath: Path
-      hasValue: unknown
+      hasValue: unknown | ((value: unknown) => boolean)
     }
+  | {
+      path: Path
+      hasValidated: boolean
+      continuousValidation?: boolean
```
langz, line 28: We should probably remember to add this property `continuousValidation` to [properties docs](https://eufemia-git-feat-visibility-validate-eufemia.vercel.app/uilib/extensions/forms/Form/Visibility/properties/)

langz:
Some thoughts/suggestions/alternatives regarding naming of the `hasValidated` property:
- `isValidated`
- `isValid` - I think I like this the best.

tujoworker:
I changed it to be `isValid` 👍

langz (approved):

PR #4042, opened 2024-10-02
fix(Forms): support validation when using object based `transformOut` and `transformIn` on fields

### LINKED ISSUE #3997: bug: transformIn and transformOut for Field.String is not working with object
## 🐛 Bug Report

I get the following error message when typing a value into a Field.String that has `transformIn` and `transformOut`, and tabbing out(blur):

`Cannot read properties of undefined (reading 'charAt')`

image

## To Reproduce

CSB https://codesandbox.io/p/sandbox/serene-cloud-rhvldn

tujoworker:
Fixes #3997

Until now, the "out" transformed value was used inside useFidldProps. But when transforming out to an object, validation can not be performed anymore. So this PR transforms the object "in" again internally, so validation can be performed.


langz (approved):
Looks great, thanks :) 
I re-made the [CSB using the build from this PR](https://codesandbox.io/p/sandbox/relaxed-river-mns963?workspaceId=9eb653cb-4406-4b8f-bc96-714927cc7fcd), and it works as expected 🚀

PR #4051, opened 2024-10-03
feat(Value.PhoneNumber and NumberFormat): displays country code using prefix `+` instead of `00`

langz:
This PR makes it so we display `+47` instead of `00 47` when using Value.PhoneNumber for values like `+47 93086048` .

I think the reason why it displays 0047 rather than +47 in main today is because of back in the days we only had Norwegian numbers and probably no country code. So then it worked. 

[See CSB for how it works/looks on main today ](https://codesandbox.io/p/sandbox/jovial-snow-z93qgz)

image

[See CSB for how it looks in this PR](https://codesandbox.io/p/sandbox/great-cohen-57qhnr?workspaceId=9eb653cb-4406-4b8f-bc96-714927cc7fcd)

image

There's a lot more work that needs to be done to support displaying/formatting phone numbers from all countries we support in Field.PhoneNumber. As there's different ways of formatting the country code based on the country, and also different ways of formatting the number based on the country. 
I believe this PR/fix is sufficient for now(mainly Norwegian, which is our main use case).


packages/dnb-eufemia/src/extensions/forms/Value/PhoneNumber/PhoneNumber.tsx
```diff
@@ -15,9 +15,16 @@ function PhoneNumber(props: Props) {
...
 
   const toInput = useCallback((value) => {
-    return format(cleanNumber(value), {
-      phone: true,
-    }).toString()
+    const splitBySpace = value?.split(' ')
+    if (splitBySpace?.length > 1) {
+      // This is good enough for norwegian country code.
+      // As we don't have to format the country code, only return +47.
+      // Other country codes, like Trinidad and Tobago
```
langz, line 22:
```suggestion
      // I believe other country codes, like Trinidad and Tobago
```

packages/dnb-eufemia/src/extensions/forms/Value/PhoneNumber/PhoneNumber.tsx
```diff
@@ -15,9 +15,16 @@ function PhoneNumber(props: Props) {
     props.label ?? (props.inline ? undefined : translations.label)
 
   const toInput = useCallback((value) => {
-    return format(cleanNumber(value), {
-      phone: true,
-    }).toString()
+    const splitBySpace = value?.split(' ')
+    if (splitBySpace?.length > 1) {
+      // This is good enough for norwegian country code.
+      // As we don't have to format the country code, only return +47.
```
langz, line 21:
```suggestion
      // As we don't have to format the country code, only return +47 number.
```

packages/dnb-eufemia/src/extensions/forms/Value/PhoneNumber/PhoneNumber.tsx
```diff
@@ -15,9 +15,16 @@ function PhoneNumber(props: Props) {
...
   const toInput = useCallback((value) => {
-    return format(cleanNumber(value), {
-      phone: true,
-    }).toString()
+    const splitBySpace = value?.split(' ')
+    if (splitBySpace?.length > 1) {
+      // This is good enough for norwegian country code.
+      // As we don't have to format the country code, only return +47 {phonenumber}.
+      // I believe other country codes, like Trinidad and Tobago
+      // Should format the country code, like so +1 868. This is not supported today.
```
langz, line 23:
```suggestion
      // Should format the country code, like so +1 868 {phonenumber}. This is not supported today.
```

joakbjerk (approved):
Works great for Norwegian numbers 👍

To bad there isn't an `Intl`-api for phone numbers embedded in the browsers 😅

tujoworker (approved):
I have added commit that changes the `+` instead of `00` in the original function. This function did not have many tests. So I added some more. 

Does that make sense?

tujoworker:
We may change the PR title to something more general. Also, it may not be a fix, but a feature.

packages/dnb-eufemia/src/components/number-format/NumberUtils.d.ts
```diff
@@ -163,3 +163,17 @@ export const roundHalfEven: (
   num: number,
   decimalPlaces?: number
 ) => number;
+
+/**
+ * Formats a number as a phone number
+ *
+ * @param {string|number} number any number
+ * @param {string} locale locale as a string. Defaults to the global LOCALE constant
```
langz, line 171:
```suggestion
 * @param {string} locale as a string. Defaults to the global LOCALE constant
```

langz:
I like this, yeah, we can rather call it a feature 🌟

PR #4056, opened 2024-10-03
feat(Forms): add `transformLabel` to all Value.* component

tujoworker:


packages/dnb-eufemia/src/extensions/forms/Value/SummaryList/__tests__/SummaryList.test.tsx
```diff
@@ -4,10 +4,6 @@ import userEvent from '@testing-library/user-event'
 import SummaryList from '../SummaryList'
 import { Field, Form, Value } from '../../..'
 
-beforeEach(() => {
-  global.console.log = jest.fn()
-})
```
langz, line 9: 👋

langz (approved):
Looks great 💯 

Just a thought, what will the following example result in?
```
      <Field.String path="/myPath" label="field label" />
      <Value.String
        path="/myPath"
        label="value label"
        inheritLabel
        transformLabel={(label: string) => label.toUpperCase()}
        showEmpty
      />
```

Will it result in `FIELD LABEL` or `VALUE LABEL`?

The answer is, `VALUE LABEL`, and it probably makes most sense yeah 👍

langz:
> Looks great 💯
> 
> Just a thought, what will the following example result in?
> 
> ```
> <Form.Handler>
>       <Field.String path="/myPath" label="field label" />
>       <Value.String
>         path="/myPath"
>         label="value label"
>         inheritLabel
>         transformLabel={(label: string) => label.toUpperCase()}
>         showEmpty
>       />
> </Form.Handler>
> ```
> 
> Will it result in `FIELD LABEL` or `VALUE LABEL`?
> 
> The answer is, `VALUE LABEL`, and it probably makes most sense yeah 👍

But I'm not sure what the answer would be if we wrap the Value in a SummaryList as well 🤔

PR #4062, opened 2024-10-04
chore: fixes false positive tests in Field.OrganizationNumber

langz:
Previous tests were false positive.
When doing:
`expect(screen.queryByRole('alert')).toBeNull()`
It would in many tests always be true, even if an error should occur.

This PR also removes a lot of not needed `fireEvent.blur(document.querySelector('input'))`, many times when removing it, the result of the test does not change.
I don't think this is needed when we test the validator prop, as the validator doesn't need a blur event to execute.

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/OrganizationNumber/__tests__/OrganizationNumber.test.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/OrganizationNumber/__tests__/OrganizationNumber.test.tsx
@@ -8,6 +8,12 @@ import nbNO from '../../../constants/locales/nb-NO'
 const nb = nbNO['nb-NO']
 
 describe('Field.OrganizationNumber', () => {
+  async function expectNever(callable: () => unknown): Promise<void> {
+    await expect(() => waitFor(callable)).rejects.toEqual(
+      expect.anything()
+    )
+  }
+
   it('should have Norwegian mask', async () => {
     render(<Field.OrganizationNumber />)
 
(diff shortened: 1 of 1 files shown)


joakbjerk (approved):

packages/dnb-eufemia/src/extensions/forms/Field/OrganizationNumber/__tests__/OrganizationNumber.test.tsx
```diff
@@ -8,6 +8,12 @@ import nbNO from '../../../constants/locales/nb-NO'
 const nb = nbNO['nb-NO']
 
 describe('Field.OrganizationNumber', () => {
+  async function expectNever(callable: () => unknown): Promise<void> {
+    await expect(() => waitFor(callable)).rejects.toEqual(
```
tujoworker, line 12: Does that mean it retries in e.g. 30 seconds?

langz: I'll wait with merging this PR, feels like I have to look into this even more.
But I'll make a PR to highlight my problems.

langz: See here https://github.com/dnbexperience/eufemia/pull/4062

langz: I'm not 100% sure how this works, I've just followed the same pattern as in String.test.tsx when we test validator there. But at least this pattern "works"(doesn't result in false-positives")

langz:
I'll reopen this PR at a later stage

PR #4065, opened 2024-10-04
demo: this should fail? (DO NOT MERGE)

langz:


packages/dnb-eufemia/src/extensions/forms/Field/OrganizationNumber/__tests__/OrganizationNumber.test.tsx
```diff
@@ -68,7 +68,7 @@ describe('Field.OrganizationNumber', () => {
         <Field.OrganizationNumber
           value={invalidOrgNo}
           validateInitially
-          validate={false}
+          validate={true}
```
langz, line 71: This should fail, just paste `987654321` into [demo](https://eufemia.dnb.no/uilib/extensions/forms/feature-fields/OrganizationNumber/#empty) here, and see.
But the test doesn't actually fail...

PR #4068, opened 2024-10-06
fix(Forms): show error on every value change when using exported validators

### LINKED ISSUE #4067: Field.OrganizationNumber pattern collides with validator/onBlurValidator
## 🐛 Bug Report

It seems like pattern collides with validator/onBlurValidator every other time. Can it co-exist? 🤔 
When entering a non valid organization number(123 331 231) in the following [CSB](https://codesandbox.io/p/sandbox/quiet-sunset-8t7q3d?workspaceId=9eb653cb-4406-4b8f-bc96-714927cc7fcd), then removing the last digit(1) to so add the same digit(1) again, I do not get the same error(I actually don't get any error) as the first time? 🤔 

I can continue to do this for ever:
- Remove digit 1.
- Add digit 1.
- Tab out to see if there's an error message or not(every other time).

https://github.com/user-attachments/assets/0324b1f2-92d2-45bd-a479-101c38ef9f1f

tujoworker:
fixes #4067

I could not find any relation with `pattern` actually.

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/hooks/__tests__/useFieldProps.test.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/hooks/__tests__/useFieldProps.test.tsx
@@ -3377,7 +3377,7 @@ describe('useFieldProps', () => {
           expect(screen.queryByRole('alert')).toHaveTextContent('bar')
         })
         expect(myValidator).toHaveBeenCalledTimes(5)
-        expect(fooValidator).toHaveBeenCalledTimes(4)
+        expect(fooValidator).toHaveBeenCalledTimes(5)
         expect(barValidator).toHaveBeenCalledTimes(4)
       })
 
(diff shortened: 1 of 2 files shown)


langz:
> fixes #4067
> 
> 
> 
> I could not find any relation with `pattern` actually.

Aha good, it was just a theory that pattern was the prop which led to what I experienced. I can retest and see if I can experience it without using the pattern prop(I don't think so, as I only think this happened when using that prop together with validator), but will have to test again🫡

langz:
> > fixes #4067
> > I could not find any relation with `pattern` actually.
> 
> Aha good, it was just a theory that pattern together with validator was the prop which led to what I experienced. I can retest and see if I can experience it without using the pattern prop(only using validator, I don't think so(just a guess), as I only think this happened when using that prop together with validator), but will have to test again🫡

I re-tested now, and I agree. It doesn't have anything to do with the `pattern` property.
I also made a [CSB](https://codesandbox.io/p/sandbox/keen-leakey-y56hzf?workspaceId=9eb653cb-4406-4b8f-bc96-714927cc7fcd) for testing this fix, and it works great, not able to reproduce it again 👍

langz (approved):

PR #4071, opened 2024-10-07
fix(Field.OrganizationNumber): should validate on all digits(not only when 9)

langz:


tujoworker (approved):
Looks good. I added another test and made a change to yours.

PR #4080, opened 2024-10-07
release of v10.52

langz:


tujoworker (approved):
We should fill inn all the features via this #4081 PR before merge.

langz:
> We should fill inn all the features via this #4081 PR before merge.

Sure, let us do so :)

I'll go for lunch now, but can do it when I come back.

tujoworker:
Not everything needs to be mentioned. I think only features and important things are worth mentioning.

PR #4087, opened 2024-10-08
fix(Forms): keep `Iterate.EditContainer` open when falsy value or empty object was given as the iterate value

### LINKED ISSUE #4046: Iterate validation bypass
## 🐛 Bug Report

When trying to add an invalid object to an iterate array, the object seems to be added while invalid. It seems to go from `PushContainer` to `EditContainer`.
Then, after typing something, the `ViewContainer` becomes active even though the object is invalid.

https://github.com/user-attachments/assets/6fd72f77-3dcf-4b19-852a-02f2975b6dfe

## To Reproduce

https://codesandbox.io/p/sandbox/iterate-validation-dwlkl3

## Expected behavior

I expect validation error messages to be displayed when trying to add an invalid object to an iterate array.
I expect the object not to be added when invalid after typing in one of the fields.

## Eufemia Version
10.51.1

tujoworker:
Fixes #4046

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Iterate/Array/Examples.tsx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Iterate/Array/Examples.tsx
@@ -6,6 +6,7 @@ import {
   Value,
   Form,
   Tools,
+  Wizard,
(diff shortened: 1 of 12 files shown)


langz (approved):
🙏👍

PR #4092, opened 2024-10-08
feat(Forms): remove internal pattern from `Field.OrganizationNumber` in favor of the internal validator

### LINKED ISSUE #4073: bug(Field.OrganizationNumber): default validation error is displayed even when replacing the validation(onBlurValidator)
## 🐛 Bug Report

When displaying error message initially(`validateInitially`), and replacing the `onBlurValidator` with a custom `onBlurValidator`, I still get the error message from the "default"/internal organizationnumbervalidator used in onBlurValidator.
image

## To Reproduce

https://codesandbox.io/p/sandbox/wispy-meadow-hpkscm?workspaceId=9eb653cb-4406-4b8f-bc96-714927cc7fcd

## Expected behavior

I would expect to see the following error message, `My Error B`.

tujoworker:
Fixes #4073

The reason for the outcome in the issue is that the internal pattern is prioritized over the custom validator. But we probably don't need the pattern. So this PR removes is.

Also, I think when first utilizing the `expectNever`, we rather should add a Jest matcher.

### DIFF
--- a/packages/dnb-eufemia/jest.d.ts
+++ b/packages/dnb-eufemia/jest.d.ts
@@ -1,5 +1,6 @@
 declare namespace jest {
   interface Matchers<R> {
     toBeType(received: string, expected?: string): R;
+    neverToResolve(): Promise<R>;
   }
 }
--- a/packages/dnb-eufemia/src/core/jest/jestSetup.js
+++ b/packages/dnb-eufemia/src/core/jest/jestSetup.js
@@ -7,6 +7,7 @@ import { axe, toHaveNoViolations } from 'jest-axe'
 import fs from 'fs-extra'
 import path from 'path'
 import sass from 'sass'
+import { waitFor } from '@testing-library/react'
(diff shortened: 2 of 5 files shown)


langz:
Thanks so much, I'll take a look tomorrow. 

Just before you opened this PR, I realized it was the internal pattern which was in control. I could of course replace that as well, allowing all patterns, but yeah. If we should use the internal pattern then we should of course document it, that the default pattern is this or that. 

Pattern and validator sort of collides (mentally in my head) and it's a bit difficult to understand when both is in play, espacially now when we provide a lot of validators.

langz (approved):
I think this looks great.
As we don't use the internal pattern to provide any other functionality(no different error message based on length of input, etc), as we always did display `"Dette er ikke et gyldig organisasjonsnummer."` regardless of whether it was
a error in pattern or validator, I think it does not make sense to use a pattern here.

PR #4099, opened 2024-10-09
feat(Forms): add `onBeforeStepChange` event handler with the current index

tujoworker:
This PR is based on #4093

It makes it possible to get the current step index, during navigation, including the current step `id`. 

If there are any better event name recommendations, let me know.


langz:
Based on the name, does it mean that this happens before the onStepChange event?

Based on the explanation in the PR description, I think the name could be getCurrentStepIndex or so. 
Will have to look into it further.

tujoworker:
Thinking over it again, we have `previousIndex` already. So we don't need this event.

Issue #4072, opened 2024-10-07
Deny navigation to the next step if there are open subforms on the page


andlbrei:
## 🚀 Feature Proposal

When the user has a subform open they should not be allowed to navigate to the next page before committing or cancelling the subform.

## Motivation

We want to make sure the user is confident that the information they want to store in the form is stored, or discarded if the user does not want it there after starting to fill it out.

## Example

If the user has started adding a new entry to a array, but not pressed "Cancel" or "Add", they should not be allowed to navigate away and lose the data they have entered. 

## Pitch

It would enhance the existing validation mechanisms of Eufemia Forms.

## Notes

I do not see this as critical functionality, but it might be a good addition, depending on how much effort it takes to implement.
It might make sense for it to be a `strict`-mode feature.
The example mentions a `Iterate.Array` use case, but I see this as something that would make sense for `Isolation` as well.
Any subform that provides a "commit" or "cancel" action to the user.

tujoworker:
Kind of `required` on a Iterate.PushContainer or Form.Isolation?

andlbrei:
Perhaps that is a good way of looking at it, yes.
The user is required to choose either the "commit" or the "cancel" action before proceeding.

tujoworker:
Currently, fields are usually responsible for informing the form when more data is needed. However, when used in isolation, they lose that capability. This feature addresses that by allowing isolated fields to pass the data requirement back to the parent form if there's an error.

- I believe this behavior should be optional (opt-in). Do you agree?
- What would you suggest as a prop name to make this opt-in work as intended?

andlbrei:
I would agree that it should be optional 👍 
After you mentioned `required` I have a hard time thinking about anything else, but I'm not sure if that is the best name for it as this functionality is a bit different 🤔 😅 

Also, my initial suggestion here was not based on fields within a subform not validating, but rather that the user should finish the process of committing the subform to the actual form (or cancelling), so that they do not fill in data in a bunch of fields, and then forget to click "Add" to actually store the data before progressing to the next step in the form.
Combining this with what you stated in letting the subform inform the parent of validation state is also natural to do, and a very good point.

tujoworker:
`propagateErrors` maybe?

tujoworker:
Its actually not only about the errors. 

- When an error is passed through, then it prevents the form from submitting ✅
- But when it gets resolve, the user may still not have committed the data.

That leads us to use a different name. Kind of back to `required`.

[...]

PR #4103, opened 2024-10-09
feat(Forms): add `bubbleValidation` to Form.Isolation and Iterate.PushContainer to prevent the form from being submitted when there are fields with errors

tujoworker:
Partially solves #4072

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Form/Isolation/Isolation.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Form/Isolation/Isolation.tsx
@@ -9,8 +9,13 @@ import React, {
 import pointer, { JsonObject } from '../../utils/json-pointer'
 import { extendDeep } from '../../../../shared/component-helper'
 import { isAsync } from '../../../../shared/helpers/isAsync'
+import useId from '../../../../shared/helpers/useId'
 import useDataValue from '../../hooks/useDataValue'
-import { Context, ContextState, Provider } from '../../DataContext'
+import {
+  Context as DataContext,
+  ContextState,
+  Provider,
+} from '../../DataContext'
 import SectionContext from '../Section/SectionContext'
 import IsolationCommitButton from './IsolationCommitButton'
 import {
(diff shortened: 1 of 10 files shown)


packages/dnb-eufemia/src/extensions/forms/Form/Isolation/Isolation.tsx
```diff
@@ -88,12 +98,25 @@ function IsolationProvider<Data extends JsonObject>(
...
+  //     return
+  //   }
+
+  //   const path = `/${id}`
+  //   outerContext.setMountedFieldState(path, {
+  //     isMounted: true,
+  //   })
+  //   outerContext.setFieldError(path, new Error('Form.Isolation'))
+  // }, [delegateValidation, id, outerContext])
+
```
langz, line 119: Should perhaps remove the lines of code above?

packages/dnb-eufemia/src/extensions/forms/Form/Isolation/IsolationDocs.ts
```diff
@@ -20,6 +20,11 @@ export const IsolationProperties: PropertiesTableProps = {
     type: 'React.Ref',
     status: 'optional',
   },
+  delegateValidation: {
+    doc: 'Prevent the form from being submitted when there are fields with errors inside the Form.Isolation.',
```
langz, line 24: Do we need to add this to `PushContainerDocs.ts` as well?

packages/dnb-eufemia/src/extensions/forms/Form/Isolation/Isolation.tsx
```diff
@@ -43,6 +48,10 @@ export type IsolationProviderProps<Data> = {
     isolatedData: JsonObject,
     handlerData: JsonObject
   ) => unknown
+  /**
+   * Propagate errors from the fields to the parent form.
```
langz, line 52: This text is different from what's documented in the props for `delegateValidation` in `packages/dnb-eufemia/src/extensions/forms/Form/Isolation/IsolationDocs.ts`.

langz (approved):

langz:
`delegateValidation`, maybe we together could come up with a better name?
When I think of delegate, I'm thinking that it delegates the validation to someone or something, but in `delegateValidation` we don't say to what/who we delegate it to?
Is delegate the correct wording, when we don't have a "recipient" 🤔 

Do you have any thoughts @andlbrei?

andlbrei:
`elevateValidation` or something similar?
Or something like `validateGlobally`, but with a different term than "global" perhaps...

langz (approved):
🫧

PR #4104, opened 2024-10-09
chore(Field.OrganizationNumber): disable default onBlurValidator using `false`

langz:
Let's say if user wants to use validator instead of onBlurValidator, then he should be able to disable the default `onBlurValidator`.

I've also fixed a few false-positive tests in this PR(tested them all manually, so the change is/was neeeded).
It could of course exist a better way of testing this, instead of using `neverToResolve`, but that's the only pattern I've found that's working now.

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/OrganizationNumber/OrganizationNumber.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/OrganizationNumber/OrganizationNumber.tsx
@@ -2,31 +2,23 @@ import React, { useCallback, useMemo } from 'react'
 import StringField, { Props as StringFieldProps } from '../String'
 import useErrorMessage from '../../hooks/useErrorMessage'
 import useTranslation from '../../hooks/useTranslation'
+import { Validator } from '../../types'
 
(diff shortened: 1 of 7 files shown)


packages/dnb-eufemia/src/extensions/forms/Field/OrganizationNumber/OrganizationNumber.tsx
```diff
@@ -37,18 +29,36 @@ function OrganizationNumber(props: Props) {
...
+  const mask = useMemo(
+    () =>
+      omitMask
+        ? [/\d/, /\d/, /\d/, /\d/, /\d/, /\d/, /\d/, /\d/, /\d/]
+        : [/\d/, /\d/, /\d/, ' ', /\d/, /\d/, /\d/, ' ', /\d/, /\d/, /\d/],
+    [omitMask]
+  )
+
+  const onBlurValidatorToUse =
+    onBlurValidator == false ? undefined : onBlurValidator
```
langz, line 50:
```suggestion
    onBlurValidator === false ? undefined : onBlurValidator
```

joakbjerk (approved):

PR #4105, opened 2024-10-09
chore(Field.NationalIdentityNumber): disable default onBlurValidator using `false`

langz:
Let's say if user wants to use validator instead of onBlurValidator, then he should be able to disable the default `onBlurValidator`.

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/NationalIdentityNumber/NationalIdentityNumber.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/NationalIdentityNumber/NationalIdentityNumber.tsx
@@ -85,7 +85,7 @@ function NationalIdentityNumber(props: Props) {
   )
 
   const onBlurValidatorToUse =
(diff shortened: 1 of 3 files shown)


packages/dnb-eufemia/src/extensions/forms/Field/NationalIdentityNumber/NationalIdentityNumber.tsx
```diff
@@ -75,18 +54,53 @@ function NationalIdentityNumber(props: Props) {
...
+            /\d/,
+            /\d/,
+            /\d/,
+            /\d/,
+          ],
+    [omitMask]
+  )
+
+  const onBlurValidatorToUse =
+    onBlurValidator == false ? undefined : onBlurValidator
```
langz, line 88:
```suggestion
    onBlurValidator === false ? undefined : onBlurValidator
```

joakbjerk (approved):

PR #4108, opened 2024-10-10
fix(Forms): Fix use of unpolyfilled structuredClone in useData hook

kimroen:
We had a spot that was using `window.structuredClone` instead of the polyfill from `@ungap/structured-clone` that is used everywhere else.

Fixed that spot, and added a rule to eslint to make sure we don't forget in the future.

Note that I don't make any judgement on whether we should be using this or not here, just enforcing consistency.

With this linting rule in place:

CleanShot 2024-10-10 at 14 30 28

### DIFF
--- a/packages/dnb-eufemia/.eslintrc
+++ b/packages/dnb-eufemia/.eslintrc
@@ -143,6 +143,13 @@
               }
             ]
           }
+        ],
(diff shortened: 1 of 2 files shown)


tujoworker (approved):
Thank you 💙

PR #4111, opened 2024-10-10
fix(Forms): ensure `emptyValue` is set in the data context when defined

### LINKED ISSUE #4070: An empty Field.String produces undefined by default, not an empty string
## 🐛 Bug Report

This could be expected and intended behavior, and if so you can close this and consider it feedback 😊 

When you empty out a `Field.String`, the resulting value in the form is `undefined`, and not `''` (an empty string). This can be confusing, especially if combined with a default data value of `''` when setting up the form.

## To Reproduce

In [this CodeSandbox](https://codesandbox.io/p/sandbox/purple-shape-tg9fdz), we find a form with one field which has the default value `''`, and a corresponding `Field.String` to edit it:

```tsx
function MyForm() {
  return (
    <Form.Handler defaultData={{ myPath: '' }}>
      <Field.String path="/myPath" />
    </Form.Handler>
  );
}
```

The value of `/myPath` when this form is rendered is `''`, as expected. However, when you type a character and then remove it, the value of `/myPath` is `undefined`.

## Expected behavior
I expected the value of `/myPath` to be an empty string after editing, not `undefined`.

## Why is this unexpected?
If you've worked with React for any amount of time, you've probably done something like this to make a controlled input:

```jsx
const [fieldValue, setFieldValue] = useState();

return (
  <input value={fieldValue} onChange={(e) => { setFieldValue(e.currentTarget.value) }} />
);
```

And if you have, then you've also seen this warning that shows up when you start typing:

> Warning: A component is changing an uncontrolled input to be controlled. This is likely caused by the value changing from undefined to a defined value, which should not happen. Decide between using a controlled or uncontrolled input element for the lifetime of the component.

So at least for me, I instinctively give the value a default value of empty string.

As a bonus, keeping it a string makes dealing with the types easier 😊 

## Eufemia Version

**Browser JS:** 10.51.2

**Browser CSS:** 10.51.2

tujoworker:
Fixes #4070

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/ArraySelection/__tests__/ArraySelection.test.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/ArraySelection/__tests__/ArraySelection.test.tsx
@@ -55,7 +55,7 @@ describe('ArraySelection', () => {
     it('handles emptyValue correctly', () => {
       const handleChange = jest.fn()
       render(
-        <Field.ArraySelection onChange={handleChange} emptyValue="empty">
+        <Field.ArraySelection onChange={handleChange} emptyValue={[]}>
           <Field.Option value="option1">Option 1</Field.Option>
           <Field.Option value="option2">Option 2</Field.Option>
         </Field.ArraySelection>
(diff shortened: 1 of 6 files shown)


langz (approved):
🫡

PR #4112, opened 2024-10-10
fnrvalidator and dnrvalidator pattern error

langz:
Todo:
 - tests
 - improvement to the code. Any single digit should give pattern error, not only when it's starting with 4(in dnr case).


packages/dnb-eufemia/src/extensions/forms/constants/locales/en-GB.ts
```diff
@@ -120,6 +120,9 @@ export default {
       errorRequired: 'You must enter a national identity number.',
       errorFnr: 'Invalid national identity number.',
       errorDnr: 'Invalid D number.',
+      errorFnrPattern:
+        'Invalid national identity number. Enter a valid 11-digit number.',
+      errorDnrPattern: 'Invalid D number. Enter a valid 11-digit number.',
```
langz, line 125: I should call these length. Not pattern.

PR #4113, opened 2024-10-11
fix(Forms): don't call internal `exportValidators` when they not are exported as an array

### LINKED ISSUE #4106: Field.OrganizationNumber when replacing validators using a function, the internal validator still runs.
## 🐛 Bug Report

Seen in the light of https://github.com/dnbexperience/eufemia/pull/4104 & https://github.com/dnbexperience/eufemia/pull/4105, the following issue applies to both `Field.OrganizationNumber` and `Field. NationalIdentityNumber`. But in this issue I've used `Field.OrganizationNumber` as the example.

`Field.OrganizationNumber` when replacing the `onBlurValidator`or `validator`, the internal validator(validators provided to `externalValidators`) still runs. I think it's something with `exportValidators`, but I don't understand how exporting validators could/should run the validators provided to it. I've tried removing/not providing `exportValidators` in the component when developing locally, and then the CSBs work as expected. So I think it has something to do with `exportValidators`.

## To Reproduce

[Here's a CSB showing that the internal validator still runs when providing a custom validator](https://codesandbox.io/p/sandbox/patient-http-66drz9?workspaceId=9eb653cb-4406-4b8f-bc96-714927cc7fcd) which is not expected. I get the internal validator error message(Dette er ikke et gyldig organisasjonsnummer.) when removing one digit and then blurring, I expect to not get any error messages at all in this scenario.

image

[Here's a CSB showing that the internal validator still runs when providing a custom `onBlurValidator`](https://codesandbox.io/p/sandbox/crazy-dawn-wtjx42?workspaceId=9eb653cb-4406-4b8f-bc96-714927cc7fcd) which is not expected. I get the internal validator error message(Dette er ikke et gyldig organisasjonsnummer.) when removing one digit and then blurring, I expect to not get any error messages at all in this scenario.

image

[Here's a CSB showing that when only removing the onBlurValidator(false) and not providing an own validator, the internal validator do not run, which is expected. ](https://codesandbox.io/p/sandbox/hopeful-gianmarco-6qrclt?workspaceId=9eb653cb-4406-4b8f-bc96-714927cc7fcd)This CSB is just showcasing that something unwanted happens when providing an validator.

image

## Expected behavior

In the following [CSB](https://codesandbox.io/p/sandbox/patient-http-66drz9?workspaceId=9eb653cb-4406-4b8f-bc96-714927cc7fcd), I get the internal validator error message(`Dette er ikke et gyldig organisasjonsnummer.`) when removing one digit and then blurring, I expect to not get any error messages at all in this scenario. And I expect the validator provided to `exportValidators` no to run.

tujoworker:
Fixes #4106

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/hooks/__tests__/useFieldProps.test.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/hooks/__tests__/useFieldProps.test.tsx
@@ -3874,7 +3874,7 @@ describe('useFieldProps', () => {
         }).neverToResolve()
       })
 
-      it('should not call internal validates when they are not returned in the publicValidator', async () => {
+      it('should not call internal validators when they are not returned in the publicValidator', async () => {
         let internalValidators, barValidator, bazValidator
 
         const MockComponent = (props) => {
--- a/packages/dnb-eufemia/src/extensions/forms/hooks/__tests__/useFieldProps.test.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/hooks/__tests__/useFieldProps.test.tsx
@@ -3372,7 +3372,7 @@ describe('useFieldProps', () => {
       })
     })
 
-    describe('exportValidators', () => {
+    describe('validators given as an array', () => {
       it('should call all validators returned as an array', async () => {
         const fooValidator = jest.fn((value) => {
           if (value.includes('foo')) {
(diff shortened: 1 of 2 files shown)


packages/dnb-eufemia/src/extensions/forms/hooks/__tests__/useFieldProps.test.tsx
```diff
@@ -3843,7 +3845,36 @@ describe('useFieldProps', () => {
...
+        render(<MockComponent validator={myValidator} />)
+
+        await expect(() => {
+          expect(
+            document.querySelector('.dnb-form-status')
+          ).toBeInTheDocument()
+        }).neverToResolve()
+      })
+
+      it('should not call internal validates when they are not returned in the publicValidator', async () => {
```
langz, line 3877:
```suggestion
      it('should not call internal validators when they are not returned in the publicValidator', async () => {
```

PR #4120, opened 2024-10-12
feat(Forms): add support for `minItems`/`maxItems` error messages to Field.ArraySelection

boyum:
With the use of JSON Schema and ajv's validation, `ArraySelection` supports setting `minItems` and `maxItems` to control how many items the user can select. The validation error message types are however not exposed, thus TypeScript won't let you override them.


tujoworker (approved):
Thank you for adding this enhancement 🫶

I will take this further and add compatibility for properties, appearance of the validaiton and translation messages.

PR #4140, opened 2024-10-17
chore(useTranslation): ensure `formatMessage` and `renderMessage` have the same instance after rerender

tujoworker:
The reason is to avoid re-renders or maximum callstack problems.

The test does fail without the changes:
Screenshot 2024-10-17 at 09 39 34


langz (approved):
🦖

PR #4142, opened 2024-10-17
feat(Forms): add support to set a custom `width` to fields and values

tujoworker:
Quick example:

```tsx
<Field.String label="Custom width" width="8rem" />
```


packages/dnb-eufemia/src/extensions/forms/FieldBlock/FieldBlock.tsx
```diff
@@ -112,6 +112,8 @@ function FieldBlock(props: Props) {
     children,
     ...rest
   } = Object.assign({}, sharedData.data, props)
+  const hasCustomWidth = /\d(ch|rem)$/.test(String(width))
```
joakbjerk, line 115: Should we allow `px` as well, or is `rem` enough in it self?

packages/dnb-eufemia/src/extensions/forms/FieldBlock/FieldBlockDocs.ts
```diff
@@ -22,12 +22,12 @@ export const fieldBlockSharedProperties: PropertiesTableProps = {
     status: 'optional',
   },
   width: {
-    doc: '`small`, `medium`, `large`, `stretch` or `false` for predefined standard widths.',
+    doc: 'Will set the width for the whole block. Use `small`, `medium`, `large` for predefined standard widths. You can also set a custom width `{number}rem` or use `stretch` or `false`.',
```
joakbjerk, line 25:
```suggestion
    doc: 'Will set the width for the whole block. Use `small`, `medium`, `large` for predefined standard widths. You can also set a custom width `{number}(rem|ch)` or use `stretch` or `false`.',
```

Don't have to write it that way, but we should mention `ch` too 🫡

packages/dnb-eufemia/src/extensions/forms/FieldBlock/FieldBlockDocs.ts
```diff
@@ -22,12 +22,12 @@ export const fieldBlockSharedProperties: PropertiesTableProps = {
...
     type: ['string', 'false'],
     status: 'optional',
   },
   contentWidth: {
-    doc: '`small`, `medium`, `large`, `stretch` or `false` for predefined standard widths.',
+    doc: 'Will set the width for its contents. Use `small`, `medium`, `large` for predefined standard widths. You can also set a custom width `{number}rem` or use `stretch` or `false`.',
```
joakbjerk, line 30: And mention `ch` here also

joakbjerk (approved):
Nice, works well 😄

packages/dnb-eufemia/src/extensions/forms/FieldBlock/FieldBlock.tsx
tujoworker: `px` should not be used here. When the font-sizes increases, we need the input to expand in width.

tujoworker, line 115:
```suggestion
  const hasCustomWidth = /\d(rem)$/.test(String(width))
```

packages/dnb-eufemia/src/extensions/forms/FieldBlock/FieldBlock.tsx
```diff
@@ -112,6 +112,8 @@ function FieldBlock(props: Props) {
...
+  const hasCustomContentWidth = /\d(ch|rem)$/.test(String(contentWidth))
```
tujoworker, line 116:
```suggestion
  const hasCustomContentWidth = /\d(rem)$/.test(String(contentWidth))
```

packages/dnb-eufemia/src/extensions/forms/FieldBlock/FieldBlockDocs.ts
tujoworker: I removed the `ch` support. I have not managed to be useful, as it will always be off because of the padding, and for some reason, it was inconsistent when used few or many chars.

PR #4144, opened 2024-10-17
fix(Field.Number): `decimalLimit={0}` when `currency` should work

langz:
I'll have to add the actual fix, not only the tests.
Not entirely sure what and where to put the fix.

### DIFF
--- a/packages/dnb-eufemia/src/components/input-masked/__tests__/InputMasked.test.tsx
+++ b/packages/dnb-eufemia/src/components/input-masked/__tests__/InputMasked.test.tsx
@@ -739,7 +739,7 @@ describe('InputMasked component with currency_mask', () => {
     expect(document.querySelector('input').value).toBe('1 234 NOK')
   })
 
-  it('should not append a comma when entering a dot', () => {
+  it('should not append a comma when entering a comma', () => {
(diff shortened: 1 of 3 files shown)


packages/dnb-eufemia/src/extensions/forms/Field/Currency/__tests__/Currency.test.tsx
```diff
@@ -124,6 +125,22 @@ describe('Field.Currency', () => {
     expect(input).toHaveAttribute('inputmode', 'decimal')
   })
 
+  it('should work with decimal limit', async () => {
```
langz, line 128:
```suggestion
  it('should work with decimal limit 0', async () => {
```

packages/dnb-eufemia/src/components/input-masked/__tests__/InputMasked.test.tsx
```diff
@@ -739,6 +739,12 @@ describe('InputMasked component with currency_mask', () => {
     expect(document.querySelector('input').value).toBe('1 234 NOK')
   })
 
+  it('should not append a comma when entering a dot', () => {
```
langz, line 742:
```suggestion
  it('should not append a comma when entering a comma', () => {
```

PR #4147, opened 2024-10-18
fix(Forms): keep Iterate.Array in sync with the data context and call onChange when `countPath` changes

tujoworker:


langz (approved):
🥳 🎆 🍾

PR #4150, opened 2024-10-18
feat(Upload): adds property `displayFileExtensionText` to hide file extension text

langz:


joakbjerk:
Should the .png after the filename be hidden in this screenshot with `displayFileExtensionText` set to `false`, or am I misunderstanding, or looking at the wrong place? 😄 
Screenshot 2024-10-20 at 09 38 30

langz:
> Should the .png after the filename be hidden in this screenshot with `displayFileExtensionText` set to `false`, or am I misunderstanding, or looking at the wrong place? 😄 
> 
> Screenshot 2024-10-20 at 09 38 30
> 
> 

It's this file extension text that will not be displayed:

image

Could of course be both?
Or maybe the naming of the prop or the docs of the prop could be better?

joakbjerk:
> > Should the .png after the filename be hidden in this screenshot with `displayFileExtensionText` set to `false`, or am I misunderstanding, or looking at the wrong place? 😄
> > Screenshot 2024-10-20 at 09 38 30
> 
> It's this file extension text that will not be displayed:
> 
> image
> 
> Could of course be both? Or maybe the naming of the prop or the docs of the prop could be better?

Hmm, not sure how to improve the text, it makes sense when you made me aware that the file extensions are shown in the list, I just didn't know that before 😄

joakbjerk (approved):

langz:
> > > Should the .png after the filename be hidden in this screenshot with `displayFileExtensionText` set to `false`, or am I misunderstanding, or looking at the wrong place? 😄
> 
> > > Screenshot 2024-10-20 at 09 38 30
> 
> > 
> 
> > It's this file extension text that will not be displayed:
> 
> > 
> 
> > image
> 
> > 
> 
> > Could of course be both? Or maybe the naming of the prop or the docs of the prop could be better?
> 
> 
> 
> Hmm, not sure how to improve the text, it makes sense when you made me aware that the file extensions are shown in the list, I just didn't know about that 😄

I'll try to improve it🤞

langz:
Turns out we don't need this now 🙏

PR #4151, opened 2024-10-18
feat(Upload): `acceptedFileTypes` is optional

langz:
Additional thoughts for some functionality:

I should probably make a different pro for displaying hiding the text/translation?
As I believe it should be possible to send in accepted file types even though not displaying the text which informs the user which file types that's accepted.


packages/dnb-eufemia/src/components/upload/__tests__/UploadFileInput.test.tsx
```diff
@@ -159,6 +159,20 @@ describe('UploadFileInput', () => {
...
+      wrapper: makeWrapper({
+        acceptedFileTypes: undefined,
+      }),
+    })
+
+    const inputElement = document.querySelector(
+      '.dnb-upload__file-input'
+    ) as HTMLInputElement
+
+    expect(inputElement.accept).toBe('')
```
langz, line 173: When not providing any `acceptedFileTypes` for now, it will render an "empty" `accept` attribute in the input element.

Is this the same as not having the `accept` attribute, or should we not set/have a `accept`property when not limiting the files based on file types.


https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/accept

joakbjerk: I'm not a 100% sure, but it seems like setting the `accept` to `""` makes it behave like everything is accepted? 🤔

joakbjerk (approved):
Nice, looks good to me 🫡

packages/dnb-eufemia/src/components/upload/__tests__/UploadFileInput.test.tsx
langz: Everything should be accepted, and it works as expected.

My thought was just if it was best to `<input accept="" />` or just `<input />`. In the code as of now, we do the first variant.

PR #4172, opened 2024-10-23
fix(Forms): only run onBlurValidator when no other errors are present

### LINKED ISSUE #4074: bug(Field.String): onBlurValidator error message is displayed instead of the required error message after removing value
## 🐛 Bug Report

`onBlurValidator` error message is displayed instead of the required error message after removing value.
I would expect to see the required error message(which I do initially), and not "My Error B".

## To Reproduce

https://codesandbox.io/p/sandbox/busy-sunset-lrshrf?workspaceId=9eb653cb-4406-4b8f-bc96-714927cc7fcd

- Initially the error message is required(Dette feltet må fylles ut.) which is expected.
- Enter 123 and tab out, the error message is from bValidator(My Error B) which is expected.
- Remove 123 and tab out, the error message is from bValidator(My Error B) which is not expected, I would expect the error message to be required(Dette feltet må fylles ut.).

Video of the steps above:

https://github.com/user-attachments/assets/ed370d1d-b32f-46e5-a1f8-eb1d2db71276

tujoworker:
Fixes #4074

With an exception for when `onChange` or `validator` is async. Then we still want to show the `onBlurValidator`.

**An example**

We have a slow `onChange` event. 

- The user types.
- `onChange` fires and is ongoing while the user types more.
- Before `onChange` has finished, the user blurs the field.
- The `onBlurValidator` fails.
- We show the error coming from the `onBlurValidator`.

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/DataContext/Provider/__tests__/Provider.test.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/DataContext/Provider/__tests__/Provider.test.tsx
@@ -1192,14 +1192,17 @@ describe('DataContext.Provider', () => {
 
     it('should evaluate sync validation, such as required, before continue with async validation', async () => {
       const onSubmit: OnSubmit = jest.fn(async () => {
+        await wait(10)
         return { info: 'Info message' } as const
       })
       const validator = jest.fn(async (value) => {
+        await wait(10)
         if (value === 'validator-error') {
           return new Error('validator-error')
         }
       })
       const onBlurValidator = jest.fn(async (value) => {
+        await wait(10)
         if (value === 'onBlurValidator-error') {
           return new Error('onBlurValidator-error')
         }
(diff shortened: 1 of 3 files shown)


langz (approved):
Great job. 
I've only tested the Field.NationalIdentity component in tve deploy preview of this PR, and that works as expected. 
However, it could also be nice to recreate the CSB from the issue and test that. I'm not on a laptop right now, so can't do it myself.

langz:
I've forked the CSB provided in the [issue](https://github.com/dnbexperience/eufemia/issues/4074), [here it is](https://codesandbox.io/p/sandbox/sharp-frog-sn4vqr?workspaceId=9eb653cb-4406-4b8f-bc96-714927cc7fcd).

langz:
PR title says, _fix(Forms): only run onBlurValidator when no other errors are present_, do we need to apply the same code for validator(onChangeValidator)?

I'll try to make a CSB and see.

PR #4176, opened 2024-10-24
fix(Forms): render multiple (combined) Ajv errors with translated messages

### LINKED ISSUE #4052: Forms: Combined error messages are inflexible and presented in English
## 🐛 Bug Report

When rendering a field with multiple validation props, a combined error message is shown. Example:

```
must NOT have more than 3 characters and must match pattern "^[a-zA-Z]$"
```

This error message has multiple problems:

- It is in English, unlike other messages
- It doesn't start with a capital letter
- It doesn't end with a period
- It is not customizable (in full or in part)

Here's an example from our application:

CleanShot 2024-10-03 at 16 29 32@2x

## To Reproduce

[Here is a CodeSandbox demonstrating the issue](https://codesandbox.io/p/sandbox/frosty-wave-s5tvmm)

## Expected behavior

The error message is displayed in Norwegian or English, depending on the set language. As a bonus, It would be nice to be able to customize (showing the pattern isn't great, for instance)

## Eufemia Version

**Browser JS:** 10.51.1

**Browser CSS:** 10.51.1

tujoworker:
Fixes #4052


langz (approved):
Looks great 😃 


Would it add value to have a visual regression test for this as well?

tujoworker:
We do have [one](https://github.com/dnbexperience/eufemia/blob/e6d3dce424bf9c5896569368e95834fbef90a7ed/packages/dnb-eufemia/src/extensions/forms/FieldBlock/__tests__/__image_snapshots__/fieldblock-for-ui-have-to-match-combined-statuses.snap.png). But I agree, it would be good to have one more, so I added one in a new commit.

PR #4162, opened 2024-10-22
feat(Forms): deprecate Ajv `validationRule` in FormError and deprecate `errorMessages` keys like `pattern` in favor of Eufemia translation keys like `Field.errorPattern`

tujoworker:
With this change, we move

PR #4177, opened 2024-10-24
feat(Forms): deprecate `useErrorMessage` hook for when creating your own fields

tujoworker:
This PR covers the deprecation (change log) of `useErrorMessage` – which was not mentioned in the PR #4162  where the implementation was done.

```tsx
// From
const MockComponent = (props) => {
  const errorMessages = useErrorMessage(
    props.path,
    props.errorMessages,
    {
      'Field.errorRequired': 'bar',
    }
  )

  return (
    <Field.String errorMessages={errorMessages} />
  )
}

// To
const MockComponent = (props) => {
  const errorMessages = useMemo(() => {
    return {{ 'Field.errorRequired': 'bar' }, ...props.errorMessages}
  }, [props.errorMessages])

  return (
    <Field.String errorMessages={errorMessages} />
  )
}
```


langz (approved):
🫡
Based on the PR description/comment _ This PR covers the deprecation (change log)_, should I expect a change in some kind of changelog file in this PR?

tujoworker:
We did [change the docs](https://github.com/dnbexperience/eufemia/pull/4162/files#diff-6dfa8fcaeaaf9b3a739398ce6b227f27cb1babfb5342defc75d8423f6942ce3e) in this PR #4162 – I think that's good enough for now.

PR #4181, opened 2024-10-25
feat: add CountryFlag component

tujoworker:
The solution:

- Use the same source as UX in their designs.
- Provide a `background-image` so the browser only downloads SVGs which are rendered/displayed.
- Only support 1x1 versions.
- Compilers still needs to "move" the SVG defined in the CSS, but much less as we don't add support for the 4:3 versions.
- That's the reason why devs need to import the extra CSS file.

## Usage

```tsx
import { CountryFlag } from '@dnb/eufemia'

// Import the flag icons as CSS
import '@dnb/eufemia/components/country-flag/style/dnb-country-flag-icons.min.css'

// ... or as SASS
import '@dnb/eufemia/components/country-flag/style/dnb-country-flag-icons.scss'

render(<CountryFlag />)
```

## Sizes

For now the sizes are the same as the Avatar component, except that here we have a `x-small` version as well. I think that's needed here, in order to use flags in a 16x16px format.

When it comes to the other sizes, should we align them really? Should other sizes be supported? 

## Example

PR Preview: https://eufemia-git-feat-country-flag-eufemia.vercel.app/uilib/components/country-flag/
Devbox: https://codesandbox.io/p/devbox/2h549y?migrateFrom=x36wyd


boyum:
Will this affect output sizes?

boyum:
Have we considered extracting country names and flags into a separate package? Changes to country names and flags happen and we sometimes want to fetch the newest country information but not care about breaking changes to other components.

tujoworker:
Meeting notes:

- [ ] Default size should be like Icon component.
- [ ] Should work in places where the Icon component work.

tujoworker:
> Will this affect output sizes?

Yes and no. 

- Not as ling as you don't include/import this file `dnb-country-flag-icons.min.css`
- When it is imported, the bundler, like Webpack, will move the SVG files to your bundle assets.
- But they will not be a part of your application bundle.
- The browser will lazy load rendered/visible flags as SVG files.
- This technique does not work in a classical Codesandbox, but does work in a [Devbox](https://codesandbox.io/docs/learn/devboxes/overview).

tujoworker:
> Have we considered extracting country names and flags into a separate package? Changes to country names and flags happen and we sometimes want to fetch the newest country information but not care about breaking changes to other components.

It would not necessarily mean less work for all parties. Also, we already have the [SelectCountry](https://eufemia.dnb.no/uilib/extensions/forms/feature-fields/SelectCountry/) field. In general, having found the correct list of county names etc. means we should not need to change this list too often anymore.

langz (approved):
I've tested it, and I think it works great 🚀

PR #4187, opened 2024-10-28
fix(Forms): updates country names in SelectCountry

Vikl5:
To have a more common approach to country names across DNB I wanted to update some of the country names in Eufemia which is different from what the KycProfileV2 API provides. I had a discussion with the KycProfileV2 API owner and we reached an agreement regarding which country names they should provide.
This way we avoid having different entities in DnB supplying their own version of country names.
 
Please let me know if some of these country names are not in accordance with your views and we could have a discussion :)

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/constants/countries.ts
+++ b/packages/dnb-eufemia/src/extensions/forms/constants/countries.ts
@@ -318,7 +318,7 @@ const countries: Array<CountryType> = [
   },
   {
     i18n: {
-      en: 'Virgin Islands (British)',
+      en: 'British Virgin Islands',
       nb: 'De britiske Jomfruøyer',
     },
     cdc: '1-284',
--- a/packages/dnb-eufemia/src/extensions/forms/constants/countries.ts
+++ b/packages/dnb-eufemia/src/extensions/forms/constants/countries.ts
@@ -2170,7 +2170,7 @@ const countries: Array<CountryType> = [
   },
   {
     i18n: {
-      en: 'United States of America',
+      en: 'United States',
       nb: 'USA',
     },
     cdc: '1',
--- a/packages/dnb-eufemia/src/extensions/forms/constants/countries.ts
+++ b/packages/dnb-eufemia/src/extensions/forms/constants/countries.ts
@@ -103,7 +103,7 @@ const countries: Array<CountryType> = [
   {
     i18n: {
       en: 'Antarctica',
-      nb: 'Antarktis',
+      nb: 'Antarktika',
     },
     cdc: '672',
     iso: 'AQ',
@@ -318,8 +318,8 @@ const countries: Array<CountryType> = [
   },
   {
     i18n: {
-      en: 'British Virgin Islands',
-      nb: 'De britiske Jomfruøyene',
+      en: 'Virgin Islands (British)',
+      nb: 'De britiske Jomfruøyer',
     },
     cdc: '1-284',
     iso: 'VG',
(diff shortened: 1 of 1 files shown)


packages/dnb-eufemia/src/extensions/forms/constants/countries.ts
```diff
@@ -103,7 +103,7 @@ const countries: Array<CountryType> = [
   {
     i18n: {
       en: 'Antarctica',
-      nb: 'Antarktis',
+      nb: 'Antarktika',
```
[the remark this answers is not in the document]
Vikl5: I agree.

packages/dnb-eufemia/src/extensions/forms/constants/countries.ts
```diff
@@ -2125,8 +2125,8 @@ const countries: Array<CountryType> = [
   },
   {
     i18n: {
-      en: 'U.S. Virgin Islands',
-      nb: 'De amerikanske Jomfruøyene',
+      en: 'Virgin Islands (U.S)',
```
Vikl5, line 2128: I suggest we revert this back to 'U.S. Virgin Islands' for consistency. [Source.](https://www.britannica.com/place/United-States-Virgin-Islands)

tujoworker: `Virgin Islands (U.S)` is probably the more correct choice with the source of https://service.unece.org/trade/locode/vi.htm and https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes

packages/dnb-eufemia/src/extensions/forms/constants/countries.ts
```diff
@@ -2161,16 +2161,16 @@ const countries: Array<CountryType> = [
   },
   {
     i18n: {
-      en: 'United Kingdom',
-      nb: 'Storbritannia',
+      en: 'United Kingdom of Great Britain and Northern Ireland',
```
tujoworker, line 2164:
```suggestion
      en: 'United Kingdom',
```

In the [UN list](https://unece.org/trade/cefact/unlocode-code-list-country-and-territory) it's only [United Kingdom](https://service.unece.org/trade/locode/gb.htm).

But then again, in the [ISO 3166 list](https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes) its `United Kingdom of Great Britain and Northern Ireland`.

That's confusing.

Vikl5: Yes it is a bit confusing. One might argue that if a user is sending money to North-Ireland they would take the closest name which is Ireland and thus submitting the wrong country name.

I think the most important part is just to have something we all agree on 🙂

tujoworker: "United Kingdom" seems to be the widely recognized short form, while "United Kingdom of Great Britain and Northern Ireland" is the official long form used in formal and legal documents, especially in standards like ISO.

Based on this, I think we should opt for "United Kingdom".

packages/dnb-eufemia/src/extensions/forms/constants/countries.ts
```diff
@@ -2170,7 +2170,7 @@ const countries: Array<CountryType> = [
   },
   {
     i18n: {
-      en: 'United States',
+      en: 'United States of America',
```
tujoworker, line 2173:
```suggestion
      en: 'United States',
```

"United States" is the widely recognized short form, while "United States of America" is the official long form used in formal and legal documents, particularly in standards like ISO.

Based on this, it makes sense to opt for "United States" in our context.

PR #4189, opened 2024-10-28
fix(Forms): fix schema validation for required paths with matching name

### LINKED ISSUE #4179: JSON schema validation is unpredictable
## 🐛 Bug Report

When using JSON schema for validation, it seems as if `required` fields are determined based on substring pattern matching, which yields some bad results.

When setting a required field at root level, fields in the form seem to match their path against the root level required field using substring pattern matching, which means that 
```json
{
  type: 'object',
  required: ['nameAndWhyIsThisCausingAValidationError'],
}
```
results in a validation error on
```jsx
<Form.Section path="/parent">
  <Field.String path="/something/name" />
</Form.Section>
```
because of a match on `name` (my assumption), even though the paths are not present in the JSON schema at all.
## To Reproduce

https://codesandbox.io/p/sandbox/schema-eager-validation-934395

## Expected behavior

I expect Eufemia Forms to respect the structure of the JSON schema when performing validation, and to match field paths with JSON Schema properties and values of the `required` array at the correct level in the tree structure of the schema.

## Eufemia Version

10.53.0

tujoworker:
Fixes #4179

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Form/Section/__tests__/Section.test.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Form/Section/__tests__/Section.test.tsx
@@ -948,7 +948,7 @@ describe('Form.Section', () => {
       ).toHaveAttribute('aria-required', 'true')
     })
 
-    it('should set "required" for firstName with nested schema', () => {
+    it('should set "required" in schema with section', () => {
       const schema: JSONSchema = {
         type: 'object',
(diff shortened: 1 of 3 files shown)


langz (approved):
Looks good, 
In the PR title is "eager" the correct word? Did it validate "too early"? 

It's just me who don't fully understand what the problem is, so eager could be a fitting word🤣👍

PR #4192, opened 2024-10-29
fix(NumberFormat): should not throw exception when providing an invalid currency

langz:
In this PR, not providing a valid currency when providing value `null` will result in `"null"`. 
And when not providing a valid currency when providing value `undefined` will result in `"undefined"`.  

I don't think it's optimal, but I think it's good enough, as I believe this is an edge-case.

Initially, I would want it to result in the same as using default currency(when not providing a valid one).
Which is `NOK 0.00`.
But to figure out that you have not provided a valid currency, one would have to check up against all valid currencies(https://en.wikipedia.org/wiki/ISO_4217), which is too much job 😅 

I guess we could just return `0` if value is `null` or `undefined`, and then don't care about the other options/properties? 🤔


tujoworker (approved):
I would check for typeof here, just because of the possibility of a falsy value.

PR #4193, opened 2024-10-30
chore: filed -> field

langz:


packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Handler/info.mdx
```diff
@@ -244,7 +244,7 @@ The callback function receives the following arguments as an object:
 - `props` The given field properties.
 - `internal` The `internal` parameter contains `{ error: Error | undefined }` you can utilize if needed.
 
-**displayValue** can be `undefined` if a filed does not support it, or it's value is not set (`emptyValue`).
+**displayValue** can be `undefined` if a field does not support it, or it's value is not set (`emptyValue`).
```
langz, line 247: Should we also add which fields that do not support it? Or are we fine without doing so?

tujoworker: Having a list will get out of sync too fast. So I think it's just good to state as so to have an argument "why" you get `undefined`.

tujoworker (approved):

PR #4198, opened 2024-10-30
chore(Portal): fix Tools.Log exception in portal example (Iterate.Array)

tujoworker:
This fixes an issue occurring in this example: https://eufemia.dnb.no/uilib/extensions/forms/Iterate/Array/#minium-one-item


langz (approved):
❤️

PR #4199, opened 2024-10-30
feat(Forms): add `onDone`, `onCancel` and `onEdit` to Form.Section containers

tujoworker:
This allows to listen on the Toolbar button events:

```tsx
<Form.Section>
  <Form.Section.EditContainer onDone={onDone} onCancel={onCancel}>
    <Field.Name required path="/name" />
  </Form.Section.EditContainer>

  <Form.Section.ViewContainer onEdit={onEdit}>
    <Value.Name path="/name" />
  </Form.Section.ViewContainer>
</Form.Section>
```

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Form/Section/EditContainer/EditContainer.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Form/Section/EditContainer/EditContainer.tsx
@@ -15,12 +15,15 @@ import { Path } from '../../../types'
 
 export type Props = {
   title?: React.ReactNode
+  onDone?: () => void
+  onCancel?: () => void
 }
 
 export type AllProps = Props & SectionContainerProps & FlexContainerProps
 
 function EditContainer(props: AllProps) {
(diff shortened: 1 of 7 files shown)


langz (approved):
Wow, that was fast ⚡ 
I just finished reading the Slack post 😀

langz:
Should we add these properties to an own events tab in our docs? As of right now, these new events is listed under properties. But usually I would expect us to list them in the events tab(which do kot exist for Edit and ViewContainer for now). 

@tujoworker

tujoworker:
We should avoid the Events tab in Eufemia forms. We may even move all events to the Properties table at some point. But I agree that we can add these in a separate table. What do you think?

langz:
> We should avoid the Events tab in Eufemia forms. We may even move all events to the Properties table at some point. But I agree that we can add these in a separate table. What do you think?

Aha, ok so we should avoid it in Eufemia forms. 
Fine by me. 
But based on the «standard approach» used in our regular/base components, one would expect a events tab. 

We can disucuss this some other time✅💚

PR #4204, opened 2024-10-31
chore(PostalCodeAndCity): adds custom matching translations to example

langz:


joakbjerk (approved):
Nice catch 🔥

joakbjerk:
> Opening this PR, as I think it's a strange error pattern message when pattern is 5 digits. See how it works in this PR: https://eufemia-git-chore-adds-docs-for-error-pattern-me-576ff9-eufemia.vercel.app/uilib/extensions/forms/feature-fields/PostalCodeAndCity/demos/#non-norwegian-postal-codes
> 
> See how it works on main today: https://eufemia.dnb.no/uilib/extensions/forms/feature-fields/PostalCodeAndCity/demos/#non-norwegian-postal-codes
> image
> 
> But I also think it's a bit strange that one would set the pattern, mask, placeholder for a 5-digit postalcode one place, and the error message for the pattern in a translation....

I agree that it's somewhat cumbersome to split up the customization like that 🤔

PR #4205, opened 2024-10-31
fix(Upload): unnecessary spacing when not providing `title` and `text`

langz:


packages/dnb-eufemia/src/components/upload/UploadInfo.tsx
```diff
@@ -57,19 +58,18 @@ const UploadInfo = () => {
     displayFilesAmountLimitItem
 
   return (
-    <>
-      {title && <Lead space="0">{title}</Lead>}
+    <Flex.Stack gap="small">
+      {(title || text) && (
+        <Flex.Stack gap="xx-small">
```
langz, line 63: You could say this Flex component is not needed if there's only a `title` or a `text` provided, and not both.
But not sure if it's disturbing to have the Flex.Stack wrapper in that scenario.

packages/dnb-eufemia/src/components/upload/__tests__/__image_snapshots__/upload-for-ui-have-to-match-file-max-size-based-on-file-type-table.snap.png
langz, line 1: I have to fix this.

langz: Fixed

tujoworker (approved):

PR #4207, opened 2024-11-01
feat(Forms): add `transformLabel` to Value.Composition

tujoworker:
We do that by moving the `transformLabel` logic from `useValueProps` to `ValueBlock`.


langz (approved):
🚀 
Tested it in the deploy preview, and it works as expected 🥇

PR #4210, opened 2024-11-04
fix(Upload): alignment when displaying single DL item

langz:


packages/dnb-eufemia/src/components/upload/__tests__/Upload.screenshot.test.ts
```diff
@@ -45,6 +45,7 @@ describe.each(['ui', 'sbanken'])('Upload for %s', (themeName) => {
 
   it('have to match disabled file max size', async () => {
     const screenshot = await makeScreenshot({
+      style: { height: '14rem' },
```
langz, line 48: I had to add this.
I think it's because the height of the component is a bit different dependent on the theme.
Not 100% sure why, perhaps font size is different based on sbanken and ui theme.
Got this error message without setting height:

Screenshot 2024-11-04 at 20 54 43

tujoworker: Did you try to use `recalculateHeightAfterSimulate: true`?

tujoworker (approved):
Nice work 👏

(Have a look at my only comment)

packages/dnb-eufemia/src/components/upload/__tests__/Upload.screenshot.test.ts
langz: I have, but will try again before merging 👍

langz: Got the same error:
image

PR #4229, opened 2024-11-06
feat(Forms): add `hideDivider` prop to Iterate.Toolbar

andlbrei:
This PR serves two purposes, based on me trying to implement a design...

1. To be able to implement the design I need to remove the divider from the toolbar. So I introduced a prop to do this, so that I can move the toolbar around.
2. I made a POC of an implementation of our design as a story in Storybook. I would like to know your thoughts on it.  
At first I tried making another version of `ViewContainer`, but I quickly figured out that when I stripped the component down, I was left with `ArrayItemArea`, which is available for consumers, but seems to be made for internal use.

I think this serves as a good example on how to make custom iterate-based components, which is nice to have IMO.

Here is the design I want to implement as well as the POC story:
image

---

image

Storybook POC
image

(I wish the height of the `Card` component was easy to configure...)

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Iterate/Toolbar/Toolbar.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Iterate/Toolbar/Toolbar.tsx
@@ -8,18 +8,26 @@ import ToolbarContext from './ToolbarContext'
 import FieldBoundaryContext from '../../DataContext/FieldBoundary/FieldBoundaryContext'
 import { useTranslation } from '../../hooks'
 
+type ToolbarProps = {
+  hideDivider?: boolean
+}
+
 export type ToolbarParams = {
   index: number
   items: Array<unknown>
   value: unknown
 }
-export type Props = Omit<SpaceAllProps, 'children'> & {
-  children?: React.ReactNode | ((params: ToolbarParams) => React.ReactNode)
-}
+export type Props = ToolbarProps &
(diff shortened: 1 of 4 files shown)


packages/dnb-eufemia/src/extensions/forms/Iterate/Toolbar/Toolbar.tsx
```diff
@@ -48,16 +56,14 @@ export default function Toolbar({
 
   return (
     <Space
-      top="medium"
+      top="large"
```
andlbrei, line 59: I moved the `x-small` top space from the toolbar `Flex.Horizontal` and "added" it to this `top` to maintain the same top space.
The reason for this is to be able to override this `top` via the `rest` props.

tujoworker:
Thank you for the PR 🫶 I'll look next week into it.

andlbrei:
@tujoworker will you have time this week? 🙏

tujoworker:
@andlbrei Yes! We will provide two new properties to make that easy for your to get.

andlbrei:
> @andlbrei Yes! We will provide two new properties to make that easy for your to get.

Thanks!

tujoworker:
I think we are better off with two new properties and some custom code to achieve this. Here is a PR # 4329.

andlbrei:
@tujoworker looks good!

andlbrei:
Solved in https://github.com/dnbexperience/eufemia/pull/4329

PR #4232, opened 2024-11-06
feat(CountryFlag): add iso to properties table

thaytharma:
Added `iso`-field to the properties table of `CountryFlag`.


langz (approved):
❤️

PR #4225, opened 2024-11-06
fix(Forms): should display error underneath fields when nested inside Field.Selection or Field.ArraySelection  

langz:
As of now, this PR just adds an example of what should be fixed/improved upon.

Here's a link to the example in the [deploy preview](https://eufemia-git-chore-add-selection-example-eufemia.vercel.app/uilib/extensions/forms/base-fields/Selection/demos/#radio-nesting-advanced), and here [locally](http://localhost:8000/uilib/extensions/forms/base-fields/Selection/demos/#radio-nesting-advanced).

The challenge/problem is that error messages for fields nested inside an `Field.Selection` is not displayed underneath each respective field, but rather merged and displayed at the bottom of the `Field.Selection`, like so:

image

PR #4236, opened 2024-11-06
fix: ensure components having `fieldset` inside still can use spacing

tujoworker:
For fields/components that uses a `fieldset` the `gap` spacing did not apply properly:

Screenshot 2024-11-06 at 14 21 27

---

I'm not 100% sure why we not did always make the margin check. But it should actually not effect other styles, as we now check if a spacing class is applied or not.

When this is in main, we may add an example to #4225


langz (approved):
Great, thanks for improving.
I did see that the styling/spacing in https://github.com/dnbexperience/eufemia/pull/4225 was a bit off 😅

PR #4245, opened 2024-11-08
chore(Anchor): make it possible to remove launch icon

langz:
I think it should be possible to remove(not display launch icon) when an anchor has target '_blank'.

It also makes the Anchor more usable, as that's the only reason why it was not used in Upload, and rather an `<a>` with class `'dnb-anchor'` was used.

### DIFF
--- a/packages/dnb-eufemia/src/components/anchor/Anchor.tsx
+++ b/packages/dnb-eufemia/src/components/anchor/Anchor.tsx
@@ -125,7 +125,7 @@ export function AnchorInstance(localProps: AnchorAllProps) {
   const _opensNewTab = opensNewTab(allProps.target, href)
   const showLaunchIcon =
     _opensNewTab &&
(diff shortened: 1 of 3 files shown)


packages/dnb-eufemia/src/components/anchor/Anchor.tsx
```diff
@@ -125,7 +125,7 @@ export function AnchorInstance(localProps: AnchorAllProps) {
   const _opensNewTab = opensNewTab(allProps.target, href)
   const showLaunchIcon =
     _opensNewTab &&
-    !className?.includes('dnb-anchor--no-icon') &&
```
langz, line 128: I should still have this line...

langz: I've added this line back again now :)

tujoworker (approved):

PR #4246, opened 2024-11-08
fix(Forms): remove extra space from Value.* components with `inline` property

tujoworker:
Before:
Screenshot 2024-11-08 at 13 12 28

After:
Screenshot 2024-11-08 at 13 12 46


langz (approved):
Thanks ❤️

PR #4244, opened 2024-11-07
feat(Card): add style for nested cards

tujoworker:
The same style as we have in the Edit- and ViewContainer's: https://eufemia.dnb.no/uilib/extensions/forms/Iterate/ViewContainer/demos/

We will get an additional approval from Hens tomorrow.

👁️ [PR Example](https://eufemia-git-feat-card-nested-eufemia.vercel.app/uilib/components/card/demos/#nested-cards).

PR #4247, opened 2024-11-08
chore: align nested Cards style

tujoworker:
This is an addition to #4244

- Different outline color (`--color-black-8`) instead of inheriting the color.
- Different outline width of 1px instead of 2px.


langz (approved):
🫡

PR #4249, opened 2024-11-09
fix(Forms): console log a warning when Value.SummaryList gets an invalid child

tujoworker:
More info [here](https://dnb-it.slack.com/archives/C07JEJVSHJR/p1730882739880279?thread_ts=1730737161.151109&cid=C07JEJVSHJR).

It verifies that only allowed children are passed. 

It does this by comparing the amount of given children components vs the amount of the actual wanted components (verified via a context call). 
When there are more other children than the wanted ones, we show a warning ⚠️ 

I'm not sure if there are smarter ways to do that. But as of now, I think this may work.

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Value/SummaryList/useVerifyChildren.ts
+++ b/packages/dnb-eufemia/src/extensions/forms/Value/SummaryList/useVerifyChildren.ts
@@ -1,0 +1,61 @@
+import {
+  Children,
+  Fragment,
+  isValidElement,
+  useCallback,
+  useEffect,
+  useRef,
+} from 'react'
+import { warn } from '../../../../shared/helpers'
+
+export function useVerifyChildren({
+  children,
+  message,
+  ignoreTypes,
+}: {
+  ignoreTypes?: Array<string>
+  children: React.ReactNode
+  message: string
+}) {
+  const verifyCount = useRef(0)
+  verifyCount.current = 0
+  const verifyChild = useCallback(() => {
+    verifyCount.current += 1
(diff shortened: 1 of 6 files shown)


packages/dnb-eufemia/src/extensions/forms/Value/SummaryList/useVerifyChildren.ts
```diff
@@ -0,0 +1,47 @@
...
+  Children,
+  Fragment,
+  isValidElement,
+  useCallback,
+  useEffect,
+  useRef,
+} from 'react'
+import { warn } from '../../../../shared/helpers'
+
+export function useVerifyChildren({
```
langz, line 11: Should we also add an own test file for this hook?

langz, line 11: When reading verifyChildren. 
I immediately think that it verifies children, but what does it mean to verify children?🤔

Some other potential names:
- useVerifyChildrenNotEmpty
- useVerifyChildren
- useVerifyChildrenHasElements
- useChildrenNotEmpty
- useVerifyChildrenHasItems
- useVerifyChildrenHasChildren

langz (approved):

packages/dnb-eufemia/src/extensions/forms/Value/SummaryList/useVerifyChildren.ts
langz: It's more understandable now when we've added a PR description🥳🫡

PR #4254, opened 2024-11-11
fix(DatePicker): ensure that `endMonth` does not fallback to `startMonth` when `endMonth` prop is defined

joakbjerk:


langz (approved):
🧷

PR #4256, opened 2024-11-11
chore: updates docs and types for lists who dont support skeleton

langz:
As of today the following elements DL, UL, OL [say they have support](https://eufemia.dnb.no/uilib/elements/lists/properties/#ol-properties) for skeleton and skeletonMethod, but they don't as they internally in the component/element set skeleton to false(`skeleton=false`).

This PR removes docs and type support for skeleton(as it is not actually supported) in:
- Dl
- Ul
- Ol

### DIFF
--- a/packages/dnb-eufemia/src/elements/ElementDocs.ts
+++ b/packages/dnb-eufemia/src/elements/ElementDocs.ts
@@ -20,14 +20,14 @@ export const ElementPropertiesWithoutSkeleton: PropertiesTableProps = {
 
 export const ElementProperties: PropertiesTableProps = {
   ...ElementPropertiesWithoutSkeleton,
-  children: {
-    doc: 'Contents.',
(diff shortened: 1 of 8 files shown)


joakbjerk (approved):

packages/dnb-eufemia/src/elements/ElementDocs.ts
```diff
@@ -1,21 +1,25 @@
...
+  '[Space](/uilib/layout/space/properties)': {
+    doc: 'Spacing properties like `top` or `bottom` are supported.',
+    type: ['string', 'object'],
     status: 'optional',
   },
+}
+
+export const ElementProperties: PropertiesTableProps = {
+  ...ElementPropertiesWithoutSkeleton,
   children: {
```
langz, line 23: This should of course be removed, will do so.

PR #4263, opened 2024-11-12
chore: enhance Value.SummaryList warning

tujoworker:
- Ignore Form.Visibility
- Show given props in message, so it is possible to find the source/cause of the message.


langz (approved):
Smart 💯 

Seems to be a failing test, but I'm sure we'll work it out 🚧

PR #4274, opened 2024-11-13
chore: improves wording in appearance

langz:
This is not correct/finished.

The motivation behind this change is that when reading the [description](https://eufemia.dnb.no/uilib/extensions/forms/Form/Appearance/info/#description), I don't easily understand which Field.X-components I can use or not with the `Form.Appearance`.

The description doesn't use commonly known names in the forms world as base fields or feature fields, etc, but it rather says  `provides theming of sizes of input fields, including Autocomplete and Dropdown`, thoughts:
- What's input fields?
- Why do we mention the base Eufemia component Autocomplete, instead of the Field equivalent?
- Why do we mention the base Eufemia component Dropdown, instead of the Field equivalent?

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Appearance/info.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Appearance/info.mdx
@@ -6,7 +6,7 @@ showTabs: true
 
 `Form.Appearance` is a provider for theming form fields.
 
-For now, it only provides theming of sizes of [base fields](/uilib/extensions/forms/base-fields/). See example below.
+For now, it only provides theming of sizes for [base fields](/uilib/extensions/forms/base-fields/) and all [feature fields](/uilib/extensions/forms/feature-fields/) that utilize them. See example below.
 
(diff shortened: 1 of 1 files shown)


packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Appearance/info.mdx
```diff
@@ -6,7 +6,7 @@ showTabs: true
 
 `Form.Appearance` is a provider for theming form fields.
 
-For now, it only provides theming of sizes of input fields, including Autocomplete and Dropdown. See example below.
+For now, it only provides theming of sizes of [base fields](/uilib/extensions/forms/base-fields/). See example below.
```
tujoworker, line 9:
```suggestion
For now, it only provides theming of sizes for [base fields](/uilib/extensions/forms/base-fields/) and all [feature fields](/uilib/extensions/forms/feature-fields/) that utilize them. See example below.
```

tujoworker (approved):
Yes, that makes more sense. We also should mention that feature fields also will be effected.

PR #4279, opened 2024-11-13
chore: adds changelog for v10.56.0

langz:

PR #4280, opened 2024-11-14
feat(Forms): add inline HelpButton to all Field.* components as default (with option to open in Dialog)

tujoworker:
This PR changes the existing behavior of the `help` prop for Fields that previously supported it. The content now opens inline instead of in a Dialog.

The `help` prop has now all of these optional props:

```tsx
<Field.String
  label="Label text"
  help={{
    title: 'Help is available',
    content:
      'Take the time to help other people without expecting a reward or gratitude is definitely important in living an optimistic life.',
    open: false,// only for the inline variant
    renderAs: 'dialog'
  }}
/>
```
 
Examples with inline "HelpButton":

- [FieldBlock](https://eufemia-git-feat-forms-inline-help-button-eufemia.vercel.app/uilib/extensions/forms/create-component/FieldBlock/demos/#inline-help-button-vertical-only) (several examples)
- [Form.MainHeading](https://eufemia-git-feat-forms-inline-help-button-eufemia.vercel.app/uilib/extensions/forms/Form/MainHeading/#with-helpbutton)
- [Form.SubHeading](https://eufemia-git-feat-forms-inline-help-button-eufemia.vercel.app/uilib/extensions/forms/Form/SubHeading/#with-helpbutton)
- [Field.String](https://eufemia-git-feat-forms-inline-help-button-eufemia.vercel.app/uilib/extensions/forms/base-fields/String/#with-help)
- [Field.Number](https://eufemia-git-feat-forms-inline-help-button-eufemia.vercel.app/uilib/extensions/forms/base-fields/Number/#with-help)
- [Field.Upload](https://eufemia-git-feat-forms-inline-help-button-eufemia.vercel.app/uilib/extensions/forms/feature-fields/more-fields/Upload/#with-help)
- ... many other fields do have an example as well.

**NB:** We do not document the internals of the inline version at this time. This can be included in another PR. The reason is that it would require additional tests, examples, and documentation. It is also an "isolated" feature specific to the HelpButton. However, as of this writing, I am unsure if we will ever document it, as integration is not straightforward. There are several considerations to address when implementing it.


langz:
https://github.com/user-attachments/assets/02f7fb2e-8c56-44ab-b7b0-84c1351405c5

When closing, the title/header of the message jumps a bit. Not very important, but it's noticeable.

langz (commented):
Looks very good. 
I've not read it all yet. 

- Have we added some docs/changelog presenting that the default behavior of the help text/property is now inline(previously a dialog). 
- And if some users still want to use the dialog variant, we could add what they need to do. 
- some docs about when we suggest using the dialog variant. You mentioned something in the lines of long texts, etc. 
- Is it only strings that can be provided to the help property? Or React.node's as well?

langz:
> https://github.com/user-attachments/assets/02f7fb2e-8c56-44ab-b7b0-84c1351405c5
> 
> When closing, the title/header of the message jumps/shifts upwards a bit. Not very important, but it's noticeable. 

Also, when pressing the help button the input Field gets focus/active state/styling. Is that intentional?

tujoworker:
> the title/header of the message jumps/shifts upwards a bit

This is on purpose. But badly executed. Added a 40ms delay, and now I think it fells perfect.

tujoworker:
> when pressing the help button the input Field gets focus/active state/styling. Is that intentional and desired?

This is a side effect. It's the "hover effect". First I have tried to have the HelpButton outside the label, but that creates other/larger layout/placement issues. I fear we have no choice. But I will take another look if there is way to avoid that.

tujoworker:
> Have we added some docs/changelog presenting that the default behavior of the help text/property is now inline(previously a dialog).

Not yet. I think we should add that in #4279 

```diff
```

> And if some users still want to use the dialog variant, we could add what they need to do.
    some docs about when we suggest using the dialog variant. You mentioned something in the lines of long texts, etc.
    Is it only strings that can be provided to the help property? Or React.node's as well?

tujoworker:
I'm not sure what happened here 🤷 and because it can't be repoened (why?) I create a new one #4282

PR #4281, opened 2024-11-14
feat(Forms): add `asyncFileHandler` to Field.Upload to support async file handling during upload

whitneymarkov:
Add support for asynchronously processing/handling files so a virus checker can be used with the Field.Upload component.


tujoworker (commented):
Very nice work so far. A test would be good to have. Can be the same case as the example.

langz (approved):

langz:
> Very nice work so far. A test would be good to have. Can be the same case as the example.

I've opened a PR for adding tests here, https://github.com/dnbexperience/eufemia/pull/4286

PR #4287, opened 2024-11-18
chore(Field.Date): add docs example on how to enable auto close

joakbjerk:


langz (approved):
To make it "auto close", why does it make sense to do showCancelButton={false}?

A property named autoClose sounds more straight forward. 

Or perhaps the only time we should allow auto close is when there is now show cancel button?

joakbjerk:
> To make it "auto close", why does it make sense to do showCancelButton={false}?
> 
> A property named autoClose sounds more straight forward.
> 
> Or perhaps the only time we should allow auto close is when there is now show cancel button?

Adding an `autoClose` prop would add more complexity for the internal handling of when the picker should close. The current code setup for that functionality in the `DatePicker` is written so that the picker will not auto close if the submit or reset button is displayed, as the user would have no way of actively confirming or canceling their date choice, if the calendar would close automatically 

Since `Field.Date` has `showCancelButton` as `true` by default, so the developers will have to set `showCancelButton` to false, to enable the auto close behaviour. 🙏

PR #4160, opened 2024-10-21
feat(Field.Date): add DatePickerProps

joakbjerk:
- [x] Add documentation
- [x] Add some tests
- [x] Double check if any of the props needs accommodate for field props
- [x] Fix startMonth and endMonth bugs
- [x] Fix date correction test
- [x] Merge `DatePicker` camelCase changes

PR #4290, opened 2024-11-18
chore: adds en-US as default in the translation object

langz:
Motivation is from the following discussion:
https://github.com/dnbexperience/eufemia/pull/4160#issuecomment-2482041756

### DIFF
--- a/packages/dnb-eufemia/src/shared/locales/index.ts
+++ b/packages/dnb-eufemia/src/shared/locales/index.ts
@@ -3,7 +3,7 @@ import nbNO from './nb-NO'
 import enUS from './en-US'
 import enGB from './en-GB'
 
-const locales = { ...nbNO, ...enUS ...enGB }
+const locales = { ...nbNO, ...enUS, ...enGB }
 
 export type TranslationsNbNO = (typeof locales)['nb-NO']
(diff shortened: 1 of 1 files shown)


packages/dnb-eufemia/src/shared/locales/index.ts
```diff
@@ -1,10 +1,12 @@
 // import default locales
 import nbNO from './nb-NO'
+import enUS from './en-US'
 import enGB from './en-GB'
 
-const locales = { ...nbNO, ...enGB }
+const locales = { ...nbNO, ...enGB, ...enUS }
```
langz, line 6:
```suggestion
const locales = { ...nbNO, ...enUS ...enGB }
```

packages/dnb-eufemia/src/shared/locales/index.ts
```diff
@@ -1,10 +1,12 @@
 // import default locales
 import nbNO from './nb-NO'
+import enUS from './en-US'
 import enGB from './en-GB'
 
-const locales = { ...nbNO, ...enGB }
+const locales = { ...nbNO, ...enUS ...enGB }
```
langz, line 6:
```suggestion
const locales = { ...nbNO, ...enUS, ...enGB }
```

PR #4300, opened 2024-11-19
chore: adds Indeterminate to list of base fields components

langz:


tujoworker:
Indeterminate is very rarely used. So I don't like the idea that it is listed by default. 
Option can never be used alone, so it is also not that important to have listed in "this" list.

tujoworker:
Yes, there we can/should list them.

langz:
> Yes, there we can/should list them.

Good, I've now set `componentType: 'base-toggle'` for Indeterminate, so that it's included in the list of base fields components(`ListBaseFieldComponents`).

tujoworker (approved):

PR #4304, opened 2024-11-20
fix(Icon): ensure icon name is rendered as `data-testid`

tujoworker:
Also, it corrects icon size of icon when given like this `icon={<Bell />}` to e.g. a button.


langz (approved):
🤝

PR #4305, opened 2024-11-20
fix(Forms): safeguard errorMessages to avoid infinite loops when not wrapped in useMemo

tujoworker:


packages/dnb-eufemia/src/extensions/forms/Field/PostalCodeAndCity/PostalCodeAndCity.tsx
```diff
@@ -96,17 +90,26 @@ function PostalCodeAndCity(props: Props) {
...
         label={postalCodeLabel ?? translations.PostalCode.label}
-        errorMessages={{
-          'Field.errorRequired': translations.PostalCode.errorRequired,
-          'Field.errorPattern': translations.PostalCode.errorPattern,
-          ...postalCodeErrorMessages,
-        }}
+        mask={postalCodeMask}
+        pattern={postalCodePattern}
+        placeholder={postalCodePlaceHolder}
+        errorMessages={useMemo(
```
langz, line 101: Should we do useMemo here or not?

Reason asking is because the pr title says `fix(Forms): enhance errorMessages comparison to avoid infinite loops when not wrapped in useMemo`, so one could think that it should be fine to do this without useMemo now?

Not sure what's optimal here.

packages/dnb-eufemia/src/extensions/forms/Field/PostalCodeAndCity/PostalCodeAndCity.tsx
```diff
@@ -121,11 +124,18 @@ function PostalCodeAndCity(props: Props) {
           cityClassName
         )}
         label={cityLabel ?? translations.City.label}
-        errorMessages={{
-          'Field.errorRequired': translations.City.errorRequired,
-          'Field.errorPattern': translations.City.errorPattern,
-          ...cityErrorMessages,
-        }}
+        errorMessages={useMemo(
```
langz, line 127: Should we do useMemo here or not?

Reason asking is because the pr title says `fix(Forms): enhance errorMessages comparison to avoid infinite loops when not wrapped in useMemo`, so one could think that it should be fine to do this without useMemo now?

Not sure what's optimal here.

langz (approved):

packages/dnb-eufemia/src/extensions/forms/Field/PostalCodeAndCity/PostalCodeAndCity.tsx
tujoworker: Agree totally. 

But we should still use memo, because it's more efficient. Then the comparison does never "hit" on renewers. So this PR is more a last resort solution. 

I will add more info in the code. 

The docs still mention the useMemo as before.

Issue #3793, opened 2024-08-06
Value.SummaryList Value without a label is not properly aligned


andlbrei:
## 🐛 Bug Report

A `Value.SummaryList` `Value` without a `label` will not be properly aligned because the `dt`element will still be present and take up space without having any content.

![bilde](https://github.com/user-attachments/assets/d827a3c1-b4c5-4d1f-8c0f-8bdeef017e11)

## To Reproduce

https://codesandbox.io/s/laughing-stitch-3mpr8j?file=/src/App.tsx

## Expected behavior

I would expect the element to have the same horizontal alignment as surrounding elements when no label is present, or there being a property that could be used to modify alignment. 

## Eufemia Version

10.42.0

PR #4311, opened 2024-11-20
fix(Forms): align Value.SummaryList when Value.* has no label

langz:
Fixes https://github.com/dnbexperience/eufemia/issues/3793


packages/dnb-eufemia/src/extensions/forms/ValueBlock/ValueBlock.tsx
```diff
@@ -116,9 +116,11 @@ function ValueBlock(props: Props) {
           value={{ ...summaryListContext, isNested: true }}
         >
           <Item>
-            <Dt className="dnb-forms-value-block__label">
-              {label && <strong>{label}</strong>}
-            </Dt>
+            {label && (
```
langz, line 119: I think this will work, since a DL must not have a `Dt` and `Dd` I believe, only one or more of one of them.

andlbrei: I'd like to share the code for the example I provided.
Basically the heading of the section works as a label for the name, and the name is part of a SummaryList for easy alignment.
This can be structured differently.
So if the answer is "won't fix", then we'll just do it a different way.

```
<Form.Section.ViewContainer
	title="Hovedkontaktperson"
	variant="basic"
>
	<Value.SummaryList layout="horizontal">
		<Value.Composition>
			<Value.String path="/firstName" showEmpty />
			<Value.String path="/lastName" showEmpty />
		</Value.Composition>
		<Value.PhoneNumber
			label={MainContactPerson.phoneNumber}
			path="/phoneNumber"
			showEmpty
		/>
		<Value.String
			label={MainContactPerson.emailAddress}
			path="/emailAddress"
			showEmpty
		/>
	</Value.SummaryList>
</Form.Section.ViewContainer>
```

packages/dnb-eufemia/src/elements/lists/style/lists-mixins.scss
```diff
@@ -162,6 +162,10 @@
         margin-top: 0;
         margin-right: calc(var(--column-gap) - 0.5rem);
         max-width: var(--dt-max-width);
+
+        &:empty {
+          display: contents;
```
langz, line 167: Not 100% sure if `display: contents;` is the best.
But when trying to just set `margin-top: 0;`, that resulted in spacing/gap above description 2 and the line above, see the following image:

list-have-to-match-dl-horizontal-list-without-dt-value snap

tujoworker: `display` changes the way a screen reader experience the content. Sure, its "empty". But I think we rather should fix the visual issue, than chaining the semantics.

This here should work:

```
.dnb-dl__layout--horizontal dt:empty {
  margin-right: 0;
}
```

langz: Isn't it what I tried, and said did not work?

tujoworker: Right, you don't warn it to take space in the vertical.

I'll add a class `dnb-sr-only` – it would ensure the height is 0.

packages/dnb-eufemia/src/extensions/forms/ValueBlock/ValueBlock.tsx
tujoworker: We will add a `labelSrOnly` property so you can provide a label that is not visible. But alternatively, you can move `Value.Composition` out of the `Value.SummaryList` I think.

tujoworker (approved):

packages/dnb-eufemia/src/extensions/forms/ValueBlock/ValueBlock.tsx
tujoworker: Here is the PR #4319

PR #4312, opened 2024-11-20
chore: removes new badge from components older than 10 months

langz:


langz:
> I think all of them are still relatively "fresh". Especially forms. We had it in beta for the longest time. But not as "new" so long.

Fine fine 🤝

PR #4321, opened 2024-11-22
chore(Card): use `outset` docs from Section docs

tujoworker:


tujoworker:
Because the Card has no `breakout` prop, we should not link them together like in this PR.

PR #4328, opened 2024-11-22
fix(Forms): ensure Form.Card supports spacing props

tujoworker:
Without, it will result in such wired cases like that:

https://github.com/user-attachments/assets/92866b4d-f602-4777-9ca8-df57bc49d688


langz (approved):
Nice find🥳✅

PR #4331, opened 2024-11-22
feat(Forms): add support for using a function references instead of a string based id

tujoworker:
This feature allows developers to use a function or a React Context as the reference instead of a string-based ID. This can be useful for ensuring a safe local scope, especially when multiple form handlers are present.

```tsx
const myReference= () => null

const MyField = () => {
  const { data } = Form.useData(myReference)
  return data.foo
} 

render(
<>
  <Form.Handler id={myReference}>
    ...
  </Form.Handler>
  
  <MyField />
</>
)
```


langz (approved):

langz:
We could perhaps add the example from the pr description to somewhere in our docs, just to demo what's now possible 🙏

tujoworker:
> We could perhaps add the example from the pr description to somewhere in our docs, just to demo what's now possible 🙏

I have updated the Getting started docs about it:  https://github.com/dnbexperience/eufemia/pull/4331/files#diff-684f9fff83faef515cc8ff2f630bce6096237323831336e5e853054312b9a2ff

langz (approved):

PR #4332, opened 2024-11-24
feat(Forms): introduce `decoupleForm` prop to Form.Handler

tujoworker:
This change allows devs to use the data context in a more flexible way:

Before:

```tsx
<App>
  <AppRelatedThings>
    <Form.Handler> 👈 Data context, form element and status message
      ...
    </Form.Handler>
  </AppRelatedThings>
</App>
```

After:

```tsx
<App>
  <Form.Handler decoupleForm> 👈 Data context
    <AppRelatedThings>
      <Form.Element> 👈 form element and status message
        ...
      </Form.Element>
    </AppRelatedThings>
  </Form.Handler>
</App>
```

- [x] Add docs

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Handler/info.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Handler/info.mdx
@@ -7,32 +7,57 @@ import AsyncChangeExample from './parts/async-change-example.mdx'
 
 ## Description
 
-The `Form.Handler` component provides a HTML form element.
+The `Form.Handler` is the root component of your form. It provides a HTML form element and handles the form data.
 
 ```jsx
 import { Form } from '@dnb/eufemia/extensions/forms'
 
-render(
-  <Form.Handler
-    data={existingData}
-    onChange={...}
-    onSubmit={...}
-  >
-    Your Form
-  </Form.Handler>,
-)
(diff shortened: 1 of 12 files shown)


tujoworker:
> Is the word Element important in decoupleFormElement? 

Since it's the element that detaches, I think it's good to have it. I struggle to come up with something better too. I also thought if we never decouple it from the `Form.Handler`, but introduce `Form.Provider` so it can be used with `Form.Element`, instead of a property. Maybe this approach is easier to communicate?

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Handler/info.mdx
```diff
@@ -7,28 +7,57 @@ import AsyncChangeExample from './parts/async-change-example.mdx'
...
+        </Form.Element>
+      </AppRelatedThings>
+    </Form.Handler>
+  )
+}
+```
+
+## Data handling
+
+You can access the mutate and filter data inside of the form context by using the `Form.useData` hook:
```
langz, line 55:
```suggestion
You can access, mutate and filter data inside of the form context by using the `Form.useData` hook:
```

Not sure if my suggestion is correct, but the original text did not make fully sense either with `You can access the mutate and filter data`

langz (approved):

PR #4338, opened 2024-11-25
chore(portal): expand button no longer below scrollbar

snorrekim:
* Also fixed not scrolling to current page when opening menu on small screens


langz (approved):

tujoworker (approved):
Nice! Wondering if we should move the badge as well:

Screenshot 2024-11-25 at 21 07 15

PR #4342, opened 2024-11-26
fix(DatePicker): make sure the picker and input only reacts to the props that have changed

joakbjerk:
- [x] Add test
- [x] FIx same issue in `range` mode
- [x] Update failing `Field.Date` test
- [x] Clean up `updateBasedOnProps` code

Should hopefully fix the issue described [here](https://dnb-it.slack.com/archives/CMXABCHEY/p1732031524606929)

Neither of us were able to recreate the issue in csb, so testing this manually might be a challenge

This also improves input deletion behaviour when the dates are prop controlled

Before:

https://github.com/user-attachments/assets/2331ac62-e758-44aa-8a93-fcd042a7fc6e

After:

https://github.com/user-attachments/assets/6c3aa9a3-1d01-415e-8028-6626d42b0e3c

### DIFF
--- a/packages/dnb-eufemia/src/components/date-picker/__tests__/DatePicker.test.tsx
+++ b/packages/dnb-eufemia/src/components/date-picker/__tests__/DatePicker.test.tsx
@@ -296,7 +296,7 @@ describe('DatePicker component', () => {
     expect(year.value).toBe('åååå')
   })
 
-  it('should delete input content one number at a time when `startDate` and `endDate` is prop controlled and in ranged mode', async () => {
+  it('should delete input content one number at a time when `startDate` and `endDate` is "prop controlled" and in ranged mode', async () => {
     const Component = () => {
       const [startDate, setStartDate] = useState('2024-05-01')
       const [endDate, setEndDate] = useState('2025-06-30')
--- a/packages/dnb-eufemia/src/components/date-picker/__tests__/DatePicker.test.tsx
+++ b/packages/dnb-eufemia/src/components/date-picker/__tests__/DatePicker.test.tsx
@@ -247,49 +247,49 @@ describe('DatePicker component', () => {
     expect(year.value).toBe('2024')
 
     await userEvent.click(year)
-    await userEvent.keyboard('{ArrowRight>4}{backspace}')
+    await userEvent.keyboard('{ArrowRight>4}{Backspace}')
 
     expect(day.value).toBe('17')
     expect(month.value).toBe('05')
     expect(year.value).toBe('202å')
 
-    await userEvent.keyboard('{backspace}')
+    await userEvent.keyboard('{Backspace}')
 
     expect(day.value).toBe('17')
     expect(month.value).toBe('05')
     expect(year.value).toBe('20åå')
 
-    await userEvent.keyboard('{backspace}')
+    await userEvent.keyboard('{Backspace}')
 
     expect(day.value).toBe('17')
     expect(month.value).toBe('05')
     expect(year.value).toBe('2ååå')
 
-    await userEvent.keyboard('{backspace}')
+    await userEvent.keyboard('{Backspace}')
 
     expect(day.value).toBe('17')
     expect(month.value).toBe('05')
     expect(year.value).toBe('åååå')
 
-    await userEvent.keyboard('{backspace>2}')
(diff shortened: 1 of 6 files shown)


langz:
I've not looked too much into this yet. Is there any potential high-risk consequences of this fix?
Like existing functionality that can stop working, etc?

joakbjerk:
> I've not looked too much into this yet. Is there any potential high-risk consequences of this fix? Like existing functionality that can stop working, etc?

Potentially not, since the prop change handling now should be the same as the `getDerivedStateFromProps` used in the old class component

packages/dnb-eufemia/src/components/date-picker/__tests__/DatePicker.test.tsx
```diff
@@ -223,6 +223,264 @@ describe('DatePicker component', () => {
...
+    const [day, month, year]: Array<HTMLInputElement> = Array.from(
+      document.querySelectorAll('input.dnb-input__input')
+    )
+
+    expect(day.value).toBe('17')
+    expect(month.value).toBe('05')
+    expect(year.value).toBe('2024')
+
+    await userEvent.click(year)
+    await userEvent.keyboard('{ArrowRight>4}{backspace}')
```
tujoworker, line 250:
```suggestion
    await userEvent.keyboard('{ArrowRight>4}{Backspace}')
```
For consistency I would write every command in PascalCase.

packages/dnb-eufemia/src/components/date-picker/__tests__/DatePicker.test.tsx
```diff
@@ -223,6 +223,264 @@ describe('DatePicker component', () => {
...
+    expect(year.value).toBe('åååå')
+
+    await userEvent.keyboard('{backspace}')
+
+    expect(day.value).toBe('dd')
+    expect(month.value).toBe('mm')
+    expect(year.value).toBe('åååå')
+  })
+
+  it('should delete input content one number at a time when `startDate` and `endDate` is prop controlled and in ranged mode', async () => {
```
tujoworker, line 299:
```suggestion
  it('should delete input content one number at a time when `startDate` and `endDate` is "prop controlled" and in ranged mode', async () => {
```

packages/dnb-eufemia/src/components/date-picker/__tests__/DatePicker.test.tsx
```diff
@@ -223,6 +223,264 @@ describe('DatePicker component', () => {
     ).not.toContain('dnb-date-picker--opened')
   })
 
+  it('should delete input content one number at a time when `date` is prop controlled', async () => {
```
tujoworker, line 226:
```suggestion
  it('should delete input content one number at a time when `date` is "prop controlled"', async () => {
```

packages/dnb-eufemia/src/components/date-picker/hooks/useDates.ts
```diff
@@ -65,21 +65,41 @@ export default function useDates(
   const hasDatePropChanges = useMemo(
     () =>
       Object.keys(dateProps).some((date) => {
-        return dateProps[date] !== previousDates[date]
+        const dateProp = dateProps[date]
+        const previousDate = previousDateProps[date]
+
+        const convertedDateProp = convertStringToDate(dateProp, {
```
tujoworker, line 71: When will it not return a Date instance?

tujoworker (commented):
I need to look into it a little more tomorrow. 

Looks great so far.

packages/dnb-eufemia/src/components/date-picker/hooks/useDates.ts
joakbjerk: When a developer for example sets the prop to be either null, undefined or a string value

tujoworker (approved):

packages/dnb-eufemia/src/components/date-picker/__tests__/DatePicker.test.tsx
langz: Improved in commit https://github.com/dnbexperience/eufemia/pull/4342/commits/ef4680a204c700f1a83a33e897fbcd278d7509fc

PR #4343, opened 2024-11-27
feat(Forms): enhance typing and add docs on how to deal with TypeScript types

tujoworker:
The main change is that we go from:

```tsx
type JsonObject = any
```

to

```tsx
type JsonObject = Record<string | number, unknown> | Array<unknown>
```

and enhance the TypeScript support docs for the "Getting Started" section the `Form.Handler` and `Form.Isolation` docs.

To disable types you can do so by:

```tsx
<Form.Handler<any>>
...
</Form.Handler>
```

or

```tsx
const { data } = Form.useData<any>()
```

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/DataContext/Provider/Examples.tsx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/DataContext/Provider/Examples.tsx
@@ -8,7 +8,7 @@ import {
 } from '@dnb/eufemia/src/extensions/forms'
 import { Flex } from '@dnb/eufemia/src'
 
-export const TestdataSchema: JSONSchema = {
+export const TestDataSchema: JSONSchema = {
   type: 'object',
   properties: {
     requiredString: { type: 'string' },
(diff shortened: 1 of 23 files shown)


langz (approved):
Looks good to me 🥇 
Maybe @snorrekim can take a quick look at this as well, as he seems to have good knowledge and thoughts around this in a recent Slack thread.

My only concern is if this change can/will blow up with a lot of TS errors for applications using Eufemia Forms, like it has done a few times previously, resulting in a bugfix release changing to any 😂

tujoworker:
Yes, it will influence type definitions in projects as we basically going from "any" to require some types to be defined.

langz:
> Yes, it will influence type definitions in projects as we basically going from "any" to require some types to be defined.

Maybe we'll have to add a description or some kind of guide of what the projects need to do when encountering the TS error (high probability this will happen).

Reason why I think this is smart is because we'll definitely get many messages in regards to "what to do now that my build does not work anymore", etc.

Worst-case if they don't want to type, can they just case/set it to be any on their side?

tujoworker:
Yes, they can. But I agree, we should make the docs better on that. Will commit something 👍

langz (approved):

PR #4344, opened 2024-11-29
draft(Forms): API draft for how fields can handle Bring API connections

tujoworker:
### The proposal:

- **Fully Treeshakeable:** Ensures no unused code is bundled.
- **Flexible and Extendable:** Designed to adapt to future needs.
- **Leverages Existing APIs:** Built on current field props and APIs.

### Example code:

```tsx
import { Form, Connectors } from '@dnb/eufemia/extensions/forms'

// 1. Create a context with the config
const { withConfig, handlerId } = Connectors.createContext({
  fetchConfig: {
    url: '...',
    headers: {
      'X-Mybring-API-Uid': '',
    },
  },
})

// 2. Use the context to create the onChangeValidator and onChange functions
const onChangeValidator = withConfig(Connectors.Bring.postalCode.onChangeValidator)

// Should we name ".postalCode.onChange" to ".postalCode.autocompleteOnChange" or something like that?
const onChange = withConfig(Connectors.Bring.postalCode.onChange, {
  cityPath: '/city',
})

render(
  <Form.Handler
    id={handlerId}
  >
    <Field.PostalCodeAndCity
      postalCode={{
        path: '/postalCode',
        onChange,
        onChangeValidator,
      }}
      city={{
        path: '/city',
      }}
    />
  </Form.Handler>
)
```


packages/dnb-eufemia/src/extensions/forms/Field/PostalCodeAndCity/stories/PostalCodeAndCity.stories.tsx
```diff
@@ -50,3 +51,138 @@ export function PostalCodeAndCityCountryCodeSelection() {
...
+    longitude: '10.482494731266165',
+    municipality: 'Asker',
+    municipalityId: '3203',
+    po_box: false,
+    postal_code: '1391',
+  }
+  return { postal_codes: [mockData] }
+
+  // Visit: https://cors-anywhere.herokuapp.com/corsdemo to enable this service
+  const url = `https://cors-anywhere.herokuapp.com/https://api.bring.com/address/api/no/postal-codes/${postalCode}`
```
langz, line 143: Nice nice, seems better/nice than spinning up a local webserver/api to avoid CORS 🙏

packages/dnb-eufemia/src/extensions/forms/Field/PostalCodeAndCity/stories/PostalCodeAndCity.stories.tsx
```diff
@@ -50,3 +51,138 @@ export function PostalCodeAndCityCountryCodeSelection() {
...
+  },
+}
+
+export function PostalCodeAPI_Draft() {
+  return (
+    <Form.Handler defaultData={{ postalCode: '123', city: 'Oslo' }}>
+      <Flex.Stack>
+        <Field.PostalCodeAndCity
+          // import { External } from '@dnb/eufemia/extensions/forms'
+          connector={External.BringConnector.postalCodeAndCity()}
```
langz, line 106: Can you send any `connector` that we expose?
I'm a bit unsure how we should name this.
- Instead of naming it `postalCodeAndCity` in `BringConnector`, should the naming rather reflect the actual API which is named [get postal code](https://developer.bring.com/api/postal-code/#get-postal-code-get)?
- How can we/the user know that this connector `postalCodeAndCity` is used for validating? Potentially it could be used for other stuff, like auto-suggest, API requests for other stuff, etc. Should this be included in the naming of the connector?
- Since this connector adds a `onBlurValidator`, what will happen if we add our own `onBlurValidator`? Will it extend or replace the `onBlurValidator`?

PR #4347, opened 2024-12-02
feat(DatePicker): render calendar in a react portal

joakbjerk:
- [x] Make calendar placement more precise
- [x] Make sure calendar inherits all the styles like in previous iteration (background color etc.)
- [x] Make sure `onClickOutside`does not fire when clicking anywhere on the calendar
- [x] Implement the `alignPicker` prop in the portal
- [x] Make sure portal is position correctly on initial render when `opened` is true
- [x] Fix positions based on input size
- [x] Add tests
- [x] Fix screenshot tests
- [x] Fix Implement animations
- [x] Correct days padding
- [x] FIx calendar position on screen resize
- [x] Fix calendar focus styling when opening with keyboard
- [x] Fix position on document scroll
- [x] add `skipPortal` test
- [x] add new test for `detectOutsideClick`
- [x] update z-index
- [x] create shared scope for calendar styles
- [x] Fix errors caused by moving shared container styles

### DIFF
--- a/packages/dnb-eufemia/src/components/date-picker/DatePicker.tsx
+++ b/packages/dnb-eufemia/src/components/date-picker/DatePicker.tsx
@@ -723,7 +723,7 @@ function DatePicker(externalProps: DatePickerAllProps) {
     [setOutsideClickHandler, onShow]
   )
 
-  // Make sure the triangle is positioned correctly after the picker is shown
+  // Make sure the triangle is positioned correctly after calendar is mounted
   useEffect(() => {
     if (!hidden) {
       setTrianglePosition()
--- a/packages/dnb-eufemia/src/components/date-picker/DatePicker.tsx
+++ b/packages/dnb-eufemia/src/components/date-picker/DatePicker.tsx
@@ -615,7 +615,7 @@ function DatePicker(externalProps: DatePickerAllProps) {
     useRef<DatePickerContextValues['getReturnObject']>()
   const hideTimeout = useRef<NodeJS.Timeout>()
   const outsideClick = useRef<DetectOutsideClickClass>()
-  const pickerContainerRef = useRef<HTMLDivElement>()
+  const calendarContainerRef = useRef<HTMLDivElement>()
 
   const translation = useTranslation().DatePicker
 
@@ -675,7 +675,7 @@ function DatePicker(externalProps: DatePickerAllProps) {
       // as that would lead to the ignoreElement being null/undefined,
       // since the portal has not been rendered yet,
       // causing the calendar to close when clicking on the calendar itself
-      [innerRef.current, pickerContainerRef],
+      [innerRef.current, calendarContainerRef],
       ({ event }: { event: MouseEvent | KeyboardEvent }) => {
         hidePicker({ ...event, focusOnHide: event?.['code'] })
       }
@@ -1005,7 +1005,7 @@ function DatePicker(externalProps: DatePickerAllProps) {
                 >
                   <span
                     className="dnb-date-picker__container"
-                    ref={pickerContainerRef}
+                    ref={calendarContainerRef}
                   >
                     <span
                       className="dnb-date-picker__triangle"
--- a/packages/dnb-eufemia/src/components/date-picker/DatePicker.tsx
+++ b/packages/dnb-eufemia/src/components/date-picker/DatePicker.tsx
@@ -977,7 +977,7 @@ function DatePicker(externalProps: DatePickerAllProps) {
                 />
                 <DatePickerPortal
                   show={!hidden}
-                  targetElement={innerRef.current}
+                  targetElementRef={innerRef}
                 >
                   <DatePickerRange
                     id={id}
--- a/packages/dnb-eufemia/src/components/date-picker/DatePicker.tsx
+++ b/packages/dnb-eufemia/src/components/date-picker/DatePicker.tsx
@@ -610,7 +610,7 @@ function DatePicker(externalProps: DatePickerAllProps) {
     useRef<DatePickerContextValues['getReturnObject']>()
   const hideTimeout = useRef<NodeJS.Timeout>()
   const outsideClick = useRef<DetectOutsideClickClass>()
-  const portalRef = useRef<HTMLDivElement>()
+  const pickerContainerRef = useRef<HTMLDivElement>()
 
   const translation = useTranslation().DatePicker
 
@@ -670,7 +670,7 @@ function DatePicker(externalProps: DatePickerAllProps) {
       // as that would lead to the ignoreElement being null/undefined,
       // since the portal has not been rendered yet,
       // causing the calendar to close when clicking on the calendar itself
-      [innerRef.current, portalRef],
+      [innerRef.current, pickerContainerRef],
       ({ event }: { event: MouseEvent | KeyboardEvent }) => {
         hidePicker({ ...event, focusOnHide: event?.['code'] })
       }
@@ -713,11 +713,17 @@ function DatePicker(externalProps: DatePickerAllProps) {
 
       onShow?.({ ...getReturnObject.current(event) })
 
-      setTrianglePosition()
       setOutsideClickHandler()
     },
-    [setTrianglePosition, setOutsideClickHandler, onShow]
+    [setOutsideClickHandler, onShow]
   )
+
+  // Make sure the triangle is positioned correctly after the picker is shown
+  useEffect(() => {
+    if (!hidden) {
+      setTrianglePosition()
+    }
+  }, [hidden, setTrianglePosition])
 
   // React to opened prop changes
   useEffect(() => {
@@ -902,6 +908,14 @@ function DatePicker(externalProps: DatePickerAllProps) {
     lang: context.locale,
   } as HTMLProps<HTMLSpanElement>
 
+  const portalClassNames = classnames(
+    opened && 'dnb-date-picker__portal--opened',
+    hidden && 'dnb-date-picker__portal--hidden',
+    showInput && 'dnb-date-picker__portal--show-input',
+    alignPicker && `dnb-date-picker__portal--${alignPicker}`,
+    size && `dnb-date-picker--${size}`
+  )
+
   const remainingDOMProps = validateDOMAttributes(props, attributes)
   const remainingSubmitProps = validateDOMAttributes(null, submitParams)
   const remainingPickerProps = validateDOMAttributes(
@@ -975,16 +989,21 @@ function DatePicker(externalProps: DatePickerAllProps) {
                 onSubmit={togglePicker}
                 {...statusProps}
               />
-              <span className="dnb-date-picker__container">
+
+              <DatePickerPortal
+                className={portalClassNames}
+                show={!hidden}
+                alignment={alignPicker}
+                targetElementRef={innerRef}
+              >
                 <span
-                  className="dnb-date-picker__triangle"
-                  ref={triangleRef}
-                />
-                <DatePickerPortal
-                  show={!hidden}
-                  portalRef={portalRef}
-                  targetElementRef={innerRef}
+                  className="dnb-date-picker__container"
+                  ref={pickerContainerRef}
                 >
+                  <span
+                    className="dnb-date-picker__triangle"
+                    ref={triangleRef}
+                  />
                   <DatePickerRange
                     id={id}
                     firstDayOfWeek={firstDay}
@@ -1020,8 +1039,8 @@ function DatePicker(externalProps: DatePickerAllProps) {
                     cancelButtonText={cancelButtonText}
                     resetButtonText={resetButtonText}
                   />
-                </DatePickerPortal>
-              </span>
+                </span>
+              </DatePickerPortal>
             </span>
             {suffix && (
               <Suffix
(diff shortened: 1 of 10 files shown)


langz:
Just an observation:

The datepicker modal/portal looks to have more padding in main branch, then in this PR:
Screenshot 2024-12-08 at 20 26 58


https://github.com/user-attachments/assets/3c1f479c-b375-4f51-874d-29ae2e815519

langz:
Just curious, why do we make this change? Which benefits/improvements does this give to the users?

joakbjerk:
> Just curious, why do we make this change? Which benefits/improvements does this give to the users?

It will prevent issues like [this](https://dnb-it.slack.com/archives/C01U1R7D729/p1732172974795299), where the calendar gets hidden inside the parent container of the DatePicker

langz:
> > Just curious, why do we make this change? Which benefits/improvements does this give to the users?
> 
> It will prevent issues like [this](https://dnb-it.slack.com/archives/C01U1R7D729/p1732172974795299), where the calendar gets hidden inside the parent container of the DatePicker

Nice nice, I've seen this this improvement when testing. Great 🚀

langz:
Tabbing/focusing seems to be different now, when using react portal than what's on main branch.

Video of me tabbing into the calendar. From main branch:


https://github.com/user-attachments/assets/ad678219-cb40-4e61-a81a-0778b26cbd7e



Video of me tabbing into the calendar. In this PR:

https://github.com/user-attachments/assets/dddf6610-722a-4ea6-b8b3-f5d83289c4f7

langz (approved):
I did some testing now, and it looks very good now 👏 🏅

langz (commented):
Do we have visual tests for this?
Perhaps there's no changes to them in this PR, since the new react portal code is so pixel perfect that they still work 🙏

packages/dnb-eufemia/src/components/date-picker/DatePickerPortal.tsx
```diff
@@ -0,0 +1,93 @@
...
+export default function DatePickerPortal({
+  className,
+  skipPortal,
+  alignment,
+  targetElementRef,
+  children,
+}: DatePickerPortalProps) {
+  const [position, setPosition] = useState({})
+
+  const positionTimeout = React.useRef(null)
```
tujoworker, line 25:
```suggestion
  const positionTimeout = useRef<NodeJS.Timeout>()
```

packages/dnb-eufemia/src/components/date-picker/DatePickerPortal.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+  useEffect(() => {
```
tujoworker, line 27:
```suggestion
  useLayoutEffect(() => {
```

I think we can set the position earlier, right after the DOM insertion.
Here we need the `useLayoutEffect` import for SSR compatibility.

packages/dnb-eufemia/src/components/date-picker/DatePickerPortal.tsx
```diff
@@ -0,0 +1,93 @@
...
+    }
+  }, [])
+
+  const setPositionDebounce = useCallback(() => {
+    if (targetElementRef.current) {
+      setPosition(getPosition(targetElementRef.current, alignment))
+    }
+
+    clearTimeout(positionTimeout.current)
+    positionTimeout.current = setTimeout(() => {
```
tujoworker, line 39: We may have a look into using our debounce method instead if that gives us less code.

packages/dnb-eufemia/src/components/date-picker/DatePickerPortal.tsx
```diff
@@ -0,0 +1,93 @@
...
+      if (targetElementRef.current) {
+        setPosition(getPosition(targetElementRef.current, alignment))
+      }
+    }, 200)
+  }, [alignment, targetElementRef])
+
+  useLayoutEffect(() => {
```
tujoworker, line 46:
```suggestion
  useEffect(() => {
```

This method is not so crucial to mount early.

packages/dnb-eufemia/src/components/date-picker/DatePickerPortal.tsx
```diff
@@ -0,0 +1,93 @@
...
+      window.addEventListener('scroll', setPositionDebounce)
+    }
+
+    return () => {
+      if (!skipPortal) {
+        window.removeEventListener('resize', setPositionDebounce)
+        window.removeEventListener('scroll', setPositionDebounce)
+      }
+    }
+  }, [])
```
tujoworker, line 58:
```suggestion
  }, [setPositionDebounce, skipPortal])
```

packages/dnb-eufemia/src/components/date-picker/__tests__/DatePicker.test.tsx
```diff
@@ -2650,6 +2650,49 @@ describe('Custom text for buttons', () => {
   })
 })
 
+describe('DatePickerPortal', () => {
```
tujoworker, line 2653:
```suggestion
describe('DatePickerPortal', () => {
```

A `skipPortal` test would be good to include.

packages/dnb-eufemia/src/components/date-picker/style/dnb-date-picker.scss
```diff
@@ -523,3 +523,66 @@
...
+
+.dnb-date-picker__portal {
+  // make variables available inside the portal
+  --date-picker-input-height: 2rem;
+  --date-picker-day-width: 2rem;
+  --date-picker-day-horizontal-spacing: 4px;
+
+  position: absolute;
+  left: 0;
+  z-index: 1000;
```
tujoworker, line 535:
```suggestion
  z-index: 1000;
```

I think we have documented the z-index number for other components. Have we?

packages/dnb-eufemia/src/components/date-picker/style/dnb-date-picker.scss
```diff
@@ -523,3 +523,66 @@
     opacity: 0;
   }
 }
+
+.dnb-date-picker__portal {
```
tujoworker, line 527: Many of the styles are a copy of existing styles. Maybe we should rather create a "shared" scope: `dnb-date-picker__shared-styles` or something like that.

packages/dnb-eufemia/src/shared/component-helper.js
```diff
@@ -553,12 +553,19 @@ export class DetectOutsideClickClass {
 
       // check the rest
       for (let i = 0, elem, l = ignoreElements.length; i < l; ++i) {
+        // Allow for comparing ref elements that are rendered conditionally,
+        // That might be `null` or ´undefined` during the construction stage of this class
+        const ignoreElement =
```
tujoworker, line 558: We have tests for `detectOutsideClick`, so we should add a case for this there as well.

packages/dnb-eufemia/src/components/date-picker/DatePicker.tsx
```diff
@@ -706,12 +717,18 @@ function DatePicker(externalProps: DatePickerAllProps) {
...
 
-      setTrianglePosition()
       setOutsideClickHandler()
     },
-    [setTrianglePosition, setOutsideClickHandler, onShow]
+    [setOutsideClickHandler, onShow]
   )
 
+  // Make sure the triangle is positioned correctly after calendar is mounted
+  useEffect(() => {
```
tujoworker, line 726:
```suggestion
  useLaoutEffect(() => {
```

packages/dnb-eufemia/src/components/date-picker/style/dnb-date-picker.scss
joakbjerk: Hmm, not sure 🤔 Will see if I can find something 👍

langz:
Looks like there's a snapshot that needs to be updated.
Could we also quickly go through all the previous comments/feedback and hide the outdated/fixed comments/feedback, just so that it will be easier to see what's left to do in this PR 🙏

joakbjerk:
> Looks like there's a [snapshot](https://github.com/dnbexperience/eufemia/actions/runs/12765304131/job/35579266052?pr=4347) that needs to be updated. Could we also quickly go through all the previous comments/feedback and hide the outdated/fixed comments/feedback, just so that it will be easier to see what's left to do in this PR 🙏

Should be ready now @langz

langz (approved):
I did a new round of testing now, and it works great 🚀 

One thing that seems breaking is that the following [example](https://eufemia.dnb.no/uilib/components/date-picker/demos/#opened-datepicker):
```
<DatePicker
  opened={true}
  preventClose={true}
  disableAutofocus={true}
  range={true}
  startDate="2019-05-05"
  endDate="2019-06-05"
/>
```

Will not work as intended anymore, some how as part of this PR we now have to also set `skipPortal`. Why so? Is it not possible to make the [existing example](https://eufemia.dnb.no/uilib/components/date-picker/demos/#opened-datepicker) work without `skipPortal`?
Maybe we could set `skipPortal` to `true` when `opened` is `true` Or just ignore portal/skipPortal when opened is true, as I guess it doesn't make sense to have `skipPortal={false}` when `opened={true}`?
Or will this lead to other issues?

joakbjerk:
> I did a new round of testing now, and it works great 🚀
> 
> One thing that seems breaking is that the following [example](https://eufemia.dnb.no/uilib/components/date-picker/demos/#opened-datepicker):
> 
> ```
> <DatePicker
>   opened={true}
>   preventClose={true}
>   disableAutofocus={true}
>   range={true}
>   startDate="2019-05-05"
>   endDate="2019-06-05"
> />
> ```
> 
> Will not work as intended anymore, some how as part of this PR we now have to also set `skipPortal`. Why so? Is it not possible to make the [existing example](https://eufemia.dnb.no/uilib/components/date-picker/demos/#opened-datepicker) work without `skipPortal`? Maybe we could set `skipPortal` to `true` when `opened` is `true` Or just ignore portal/skipPortal when opened is true, as I guess it doesn't make sense to have `skipPortal={false}` when `opened={true}`? Or will this lead to other issues?

The calendar will overlap the live code edit field/block/thingy if `skipPortal` is not `true`, for that example, it will also break the screenshot test that relies on that example. We could perhaps circumvent that issue if we could set a height that equals the height of the calendar, to the example area above the code edit area 🤔

langz:
> > I did a new round of testing now, and it works great 🚀
> > One thing that seems breaking is that the following [example](https://eufemia.dnb.no/uilib/components/date-picker/demos/#opened-datepicker):
> > ```
> > <DatePicker
> >   opened={true}
> >   preventClose={true}
> >   disableAutofocus={true}
> >   range={true}
> >   startDate="2019-05-05"
> >   endDate="2019-06-05"
> > />
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Will not work as intended anymore, some how as part of this PR we now have to also set `skipPortal`. Why so? Is it not possible to make the [existing example](https://eufemia.dnb.no/uilib/components/date-picker/demos/#opened-datepicker) work without `skipPortal`? Maybe we could set `skipPortal` to `true` when `opened` is `true` Or just ignore portal/skipPortal when opened is true, as I guess it doesn't make sense to have `skipPortal={false}` when `opened={true}`? Or will this lead to other issues?
> 
> The calendar will overlap the live code edit field/block/thingy if `skipPortal` is not `true`, for that example, it will also break the screenshot test that relies on that example. This i happening since the calendar now is rendered inside a portal, which is placed somewhere else on the page, with an absolute position, causing the overlap. In the portal-less `DatePicker` version, the calendar is rendered inside the component wrapper, causing the example area to have a height matching the `DatePicker` component, and no overlap.
> 
> We could perhaps circumvent that issue if we could set a height that equals the height of the calendar, to the example area above the code edit area 🤔

Aha, thanks for explaining. So it's sort of only breaking for the portal, not for actual apps?
In actual apps `opened={true}` would still work without the need for adding `skipPortal={true}`, am I correct?

joakbjerk:
> > > I did a new round of testing now, and it works great 🚀
> > > One thing that seems breaking is that the following [example](https://eufemia.dnb.no/uilib/components/date-picker/demos/#opened-datepicker):
> > > ```
> > > <DatePicker
> > >   opened={true}
> > >   preventClose={true}
> > >   disableAutofocus={true}
> > >   range={true}
> > >   startDate="2019-05-05"
> > >   endDate="2019-06-05"
> > > />
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Will not work as intended anymore, some how as part of this PR we now have to also set `skipPortal`. Why so? Is it not possible to make the [existing example](https://eufemia.dnb.no/uilib/components/date-picker/demos/#opened-datepicker) work without `skipPortal`? Maybe we could set `skipPortal` to `true` when `opened` is `true` Or just ignore portal/skipPortal when opened is true, as I guess it doesn't make sense to have `skipPortal={false}` when `opened={true}`? Or will this lead to other issues?
> > 
> > The calendar will overlap the live code edit field/block/thingy if `skipPortal` is not `true`, for that example, it will also break the screenshot test that relies on that example. This i happening since the calendar now is rendered inside a portal, which is placed somewhere else on the page, with an absolute position, causing the overlap. In the portal-less `DatePicker` version, the calendar is rendered inside the component wrapper, causing the example area to have a height matching the `DatePicker` component, and no overlap.
> > 
> > We could perhaps circumvent that issue if we could set a height that equals the height of the calendar, to the example area above the code edit area 🤔
> 
> Aha, thanks for explaining. So it's sort of only breaking for the portal, not for actual apps?
> In actual apps `opened={true}` would still work without the need for adding `skipPortal={true}`, am I correct?

Yeah, that is correct 😄

langz:
@tujoworker, do you have time to look through this?
Hopefully we can include this in [release of v10.63.0](https://github.com/dnbexperience/eufemia/pull/4450) this week 🙏

tujoworker (approved):
Very nice work 🫶

PR #4349, opened 2024-12-02
feat(Forms): add `wrapper` property to Form.Handler

tujoworker:
This allows a dev to wrap multiple providers or other wrappers, without creating a Christmas tree of wrappers/providers.

This is my use-case coming in a new PR:

```tsx
<Form.Handler
  wrapper={[Form.Status.Error, Form.Status.Success]}
>
  content
</Form.Handler>
```

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Form/Handler/Handler.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Form/Handler/Handler.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef } from 'react'
+import React, { createElement, useEffect, useRef } from 'react'
 import { JsonObject } from '../../utils/json-pointer'
 import { warn } from '../../../../shared/helpers'
 import DataContextProvider, {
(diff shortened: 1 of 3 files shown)


packages/dnb-eufemia/src/extensions/forms/Form/Handler/Handler.tsx
```diff
@@ -17,6 +17,14 @@ export type Props = FormElementProps & {
    * Will decouple the form element from rendering
    */
   decoupleForm?: boolean
+
+  /**
+   * A wrapper component to wrap the form element.
+   * You can provide an array of components to wrap the form element from right to left.
```
langz, line 23:
```suggestion
   * You can provide an array of components to wrap the form element from left to right.
```

Isn't this correct? Wouldn't right to left result in the last item in the array being at the top of the hierarchy?
Based on the tests, it seems like it works opposite. First item in array at top.

langz (approved):

packages/dnb-eufemia/src/extensions/forms/Form/Handler/HandlerDocs.ts
```diff
@@ -21,6 +21,11 @@ export const HandlerProperties: PropertiesTableProps = {
     type: 'boolean',
     status: 'optional',
   },
+  wrapper: {
+    doc: 'A wrapper component to wrap the form element. You can provide an array of components to wrap the form. Use it instead of wrapping your form with multiple providers.',
```
langz, line 25:
```suggestion
    doc: 'A wrapper component to wrap the form element. You can provide an array of components to wrap the form in. Use it instead of wrapping your form with multiple providers.',
```

tujoworker:
Thinking of it, we may avoid this pattern because it can introduce problems later on. Such as typing issue, where a prop is required etc.

PR #4350, opened 2024-12-02
feat(Forms): add `onVisible` property to Form.Visibility

tujoworker:
We forward the props to HeightAnimation, so we just need the types and the docs to make this work.

Maybe we should call it `onVisible`? But on the other hand, it's meant to be used only when `animate` is set to true.

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Form/Visibility/Visibility.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Form/Visibility/Visibility.tsx
@@ -76,6 +76,7 @@ export type Props = {
   animate?: boolean
   /** Keep the content in the DOM, even if it's not visible */
   keepInDOM?: boolean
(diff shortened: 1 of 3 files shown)


packages/dnb-eufemia/src/extensions/forms/Form/Visibility/Visibility.tsx
```diff
@@ -76,6 +76,7 @@ export type Props = {
   animate?: boolean
   /** Keep the content in the DOM, even if it's not visible */
   keepInDOM?: boolean
+  onOpen?: HeightAnimationProps['onOpen']
```
langz, line 79:
```suggestion
  /** Callback when the content is visible. Only for when `animate` is true. */
  onOpen?: HeightAnimationProps['onOpen']
```

langz (approved):
I vote for `onVisible` as well.

PR #4351, opened 2024-12-02
feat(Upload): adds support for async `onFileDelete`

langz:
Can be demoed here  https://eufemia-git-feat-on-file-delete-async-eufemia.vercel.app/uilib/components/upload/demos/#upload-with-async-onfiledelete


langz:
I think it works as intended now.
Would be great if you can take a look at the code @tujoworker, as I added `useRef` to handle getting the updated/latest files available in the async function. When not doing so, it would update the `files` using outdated `files`.

tujoworker (approved):
Looks good. I kind of miss `useCallback` in all the functions inside `UploadFileList`, but it should work OK there, as there are no deps relaying on the same instance.

Issue #4352, opened 2024-12-02
Unhandled runtime error with FormLabel.tsx


Monkeyboyen:
Works in Eufemia package 10.53.0, updated to 10.57.1 and then I get a error.

Using Formset with Formrows and FormLabel as Label to populate form I get the following error :

![Screenshot 2024-12-02 141636](https://github.com/user-attachments/assets/329a91f8-7f14-4997-9771-4083823c8c38)

langz:
Thanks so much for reporting 🙇 
I've tried making a [CSB](https://codesandbox.io/p/sandbox/inspiring-framework-n7ptzt) to reproduce this error, but have not been able to get it right yet.
Do you see anything I'm missing in the code, to be able to reproduce it?

Further, I believe the error is related to the changes done in https://github.com/dnbexperience/eufemia/pull/4282, more specifically the following line https://github.com/dnbexperience/eufemia/blob/85445b484f3b7eb85c07f4e396c9514a80275055/packages/dnb-eufemia/src/components/form-label/FormLabel.tsx#L181

langz:
@Monkeyboyen, I think and hope we were able to fix it as part of https://github.com/dnbexperience/eufemia/pull/4354
But we never really managed to reproduce the error you reported.
Could you share some more of the code you've ran which resulted in the error?

Thanks 🙇

Monkeyboyen:
@langz Hi, tested it now with 10.59.0 and its working as intended :)

PR #4354, opened 2024-12-02
fix(FormLabel): unhandled runtime error, elem is null

langz:
I think it makes sense to add the and statement: https://github.com/dnbexperience/eufemia/pull/4354/files#diff-f476207b8a837f2c0dea83ca374c5d383bcd895b0f6f10e033e12d7c0a7a3f84R142
As all following code in the if statement is dependent on that `ref.current` is existing.

Fixes https://github.com/dnbexperience/eufemia/issues/4352


tujoworker (approved):
I tried (hard) to create a test that replicates this issue, but was not able to. I really wonder how this is possible 😄 Hope we get more feedback, so we can create a test.

PR #4357, opened 2024-12-04
feat(Forms): add `Form.InfoOverlay` to display error, success (receipt), or custom messages to users

tujoworker:
PR [Preview](https://eufemia-git-feat-forms-status-eufemia.vercel.app/uilib/extensions/forms/Form/InfoOverlay/).


packages/dnb-eufemia/.releaserc2.json
langz, line 1: I believe we can remove this file 🙏

packages/dnb-eufemia/release.config.cjs
langz, line 1: I believe we can remove this file 🙏

andlbrei:
I did a quick test in CSB, looks good! 👍 There are some quirks.
- The title of the status page takes a few "tries" it seems to render in the proper style.
- I need to press submit twice to get the status page up.

And I wonder if it is possible to customize the success and error pages?

https://github.com/user-attachments/assets/50646111-ec70-4c3b-9c57-131be2db7f6d

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Status/info.mdx
```diff
@@ -0,0 +1,56 @@
...
+
+You can show the success or error message by using the `Form.Status.setStatus` method:
+
+```tsx
+Form.Status.setStatus(id, 'success')
+// or
+Form.Status.setStatus(id, 'error')
+```
+
+You can call it when ever you need to show the success message. Here is an example of how to use it:
```
langz, line 34:
```suggestion
You can call it whenever you need to show the success message. Here is an example of how to use it:
```

langz:
When "trying again", should I as a user get updated feedback that something went wrong again?
When just pressing "Prøv igjen"-button even though it nicely displays a loader in the button, I don't feel like the error message "Vi klarte ikke å sende inn skjemaet." is based on my latest action(pressing "Prøv igjen"-button).


https://github.com/user-attachments/assets/4d44887d-540a-41fa-a1a5-bd254784ef74



Perhaps we could remove the existing error message when displaying the loader in the button?
That's just a suggestion. Could possible be solved in other ways.

packages/dnb-eufemia/release.config.cjs
tujoworker: Yes 😄

tujoworker:
> but scrolls down the page, not very important though.

It is actually important that this not happens. The focus gets set initially. Which should it not. Let me fix that.

tujoworker:
@andlbrei 

> The title of the status page takes a few "tries" it seems to render in the proper style.

I think its because the heading font is a different font and needs to be downloaded. If you would have a heading on the first view, it should not happen.

> I need to press submit twice to get the status page up.

I will look into this asap.
   
And I wonder if it is possible to customize the success and error pages?

Just the texts. 

> Our design has some images in addition to the text, but I don't think that is important.

Yes. We will add the illustrations in a followup PR.

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Status/info.mdx
```diff
@@ -0,0 +1,56 @@
+---
+showTabs: true
+hideInMenu: true
+---
+
+## Description
+
+`Form.Status` shows a success or error message.
```
langz, line 8:
```suggestion
`Form.Status` shows a success or error message, and can be used when...
```

When could/should this component be used? To standardize handling successful and failing submission of a form?

langz:
Thoughts in regards to naming, `Form.Status`:
- I'm thinking about the status of the form, sort of displaying the status of the form. 
- Or something related to [GlobalStatus](https://eufemia.dnb.no/uilib/components/global-status/) or [FormStatus](https://eufemia.dnb.no/uilib/components/form-status/) 🤔

packages/dnb-eufemia/.releaserc2.json
langz, line 1: I think this can be removed 🙏

packages/dnb-eufemia/.releaserc.dry-run.json
langz, line 1: I think this can be removed 🙏

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Status/properties.mdx
```diff
@@ -0,0 +1,25 @@
...
+
+## Properties
+
+### Error
+
+<PropertiesTable props={StatusErrorProperties} />
+
+### Success
+
+<PropertiesTable props={StatusSuccessProperties} />
```
langz, line 21: We could perhaps add an example which shows how to use these properties.
I'm not 100% sure how to do so when reading the docs my self, as the examples as of now only use `Form.Status.setStatus`

packages/dnb-eufemia/src/extensions/forms/Form/Status/Status.tsx
```diff
@@ -0,0 +1,167 @@
...
+import { useTranslation } from '../../hooks'
+import MainHeading from '../MainHeading'
+import SubmitButton from '../SubmitButton'
+
+export type Props = {
+  success?: {
+    title?: React.ReactNode
+    description?: React.ReactNode
+    buttonText?: React.ReactNode
+    buttonHref?: string
```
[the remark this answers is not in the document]
tujoworker: That should now be documented.

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Status/properties.mdx
tujoworker: We don't have a prop to show the status at the moment. What could be call it? I guess just `status="error"`.

tujoworker:
> Perhaps we could remove the existing error message when displaying the loader in the button?
> That's just a suggestion. Could possible be solved in other ways.

I have changed the text. Maybe it's better now?

langz:
> > Perhaps we could remove the existing error message when displaying the loader in the button?
> > That's just a suggestion. Could possible be solved in other ways.
> 
> I have changed the text. Maybe it's better now?

I don't feel it's about the text, but that I as a user don't get any feedback when clicking "prøv igjen", and it fails over and over again. It just displays the same text. No "flashing" of the text and/or feedback. I'm not sure if I see the latest error message or the first error message.


https://github.com/user-attachments/assets/b5645da4-cb37-4418-b06a-9f51fe095996

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Status/properties.mdx
langz: I'm not sure if we are following each other here 😅 

To me, when I read the property docs https://eufemia-git-feat-forms-status-eufemia.vercel.app/uilib/extensions/forms/Form/Status/properties/#error, I would think that I as a consumer/user could use these props. How can I use the `cancelButton` property? Could we add an example or so, so that I can see how to use the prop?

packages/dnb-eufemia/src/extensions/forms/constants/locales/nb-NO.ts
```diff
@@ -34,6 +34,18 @@ export default {
...
     },
+    StatusSuccess: {
+      title: 'Takk skal du ha',
+      description: 'Vi har mottatt din informasjon.',
+      buttonText: 'Tilbake til forsiden',
+    },
+    StatusError: {
+      title: 'Beklager, noe gikk galt',
+      description:
+        'Prov igjen eller ta kontakt med oss om feilen vedstår.',
```
langz, line 45:
```suggestion
        'Prøv igjen eller ta kontakt med oss om feilen vedstår.',
```

Not sure about the "ta kontakt med oss"-text. If so, then we should perhaps provide a way of contacting us? :P

langz:
> Regarding the naming. I think I want to avoid to use submit in it. Just because it can be used in other cases as well. Like in between a Wizard step. In many ways, it acts like a "modal".
> 
> 1. And we can make it possible to render something totally custom as well (`status="custom").
> 2. And maybe we should change the "success" status to "complete", just because we then go ever further away from thinking it is just "submit" related.
> 
> Something like this?
> 
> * `Form.Message`
> * `Form.ViewMessage`
> * `Form.InfoMessage`
> * `Form.StatusView`
> * `Form.InformationView`
> * `Form.InfoPanel`
> * `Form.InfoOverlay`

I see, sounds smart :) 

Perhaps `Form.Message`? And `Form.Message.setMessage`?

But it doesn't feel like a normal component to me? Sort of like a Provider (probably not) or something else?
It's like you wrap your whole form in it, and then use Form.Message.setMessage to replace children of it with one of many messages.

packages/dnb-eufemia/src/extensions/forms/constants/locales/nb-NO.ts
tujoworker: Yeah, the original message I have seen, provides a phone number. I don't want to put that in there. But I think that this is good enough. It can vary much where and how to get in contact with support. Also, what about internal apps? Or DNB Eiendom? etc.

langz: Agree 🤝

tujoworker:
- Renamed to `InfoOverlay` with `setContent`.
- Add support to be used without Form.Handler.
- Enhanced docs with info about how to use the props and how to customize.

langz (approved):
Looks good :) 
My last thought is regarding the default texts/translations that we provide.
Should we provide them? Do we think they'll ever be used? 
Perhaps most of them will be used, so that gives some value. But I can't imagine `Vi har mottatt din informasjon.` to ever be used in an app, as it just seems to generic for me. Same for `Tilbake til forsiden`, is it that likely that they'll be navigated to the home page?

Just keep it as it is, just a few last thoughts 💚

tujoworker:
Yes, good thoughts. 
They are based on designs by Erik I got via @andlbrei. But the defaults can be adjusted for sure when ever needed. Either by us or the projects.

andlbrei:
> Yes, good thoughts. They are based on designs by Erik I got via @andlbrei. But the defaults can be adjusted for sure when ever needed. Either by us or the projects.

@langz can confirm that those are what we are going to use. 😅

PR #4359, opened 2024-12-05
feat: host fonts in the public directory

tujoworker:
- Change `playwright.config.ts` so it runs locally on post 8000.
- Copy fonts folder in /public directory so it can be used by third party apps without getting hashed.
- Document the hosted fonts on [this page](https://eufemia-git-feat-font-cdn-eufemia.vercel.app/uilib/typography/) (could not find a better place, any better ideas out there?).
- Added a e2e test to verify the fonts are located properly.
- Moved `Typography.e2e.spec.ts` from the Portal workspace to the Eufemia Workspace, because I think it's the correct place.

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/typography.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/typography.mdx
@@ -74,22 +74,22 @@ Use it either by a CSS class `.dnb-t__family--monospace` or define your own like
 }
 ```
 
-## Hosted Fonts
+## Hosted Fonts (CDN)
 
 The font files are hosted under the following URLs:
 
 ### DNB
 
-- https://eufemia.dnb.no/fonts/dnb/DNB-Regular.woff2
-- https://eufemia.dnb.no/fonts/dnb/DNB-Medium.woff2
-- https://eufemia.dnb.no/fonts/dnb/DNB-Bold.woff2
-- https://eufemia.dnb.no/fonts/dnb/DNBMono-Regular.woff2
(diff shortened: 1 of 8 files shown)


packages/dnb-design-system-portal/src/docs/uilib/typography.mdx
```diff
@@ -73,3 +73,23 @@ Use it either by a CSS class `.dnb-t__family--monospace` or define your own like
...
+- https://eufemia.dnb.no/fonts/dnb/DNBMono-Regular.woff2
+- https://eufemia.dnb.no/fonts/dnb/DNBMono-Medium.woff2
+- https://eufemia.dnb.no/fonts/dnb/DNBMono-Bold.woff2
+
+### Sbanken
+
+- https://eufemia.dnb.no/fonts/sbanken/MaisonNeue.woff2
+- https://eufemia.dnb.no/fonts/sbanken/Roboto-Regular.woff2
+- https://eufemia.dnb.no/fonts/sbanken/Roboto-Medium.woff2
+- https://eufemia.dnb.no/fonts/sbanken/Roboto-Bold.woff2
```
langz, line 95: Should all these links be absolute or could they be relative?
Would be easier to test locally and/or in deploy preview if relative.

langz (approved):
👍

packages/dnb-design-system-portal/src/docs/uilib/typography.mdx
tujoworker: I could not find a good solution at the moment. The thing is that we really just want this particular URL to be used. But I changed to be copy&paste able, instead of being links.

PR #4360, opened 2024-12-05
fix(Field.Upload): handling of multiple async file uploads

langz:
From:

https://github.com/user-attachments/assets/3fdc65d0-691e-44b0-bd9b-e97c1d359393

To:

https://github.com/user-attachments/assets/c27a0977-8553-4348-a1ee-4d0aaa6524b5

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/Upload/Upload.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/Upload/Upload.tsx
@@ -117,7 +117,7 @@ function UploadComponent(props: Props) {
 
   const { files: fileContext, setFiles } = useUpload(id)
 
-  const filesRef = useRef(null)
+  const filesRef = useRef<Array<UploadFile>>()
 
   useEffect(() => {
     filesRef.current = fileContext
--- a/packages/dnb-eufemia/src/extensions/forms/Field/Upload/Upload.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/Upload/Upload.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useMemo } from 'react'
+import React, { useCallback, useEffect, useMemo, useRef } from 'react'
 import classnames from 'classnames'
 import FieldBlock, {
   Props as FieldBlockProps,
@@ -117,6 +117,12 @@ function UploadComponent(props: Props) {
 
   const { files: fileContext, setFiles } = useUpload(id)
 
+  const filesRef = useRef(null)
+
+  useEffect(() => {
+    filesRef.current = fileContext
+  }, [fileContext])
+
   useEffect(() => {
     // Files stored in session storage will not have a property (due to serialization).
     const hasInvalidFiles = value?.some(({ file }) => !file?.name)
@@ -126,10 +132,10 @@ function UploadComponent(props: Props) {
   }, [setFiles, value])
 
   const handleChangeAsync = useCallback(
-    async (files: UploadValue) => {
+    async (existingFiles: UploadValue) => {
       // Filter out existing files
       const existingFileIds = fileContext?.map((file) => file.id) || []
-      const newFiles = files.filter(
+      const newFiles = existingFiles.filter(
         (file) => !existingFileIds.includes(file.id)
       )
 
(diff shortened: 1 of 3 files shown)


packages/dnb-eufemia/src/extensions/forms/Field/Upload/Upload.tsx
```diff
@@ -117,6 +117,12 @@ function UploadComponent(props: Props) {
 
   const { files: fileContext, setFiles } = useUpload(id)
 
+  const filesRef = useRef(null)
```
tujoworker, line 120:
```suggestion
  const filesRef = useRef<Array<UploadFile>>()
```

When we define the expected type, we don't need a default value.

packages/dnb-eufemia/src/extensions/forms/Field/Upload/Upload.tsx
```diff
@@ -117,6 +117,12 @@ function UploadComponent(props: Props) {
...
+
+  useEffect(() => {
+    filesRef.current = fileContext
+  }, [fileContext])
+
```
tujoworker, line 125:
```suggestion
```

So the trick here is to use `useEffect`, right? Else I think we don't need the `useEffect`.

packages/dnb-eufemia/src/extensions/forms/Field/Upload/Upload.tsx
```diff
@@ -140,15 +146,26 @@ function UploadComponent(props: Props) {
...
-        const uploadedFiles = updateFileLoadingState(
-          await fileHandler(newFiles),
-          { isLoading: false }
+        const incomingFiles = await fileHandler(newFiles)
+
+        const uploadedFiles = updateFileLoadingState(incomingFiles, {
+          isLoading: false,
+        })
+
+        const indexOfFirstNewFile = filesRef.current.findIndex(
```
tujoworker, line 155: What is the reason we do it that way? What if there is one file added already. And later the user adds two. Will the second one ever be `newFiles[0]`?

packages/dnb-eufemia/src/extensions/forms/Field/Upload/__tests__/Upload.test.tsx
```diff
@@ -1154,6 +1154,84 @@ describe('Field.Upload', () => {
       })
     })
 
+    it('should add new files from fileHandler with async function with multiple actions', async () => {
```
tujoworker, line 1157: Nice test 👏

packages/dnb-eufemia/src/extensions/forms/Field/Upload/Upload.tsx
langz: We can probably do it in a different way, somehow this is the only one that made sense to me at the given time.
What I do is that I find the first "new file", and use that to find the index of where the "new files" is located in the files. I so remove/replace these new files with the async received files.

So what I want to do is to update the latest files by replacing the "new files" (the files newly added by the user) with the async received files, and keeping the same order.

packages/dnb-eufemia/src/extensions/forms/Field/Upload/Upload.tsx
langz: Yes, correct.
Without it wouldn't give me the latest/synced files when the async function executes.

packages/dnb-eufemia/src/extensions/forms/Field/Upload/Upload.tsx
tujoworker: Right. Can you try if this works as well? I think it should do the same as the slicing: 

```ts
const updatedFiles = filesRef.current.map((file) => {
  const matchingNewFile = uploadedFiles.find(({ id }) => id === file.id)
  return matchingNewFile || file
})
```

langz: Thanks for the great suggestion.
The problem is that the id of the uploaded file can be something other than the local id, as it can be server generated, as in the [example](https://eufemia.dnb.no/uilib/extensions/forms/feature-fields/more-fields/Upload/demos/#with-asynchronous-file-handler) setting `server_generated_id`.

packages/dnb-eufemia/src/extensions/forms/Field/Upload/Upload.tsx
langz: I tried removing this locally, and that did not work out 🦯

tujoworker: As long as a test breaks without, I'm happy 🤗

langz: It does break. I did update the test now, so that it also takes into account loading(never ending promises).

packages/dnb-eufemia/src/extensions/forms/Field/Upload/__tests__/Upload.test.tsx
```diff
@@ -1154,6 +1154,91 @@ describe('Field.Upload', () => {
...
+                  id: 'server_generated_id_' + id,
+                  exists: false,
+                },
+              ]),
+            1
+          )
+        )
+
+      const asyncValidatorNeverResolving = () =>
+        new Promise<UploadValue>(() => {})
```
langz, line 1187:
```suggestion
        new Promise<UploadValue>(() => undefined)
```

packages/dnb-eufemia/src/extensions/forms/Field/Upload/Upload.tsx
tujoworker: Nice 🫶

tujoworker (approved):

packages/dnb-eufemia/src/extensions/forms/Field/Upload/Upload.tsx
langz: I'll merge now, then try this improvement later.
I see that this again also depends on the id, which I again think can be different in what's coming from the "server".

Issue #4355, opened 2024-12-03
Extend <Field.Upload /> with `onFileDownload` handler


torryt:
## 🚀 Feature Proposal

Extend the [<Field.Upload />](https://eufemia.dnb.no/uilib/extensions/forms/feature-fields/more-fields/Upload/) with a `onFileDownload` handler so that the full file can be downloaded on demand instead of pre-loaded in the form data.

## Motivation
The [<Field.Upload />](https://eufemia.dnb.no/uilib/extensions/forms/feature-fields/more-fields/Upload/) component expects input as an array containing the full [File](https://developer.mozilla.org/en-US/docs/Web/API/File) object. This means that when we preload a large form with several steps, and possibly several `<Field.Upload />` components, we also have to load all the files already uploaded in the form and preload them into the form state, which can significantly delay the first page load.

## Example
```tsx
async function onFileDownload(fileId: string): Promise<Blob> {
  const response = await getFileFromServer(fileId);

  return await response.blob();
}

....

<Field.Upload
  onFileDownload={onFileDownload}
/>
```

`onFileDownload` expects an async function that returns a file blob. When the function is provided, it overrides the default behavior of the file link and downloads from server instead of downloading the File object in the form state.

## Pitch

Reduces initial load time of forms with large files or many files and reduce unnecessary loading of documents that will probably not be downloaded.

langz:
Thanks for the detailed feature request, @torryt 🚀 

I like the idea of a `onFileDownload` event or so, to download a given file from the server with a api request, and can definitely see the need for this.

I'll have to look more into how this works as of today.

I just took a quick search in the codebase, and as of today the code in Eufemia use multiple "file" properties, like `name`, `size`, `lastModified`, `id`, `type`, etc. Not sure what would happen if we don't provide these before the user clicks the link.

langz:
I think this feature request is solved with the following PRs which is included in [v10.60.0](https://github.com/dnbexperience/eufemia/releases/tag/v10.60.0) of Eufemia:
- https://github.com/dnbexperience/eufemia/pull/4370
- https://github.com/dnbexperience/eufemia/pull/4369
- https://github.com/dnbexperience/eufemia/pull/4367

Please report back, and reopen the feature request if you find anything missing @torryt 🙇

torryt:
This works just as expected @langz . Thank you very much! 💯 🙏

langz:
> This works just as expected @langz . Thank you very much! 💯 🙏
> 
[...]

PR #4362, opened 2024-12-05
feat(Upload): adds custom `href` property to file object

langz:


torryt:
This looks like an elegant solution 🤩 Question: if the file URL resource requires authentication through cookies, will the cookies be forwarded with this approach?

torryt:
@langz We've done some searching on this, and cookies are not included in CORS requests. In our frontend API client we add `credentials: 'include'` to ensure cookies are passed along. Therefore this approach would not work for us, since the `href` don't allow any custom HTTP client logic as I understand it.

langz:
> @langz We've done some searching on this, and cookies are [not included in CORS requests.](https://jakearchibald.com/2021/cors/#adding-credentials) In our frontend API client we add `credentials: 'include'` to ensure cookies are passed along. Therefore this approach would not work for us, since the `href` don't allow any custom HTTP client logic as I understand it.

Thanks for checking 🤝 
I can quite quickly change this to a `onFileClick` handler which you can use programatically to do what you want.

However in regards to https://github.com/dnbexperience/eufemia/issues/4355 and not having to pre-loaded all files in the form data, I think that I would have to make changes other places in the code base as well so the form will still be valid even though the form does not have the full file, etc, etc.

langz:
I've replaced this PR with the following PR https://github.com/dnbexperience/eufemia/pull/4365

PR #4363, opened 2024-12-05
fix(Forms): avoid unnecessary rerenders in Form.Handler

tujoworker:
The reason behind this is that I saw a negative side-effect in PR #4357 when it comes to the initial render, where we don't want an animation, but it did animate.


langz (approved):
🤌

PR #4373, opened 2024-12-06
chore(Upload): improves readability in examples

langz:
Better to duplicate, as it makes it easier to read the code examples.
 
 Previously, one would read `useMockFiles(setFiles, { errorMessage: 'This is no real file!' })` in the example code, but not be able to see what `useMockFiles` would do.


tujoworker (approved):
Clean Code 🤩

PR #4380, opened 2024-12-11
fix(Upload): avoid scrolling when removing file from list

tujoworker:


langz (approved):
Seems great 👏 
I've never experienced it myself, or at least, I've not thought about it 😅

PR #4387, opened 2024-12-12
feat(Value.Upload): add async `onFileClick` event

langz:


packages/dnb-eufemia/src/components/upload/UploadFileListCell.tsx
```diff
@@ -226,3 +196,39 @@ function useExistsHighlight(id: string, file: File) {
...
   return exists
 }
+
+export function getFileIcon(
+  file: File,
+  isLoading?: UploadFile['isLoading'],
+  hasWarning?: boolean
+) {
+  if (isLoading) {
+    return <ProgressIndicator size="medium" />
```
langz, line 206: Not sure if we want this `medium` size for both places(in `Field.Upload` and `Field.Value`).
Might get too small for `Field.Upload`?

langz: I made it so that we support different sizes of the `ProgressIndicator`, depending on if used in `Field.Upload` and `Value.Upload`

langz:
TODO:
Have to improve this when clicking.
The progress indicator makes it jump a bit, file name changes position, etc:



https://github.com/user-attachments/assets/9d7fde55-7fa4-4529-9539-6f74408eb4a6

PR #4390, opened 2024-12-12
fix(Upload): display files without anchor when their size is not given

tujoworker:
Example for Value.Upload(https://eufemia-git-fix-forms-upload-empty-file-eufemia.vercel.app//uilib/extensions/forms/Value/Upload/#display-files-as-non-clickable)


langz (commented):
Looks good to me 🎸

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Value/Upload/demos.mdx
```diff
@@ -53,3 +53,9 @@ import * as Examples from './Examples'
 ### Using `onFileClick`
 
 <Examples.OnFileClick />
+
+### Display files as non-clickable
+
+When file size is 0, the file is displayed as a span instead of an anchor.
```
langz, line 59: Why?
Because there's no file if file size is 0?

langz (approved):

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Value/Upload/demos.mdx
tujoworker: `new File([], name, { type })` will create a file with the size 0 when downloading it. But a file with the size of 0 is not valid. So then we "show" just the name, but do now allow it to be downloaded or displayed.

langz: Makes sense. We could perhaps state this reason in the docs as well🤌

langz: And perhaps something about that if providing `onFileClick`, it will display a button no matter what the file size is.

langz:
I rebased this with latest main now.

PR #4394, opened 2024-12-12
feat(Forms): add missing support for `defaultValue` for Field.Slider

tujoworker:


packages/dnb-eufemia/src/extensions/forms/Field/Slider/Slider.tsx
```diff
@@ -56,7 +56,9 @@ function SliderComponent(props: Props) {
     [getSourceValue]
   )
 
-  const value = getValues(props.paths ?? props.path ?? props.value)
+  const value = getValues(
+    props.paths ?? props.path ?? props.value ?? props.defaultValue
```
langz, line 60:
```suggestion
    props.path ?? props.value ?? props.defaultValue
```

Do we need `props.paths ?? props.path`? 🤔

langz (approved):

packages/dnb-eufemia/src/extensions/forms/Field/Slider/Slider.tsx
tujoworker: yeah, we actually have a `path` and a `paths` prop here. Just to support an array with paths, which results in several slider thumbs. I see the docs states its required. But it is not. Will correct it with an commit.

PR #4396, opened 2024-12-12
add example

langz:


tujoworker (approved):

kimroen:
Thanks for fixing!

> I assume the calendar needs to display the new/updated date(2024-12-12) instead?

Yeah I'd hope so, but not sure either 🤔

PR #4401, opened 2024-12-13
docs(Upload): add `id` property

langz:
`id` property is [required](https://github.com/dnbexperience/eufemia/blob/ee948ede2201b8bcc489e49874be400187809c85/packages/dnb-eufemia/src/components/upload/types.ts#L19), so makes sense to document it.

### DIFF
--- a/packages/dnb-eufemia/src/components/upload/UploadDocs.ts
+++ b/packages/dnb-eufemia/src/components/upload/UploadDocs.ts
@@ -1,6 +1,11 @@
 import { PropertiesTableProps } from '../../shared/types'
 
 export const UploadProperties: PropertiesTableProps = {
+  id: {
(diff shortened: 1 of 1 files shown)


packages/dnb-eufemia/src/components/upload/UploadDocs.ts
```diff
@@ -1,6 +1,11 @@
 import { PropertiesTableProps } from '../../shared/types'
 
 export const UploadProperties: PropertiesTableProps = {
+  id: {
+    doc: 'Unique id used with the useUpload hook to manage the files.',
+    type: 'string',
```
tujoworker, line 6:
```suggestion
    type: ['string', 'Function', 'Object', 'React.Context'],
```

tujoworker (approved):

PR #4403, opened 2024-12-13
chore(Upload): improve typing of `id` property

langz:


packages/dnb-eufemia/src/components/upload/types.ts
```diff
@@ -16,7 +16,12 @@ export type UploadProps = {
   /**
    * unique id used with the useUpload hook to manage the files
    */
-  id: string
+  id:
+    | string
+    | (() => void)
+    | Promise<() => void>
+    | React.Context<any>
+    | Record<string, unknown>
```
tujoworker, line 24:
```suggestion
  id: SharedStateId
```

NB: The import is missing.

tujoworker (approved):
Nice. I think we can used/import the this type `SharedStateId` for simplification. We do that other places too.

packages/dnb-eufemia/src/components/upload/types.ts
langz: Yeah, of course. 
Not sure why I did not do it, I think I thought it was from Forms, but it's clearly not 👍

langz: Updated now :)

PR #4408, opened 2024-12-13
feat(Typography): change iOS Safari basis font-size from 17px to 15px for DNB

tujoworker:
**Why?**

In order to let Safari know the font-size of iOS, we use `font: -apple-system-body`. iOS uses 17px as its default font-size. But because the DNB body text size is `18px` we get `19.125px`, which is extensively larger than e.g. the iOS default of 17px. So with this change, the DNB body text size will be `16.875px`. Which is much closer to the native iOS body text size.

- [ ] Sort out any negative side effects of this change.

### DIFF
--- a/packages/dnb-eufemia/src/style/core/scopes.scss
+++ b/packages/dnb-eufemia/src/style/core/scopes.scss
@@ -69,14 +69,17 @@
 
 @mixin htmlDefault() {
   html {
+    font-size: 100%;
+
     &:not([data-visual-test]) {
       scroll-behavior: smooth;
     }
-    font-size: 100%;
 
-    // IS_SAFARI_MOBILE
-    @supports (-webkit-touch-callout: none) {
-      @supports (font: -apple-system-body) {
+    // In order to let Safari know the font-size set on the system level
+    @supports (font: -apple-system-body) {
+      // Safari on iOS
+      @supports (-webkit-touch-callout: none) {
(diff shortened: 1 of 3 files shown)


tujoworker:
Closing this approach because switching the root from `-apple-system-body` to `-apple-system-subheadline` has some important downsides:

- It does not match macOS at the default setting. DNB body text becomes 16.875px instead of 18px.
- It changes every root-relative dimension from a 17px basis to a 15px basis, including spacing, sizing, and typography.
- Subheadline has a different, non-linear Dynamic Type scaling curve from Body, so the difference changes across accessibility text sizes.

A more promising direction is to retain `font: -apple-system-body` and apply an iOS-only `-webkit-text-size-adjust: 94.117647%` correction, based on `16 / 17`, at the Eufemia typography/body scope. Tests in Safari on iOS 18.2 and iOS 26.3 simulators produced exactly 18px DNB body text at the default setting, continued to respond to every tested Dynamic Type setting, and left root-relative layout dimensions unchanged.

That alternative still needs focused real-device coverage across Dynamic Type categories, nested relative typography and line heights, form controls, and representative components before it should become a framework default.

The longer-term standards-based direction is `env(preferred-text-scale)` together with `<meta name="text-scale" content="scale">`. Those APIs are standard-track, but Safari and iOS Safari do not support them yet. We can revisit that route when WebKit implements them.

PR #4412, opened 2024-12-17
chore(Drawer): use `none` instead of `auto` for min and max width

tujoworker:
I don't think this has any effect, as `auto` was a non-valid value. Here are the [docs](https://developer.mozilla.org/en-US/docs/Web/CSS/min-width).


langz (approved):
🤝

Issue #4293, opened 2024-11-19
Autocomplete: In search_numbers mode, non-alphanumeric characters are not matched


kimroen:
## 🐛 Bug Report

When using an Autocomplete component with `search_numbers` set to `true`, non-alphanumeric characters like "ø" are not matched.

## To Reproduce

1. Render `<Autocomplete />` with `search_numbers={true}`, and a data set that contains a string with `ø` inside
2. Type something that should match

Here's a code sandbox demonstrating this issue:

https://codesandbox.io/p/sandbox/stoic-feather-8ymh7s

The top input does not use numbers mode. Observe that typing "Røen" matches one item
![CleanShot 2024-11-19 at 13 35 10](https://github.com/user-attachments/assets/9a18902b-e52b-4752-bbad-afaa5b80e39a)

The bottom input uses numbers mode. Typing "Røen" shows no matches.
![CleanShot 2024-11-19 at 13 35 44](https://github.com/user-attachments/assets/f11d60fb-ed82-45a2-8bd8-625830976560)

## Expected behavior

The matching name is highlighted and is part of the filtered result set.

The codesandbox example also includes an account number formatted in different ways. This is to illustrate why we are using numbers mode here. In the top version typing "12345678901" only matches the first account, but the bottom one matches all three.

![CleanShot 2024-11-19 at 13 35 31](https://github.com/user-attachments/assets/ce8995d5-47be-4832-913f-7224dcfbc1d5)
![CleanShot 2024-11-19 at 13 35 56](https://github.com/user-attachments/assets/07a74932-5294-4d38-bc0c-8d4b744d21ad)

## Eufemia Version

**Browser JS:** 10.56.0  
**Browser CSS:** 10.56.0

kimroen:
I see the code that handles highlighting and the code that handles filtering are separate, yet both are fairly entangled in the code of the autocomplete component.

How would you feel about a PR making the algorithm used for searching/filtering replaceable, or maybe starting by just separating it out a little more? There are a few approaches we could look at to improve this, but step one is making it a little more modular regardless.

(Maybe something like the command score function used by a library like cmdk ([source](https://github.com/pacocoursey/cmdk/blob/main/cmdk/src/command-score.ts)) could work, for example).

PR #4419, opened 2024-12-19
fix(Autocomplete): support query with unicodes when using `search_numbers`

langz:
fixes https://github.com/dnbexperience/eufemia/issues/4293

As of now, I've just added support for the characters ÆØÅ, not sure if there's other non-alphanumeric characters we should support, and perhaps improve the regex to something better than just`\wÆØÅæøå`.

[Original CSB from the issue](https://codesandbox.io/p/sandbox/stoic-feather-8ymh7s)
[CSB with the fix](https://codesandbox.io/p/sandbox/relaxed-paper-gyxyf8?workspaceId=ws_LbiDwJdEBdbwxH2AkyAyVE) using commit https://github.com/dnbexperience/eufemia/pull/4419/commits/aa28f0c632f1baf57ff2b234157f88ebbe49c043
[CSB with the fix](https://codesandbox.io/p/sandbox/flamboyant-feistel-klw9c2?workspaceId=ws_LbiDwJdEBdbwxH2AkyAyVE) using commit https://github.com/dnbexperience/eufemia/pull/4419/commits/2334c879516dd658db048778439e4efbb9f62cd5

### DIFF
--- a/packages/dnb-eufemia/src/components/autocomplete/Autocomplete.js
+++ b/packages/dnb-eufemia/src/components/autocomplete/Autocomplete.js
@@ -1353,7 +1353,7 @@ class AutocompleteInstance extends React.PureComponent {
         .filter(({ word, wordIndex }) => {
           if (searchNumbers) {
             // Remove all other chars, except numbers, so we can compare
-            word = word.replace(/[^\d\wÆØÅæøå]/g, '')
+            word = word.replace(/[^\p{L}\p{N}]+/gu, '')
           } else {
             // To ensure we escape regex chars
             word = escapeRegexChars(word)
--- a/packages/dnb-eufemia/src/components/autocomplete/Autocomplete.js
+++ b/packages/dnb-eufemia/src/components/autocomplete/Autocomplete.js
@@ -1353,7 +1353,7 @@ class AutocompleteInstance extends React.PureComponent {
         .filter(({ word, wordIndex }) => {
           if (searchNumbers) {
             // Remove all other chars, except numbers, so we can compare
-            word = word.replace(/[^\d\w\æøå]/g, '')
+            word = word.replace(/[^\d\wÆØÅæøå]/g, '')
           } else {
             // To ensure we escape regex chars
             word = escapeRegexChars(word)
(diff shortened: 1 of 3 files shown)


packages/dnb-eufemia/src/components/autocomplete/Autocomplete.js
```diff
@@ -1353,7 +1353,7 @@ class AutocompleteInstance extends React.PureComponent {
         .filter(({ word, wordIndex }) => {
           if (searchNumbers) {
             // Remove all other chars, except numbers, so we can compare
-            word = word.replace(/[^\d\w]/g, '')
+            word = word.replace(/[^\d\wÆØÅæøå]/g, '')
```
tujoworker, line 1356:
```suggestion
            word = word.replace(/[^\p{L}\p{N}]+/gu, '')
```

- [^\p{L}\p{N}] means "any character not a letter or digit."
- u flag makes it a Unicode-aware search, and allows you to use \p{...} property escapes.

packages/dnb-eufemia/src/components/autocomplete/Autocomplete.js
```diff
@@ -1474,7 +1474,7 @@ class AutocompleteInstance extends React.PureComponent {
 
               if (searchNumbers) {
                 word.split('').forEach((char) => {
-                  if (/[\d\w]/.test(char)) {
+                  if (/[\d\wÆØÅæøå]/.test(char)) {
```
tujoworker, line 1477:
```suggestion
                  if (/[\p{L}\p{N}]/u.test(char)) {
```

- [\p{L}\p{N}] matches any Unicode letter (\p{L}) or any Unicode number (\p{N}).
- The u (Unicode) flag enables Unicode-aware matching and allows the use of \p{...} properties.

packages/dnb-eufemia/src/components/autocomplete/Autocomplete.js
kimroen: TIL! Cool 😊 
- 📚 [Unicode mode (`u`)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/unicode)
- 📚 [Unicode character class escape](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Unicode_character_class_escape)
- "L" and "N" are "General_Category" property values - they are [listed on the Unicode website](https://unicode.org/reports/tr18/#General_Category_Property)
  - From what I read here, this regex could also be expressed using the long form names "Letter" and "Number" (which might be easier to read, but could also just be unfamiliarity?)

kimroen: The comment above this line seems a little outdated now, by the way.

kimroen (approved):
It works! ✨ Thank you!

PR #4420, opened 2024-12-19
fix(Validator): triggers `onChangeValidator` when changing connected field

langz:
Todo: the actual fix

Reported in this [Slack thread](https://dnb-it.slack.com/archives/CMXABCHEY/p1734603582493269)


langz:
Closing this, as I believe it should not work as documented in the tests.

PR #4421, opened 2024-12-20
feat(Forms): add support for conditional function based `info`, `warning` and `error` props to all `Field.*` components

tujoworker:
The PR updates the `info`, `warning`, and `error` props to accept functions. This makes it easier to create conditional messages without relying on making a field "controlled."

Additionally, it ensures the same user experience for displaying these messages (by using the `conditionally` function), consistent with how error messages are handled by default today.

```tsx
<Field.Number info={(value, { conditionally, getValueByPath }) => {
  return conditionally(() => 'Show this message, with this value ' + getValueByPath('/otherField'))
}} />
```

Here's an [example](https://eufemia-git-feat-forms-conditional-infos-eufemia.vercel.app/uilib/extensions/forms/base-fields/Number/#displaying-messages---conditional-info-message).

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/base-fields/Number/demos.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/base-fields/Number/demos.mdx
@@ -108,7 +108,6 @@ Optionally, use the `interactive` higher order function to show the message only
 You can also pass options to the `interactive` function:
 
 - `showInitially` – Display the message when the field is first rendered.
-- `showContinuously` – Continuously update and show the message on every field change.
 
 ```tsx
 <Field.String
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/base-fields/Number/demos.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/base-fields/Number/demos.mdx
@@ -92,29 +92,29 @@ You can provide a function to the `info`, `warning` or `error` props that return
 />
 ```
 
-Optionally, use the `interactive` higher order function to show the message only when the field got changed (onChange) and blurred (onBlur).
+Optionally, use the `conditionally` higher order function to show the message only when the field got changed (onChange) and blurred (onBlur).
 
 ```tsx
 <Field.String
-  info={(value, { interactive, getValueByPath, getFieldByPath }) => {
+  info={(value, { conditionally, getValueByPath, getFieldByPath }) => {
     if (value === '123') {
       // Show this message only when the field got changed and blurred.
-      return interactive(() => 'The value is 123')
+      return conditionally(() => 'The value is 123')
     }
   }}
 />
 ```
 
-You can also pass options to the `interactive` function:
+You can also pass options to the `conditionally` function:
 
 - `showInitially` – Display the message when the field is first rendered.
 
 ```tsx
 <Field.String
-  info={(value, { interactive, getValueByPath, getFieldByPath }) => {
+  info={(value, { conditionally, getValueByPath, getFieldByPath }) => {
     if (value === '123') {
       // Show this message only when the field got changed and blurred.
-      return interactive(() => 'The value is 123', {
+      return conditionally(() => 'The value is 123', {
         showInitially: true,
       })
     }
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/base-fields/Number/demos.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/base-fields/Number/demos.mdx
@@ -107,7 +107,7 @@ Optionally, use the `conditionally` higher order function to show the message on
 
 You can also pass options to the `conditionally` function:
 
-- `showInitially` – Display the message when the field is first rendered.
+- `showInitially` – display the message when the field is first rendered.
 
 ```tsx
 <Field.String
@@ -121,6 +121,10 @@ You can also pass options to the `conditionally` function:
   }}
 />
 ```
+
+Down below you can see an example of how to use the `conditionally` function. There are two input fields which depend on each other. Here we use `info` to show a message when the value of the first field is too low. While we use an error message when the value of the second field is more than what the first field has. The `info` on the first field will only be shown when the user has changed the value and blurred the field.
+
+Read more about [validation and the user experience](/uilib/extensions/forms/getting-started/#validation-and-the-user-experience-ux).
 
 <Examples.ConditionalInfo />
 
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/base-fields/Number/demos.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/base-fields/Number/demos.mdx
@@ -96,7 +96,7 @@ Optionally, use the `interactive` higher order function to show the message only
 
 ```tsx
 <Field.String
-  info={(value, { interactive }) => {
+  info={(value, { interactive, getValueByPath, getFieldByPath }) => {
     if (value === '123') {
       // Show this message only when the field got changed and blurred.
       return interactive(() => 'The value is 123')
@@ -112,7 +112,7 @@ You can also pass options to the `interactive` function:
 
 ```tsx
 <Field.String
-  info={(value, { interactive }) => {
+  info={(value, { interactive, getValueByPath, getFieldByPath }) => {
     if (value === '123') {
       // Show this message only when the field got changed and blurred.
       return interactive(() => 'The value is 123', {
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/base-fields/Number/demos.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/base-fields/Number/demos.mdx
@@ -82,12 +82,46 @@ You can also use a function as a prefix or suffix.
 
 You can provide a function to the `info`, `warning` or `error` props that returns a message based on your conditions.
 
-Optionally, use the `showMessage` modifier function to control when the message is shown:
+```tsx
+<Field.String
+  info={(value) => {
+    if (value === '123') {
+      return 'The value is 123'
+    }
+  }}
+/>
+```
 
-- `onBlur` – The **preferred** UX behavior. It will show the message only when the field is blurred (onBlur).
-- `initially` – It will show the message initially, and later like `onBlur`.
-- `continuously` – It will show and update the message on every field value change.
-- `always` – Same as `initially` and `continuously` together.
+Optionally, use the `interactive` higher order function to show the message only when the field got changed (onChange) and blurred (onBlur).
+
+```tsx
+<Field.String
+  info={(value, { interactive }) => {
+    if (value === '123') {
+      // Show this message only when the field got changed and blurred.
+      return interactive(() => 'The value is 123')
+    }
+  }}
+/>
+```
+
+You can also pass options to the `interactive` function:
+
+- `showInitially` – Display the message when the field is first rendered.
+- `showContinuously` – Continuously update and show the message on every field change.
+
+```tsx
+<Field.String
+  info={(value, { interactive }) => {
+    if (value === '123') {
+      // Show this message only when the field got changed and blurred.
+      return interactive(() => 'The value is 123', {
+        showInitially: true,
+      })
+    }
+  }}
+/>
+```
 
 <Examples.ConditionalInfo />
 
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/base-fields/Number/demos.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/base-fields/Number/demos.mdx
@@ -86,7 +86,7 @@ Optionally, use the `visibleWhen` modifier function to control when the message 
 
 - `onBlur` – The **preferred** UX behavior. It will show the message only when the field is blurred (onBlur).
 - `initially` – It will show the message initially, and later like `onBlur`.
-- `continuously` – It will show the message on every change.
+- `continuously` – It will show and update the message on every field value change.
 - `always` – Same as `initially` and `continuously` together.
 
 <Examples.ConditionalInfo />
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/base-fields/Number/demos.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/base-fields/Number/demos.mdx
@@ -82,12 +82,12 @@ You can also use a function as a prefix or suffix.
 
 You can provide a function to the `info`, `warning` or `error` props that returns a message based on your conditions.
 
-Optionally, use the `renderMode` to control when the message is shown:
+Optionally, use the `visibleWhen` modifier function to control when the message is shown:
 
-- `interactive` – The **preferred** UX behavior. It will show the message only when the field is blurred (onBlur).
-- `initially` – It will show the message initially, and later like `interactive`.
-- `continuously` – It will show the message initially and on every change.
-- `always` – The **default** behavior. It will show the message always.
+- `onBlur` – The **preferred** UX behavior. It will show the message only when the field is blurred (onBlur).
+- `initially` – It will show the message initially, and later like `onBlur`.
+- `continuously` – It will show the message on every change.
+- `always` – Same as `initially` and `continuously` together.
 
 <Examples.ConditionalInfo />
 
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/base-fields/Number/demos.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/base-fields/Number/demos.mdx
@@ -82,6 +82,13 @@ You can also use a function as a prefix or suffix.
 
 You can provide a function to the `info`, `warning` or `error` props that returns a message based on your conditions.
 
+Optionally, use the `renderMode` to control when the message is shown:
+
+- `interactive` – The **preferred** UX behavior. It will show the message only when the field is blurred (onBlur).
+- `initially` – It will show the message initially, and later like `interactive`.
+- `continuously` – It will show the message initially and on every change.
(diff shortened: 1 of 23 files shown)


packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/base-fields/Number/demos.mdx
```diff
@@ -78,6 +78,19 @@ You can also use a function as a prefix or suffix.
...
+### Validation - Conditional info message
+
+You can provide a function to the `info`, `warning` or `error` props that returns a message based on your conditions.
+
+Optionally, use the `renderMode` to control when the message is shown:
+
+- `interactive` – The **preferred** UX behavior. It will show the message only when the field is blurred (onBlur).
+- `initially` – It will show the message initially, and later like `interactive`.
+- `continuously` – It will show the message initially and on every change.
+- `always` – The **default** behavior. It will show the message always.
```
langz, line 90: Why is not the the preferred way(interactive) default? 🤔

packages/dnb-eufemia/src/extensions/forms/hooks/useFieldProps.ts
```diff
@@ -324,12 +334,111 @@ export default function useFieldProps<Value, EmptyValue, Props>(
...
+          getFieldByPath,
+        })
+
+        const isError =
+          msg instanceof Error ||
+          msg instanceof FormError ||
+          (Array.isArray(msg) && checkForError(msg))
+
+        if (
+          // Remove the message if, if it gets update from outside to have no message anymore
```
langz, line 372:
```suggestion
          // Remove the message if it gets update from outside to have no message anymore
```

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/base-fields/Number/demos.mdx
tujoworker: I remove it from the list. Because it's not needed. 
My thoughts are to align on the default behavior of the props `info="string"` etc. It will show the message regardless. When using a function, we also can assume the dev wants to show an info always. 
But when opting in for an "interactive" approach, the dev needs to opt-in by defining the appearance.

packages/dnb-eufemia/src/extensions/forms/hooks/DataValueWritePropsDocs.ts
```diff
@@ -17,18 +17,18 @@ export const DataValueWritePropsProperties: PropertiesTableProps = {
     status: 'optional',
   },
   info: {
-    doc: 'Info message shown below / after the field.',
-    type: ['React.Node', 'Array<React.Node>'],
+    doc: 'Info message shown below / after the field. When provided as a function, the function will be called with the current value as argument. The second parameter is an objet with `{ visibleWhen, getValueByPath, getFieldByPath }`. You can call `visibleWhen` with `onBlur`, `initially`, or `conditionally`.',
```
langz, line 20:
```suggestion
    doc: 'Info message shown below / after the field. When provided as a function, the function will be called with the current value as argument. The second parameter is an object with `{ visibleWhen, getValueByPath, getFieldByPath }`. You can call `visibleWhen` with `onBlur`, `initially`, or `conditionally`.',
```

packages/dnb-eufemia/src/extensions/forms/hooks/DataValueWritePropsDocs.ts
```diff
@@ -17,18 +17,18 @@ export const DataValueWritePropsProperties: PropertiesTableProps = {
...
+    type: ['React.Node', 'Array<React.Node>', 'function'],
     status: 'optional',
   },
   warning: {
-    doc: 'Warning message shown below / after the field.',
-    type: ['React.Node', 'Array<React.Node>'],
+    doc: 'Warning message shown below / after the field. When provided as a function, the function will be called with the current value as argument. The second parameter is an objet with `{ visibleWhen, getValueByPath, getFieldByPath }`. You can call `visibleWhen` with `onBlur`, `initially`, or `conditionally`.',
```
langz, line 25:
```suggestion
    doc: 'Warning message shown below / after the field. When provided as a function, the function will be called with the current value as argument. The second parameter is an object with `{ visibleWhen, getValueByPath, getFieldByPath }`. You can call `visibleWhen` with `onBlur`, `initially`, or `conditionally`.',
```

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/base-fields/Number/demos.mdx
```diff
@@ -78,6 +78,19 @@ You can also use a function as a prefix or suffix.
 
 <Examples.ValidateMaximumCustomError />
 
+### Validation - Conditional info message
```
langz, line 81:
```suggestion
### Displaying messages - Conditional info message
```

When using the word `validation`, won't it potentially collide with our validator concept, which is `onBlurValidator` and `onChangeValidator`. To me, info, warning and error is just message, right? And doesn't have anything to do with validation(validator)? 🤔

langz:
Can a Field have a visible `info`, `warning`, and `error` message?
If not, which message takes precedence?

langz:
Will this functionality, like `error`, collide with validators? Can it both display an error message from the `error` and from the `validator`(onChangeValidator or onBlurValidator) at the same time?

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/base-fields/Number/demos.mdx
```diff
@@ -78,6 +78,19 @@ You can also use a function as a prefix or suffix.
...
+
+You can provide a function to the `info`, `warning` or `error` props that returns a message based on your conditions.
+
+Optionally, use the `visibleWhen` modifier function to control when the message is shown:
+
+- `onBlur` – The **preferred** UX behavior. It will show the message only when the field is blurred (onBlur).
+- `initially` – It will show the message initially, and later like `onBlur`.
+- `continuously` – It will show and update the message on every field value change.
```
langz, line 89: Could also be called `instant`? 🤔

langz:
> In a meeting, we agreed to go with something like this:
> 
> ```tsx
> import { Field, renderMode } from '@dnb/eufemia/extensions/forms'
> 
> render(<Field.Number info={renderMode((value) => 'The message', { mode: 'onBlur' })} />)
> ```
> 
> However, this approach isn’t feasible because the function needs to be executed within the "field scope/context" (`useFieldProps.ts`).
> 
> To address this, I changed the name from `renderMode` to `visibleWhen` to make it more declarative.
> 
> What do you think?

Thanks for improving upon this @tujoworker 🙇 

I do find it a bit difficult to understand:
- `visibleWhen('onBlur')`
- `visibleWhen('initially')`
- `visibleWhen('continuously')`
- `visibleWhen('always')`

The options onBlur, initially, continuously, and always made sense to me when the function was named `renderMode`, `validationMode` or so. As I feel all these option names are sort of "modes".

If we would use `visibleWhen` we could consider:
- `visibleWhen('lossOfFocus')`
- `visibleWhen('rendered')`
- `visibleWhen('changed')`
- `visibleWhen('renderedAndChanged')`

But I do sort of like that the option names onBlur, initially, continuously, and always sort of match the "validation modes" we have, https://eufemia.dnb.no/uilib/extensions/forms/create-component/useFieldProps/info/#validation
So that when you understand these, you'll understand these options as well.

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/base-fields/Number/demos.mdx
langz: But makes sense that the word used  matches with this https://eufemia.dnb.no/uilib/extensions/forms/create-component/useFieldProps/info/#validation

langz:
On a sidenote, in regards to [validation](https://eufemia.dnb.no/uilib/extensions/forms/create-component/useFieldProps/info/#validation) where we have:
- validateRequired
- validateInitially
- validateUnchanged
- continuousValidation

Should we rename `continuousValidation` to `validateContinuously`?

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/base-fields/Number/demos.mdx
tujoworker: Yes, I agree 🙏

tujoworker:
> Can a Field have a visible info, warning, and error message?

Yes, all three are displayed.

tujoworker:
> Will this functionality, like error, collide with validators? Can it both display an error message from the error and from the validator(onChangeValidator or onBlurValidator) at the same time?

Yes, they will collide. When providing an `error` it is handled like an error. From a user perspective, it looks the same, so I think we should not handle it differently. But I agree that the handling of the `visibleWhen` with an error is not needed, because it can and will collide when used together. 

We could maybe throw an error when both are provided? (I mean when e.g. `onBlurValidator` and `visibleWhen('onBlur')` is used).

tujoworker:
> Should we rename continuousValidation to validateContinuously?

I agree that this makes sense. Do you take this as a task?

tujoworker:
> I do find it a bit difficult to understand:

So it sounds `renderMode` is a better name for this "appearance" change? Could it be something like: 

- `changeAppearanceTo('onBlur')`

I mean, I'm fine with `renderMode` too.

langz:
> > I do find it a bit difficult to understand:
> 
> So it sounds `renderMode` is a better name for this "appearance" change? Could it be something like:
> 
> * `changeAppearanceTo('onBlur')`
> 
> I mean, I'm fine with `renderMode` too.

Yeah, I don't think renderMode is really describing what this is or does either.
I'll try to think of a few alternative 🤔

langz:
> > > I do find it a bit difficult to understand:
> > 
> > 
> > So it sounds `renderMode` is a better name for this "appearance" change? Could it be something like:
> > 
> > * `changeAppearanceTo('onBlur')`
> > 
> > I mean, I'm fine with `renderMode` too.
> 
> Yeah, I don't think `renderMode` is really describing what this is or does either. Perhaps `changeAppearanceTo` is better 🤔 I'll also try to think of a few alternative 🤔
> 
> If this would have been validation, I think validationMode would make sense.

Some alternatives:

`displayMessage`
- `displayMessage('onBlur')`
- `displayMessage('initially')`
- `displayMessage('continuously')`
- `displayMessage('always')`

`display`
- `display('onBlur')`
- `display('initially')`
- `display('continuously')`
- `display('always')`

`showMessage`
- `showMessage('onBlur')`
- `showMessage('initially')`
- `showMessage('continuously')`
- `showMessage('always')`

`show`
- `show('onBlur')`
- `show('initially')`
- `show('continuously')`
- `show('always')`

`visible`
- `visible('onBlur')`
- `visible('initially')`
- `visible('continuously')`
- `visible('always')`

packages/dnb-eufemia/src/extensions/forms/types.ts
```diff
@@ -265,6 +266,27 @@ export type DataValueReadWriteComponentProps<
   DataValueReadProps<Value> &
   DataValueWriteProps<Value, EmptyValue>
 
+export type ShowMessage =
+  | 'onBlur' // It will show the message only when the field is blurred (onBlur).
+  | 'initially' // It will show the message initially, and later like 'onBlur'.
+  | 'continuously' // It will show the message on every change.
+  | 'always' // Same as `initially` and `continuously` together
```
langz, line 273:
```suggestion
  | 'always' // Same as `initially` and `continuously` together.
```

packages/dnb-eufemia/src/extensions/forms/hooks/DataValueWritePropsDocs.ts
```diff
@@ -17,18 +17,18 @@ export const DataValueWritePropsProperties: PropertiesTableProps = {
     status: 'optional',
   },
   info: {
-    doc: 'Info message shown below / after the field.',
-    type: ['React.Node', 'Array<React.Node>'],
+    doc: 'Info message shown below / after the field. When provided as a function, the function will be called with the current value as argument. The second parameter is an object with `{ visibleWhen, getValueByPath, getFieldByPath }`. You can call `visibleWhen` with `onBlur`, `initially`, or `conditionally`.',
```
langz, line 20:
```suggestion
    doc: 'Info message shown below / after the field. When provided as a function, the function will be called with the current value as argument. The second parameter is an object with `{ showMessage, getValueByPath, getFieldByPath }`. You can call `showMessage` with `onBlur`, `initially`, or `conditionally`.',
```

packages/dnb-eufemia/src/extensions/forms/hooks/DataValueWritePropsDocs.ts
```diff
@@ -17,18 +17,18 @@ export const DataValueWritePropsProperties: PropertiesTableProps = {
...
+    type: ['React.Node', 'Array<React.Node>', 'function'],
     status: 'optional',
   },
   warning: {
-    doc: 'Warning message shown below / after the field.',
-    type: ['React.Node', 'Array<React.Node>'],
+    doc: 'Warning message shown below / after the field. When provided as a function, the function will be called with the current value as argument. The second parameter is an object with `{ visibleWhen, getValueByPath, getFieldByPath }`. You can call `visibleWhen` with `onBlur`, `initially`, or `conditionally`.',
```
langz, line 25:
```suggestion
    doc: 'Warning message shown below / after the field. When provided as a function, the function will be called with the current value as argument. The second parameter is an object with `{ showMessage, getValueByPath, getFieldByPath }`. You can call `showMessage` with `onBlur`, `initially`, or `conditionally`.',
```

packages/dnb-eufemia/src/extensions/forms/hooks/DataValueWritePropsDocs.ts
```diff
@@ -17,18 +17,18 @@ export const DataValueWritePropsProperties: PropertiesTableProps = {
...
+    type: ['React.Node', 'Array<React.Node>', 'function'],
     status: 'optional',
   },
   error: {
-    doc: 'Error message shown below / after the field.',
-    type: ['Error', 'FormError', 'Array<Error | FormError>'],
+    doc: 'Error message shown below / after the field. When provided as a function, the function will be called with the current value as argument. The second parameter is an object with `{ visibleWhen, getValueByPath, getFieldByPath }`. You can call `visibleWhen` with `onBlur`, `initially`, or `conditionally`.',
```
langz, line 30:
```suggestion
    doc: 'Error message shown below / after the field. When provided as a function, the function will be called with the current value as argument. The second parameter is an object with `{ showMessage, getValueByPath, getFieldByPath }`. You can call `showMessage` with `onBlur`, `initially`, or `conditionally`.',
```

langz (approved):

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/base-fields/Number/demos.mdx
```diff
@@ -78,6 +78,53 @@ You can also use a function as a prefix or suffix.
...
+<Field.String
+  info={(value) => {
+    if (value === '123') {
+      return 'The value is 123'
+    }
+  }}
+/>
+```
+
+Optionally, use the `interactive` higher order function to show the message only when the field got changed (onChange) and blurred (onBlur).
```
langz, line 95:
```suggestion
Optionally, use the `interactive` higher order function to show the message only when the field was changed (onChange) and blurred (onBlur).
```

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/base-fields/Number/demos.mdx
```diff
@@ -78,6 +78,53 @@ You can also use a function as a prefix or suffix.
...
+/>
+```
+
+Optionally, use the `interactive` higher order function to show the message only when the field was changed (onChange) and blurred (onBlur).
+
+```tsx
+<Field.String
+  info={(value, { interactive }) => {
+    if (value === '123') {
+      // Show this message only when the field got changed and blurred.
```
langz, line 101:
```suggestion
      // Show this message only when the field was changed and blurred.
```

langz:
In the ["maximum for amount" example](https://eufemia-git-feat-forms-conditional-infos-eufemia.vercel.app/uilib/extensions/forms/base-fields/Number/demos/#displaying-messages---conditional-info-message) the tab order, when tabbing using keyboard, is not logical(the anchor/link is not following the vertical order):


https://github.com/user-attachments/assets/dc701b97-b0a7-4a1b-97ca-f14f8b7209d9

tujoworker:
> In the ["maximum for amount" example](https://eufemia-git-feat-forms-conditional-infos-eufemia.vercel.app/uilib/extensions/forms/base-fields/Number/demos/#displaying-messages---conditional-info-message) the tab order, when tabbing using keyboard, is not logical(the anchor/link is not following the vertical order):

Correct, FormStatus messages are semantically placed above fields for a11y reasons.

tujoworker:
During a meeting, we considered using an object as the foundation for the new API. However, we later realized that this approach would be less flexible. As a result, the new API is designed around a function instead. 

For implementing the `onBlur` UX, the `interactive` higher-order function can be utilized. However, it's not limited to just `onBlur`. The user needs to make a change first (`onChange`) and then blur the field before the message is displayed. This behavior is effectively "onChangeAndOnBlur."

tujoworker:
So when using `interactive`, the UX will be "onChangeAndOnBlur". In addition, there are these options:

- `showInitially` – Display the message when the field is first rendered.
- `showContinuously` – Continuously update and show the message on every field change.

langz (approved):
I like the API, and agree with that 👏 
However, I'm not sure myself that I would understand that by using `interactive`, I would get a message that's displayed on change and on blur(onChangeAndOnBlur). But I don't have any better naming suggestions as of right now 🤔

snorrekim:
I believe `interactive` should be called `conditionally`. I don't think we can come up with a name that will fully explain what it does, but `conditionally` at least does a much better job than `interactive`. And it "flows" well: `return conditionally(()=>"This is the answer")` makes  a sort of sense.

And it highlights where the "real" conditions are. It's already implicit that you can use `if` or `else` in a function, the tricky part is to make it clear that there are other conditions than just that.

PR #4424, opened 2024-12-20
fix(Forms): add support for `sessionStorageId` in Field.Upload

tujoworker:
In this PR #4339 we did not render the stored files.
This time, we do support rendering the file names stored in the session storage. The test dedicated to session storage got aligned.

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/feature-fields/more-fields/Upload/info.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/feature-fields/more-fields/Upload/info.mdx
@@ -165,4 +165,4 @@ function MyForm() {
 
 The `sessionStorageId` property can be used to store the files in the session storage so they persist between page reloads.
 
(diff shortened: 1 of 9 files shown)


langz (approved):
🎄🎅

langz:
Do you think we should release it now?
v10.62.1? Or wait to after the holidays?

tujoworker:
I think we should first look into Value.Upload and add support for it there as well.

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/feature-fields/more-fields/Upload/info.mdx
```diff
@@ -160,3 +160,9 @@ function MyForm() {
   )
 }
 ```
+
+### Persist files in session storage
+
+The `sessionStorageId` property can be used to store the files in the session storage so they persist between page reloads.
+
+But the persisted files only render the file name, and not the file itself. The file blog got lost during the serialization process.
```
langz, line 168:
```suggestion
But the persisted files only render the file name, and not the file itself. The file blob will be lost during the serialization process.
```

PR #4437, opened 2025-01-08
fix(Card): prioritize `gap` property over `stack` spacing

tujoworker:
After:

Screenshot 2025-01-08 at 10 02 19

Before:

Screenshot 2025-01-08 at 10 03 10


langz (approved):
🤝

PR #4441, opened 2025-01-08
feat(Forms): deprecate `continuousValidation` in favor of `validateContinuously`

langz:


tujoworker (approved):
Nice work 🚀

tujoworker:
We need to add an entry in to the `v11.mdx` file before merge.

Issue #3134, opened 2023-12-20
Improve docs or types of InputMasked


langz:
## 🐛 Bug Report

As of now, all the properties of InputMasked is optional, so I would assume that using `<InputMasked />` without any props, would result in a default working version of the `InputMasked` component. But when just doing `<InputMasked />`, the result I get can't be used for anything? 🤔  Or at least, I'm not able to enter any data in the input field, etc.

## To Reproduce

https://codesandbox.io/p/sandbox/gallant-bird-p2lhdf?file=%2Fsrc%2FApp.tsx%3A1%2C30

PR #4446, opened 2025-01-10
fix(InputMasked): should work without any properties

langz:
This PR aims to make InputMasked work without providing any props.
Reason why it should work without providing any props, is because all its props is optional.

Defaults to using number mask, as a mask is required.

fixes https://github.com/dnbexperience/eufemia/issues/3134


langz:
I'll add a test

tujoworker (approved):
I think that this should work fine 🚀

Issue #4145, opened 2024-10-17
Annoying defaultProps deprecation warnings


andlbrei:
React 18.3.1 (I think) introduced warnings for deprecation of defaultProps.
These are quite annoying especially in test output, and are not possible to silence without silencing other warnings, as far as I know.

`Warning: Content: Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead.`

If you know how to get rid of only these warnings, please let me know.

PR #4447, opened 2025-01-10
chore: replace defaultProps in function components

langz:
Fixes a few warnings reported here, not all I believe:
https://github.com/dnbexperience/eufemia/issues/4145

PR #4448, opened 2025-01-13
fix: refactor defaultProps to ES6 default parameters

langz:
- Fixes https://github.com/dnbexperience/eufemia/issues/4145
- In relation to #4447

### DIFF
--- a/packages/dnb-eufemia/src/components/input-masked/InputMaskedUtils.js
+++ b/packages/dnb-eufemia/src/components/input-masked/InputMaskedUtils.js
@@ -65,7 +65,12 @@ export const correctNumberValue = ({
   locale,
   maskParams,
 }) => {
-  let value = props.value === null ? null : String(props.value)
+  let value =
+    props.value === null
+      ? null
+      : props.value === undefined
+      ? undefined
+      : String(props.value)
 
   if (isNaN(parseFloat(value))) {
     return value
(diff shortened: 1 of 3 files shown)


packages/dnb-eufemia/src/components/input-masked/InputMaskedUtils.js
```diff
@@ -65,7 +65,12 @@ export const correctNumberValue = ({
   locale,
   maskParams,
 }) => {
-  let value = props.value === null ? null : String(props.value)
+  let value =
+    props.value === null
+      ? null
+      : props.value === undefined
+      ? undefined
```
langz, line 72: I've added this code, because:
- Previously when we used defaultProps, user providing `value={undefined}` would result in value getting the default value `"initval"`.
- But now since we merge props(user provided) and defaultProps, the user provided `value={undefined}` will and should take presence over the default value(`"initval"`). Hence adding this line, to support getting a `undefined` value in this function(previously it would never happen because of the default value "initval")

tujoworker (approved):
Looks good 🤗

tujoworker:
just an idea: Should we name this PR to effect all of the changes, so it appears in the release logs? In regards [of this comment](https://github.com/dnbexperience/eufemia/pull/4447#issuecomment-2587077590).

PR #4449, opened 2025-01-13
chore(Accordion): replace defaultProps

langz:
- Fixes https://github.com/dnbexperience/eufemia/issues/4145
- In relation to #4447 and #4448

### DIFF
--- a/packages/dnb-eufemia/src/components/accordion/Accordion.tsx
+++ b/packages/dnb-eufemia/src/components/accordion/Accordion.tsx
@@ -435,9 +435,6 @@ function Accordion({
     </Context.Consumer>
   )
 }
-// TEMPORARY SOLUTION (defaultProps will be deprecated at one point). Needs to replacement with default prop parameters for example "({expanded: null})"
-// Only solved this way to prevent tests from failing, for when expanded is undefined instead of null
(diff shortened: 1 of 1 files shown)


packages/dnb-eufemia/src/components/accordion/Accordion.tsx
```diff
@@ -435,9 +435,6 @@ function Accordion({
     </Context.Consumer>
   )
 }
-// TEMPORARY SOLUTION (defaultProps will be deprecated at one point). Needs to replacement with default prop parameters for example "({expanded: null})"
-// Only solved this way to prevent tests from failing, for when expanded is undefined instead of null
-Accordion.defaultProps = accordionDefaultProps
```
langz, line 440: Not entirely sure if this is just safe to remove or not?
If not, what else can we do to keep the same functionality, but go away from defaultProps?

Was introduced in https://github.com/dnbexperience/eufemia/pull/2529

@joakbjerk

tujoworker: I also think it's safe to remove. The only props which where defined here was:

variant: 'outlined'
icon_size: 'medium'

Issue #3116, opened 2023-12-20
Tabs in portal doesn't change page


langz:
## 🐛 Bug Report

The Tabs in the Eufemia Portal doesn't seem to change the actual page/tab that's displayed when pressing the arrow button. But when pressing, it visually seems like a new page/tab is active. 

Take a look at [these docs](https://eufemia.dnb.no/uilib/extensions/payment-card/)(I expect the same happening everywhere we ise Tabs in docs, and not just this example), and the following video/demo:

https://github.com/dnbexperience/eufemia/assets/1359205/49639641-2f03-4a18-af81-a7b9731dea9a

Sometimes it also just "glitches", and goes "back and forth"(between info and demos in the following example):

https://github.com/dnbexperience/eufemia/assets/1359205/e1973404-34bc-4471-880b-d5e142c1e453

PR #4455, opened 2025-01-14
chore(TabBar): fix navigation on next and prev button click

langz:
fixes https://github.com/dnbexperience/eufemia/issues/3116

Removes `use_hash` which was an internally used property(but was not used anywhere), which was not publicly documented.

Added a new internal property `onOpenTabNavigationFn ` to be able to use Gatsby `navigate` when user presses the next and prev buttons in Tabs:
image

### DIFF
--- a/packages/dnb-design-system-portal/src/shared/tags/TabBar.tsx
+++ b/packages/dnb-design-system-portal/src/shared/tags/TabBar.tsx
@@ -111,7 +111,7 @@ export default function TabBar({
       )}
       <Tabs
         id="tab-bar"
-        on_open_tab_navigation_fn={navigate}
+        onOpenTabNavigationFn={navigate}
         tab_element={Link}
         data={preparedTabs}
         selected_key={selectedKey}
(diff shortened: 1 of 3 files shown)


packages/dnb-design-system-portal/src/shared/tags/TabBar.tsx
```diff
@@ -110,6 +111,7 @@ export default function TabBar({
       )}
       <Tabs
         id="tab-bar"
+        on_open_tab_navigation_fn={navigate}
```
tujoworker, line 114:
```suggestion
        onOpenTabNavigationFn={navigate}
```

Can we camelCase this new prop?

langz: Updated in commit https://github.com/dnbexperience/eufemia/pull/4455/commits/ea58a95c19c3cf957c54d9c24fb5ad7ae3142200

tujoworker (approved):
Nice! Thank you 🌟

PR #4473, opened 2025-01-17
feat(Forms): fix inactive `Iterate.PushContainer` blocking Wizard navigation when `bubbleValidation` is true

### LINKED ISSUE #4460: Required Field in inactive PushContainer with bubbleValidation prevents navigation
## 🐛 Bug Report

Cannot navigate to next step in wizard when the form has a `PushContainer` with `bubbleValidation` with a `required` `Field`, even though the `PushContainer` is not active.

https://github.com/user-attachments/assets/568740fa-f8ab-4c0c-90c9-dbc10b27b91c

## To Reproduce

https://codesandbox.io/p/sandbox/push-container-required-znjrzy

## Expected behavior

Expect validation inside a inactive `PushContainer` to not have any effect until the container is actually open.

## Eufemia Version

10.64.2

tujoworker:
Previously, if a required field was placed inside a PushContainer with `bubbleValidation`, it prevented the Wizard from advancing—even when the container was inactive (not opened). This change ensures that the inactive PushContainer no longer blocks navigation, allowing users to proceed to the next step as intended.

Fixes #4460

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Form/Isolation/Isolation.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Form/Isolation/Isolation.tsx
@@ -285,6 +285,13 @@ function BubbleValidation({
     }
 
     setFieldError?.(path, errors ? new Error('Form.Isolation') : undefined)
+
+    return () => {
+      setFieldError?.(path, undefined)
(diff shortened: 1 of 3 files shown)


joakbjerk (approved):
Looks good, seems to have fixed the issue 🙏

PR #4154, opened 2024-10-18
feat(DrawerList, Dropdown, Autocomplete, Field.Selection, Field.ArraySelection): disabled options

snorrekim:
* Individual entries in `DrawerList` can now be disabled
* Works with `Dropdown`, `Autocomplete`, `Field.Selection`, and `Field.ArraySelection`

Other changes:
* `selected_key` prop deprecated and can be replaced by `selectedKey`
* Updated types for `DrawerList` props
* `value` also checks index if no `selectedKey` is a match
* `title` overrides `children` in `Field.Option` used by `Field.Selection` and `Field.ArraySelection`

PR #4480, opened 2025-01-21
feat(DrawerList): add `overflowingTextOptions` for CSS `hyphens`, `overflow-wrap`, `word-break`, `overflow`

langz:
Initial feature request here: https://dnb-it.slack.com/archives/CMXABCHEY/p1737113652911309


haigol:
Thanks for checking this out! 

Thought it made sense for long words without spaces, as my use case needs, but seeing that it also breaks sentences that simply could span several lines makes me unsure 🤔 I like the idea of being able to send it as prop

langz:
> Thanks for checking this out! 
> 
> 
> 
> Thought it made sense for long words without spaces, as my use case needs, but seeing that it also breaks sentences that simply could span several lines makes me unsure 🤔 I like the idea of being able to send it as prop

I can make a new property called something like `itemHyphenation`, which will have the possible values from the hyphens CSS property https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens, and will default to the value none if not specified. 

Or maybe just ```
hyphenation
```?

**Comment by haigol** [comment]
I think this sounds like a good idea! 
`optionHyphens` `optionItemsHyphenation`? always tricky...

 Thank you!

**Comment by langz** [comment]
> I think this sounds like a good idea! `optionHyphens` `optionItemsHyphenation`? always tricky...
> 
> Thank you!

Thanks, for now in the updated commit, I just called it `hyphenation`.
Any thoughts regarding the naming @tujoworker ?

File: packages/dnb-design-system-portal/src/docs/uilib/components/fragments/drawer-list/Examples.tsx [hunk 1]
```diff
@@ -106,7 +106,7 @@ export const DrawerListExampleOnlyToVisualize = () => (
 
 export const DrawerListExampleDefault = () => (
   <Wrapper>
-    <ComponentBox scope={{ data }} data-visual-test="drawer-list-default">
```
langz, line 109: This data-visual-test value was not used, hence removing it.

packages/dnb-eufemia/src/fragments/drawer-list/DrawerList.d.ts
```diff
@@ -75,6 +76,10 @@ export interface DrawerListProps {
    * Defines the direction of how the drawer-list shows the options list. Can be 'bottom' or 'top'. Defaults to 'auto'.
    */
   direction?: DrawerListDirection;
+  /**
+   * Defines how words should be hyphenated when text of an item wraps across multiple lines. Can be 'none', 'manual' or 'auto'. Defaults to 'null'.
```
langz, line 80:
```suggestion
   * Defines how words should be hyphenated when text of an item wraps across multiple lines. Can be 'none', 'manual' or 'auto'.
```

packages/dnb-eufemia/src/fragments/drawer-list/__tests__/__image_snapshots__/drawerlist-for-sbanken-have-to-match-the-disabled-option.snap.png
langz, line 1: Not entirely sure what can have caused this 🤔

langz, line 1: Not entirely sure what can have caused this 🤔

langz: I think I found the culprit

langz: No, I did not 🤔 
But the tests does perhaps seem to be a bit shaky, as it was added padding-top in screenshot tests in https://github.com/dnbexperience/eufemia/pull/4154

packages/dnb-eufemia/src/fragments/drawer-list/DrawerList.js
```diff
@@ -325,7 +326,11 @@ class DrawerListInstance extends React.PureComponent {
         }
 
         return (
-          <DrawerList.Item key={hash} {...liParams}>
+          <DrawerList.Item
+            key={hash}
+            hyphenation={hyphenation || dataItem.hyphenation}
```
langz, line 331:
```suggestion
            hyphenation={dataItem.hyphenation || hyphenation}
```

langz:
This feature is now working as per my expectations, and can be reviewed 🙏

joakbjerk (approved):
Nice, looks good to me 🔥

tujoworker:
Instead of "overflowing", are one of these better alternatives?

- textOverflowConfig
- overflowTextProps
- textStyleOptions
- overflowStyle

cc @joakbjerk @snorrekim

langz:
> * overflowTextProps

`overflowTextProps`?

tujoworker:
I think we should rethink a better model. I don't think it's worth to document so many options. 

So if we could make it work with "all" styles, by using "inline" styles, we are probably better off. 

Less code. Less to document and test and keep up to date.

So then we probably should also mention "item" in the prop. 

Like "itemTextStyle" or "itemStyle".

We can also consider a totally independent render prop to let devs manipulate each item, before it gets inserted in the dom. But that's probably too far away from the original intention.

langz:
> I think we should rethink a better model. I don't think it's worth to document so many options.
> 
> So if we could make it work with "all" styles, by using "inline" styles, we are probably better off.
> 
> Less code. Less to document and test and keep up to date.
> 
> So then we probably should also mention "item" in the prop.
> 
> Like "itemTextStyle" or "itemStyle".
> 
> We can also consider a totally independent render prop to let devs manipulate each item, before it gets inserted in the dom. But that's probably too far away from the original intention.

Fine by me, not sure how to take this further at the given time.
The initial thought was just to provide @haigol the possibility of setting the `hyphens` property using JS instead of adding custom CSS on her end 🙏

tujoworker:
Lets have a chat tomorrow about this.

langz:
This PR is replaced with the following PR, which rather just documents and adds tests for existing inline styling functionality.
https://github.com/dnbexperience/eufemia/pull/4513

PR #4481, opened 2025-01-21
feat(Forms): add `transformSelection` to Field.Selection

tujoworker:
This is only for the Dropdown and Autocomplete.

[Demo](https://eufemia-git-feat-forms-selection-transform-eufemia.vercel.app/uilib/extensions/forms/base-fields/Selection/#dropdown-with-a-transformed-selection-text).

Screenshot 2025-01-21 at 12 51 22
Screenshot 2025-01-21 at 12 49 00

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/Selection/SelectionDocs.ts
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/Selection/SelectionDocs.ts
@@ -11,6 +11,11 @@ export const SelectionProperties: PropertiesTableProps = {
     type: ['number', 'string'],
     status: 'optional',
   },
+  transformSelection: {
+    doc: 'Transform the displayed selection for Dropdown and Autocomplete variant. Use it to display a different value than the one in the data set. The first parameter is the props of the Option component or data item. You can return a React.Node that will be displayed in the selection.',
+    type: 'function',
+    status: 'optional',
+  },
   optionsLayout: {
     doc: 'Layout for the list of options. Can be `horizontal` or `vertical`.',
     type: 'string',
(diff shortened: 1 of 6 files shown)


langz (approved):
We should probably say to the users somewhere in our docs that it's only usable with variants dropdown and autocomplete. 

What happens if this prop is used with a different variant? Will it break?

tujoworker:
> We should probably say to the users somewhere in our docs that it's only usable with variants dropdown and autocomplete.

We do it in the property description. Is that enough you think?

_Transform the displayed selection for Dropdown and Autocomplete variant._ 

> What happens if this prop is used with a different variant? Will it break?

The function does not run for other variants.

packages/dnb-eufemia/src/extensions/forms/Field/Selection/SelectionDocs.ts
```diff
@@ -11,6 +11,11 @@ export const SelectionProperties: PropertiesTableProps = {
     type: ['number', 'string'],
     status: 'optional',
   },
+  transformSelection: {
+    doc: 'Transform the displayed selection for Dropdown and Autocomplete variant. Use it to display a different value than the one in the data set. The first parameter is the props of the Option component or data item. You can return a React.Node that will be displayed in the selection.',
```
langz, line 15: 🫡

langz (approved):

snorrekim:
I like it 👍

tujoworker:
> We might want to consider merging `children and `title`. Normally, `children`and`title`are mutually exclusive. If you use`title`, `children` is ignored. So perhaps we should use the same logic in the transform?
> 
> So `title` returns `title || children`. Should the transform callback match the props in `Field.Option` or the fields in the presentation? I don't really know which is preferred here. It can be very useful to just get all the props like it is now.

Yes maybe we should do something with it. But not in this PR.

langz (approved):

PR #4484, opened 2025-01-22
chore(DrawerList): removes unused `data-visual-test` props

langz:


packages/dnb-eufemia/src/fragments/drawer-list/__tests__/DrawerList.screenshot.test.ts
```diff
@@ -28,7 +28,6 @@ describe.each(['ui', 'sbanken'])('DrawerList for %s', (themeName) => {
     const screenshot = await makeScreenshot({
       style: {
         width: '14rem',
-        'padding-top': '3rem',
```
langz, line 31: I'm not 100% sure why this was added in https://github.com/dnbexperience/eufemia/pull/4154
But I'm pretty confident that it's safe to remove :)

PR #4485, opened 2025-01-22
chore: use calc with var(--modal-z-index)

langz:


packages/dnb-eufemia/src/components/date-picker/style/dnb-date-picker.scss
```diff
@@ -515,7 +515,7 @@
     position: absolute;
     left: 0;
     // Over modal, but below tooltip
-    z-index: var(--modal-z-index + 10);
+    z-index: calc(var(--modal-z-index) + 10);
```
langz, line 518: Or maybe the intention was not to use calc? But rather add 10 to `var(--modal-z-index)`(3200) whoich would be 3200+10 = 320010?

tujoworker: No, this is a bug 🐞

packages/dnb-eufemia/src/fragments/drawer-list/style/dnb-drawer-list.scss
```diff
@@ -83,7 +83,7 @@
     }
 
     // higher than modal --modal-z-index:
-    z-index: 3200;
+    z-index: calc(var(--modal-z-index) + 200);
```
tujoworker, line 86:
```suggestion
    z-index: calc(var(--modal-z-index, 3200) + 200);
```

We probably need to still have a fallback, in case the variable is not present.

joakbjerk (approved):

PR #4488, opened 2025-01-22
chore(Portal): allow links in types and make primitive types purple

snorrekim:


snorrekim:
The idea with the new colors is that any type that isn't a primitive should have a different color. Making them more noticeable.

Primitive types are `boolean`, `number`, `bigint`, `string`, `symbol` and `object` as well as any array of them like `string[]` or `Array<number>`

Comparison to see what I mean:
image

snorrekim:
`object` should probably also be purple. Like `function` it doesn't tell the full story. (and it isn't technically a primitive type)

langz (approved):
Looks good to me :) 

The pull request title, `chore(Portal): allow links in types and make primitive types purple` doesn't seem to align with a statement in the PR description `Make non-primitive types purple`. Should primitive or non-primitive types be purple? 🤔 
Consider updating either description or title 🙏

snorrekim:
> Looks good to me :)
> 
> The pull request title, `chore(Portal): allow links in types and make primitive types purple` doesn't seem to align with a statement in the PR description `Make non-primitive types purple`. Should primitive or non-primitive types be purple? 🤔 Consider updating either description or title 🙏
> 
> In the code, I see `const colorPrimitive = 'var(--color-success-green)'` so primitive types is green and not purple, right? 💭

Right, the title is wrong, I'll fix it. Primitive types should be green. I'm also going to add `false` and `true` to the list of primitives.

tujoworker (changes requested):
I believe the property name plays a much more important role during the process of scanning and consuming the documentation. However, the use of colors can be distracting and take away from effectively focusing on the content. 

I think we should evaluate other ways to communicate primitives, if it is important to distinguish them.

snorrekim:
Changed the primitives color back to green for now to finish the PR. The option is still there for the future.

tujoworker (approved):
Nice 👏 Thank you for this 🎉

PR #4496, opened 2025-01-24
feat(Flex): improve `gap` and `rowGap` behaviour

snorrekim:
## Changes
* give `rowGap` the same sizes as `gap` by adding `x-small`, `xx-small`, `x-large` and `xx-large`
* stop adding both `gap` and `rowGap` to at the same time in vertical layout. Only use the prop set by the user. If both are set, `rowGap` will override `gap`.
* fix right margin not scaling with `gap` on horizontal layout

## Minor changes/details
* In vertical layout, we no longer use flex-gap. Margins are always used so individual child items can override their gap.
* remove value `true` from `rowGap`, it will treat it the same as `undefined` (results in `'small'`) (`true` will be removed in v11)
* `rowGap` is no longer affected by the use of `size` or a heading in children items
* Bugfix: The css variable `--gap` will no loner be erroneously inherited by parent `<Flex>` components
* Portal: allow strikethrough markdown in properties description


snorrekim:
Perhaps we should just set the ccs variable `--gap: 0` as default to avoid having the `.dnb-flex-container--row-gap-off` class everywhere?

snorrekim:
The issues only really affect vertical direction. So heres some examples for discussion:

Today: `gap` and `rowGap` is added, but `rowGap` is zero by default
PR: `rowGap` is ignored
REV: `gap` is ignored
SET: unset props are zero, unless no prop is set, then `gap` or `rowGap` is `small`

Always makes sense
```
<Flex.Vertical >
// Today: small (margin)
// PR: small (margin)
// REV: small (flex)
// SET: small (flex/margin)
```

Is always strange, this is the one we should decide on
```
<Flex.Vertical gap="large" rowGap="medium" >
// Today: large (margin) + medium (flex) 
// PR: large (margin) 
// REV: medium (flex)
// SET: large (margin) + medium (flex) 
```

"REV" and "SET" makes sense
```
<Flex.Vertical rowGap="medium" >
// Today: small (margin) + medium (flex)
// PR: small (margin)
// REV: medium (flex)
// SET: medium (flex)
```

"PR" and "SET" makes sense
```
<Flex.Vertical gap="large" >
// Today: large (margin)
// PR: large (margin)
// REV: small (flex)
// SET: large (margin)
```

snorrekim:
## Margin or flex-gap?

Right now the `gap` and `rowGap` are trying to answer two different questions at the same time. "Should we use margin or flex-gap?" and "what should decide row gap?`

If we can answer one of those questions, then we could more easily fix the code.

* Do we need to give the option of using both margin and flex-gap?
* Do we want `colGap` as well perhaps?
* Is margin only needed for horizontal gaps?

snorrekim:
Some new issues I've found regarding wrapping children, left right spacing, and dividers of items:

### left and right spacing
On horizontal layout: 
If you add left spacing to a child it is applied to both left and right
If you add right spacing it is also applied to the next element

### wrapping children

We handle fragments and `_supportsSpacingProps='children'` sometimes the same and sometimes different. Sometimes it creates a new flex.container, and sometimes it adds the wrapper to each child

This applies the Wrapper around every Item:
```
<Flex.Container>
  <>
    <Wrapper>
      <Item>1</Item>
      <Item>2</Item>
      <Item>3</Item>
    </Wrapper>
  </>
</Flex.Container>

```

This puts the items inside another Flex.Container inside the Wrapper
```
<Flex.Container>
    <Wrapper>
      <Item>1</Item>
      <Item>2</Item>
      <Item>3</Item>
    </Wrapper>
</Flex.Container>
```

there are also a lot of other weird behaviour.

Should we not be more clear on exactly what a fragment does vs a `_supportsSpacingProps='children'`? Fragments should always just be ignored. But  what exactly should `_supportsSpacingProps='children'` do? And it's use case in vertical is different than in horizontal? We should have a clear definition so we can make the code work. Because right now it's doing a lot of unpredictable things.

theres also issues with inheriting spacing form parents vs itself. Itself should override, but sometimes it doesn't because of the wrapper logic.

Wrapper with `_supportsSpacingProps='children'` works differently for vertical than horizontal. On horizontal, it can act as if the children are just part of the flex layout. But on horizontal it just creates a new Flex.Container that (buggily) inherits the gap of the parent Flex.Container. You might as well just use Flex.Container with the same gap. 

It also does something different depending on what level it is on. 
* On the first level, it wraps and creates a new Flex.Container with its children (space-wrapped) inside
* On the second level (in a fragment) it wraps **each** of it's children (space-wrapped). This is the oddest.

Or said differently:
* On the first level, a fragment is ignored, and a Wrapper creates a new Flex.Container
* On the second level, a fragment is treated as a Wrapper. But since a fragment

We need to define:
* What is the purpose of the Wrapper?
* Should it wrap all the children, or each of the children? Not either depending on level like now.

### Dividers only really work for vertical layout

The code, tests and examples suggests that we think it should work on horizontal, but it really doesn't. Is this a mistake? As I see it, it should only work on vertical, or it should change the direction to vertical.

tujoworker (approved):
Looks good to me. Great work 🚀

PR #4499, opened 2025-01-26
feat(Forms): add `itemPath` property support to `Iterate.PushButton` and `Iterate.PushContainer`

### LINKED ISSUE #4417: Dynamic list within `Iterate`
## 🚀 Feature Proposal

A dynamic list, where user can add and remove objects easily (like `Iterate.Array`) which works inside of the `Iterate` context, both in `Iterate.Array` and `Iterate.PushContainer`.

## Motivation

`Iterate.Array` works well for both "large" objects, with the possibility of making custom view and edit containers, as well as with simple single value lists.  
It is however not possible to use `Iterate` within `Iterate`.  
It would be great to be able to make a list of some data while in the context of an object already in `Iterate`.

## Example
We have a design where we are adding people to a list.
These people can have one or more citizenships.
Currently there is no easy way of handling this use case. `Iterate.Array` works great for making simple dynamic lists with `PushButton`, but it does not work within `Iterate`

bilde

## Pitch
It seems to me to be a natural progression of `Iterate`.

## CSB
Here is a CSB which shows how usage of this could look.  
Note that this is (obviously) not working right now. The code might be of interest, not the result of running the code.

https://codesandbox.io/p/sandbox/show-iterate-within-iterate-wq9snc

tujoworker:
Closes #4417

Here is a [working example](https://eufemia-git-feat-iterate-push-container-item-path-eufemia.vercel.app/uilib/extensions/forms/Iterate/demos/#nested-iterate-with-pushcontainer) based on the issue reprod from @andlbrei .

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Iterate/PushButton/PushButtonDocs.ts
+++ b/packages/dnb-eufemia/src/extensions/forms/Iterate/PushButton/PushButtonDocs.ts
@@ -6,6 +6,11 @@ export const PushButtonProperties: PropertiesTableProps = {
     type: 'string',
     status: 'required',
   },
+  itemPath: {
+    doc: 'The path to the item in a nested array to add the new item to.',
+    type: 'string',
+    status: 'optional',
+  },
   pushValue: {
     doc: 'The element to add to the array when the button is clicked. Can be a function to returns the push value.',
     type: 'unknown',
--- a/packages/dnb-eufemia/src/extensions/forms/Iterate/PushButton/PushButtonDocs.ts
+++ b/packages/dnb-eufemia/src/extensions/forms/Iterate/PushButton/PushButtonDocs.ts
@@ -7,7 +7,7 @@ export const PushButtonProperties: PropertiesTableProps = {
     status: 'required',
   },
   itemPath: {
-    doc: 'The path to the item in a nested array to add the new item to.',
+    doc: 'The path to the item in a nested array, to add the new item to.',
     type: 'string',
     status: 'optional',
   },
--- a/packages/dnb-eufemia/src/extensions/forms/Iterate/PushContainer/PushContainer.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Iterate/PushContainer/PushContainer.tsx
@@ -17,18 +17,43 @@ import OpenButton from './OpenButton'
 import { Flex, HeightAnimation } from '../../../../components'
 import { OnCommit, Path } from '../../types'
 import { SpacingProps } from '../../../../shared/types'
-import { useArrayLimit, useSwitchContainerMode } from '../hooks'
+import {
+  useArrayLimit,
+  useSwitchContainerMode,
+  useItemPath,
+} from '../hooks'
 import Toolbar from '../Toolbar'
 import { useTranslation } from '../../hooks'
 import { ArrayItemAreaProps } from '../Array/ArrayItemArea'
 import { clearedData } from '../../DataContext/Provider'
 
-export type Props = {
+/**
+ * Deprecated, as it is supported by all major browsers and Node.js >=v18
+ * So it's a question of time, when we will remove this polyfill
+ */
+import structuredClone from '@ungap/structured-clone'
+
+type OnlyPath = {
   /**
    * The path to the array to add the new item to.
    */
   path: Path
 
+  /** The sub path to the array to add the new item to. */
+  itemPath?: Path
+}
+
+type OnlyItemPath = {
+  /**
+   * The path to the array to add the new item to.
+   */
+  path?: Path
+
+  /** The sub path to the array to add the new item to. */
+  itemPath: Path
+}
+
+export type Props = (OnlyPath | OnlyItemPath) & {
   /**
    * The title of the container.
    */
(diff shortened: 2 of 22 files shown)


packages/dnb-eufemia/src/extensions/forms/Iterate/PushButton/PushButtonDocs.ts
```diff
@@ -6,6 +6,11 @@ export const PushButtonProperties: PropertiesTableProps = {
     type: 'string',
     status: 'required',
   },
+  itemPath: {
+    doc: 'The path to the item in a nested array to add the new item to.',
```
langz, line 10:
```suggestion
    doc: 'The path to the item in a nested array, to add the new item to.',
```

packages/dnb-eufemia/src/extensions/forms/Iterate/PushContainer/PushContainer.tsx
```diff
@@ -17,18 +17,43 @@ import OpenButton from './OpenButton'
...
   /**
    * The path to the array to add the new item to.
    */
   path: Path
 
+  /** The sub path to the array to add the new item to. */
+  itemPath?: Path
+}
+
+type OnlyItemPath = {
```
langz, line 46:
```suggestion
type OnlyItemPath = {
```

If the type is "only item path", does it still make sense to have the path property here as well? As the type says "only item path"?

Or perhaps it could be named `OnlyItemPathRequired` or so?
Same for `OnlyPath`, perhaps `OnlyPathRequired`?

packages/dnb-eufemia/src/extensions/forms/Iterate/PushContainer/PushContainerDocs.ts
```diff
@@ -7,6 +7,11 @@ export const PushContainerProperties: PropertiesTableProps = {
     type: 'string',
     status: 'required',
   },
+  itemPath: {
+    doc: 'The path to the item in a nested array to add the new item to.',
```
langz, line 11:
```suggestion
    doc: 'The path to the item in a nested array, to add the new item to.',
```

packages/dnb-eufemia/src/extensions/forms/Iterate/PushContainer/PushContainer.tsx
```diff
@@ -17,18 +17,43 @@ import OpenButton from './OpenButton'
...
+ */
+import structuredClone from '@ungap/structured-clone'
+
+type OnlyPath = {
   /**
    * The path to the array to add the new item to.
    */
   path: Path
 
+  /** The sub path to the array to add the new item to. */
```
langz, line 42:
```suggestion
  /** The sub path to the array, to add the new item to. */
```

packages/dnb-eufemia/src/extensions/forms/Iterate/PushContainer/PushContainer.tsx
```diff
@@ -17,18 +17,43 @@ import OpenButton from './OpenButton'
...
    */
   path: Path
 
+  /** The sub path to the array, to add the new item to. */
+  itemPath?: Path
+}
+
+type OnlyItemPath = {
+  /**
+   * The path to the array to add the new item to.
```
langz, line 48:
```suggestion
   * The path to the array, to add the new item to.
```

packages/dnb-eufemia/src/extensions/forms/Iterate/PushContainer/PushContainer.tsx
```diff
@@ -17,18 +17,43 @@ import OpenButton from './OpenButton'
...
+   */
+  path?: Path
+
+  /** The sub path to the array to add the new item to. */
```
langz, line 52:
```suggestion
  /** The sub path to the array, to add the new item to. */
```

langz (approved):
🎸 👨‍🎤

tujoworker:
I have added an [working example](https://eufemia-git-feat-iterate-push-container-item-path-eufemia.vercel.app/uilib/extensions/forms/Iterate/demos/#nested-iterate-with-pushcontainer).

andlbrei:
Great work 👍 
Tested this with my CSB, which might be "outdated" or it might have exposed some issues.

Seems like it has some trouble in a PushContainer when there is not previous value.
Doesn't seem like it crashes completely though, it just complains.

Also for some reason when there are existing values in the citizenships array, when going to edit mode and adding citizenships it just closes the edit container.

https://github.com/user-attachments/assets/64468eae-10e2-4882-8e05-7da849ef2352

tujoworker:
I see your PushButton has no pushValue, but that was not the issue. I pushed a new commit, which should fix it. You may wait until codesandbox-ci has created a new version and test it again.

```tsx
<Iterate.PushButton
  itemPath="/citizenships"
  text="Add citizenship"
  pushValue="Four of three"
/>
```

tujoworker:
Nice findings. Thank you. It should stay in edit mode now.

> BTW, should pushValue be a required prop, or will it default to a sensible value?

It pushes an undefined value. So I think its just fine as is.

andlbrei:
> Nice findings. Thank you. It should stay in edit mode now.
> 
> > BTW, should pushValue be a required prop, or will it default to a sensible value?
> 
> It pushes an undefined value. So I think its just fine as is.

Yea, looks good now 👍

PR #4502, opened 2025-01-27
feat(Forms): add `Iterate.Visibility` to be used within `Iterate.Array` (relative paths)

### LINKED ISSUE #4472: `Visibility` with `pathUndefined` etc. is not working for `itemPath` in `Iterate`
## 🐛 Bug Report

Not sure if this is a bug, or just not supported. But anyways...

Using `Visibility` with `pathxxx` like `pathUndefined` does not seem to work on `itemPath`s in `Iterate`.  
`Visibility` with `pathUndefined="/someItemPath"` will always show the content, because the `itemPath` will be considered `undefined`.  

We have some data we fetch from the server. Sometimes certain data fields are missing, and if they are missing we would like to show a `Field` to insert the data. If the data is there we show a `Value`.  
Using `pathUndefined` for this seems perfect, but it's not working for `itemPath`/within `Iterate`.

## To Reproduce

https://codesandbox.io/p/sandbox/itempath-visibility-rgnjyt

## Expected behavior

I expect (or would like) it to be possible to use `Visibility` for `itemPath` in the exact same way as for `path`.

## Eufemia Version

10.62.4

tujoworker:
Closes #4472

- [x] We need to discuss the API.
- [x] Add example/docs.


packages/dnb-eufemia/src/extensions/forms/Form/Visibility/__tests__/Visibility.test.tsx
```diff
@@ -982,6 +982,384 @@ describe('Visibility', () => {
...
+              Child
+            </Visibility>
+          </Iterate.Array>
+        </Provider>
+      )
+      expect(screen.queryByText('Child')).not.toBeInTheDocument()
+    })
+  })
+
+  describe('pathUndefined', () => {
```
andlbrei, line 1049: The names in this describe seems to indicate the opposite of the test

tujoworker: Very true indeed.

langz (approved):

PR #4503, opened 2025-01-28
chore: ReferenceError: Tools is not defined

langz:
Removes ReferenceError: Tools is not defined in https://eufemia.dnb.no/uilib/extensions/forms/feature-fields/more-fields/Upload/demos/#session-storage-support


tujoworker (approved):
We may add Tools as a default import as well 🐮

PR #4513, opened 2025-01-28
chore(DrawerList): add test and example for inline styling

langz:


langz:
TODO in an other PR:
- Add tests for inline style prop in Eufemia forms as well, for Field.Selection variants dropdown and autocomplete

tujoworker (approved):
That looks lovely 😍

PR #4516, opened 2025-01-29
chore(Field.Option): remove unused code

langz:
I don't see that the `Field.Option` code is actually used at all 🤔 

It's only the types that's used, as the actual JSX code that `Field.Option` had before this PR would never be ran, it would just be overridden in `Field.Selection` and `Field.ArraySelection`. And as it says in [docs](https://eufemia.dnb.no/uilib/extensions/forms/base-fields/Option/#description): _Field.Option is a pseudo-component that is used by parent components to configure options. It has no use on its own._

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/Option/Option.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/Option/Option.tsx
@@ -1,5 +1,4 @@
 import React from 'react'
-import classnames from 'classnames'
 import type { FieldProps } from '../../types'
 
 export type Props = FieldProps<number | string> & {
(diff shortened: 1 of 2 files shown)


packages/dnb-eufemia/src/extensions/forms/Field/Selection/Selection.tsx
```diff
@@ -435,6 +435,7 @@ function renderDropdownItems(
   data: Data,
   transformSelection?: Props['transformSelection']
 ) {
+  console.log(data)
```
langz, line 438:
```suggestion
```

tujoworker (approved):
Yes, that's true. I think the intention from Henrik was that it may be used one time. But I'm fine of removing it.
We just "pick" the props from inside Selection.

PR #4520, opened 2025-01-30
fix(Skeleton): move CSS variables for Skeleton from a global scope to themed skeleton files

langz:


packages/dnb-eufemia/src/components/skeleton/style/themes/dnb-skeleton-theme-sbanken.scss
```diff
@@ -2,6 +2,6 @@
  * No theme is provided
  *
  * Theming variables for the Skeleton component is required globally
- * and is thus defined in src/style/themes/theme-sbanken/customisations.scss
```
langz, line 5: This path is not existing, what is the correct path?

tujoworker: In the `globals.scss` file there is `--skeleton-color` and `--skeleton-color--contrast`.

tujoworker: But these could actually be moved into the empty theme files instead. Would work just fine. Probably because of the new setup we got with Sbanken.

packages/dnb-eufemia/src/components/skeleton/style/themes/dnb-skeleton-theme-sbanken.scss
```diff
@@ -2,6 +2,6 @@
...
+ * and is thus defined in what's the correct path?
```
langz, line 5:
```suggestion
 * and is thus defined in src/style/themes/theme-sbanken/sbanken-theme-basis.scss?
```

PR #4523, opened 2025-01-30
fix(types): bump @types/react to v18.2.21

langz:
Not sure if it will work or not, will have to validate in a CSB.
Motivation is to update types, so that React.CSSProperties will have the latest/most updated properties available.
In v10.63.0 of Eufemia, the React.CSSProperties type does not accept hyphens as property, see this [CSB](https://codesandbox.io/p/sandbox/elated-frog-xr2x86?workspaceId=ws_LbiDwJdEBdbwxH2AkyAyVE).

Hopefully that will work with this bump, will have to make a new csb and validate.

### DIFF
--- a/packages/dnb-design-system-portal/package.json
+++ b/packages/dnb-design-system-portal/package.json
@@ -61,7 +61,7 @@
     "@emotion/cache": "11.11.0",
     "@playwright/test": "1.49.1",
     "@types/json-schema": "7.0.12",
-    "@types/react": "18.2.13",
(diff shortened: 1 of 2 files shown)


tujoworker (approved):

langz:
No need, it's just a devdependency, can be solved like this https://codesandbox.io/p/sandbox/xenodochial-volhard-xrlcww?workspaceId=ws_LbiDwJdEBdbwxH2AkyAyVE

PR #4529, opened 2025-02-03
fix(Forms): ensure `setFieldStatus` by useValidation does not throw when used without id

tujoworker:
More [info](https://dnb-it.slack.com/archives/CMXABCHEY/p1738590820694909).


langz (approved):

tujoworker:
The issue with this PR is that it was not an issue as long as the hook was used as documented. But the reported issue is much more complicated to solve. So we do it with another PR. Probably even with a new hook.

PR #4533, opened 2025-02-04
feat(Forms): add `valueDefined` to Form.useVisibility

tujoworker:
Instead of:

```tsx
<Form.Visibility
  visibleWhen={{
    path: '/county',
    hasValue: (value) => value !== undefined,
  }}
  animate
>
```

It could now be written like so:

```tsx
<Form.Visibility
  valueDefined="/county"
  animate
>
```

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Form/Visibility/Visibility.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Form/Visibility/Visibility.tsx
@@ -74,6 +74,8 @@ export type Props = {
   pathTrue?: Path
   /** Given data context path must be false to show children */
   pathFalse?: Path
+  /** Given data context path must have a value other than "undefined" to show children */
+  valueDefined?: Path
   /** Provide a `path` or `itemPath` and a `hasValue` method that returns a boolean or the excepted value in order to show children. The first parameter is the value of the path. */
   visibleWhen?: VisibleWhen
   /** Same as `visibleWhen`, but with inverted logic. */
@@ -117,6 +119,7 @@ function Visibility(props: Props) {
     pathFalse,
     pathValue,
     whenValue,
+    valueDefined,
     visibleWhen,
     visibleWhenNot,
     inferData,
@@ -148,6 +151,7 @@ function Visibility(props: Props) {
     pathTrue,
     pathFalse,
     pathValue,
+    valueDefined,
     whenValue,
     visibleWhen,
     visibleWhenNot,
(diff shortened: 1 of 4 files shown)


packages/dnb-eufemia/src/extensions/forms/Form/Visibility/Visibility.tsx
```diff
@@ -74,6 +74,8 @@ export type Props = {
   pathTrue?: Path
   /** Given data context path must be false to show children */
   pathFalse?: Path
+  /** Given data context path must have a value other than "undefined" to show children */
+  valueDefined?: Path
```
langz, line 78: I don't initially understand how this prop `valueDefined` is different from `pathDefined` 🤔 
And for the comment/description of the prop the word value is mentioned, but that's not mentioned in any other `pathX`-prop. Is the value only used/meaningful for the new prop `valueDefined`? 🤔

langz:
> Instead of:
> 
> ```tsx
> <Form.Visibility
>   visibleWhen={{
>     path: '/county',
>     hasValue: (value) => value !== undefined,
>   }}
>   animate
> >
> ```
> 
> It could now be written like so:
> 
> ```tsx
> <Form.Visibility
>   valueDefined="/county"
>   animate
> >
> ```

Couldn't we write the same today using `pathDefined`?

```
<Form.Visibility
  pathDefined="/county"
  animate
>
```

packages/dnb-eufemia/src/extensions/forms/Form/Visibility/Visibility.tsx
tujoworker: Yes. All with "part" i. The word targets the status of the path itself and not the value. While this one does target only the value. 

We can have a chat about it tomorrow.

tujoworker:
I did have a comparison. It results not in the same. When using `pathDefined`, the data context can not be e.g. `{ thisExistsButIsUndefined: undefined }`. So it results in `true`. While `valueDefined` in `false`. That said, I think we can close this PR and rather fix `pathDefined` to result in `true`, with such a data context: `{ thisExistsButIsUndefined: undefined }`.

langz:
> I did have a comparison. It results not in the same. When using `pathDefined`, the data context can not be e.g. `{ thisExistsButIsUndefined: undefined }`. So it results in `true`. While `valueDefined` in `false`. That said, I think we can close this PR and rather fix `pathDefined` so it results in `true`, with such a data context: `{ thisExistsButIsUndefined: undefined }`.

Thanks, sounds perfect 🙇

tujoworker:
Closing this in favor of #4539 and 4540

Issue #4461, opened 2025-01-15
Visibility does not render correctly when manipulating data with `useData` on form with `id`


andlbrei:
## 🐛 Bug Report

When using `Visibility` to show/hide fields when also doing `useData.remove()` or `useData.update`, there are issues in updating the values.  
Our form has a boolean field which controls the visibility of a set of fields. `onChange` on this field will remove/update to reset an array of values when the value is `false`. A `false` value will also hide the set of fields, using `Visibility`.  

This does not work well when the form has an `id`. `update` and `remove` works, but `Visibility` does not, seemingly because the value for the path used to control visibility is not updated in the form data context.

https://github.com/user-attachments/assets/d95b7bac-a9a1-4bcc-a5bd-b0f5a03ce3e7

And another video, with a form without `id`:

https://github.com/user-attachments/assets/b704367b-c010-47de-95d5-2288ab7c4f09

## To Reproduce

https://codesandbox.io/p/sandbox/visibility-with-id-6d7yzx

- There are two similar components, one uses `update` and the other uses `remove`. There are console logs showing inconsistency in the path value from `data` vs the value in `onChange`.  Comment out one of the components to make the logs easier to read.

## Expected behavior

I expect path values to be updated consistently when using `id` in a `Form.Handler`, like when not using `id`.

## Eufemia Version

10.62.4

tujoworker:
It seems that there could be a timing issue. When I delay the `update` routine, it looks like it works fine. Are you able to check that out?

```tsx
setTimeout(() => {
  update('/beneficialOwners/otherBeneficialOwners2', [])
}, 1e3)
```

tujoworker:
The `remove` method will probably always have racing conditions. Because a "Field.*" wants to set/sync its internal value with the data context. So it will set `undefined` after it got remove from the data context. 

Do you have any thoughts about this?

andlbrei:
Hmm, I don't know.
I wonder why it works fine without an `id`, but not with an `id`?
Is the "race" that much different when an `id` is set?

tujoworker:
Currently there is an issue. I have fix so everything works. It's just that the value never gets removed. But that does not negatively influence the field. So I think that is the fine.

tujoworker:
Yes

andlbrei:
Ok, that will probably be fine. I think ajv treats `undefined` different from `null`, which would be a problem with validation.

tujoworker:
[...]

PR #4535, opened 2025-02-04
chore: upgrade Storybook to v8

tujoworker:
To work on issue #4461 it would be preferable to upgrade Storybook from v7 to v8.


langz (approved):
I can give it a test locally tomorrow 🙏

langz:
I did find a error related to the upgrade:
Screenshot 2025-02-05 at 09 24 41

It exists in multiple stories.
Looking into fixing it now.

tujoworker:
Yes, I do. We should look into the config to enable `runtime: 'automatic'`.

packages/dnb-eufemia/.storybook/main.ts
```diff
@@ -13,6 +14,16 @@ const config: StorybookConfig = {
     name: getAbsolutePath('@storybook/react-webpack5'),
     options: {},
   },
+  swc: (config: Options, options): Options => {
+    if (config.jsc) {
```
langz, line 18: Nice, I never thought of adding the if statement, and got other errors then 😓

langz:
http://localhost:8002/?path=/story/eufemia-components-sectionlayout--section-layout

image

Not sure if related to upgrade or just an existing bug, will check later.

langz:
http://localhost:8002/?path=/story/eufemia-extensions-forms-visibility--filter-data-handler

image

Not sure if related to upgrade or just an existing bug, will check later.

tujoworker:
Yeah, looks like SWC and Storybook are both more stricter than v7 together with Babel. So we need to adjust all the findings. I did commit two of them. Not looking further at the moment.

langz (approved):

PR #4537, opened 2025-02-05
chore: add test for inline `style` property

langz:


tujoworker (approved):
wow, that was a lot of tests 🚀

PR #4542, opened 2025-02-05
feat(Forms): add `createMinimumAgeVerifier` to Field.NationalIdentityNumber

tujoworker:
Not sure if we really need the hook `useMinimumAge`. But it may save some lines of code?

```tsx
import {
  createMinimumAgeVerifier,
} from '@dnb/eufemia/extensions/forms/Field/NationalIdentityNumber'

const isAdult = createMinimumAgeVerifier(18)

isAdult('123') // false
isAdult('10072476609') // false
isAdult('09100654021') // true
```


langz (approved):

langz (approved):

langz:
Could the following naming convention work, `minimumAge` and `useMinimumAge`?


```tsx
import {
  minimumAge,
  useMinimumAge,
} from '@dnb/eufemia/extensions/forms/Field/NationalIdentityNumber'

const isAdult = minimumAge(18)
isAdult('123') // false
isAdult('10072476609') // false
isAdult('09100654021') // true

function Component() {
  const isAdult = useMinimumAge(18)

  isAdult('123') // false
  isAdult('10072476609') // false
  isAdult('09100654021') // true
}
```

tujoworker:
In the meeting we decided to remove the hook, as it is redundant. 
Regarding the naming, what about "verifier"? I made a commit to name it `createMinimumAgeVerifier`. The idea behind extra name is, that it "creates" a function you can use later. So describes the purpose, instead of just using `minimumAge`.

stenburger1:
Thank you!

PR #4552, opened 2025-02-11
fix(ePlatform): adjust ul list style and wrapper style (`max-width` and `padding`)

### LINKED ISSUE #4524: E-platform styling issues
## 🐛 Bug Report

(Not really a bug)

In `src/style/core/reset.scss` certain styling is set, seemingly to help with e-platform styling.  
Several users (including me) who are developing new apps have needed to override some of these to get their apps to look correct on e-platform.  

Now, I don't know why this was introduced, I am sure it was needed at the time, but it might be time to consider removing/deprecating some of this CSS. Or finding some other solution so that not every new e-platform app needs to override e-platform specific Eufemia styles.

Eufemia CSS:
```css
// reset ePlatform css
@mixin resetLegacyStyles() {
...
    body#dnbLayoutDefault #wrapper,
    body#dnbLayoutDefaultStartPage #wrapper {
      width: 100%;
      max-width: 60rem;
    }

    ul {
      list-style: initial;
    }
}
```

Needed overrides:
```css
// E-platform styles
// NB: These styles fix some layout issues when deploying to e-platform
div#wrapper {
  max-width: none !important;
  padding: 0px;
}

ul.quickMenuList,
ul[role='presentation'],
ul.ui-corner-all {
  list-style-type: none !important;
}

// - End of e-platform styles

```

## Eufemia Version
10.63.0

tujoworker:
Fixes #4524

### DIFF
--- a/packages/dnb-eufemia/src/style/core/reset.scss
+++ b/packages/dnb-eufemia/src/style/core/reset.scss
@@ -483,11 +483,6 @@
 @mixin resetLegacyStyles() {
   html[xmlns="http://www.w3.org/1999/xhtml"]
   {
-    // reset underline during hover
-    a:hover {
-      text-decoration: none;
-    }
-
     .dnb-anchor--active {
       color: var(--color-mint-green) !important;
     }
@@ -501,17 +496,16 @@
     }
 
     // reset lists
-    ul {
+    .dnb-ul {
       list-style: initial;
     }
 
     // reset paragraphs
-    p {
+    .dnb-p {
       overflow: initial;
       text-align: inherit;
       background-color: initial;
     }
-
     i {
       font-style: italic;
     }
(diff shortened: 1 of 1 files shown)


packages/dnb-eufemia/src/style/core/reset.scss
```diff
@@ -504,6 +504,11 @@
     ul {
       list-style: initial;
     }
+    ul.quickMenuList,
```
andlbrei, line 507: I've found that `FileUpload` was affected by this as well, and we had to add
```
ul.dnb-upload__file-list,
```

tujoworker: I think we can rather take care of the Eufemia classes, instead going globally in. So instead of using `ul` we can target `.dnb-ul`. So then we don't need to find all other possible lists to reverse it again.

langz (approved):
Let's see if this improves it🤞

PR #4556, opened 2025-02-11
feat(Forms): add `reverse` to Iterate.Array

tujoworker:


langz (approved):
Looks fine to me :) 
When adding `reverse` do we have to make any changes to the animation, which I assume is made for adding items at the bottom?  Or maybe no change is needed 🙏

tujoworker:
> do we have to make any changes to the animation

No, the array animation does not know whats going on inside.

PR #4566, opened 2025-02-12
chore: postcode -> postal code

langz:
Motivation is just to continue using postal code instead of postcode.
We already have most occurrences of postal code, hence continuing that trend...

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/feature-fields/PostalCodeAndCity.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/feature-fields/PostalCodeAndCity.mdx
@@ -1,6 +1,6 @@
 ---
 title: 'PostalCodeAndCity'
-description: '`Field.PostalCodeAndCity` is a wrapper component for input of two separate values with user experience tailored for postcode and city values.'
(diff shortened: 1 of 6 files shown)


joakbjerk:
Isn't postcode the correct British spelling? 🤔

[slack](https://dnb-it.slack.com/archives/C05VCNNHG9K/p1705924373470199)

[Wikipedia](https://en.wikipedia.org/wiki/Postcodes_in_the_United_Kingdom)

langz:
> Isn't postcode the correct British spelling? 🤔
> 
> [slack](https://dnb-it.slack.com/archives/C05VCNNHG9K/p1705924373470199)
> 
> [Wikipedia](https://en.wikipedia.org/wiki/Postcodes_in_the_United_Kingdom)

Could be, when first implementing this I think the component was called PostalCodeAndCity because of:
- Postal codes was used in [Wikipedia](https://en.wikipedia.org/wiki/Postal_codes_in_Norway)
- Postal codes was used in [Posten](https://adressesok.posten.no/en/postal_codes/search)

But I'm fine with whatever, as long as we use the same-ish. But to me both terms mean the same and is understandable.
In translations for the user, we should probably use the same, and not mix postcode and postal code.

tujoworker:
In US English, it's postal code.
In British English, it's postcode.

We use US English in code and documentation and British English in the translation strings.

tujoworker (approved):
Nice 👏

PR #4567, opened 2025-02-13
fix(Drawer): ensure content on very small screen sizes doesn't get cut off

tujoworker:
More [info](https://dnb-it.slack.com/archives/CMXABCHEY/p1739367410251329).


langz (approved):
🤗

packages/dnb-design-system-portal/src/docs/uilib/components/drawer/properties.mdx
```diff
@@ -22,6 +22,6 @@ The Drawer is responsive with the following properties:
 
 ```css
 --drawer-width: 40vw;
---drawer-min-width: 25rem;
+--drawer-min-width: 384px;
 --drawer-max-width: 40rem;
```
langz, line 26: Should we change this to px as well, or does it make sense that one is en and the other px?

tujoworker: Yes, it makes sense when the rem font-size gets larger, we don't want to "cut" (min-width) when the content grows, but we want to expand the max to give room for the content to grow.

PR #4568, opened 2025-02-13
docs: removes duplicated docs for help prop

langz:
`help` property is already listed under General properties, no need to list it twice? 🤔


joakbjerk (approved):
Nice! No that seems like overkill 😁

PR #4582, opened 2025-02-17
fix(TextArea): ResizeObserver loop completed with undelivered notifications

langz:
Motivation https://dnb-it.slack.com/archives/CMXABCHEY/p1739800350978059


langz:
I've tested it a few times now, and it does not destroy anything at least.
So let us hope this works 🤞

PR #4601, opened 2025-02-20
feat(Toggle): add `size` property

langz:


packages/dnb-eufemia/src/extensions/forms/Field/Toggle/ToggleDocs.tsx
```diff
@@ -21,6 +21,11 @@ export const ToggleProperties: PropertiesTableProps = {
     type: 'string',
     status: 'optional',
   },
+  size: {
+    doc: 'The sizes you can choose is small (1.5rem), default (2rem), medium (2.5rem) and large (3rem) are supported component sizes. Defaults to default / null. Also, if you define a number like size="2" then it will be forwarded as the input element attribute.',
```
joakbjerk, line 25: Does it `default` the actual value you have to give the property, or does it default to `some value`and `null` 🤔

joakbjerk (approved):

packages/dnb-eufemia/src/extensions/forms/Field/Toggle/ToggleDocs.tsx
langz: Not quite sure, we'll have to check, it's implemented the same in all Field components, basing off and existing example. 

I do believe it defaults to either `"default"` or `null` but that's decided in the actual eufemia component, for instance the Toggle has different variants and use different Eufemia components, like Toggle and Checkbox, and the default size value there is different.

It's optional, so you do not have to or need to set it.

PR #4606, opened 2025-02-20
feat(Wizard.Step): add inactive prop to handle a step as inactive

snorrekim:


tujoworker:
We already have "active" and "activeWhen". I think having "inactive" in addition will create confusion about what should be used when.

tujoworker (approved):
Nice work 👏

snorrekim:
For posterity: we are renaming the `active` and `activeWhen` props in this PR: https://github.com/dnbexperience/eufemia/pull/4612

PR #4626, opened 2025-02-26
fix(Forms): always run validation on Wizard step changes

tujoworker:


langz (approved):
Looking good 👍 
I can't think of a realistic scenario where this would be "breaking", but I guess we'll find out soon enough 😅

PR #4631, opened 2025-02-26
feat(Forms): add `validationMode` property with `bypassOnNavigation` to Wizard.Container

tujoworker:


langz (approved):
I think looks good :)
Do we need the same in Form.Handler as well?
To be able to bypass validation of fields in a form?

Also, is this PR even necessary if the `Wizard.Container` would be nested in a `Form.Handler` which bypasses validation? Or would one need to do both (bypass validation in `Form.Handler` and in `Wizard.Container`)?

tujoworker:
> Do we need the same in Form.Handler as well?

Maybe, but then we do it on a feature request.

> Also, is this PR even necessary if the Wizard.Container would be nested in a Form.Handler which bypasses validation? Or would one need to do both (bypass validation in Form.Handler and in Wizard.Container)?

I think so. One would be for navigation, the other would be for submitting with errors.

langz:
> > Do we need the same in Form.Handler as well?
> 
> 
> 
> Maybe, but then we do it on a feature request.
> 
> 
> 
> > Also, is this PR even necessary if the Wizard.Container would be nested in a Form.Handler which bypasses validation? Or would one need to do both (bypass validation in Form.Handler and in Wizard.Container)?
> 
> 
> 
> I think so. One would be for navigation, the other would be for submitting with errors.

Maybe we should check with @carlodiaz, whom I think requested this feature in Wizard. I can envision a scenario where editors would need/want to navigate to something after a form(Form.Handler) like success message or other screens, without wanting to fill in a complete valid form.

PR #4625, opened 2025-02-25
feat(Forms): return existing errors in `onSubmitRequest` before Form.Handler submit or Wizard step change

tujoworker:
> You can use the `onSubmitRequest` prop on the [Form.Handler](/uilib/extensions/forms/Form/Handler/) to get visible errors before the form is submitted.

```tsx
const onSubmitRequest: OnSubmitRequest = ({ errors }) => {
  errors.forEach(({ path, value, label, error }) => {
    // Do something with the error
    console.log(label, error.message)
  })
}
```

PR #4633, opened 2025-02-27
chore: return `getErrors` function in `onSubmitRequest` instead of errors

tujoworker:
This is an addition to #4625.
I don’t think we should run the code to generate the errors with all the info unless it’s really necessary. That’s why I suggest using a function. Not that it would have a big performance impact, but still.


langz (approved):
🤝

PR #4634, opened 2025-02-27
fix(Forms): hide `info`, `warning` or `error` prop when `undefined` is given to a Field

### LINKED ISSUE #4611: Messages for a field sticks around when changing to undefined
## 🐛 Bug Report

If you render a field with a message (like `info`, `warning` or `error`) and then change the message to `undefined`, the previous message is still displayed.

## To Reproduce

1. Open [the codesandbox](https://codesandbox.io/p/devbox/ecstatic-fire-q7dnqw)
2. Observe that the info message is shown under each field
3. Click "Toggle messages" to stop passing the message
4. Observe that two of the fields are still showing the message

Bonus:
1. Click "Next message" while messages are hidden
2. Click "Toggle messages" to show messages again

## Expected behavior
The message should be hidden for all the inputs when it is not passed to the component any more.

## Eufemia Version

**Browser JS:** 10.64.1
**Browser CSS:** 10.64.1

tujoworker:
Fixes #4611

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/hooks/useFieldProps.ts
+++ b/packages/dnb-eufemia/src/extensions/forms/hooks/useFieldProps.ts
@@ -125,7 +125,7 @@ export default function useFieldProps<Value, EmptyValue, Props>(
     disabled: disabledProp,
     info: infoProp,
     warning: warningProp,
-    error: errorProp,
+    error: initialErrorProp = 'initial',
     errorMessages,
     onFocus,
     onBlur,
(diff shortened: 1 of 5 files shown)


langz (approved):
Looks great 👏 

I do find the it statements, `it('a field (with useFieldProps) ...`, a bit difficult to read.
I like the format` it('should ...`, but might be that it's not easy to refactor/rephrase them to that format.

packages/dnb-eufemia/src/extensions/forms/hooks/useFieldProps.ts
```diff
@@ -713,13 +717,16 @@ export default function useFieldProps<Value, EmptyValue, Props>(
...
   }
 
-  const bufferedError = revealErrorRef.current
-    ? prepareError(error) ??
-      localErrorRef.current ??
-      contextErrorRef.current
-    : error === null
-    ? null
-    : undefined
+  const bufferedError =
```
langz, line 720: Maybe this could be more readable if we would refactor it as being a function or so, with if sentences or so?

langz: Thanks, https://github.com/dnbexperience/eufemia/pull/4634/commits/b63a0a74755d6344e241ce4538bade4e118163fa, looks great :)

PR #4635, opened 2025-02-27
docs(Name): remove Field.SummaryList from example

langz:
I don't think this docs/info is correct, as we don't have a field named `Field.SummaryList`, and it doesn't make sense to use `Value.SummaryList` in the example either, so changing to `Form.Handler` instead.


tujoworker (approved):
👏

PR #4637, opened 2025-02-27
feat(DatePicker): add invalid dates to return object

joakbjerk:
Adding invalid dates, typed in by user, to the return object, to make it possible to validate if the given date is invalid. As of today, the date is set to null of the date is invalid, making it impossible to tell if the date is null because the input field is empty, or if the date is invalid

### DIFF
--- a/packages/dnb-eufemia/src/components/date-picker/DatePickerInput.tsx
+++ b/packages/dnb-eufemia/src/components/date-picker/DatePickerInput.tsx
@@ -348,6 +348,8 @@ function DatePickerInput(externalProps: DatePickerInputProps) {
           hoverDate: null,
         },
         (dates) => {
+          // Should fire if user has filled out an invalid date,
+          // or of the date was valid, but the user has pressed backsapce or removed it
           if (isDateFullyFilledOutRef.current || hasHadValidDate) {
             const { startDate, endDate, event } = {
               ...state,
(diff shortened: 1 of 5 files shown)


packages/dnb-eufemia/src/components/date-picker/DatePickerInput.tsx
```diff
@@ -141,11 +147,18 @@ function DatePickerInput(externalProps: DatePickerInputProps) {
   } = props
 
   const [focusState, setFocusState] = useState<string>('virgin')
+  // TODO: Turn into a ref, as these values should not trigger a rerender
```
langz, line 150: TODO, as something you aim to do in this PR, or in something in the near future? 🙏

packages/dnb-eufemia/src/components/date-picker/DatePickerInput.tsx
```diff
@@ -335,12 +348,19 @@ function DatePickerInput(externalProps: DatePickerInputProps) {
           hoverDate: null,
         },
         (dates) => {
-          if (hasHadValidDate) {
+          // Should fire if user has filled out an invalid date,
+          // or of the date was valid, but the user has pressed backsapce or removed it
```
langz, line 352:
```suggestion
          // or if the date was valid. Like if the user has pressed backspace or removed the valid date.
```

langz (approved):

packages/dnb-eufemia/src/components/date-picker/hooks/useDates.ts
```diff
@@ -39,6 +39,7 @@ export type DatePickerDates = {
   maxDate?: Date
   startMonth?: Date
   endMonth?: Date
+  // TODO: Move partial dates to own types, as they are not really only used on `getReturnObject`, and not in this hook
```
langz, line 42: TODO, as something you aim to do in this PR, or in something in the near future? 🙏

packages/dnb-eufemia/src/components/date-picker/DatePickerInput.tsx
joakbjerk: Near future 🫡

PR #4645, opened 2025-03-03
chore(StepIndicator): add `triggerButtonProps` for internal Wizard use to show a state on the trigger button when in small view

tujoworker:
We need this prop in order to be able to show an error status on the Wizard when the screen has a small width. Its used internally only. And may even be removed in v3 of the StepIndicator.

### DIFF
--- a/packages/dnb-eufemia/src/components/step-indicator/StepIndicatorTriggerButton.tsx
+++ b/packages/dnb-eufemia/src/components/step-indicator/StepIndicatorTriggerButton.tsx
@@ -41,6 +41,7 @@ function StepIndicatorTriggerButton(
 
   const {
     data, // eslint-disable-line
+    triggerButtonProps, // eslint-disable-line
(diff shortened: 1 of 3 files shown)


packages/dnb-eufemia/src/components/step-indicator/StepIndicatorTriggerButton.tsx
```diff
@@ -55,6 +56,7 @@ function StepIndicatorTriggerButton(
   } as React.HTMLProps<HTMLElement>
 
   const buttonParams = {
+    ...triggerButtonProps,
```
langz, line 59: Shouldn't this complain about that the potential property `sidebar_id` in triggerButtonProps(`StepIndicatorTriggerButtonProps`) can't be used in `Button` component?

Might be that I misunderstand this though🙏

langz (approved):
🟢

PR #4646, opened 2025-03-04
chore(Field.Slider): incorrect use of label `for`

langz:
Hopefully fixes the following error, seen in the [example](https://eufemia.dnb.no/uilib/extensions/forms/feature-fields/more-fields/Slider/demos/): 

```
The label's for attribute doesn't match any element id. This might prevent the browser from correctly autofilling the form and accessibility tools from working correctly.

To fix this issue, make sure the label's for attribute references the correct id of a form field.
```

Fix can be tested/validated here: https://eufemia-git-chore-incorrect-use-of-label-eufemia.vercel.app/uilib/extensions/forms/feature-fields/more-fields/Slider/


langz:
Built upon and fixed in https://github.com/dnbexperience/eufemia/pull/4650

PR #4648, opened 2025-03-04
chore(DatePicker): refactor partial dates

joakbjerk:
- [x] Update docs with correct partial string structure/ value
- [x] Move partial dates to own hook for reduced complexity
- [x] Make partial values null when valid date is filled out

### DIFF
--- a/packages/dnb-eufemia/src/components/date-picker/hooks/usePartialDates.ts
+++ b/packages/dnb-eufemia/src/components/date-picker/hooks/usePartialDates.ts
@@ -1,4 +1,4 @@
-import { useRef } from 'react'
+import { useCallback, useRef } from 'react'
 
 export type PartialDates = {
   partialDate?: string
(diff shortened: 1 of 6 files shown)


packages/dnb-eufemia/src/components/date-picker/hooks/usePartialDates.ts
```diff
@@ -0,0 +1,39 @@
...
+  partialEndDate?: string
+}
+
+export default function usePartialDates() {
+  const partialDatesRef = useRef<PartialDates>({
+    partialStartDate: null,
+    partialEndDate: null,
+  })
+
+  function setPartialDates(partialDates: PartialDates) {
```
tujoworker, line 15:
```suggestion
  const setPartialDates = useCallback((partialDates: PartialDates) => {
  etc.
```

Using a function without useCallback will create a new instance and can create issues when used in other hooks later.

tujoworker (approved):
The code is much cleaner now 🧹

PR #4655, opened 2025-03-04
fix(Field.Upload): set `aria-required` and `required` props in input rather than div

langz:


packages/dnb-eufemia/src/components/upload/Upload.tsx
```diff
@@ -144,11 +144,10 @@ const Upload = (localProps: UploadAllProps) => {
       <Provider skeleton={skeleton}>
         <UploadDropzone
           className={classnames('dnb-upload', spacingClasses, className)}
-          {...props}
         >
           <UploadInfo />
 
-          <UploadFileInput />
+          <UploadFileInput {...props} />
```
langz, line 150: This is potentially breaking(as rest props previously was sent to UploadDropzone), but also fixing issues with Field.Upload which would pass required and aria-required to UploadDropzone which would result in a `div`(without `role`) with `required` and `aria-required`.

tujoworker: yeah, this PR looks like a fix for something (not a simple chore). What does it solve?

langz: When I added the aria test to a required Field.Upload, the test would fail saying it's not allowed to have aria-required in a div (without role), and I think it would be better to pass both required and aria-required props to the `input` instead of the `div`.

Screenshot 2025-03-05 at 12 41 31

tujoworker: Thanks for the good details 🙏

tujoworker (approved):
Nice catch 🪰

PR #4667, opened 2025-03-05
fix(Value.SummaryList): spacing of `help` when wrapped in `Form.InfoOverlay`

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1741186089782099?thread_ts=1740042283.488439&cid=CMXABCHEY
CSB of the issue: https://codesandbox.io/p/devbox/value-padding-forked-jsz5nm?workspaceId=ws_LbiDwJdEBdbwxH2AkyAyVE

Issue is happening because the [existing styling](https://github.com/dnbexperience/eufemia/pull/4667/files#diff-281ca53e6a78ae24056e8fc94bf0274576aef89044ddcc613fb09b0844023e13R50) does not work when it shows up a `div` like so:
`<div class="dnb-space dnb-height-animation dnb-forms-visibility dnb-height-animation--is-visible dnb-height-animation--is-in-dom dnb-height-animation--parallax" aria-hidden="false">VALUE CONTENT HERE</div>`

Screenshot from [CSB](https://codesandbox.io/p/devbox/value-padding-forked-jsz5nm?workspaceId=ws_LbiDwJdEBdbwxH2AkyAyVE):
Screenshot 2025-03-05 at 19 04 20

I've added a "quickfix" for this now, but doesn't feel 100% correct, as I'm just adding a specific selector to apply the existing css with margin when there's a `div` there as well.

I'm not sure why the `div` is present when wrapped in a `Form.InfoOverlay`, the fix could perhaps be related to not applying that `div`(if possible).


tujoworker:
ohh, it looks like we need to fix this (in another PR). A div inside a dt is not valid HTML. 

```
<dl>
  <dt>
    <div>Invalid Term</div> <!-- ❌ Not valid -->
  </dt>
  <dd>Definition</dd>
</dl>
```

I will commit a selector change to make it more save anyways.

tujoworker (approved):

PR #4695, opened 2025-03-07
chore: bump @navikt/fnrvalidator to 2.1.5

langz:


tujoworker:
Looking at the [change log](https://github.com/navikt/fnrvalidator/releases), some essential parts have changes since last version. So I think we need to make it clear in the change log, therefor the "feat" decoration in the PR title.

Here is an example:

Screenshot 2025-03-07 at 10 30 31

Happily, we have a good amount of tests ourselves.

tujoworker (approved):

PR #4697, opened 2025-03-07
docs(DatePicker): correct `range` description

joakbjerk:


langz (approved):
When looking at the change up against the commit message, I don't fully understand the change.

The change changes the `DatePicker` docs for `range`, potentially also change the `Field.Date` `range` docs.

The commit message says that `Field.Date ``range` description/docs is replaced with `DatePicker` description/docs. If so, shouldn't it somehow be a change in the `Field.Date` docs as part of this PR? 🤔

joakbjerk:
> When looking at the actual code change up against the commit message, I don't fully understand the change.
> 
> The change changes the `DatePicker` docs for `range`, potentially also change the `Field.Date` `range` docs if it's already used there.
> 
> The commit message says that `` Field.Date ``range `` description/docs is replaced with `DatePicker` description/docs. If so, shouldn't it somehow be a change in the `Field.Date` docs as part of this PR? 🤔

Need to find a new title then 😝 The issue was that the `DatePicker` docs was using the `Field.Date` docs description for `range`

langz:
> > When looking at the actual code change up against the commit message, I don't fully understand the change.
> > The change changes the `DatePicker` docs for `range`, potentially also change the `Field.Date` `range` docs if it's already used there.
> > The commit message says that `Field.Date`range `` description/docs is replaced with `DatePicker` description/docs. If so, shouldn't it somehow be a change in the `Field.Date` docs as part of this PR? 🤔
> 
> Need to find a new title then 😝 The issue was that the `DatePicker` docs was using the `Field.Date` docs description for `range`

That can be right? _That the `DatePicker` docs was using the `Field.Date` docs description for `range`_? Most be opposite way, or?

PR #4698, opened 2025-03-07
fix(DatePicker): using custom locale `sv-SE` breaks calendar

langz:
This PR fixes that the application breaks when providing a locale of `sv-SE`.
Reason this was failing is because in the code for formatting a date, we don't import and support swedish(`sv-SE`) for date-fns format. We only support `nb-NO`, `en-GB`, `en-US`.

This PR makes it so that when providing a non valid locale for formatting date, it defaults to use norwegian.

There's a different PR https://github.com/dnbexperience/eufemia/pull/4706 for supporting adding any locale for formatting date using date-fns.

Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1741273915008559
CSB: https://codesandbox.io/p/devbox/lucid-swartz-tj9v2c?workspaceId=ws_LbiDwJdEBdbwxH2AkyAyVE

Opening DatePicker's calendar when locale is sv-SE results in:
```
RangeError
locale must contain localize property
format
https://tj9v2c.csb.app/node_modules/date-fns/format/index.js:382:11
```


joakbjerk (approved):
Looks good! 🔥

PR #4699, opened 2025-03-07
chore: bump classnames to v2.5.1

langz:


joakbjerk (approved):

tujoworker (approved):
There is at least one feature that makes it interesting to upgrade 👍

> Enhanced performance by using string concatenation.

PR #4704, opened 2025-03-07
feat: use @dnb/browserslist-config for when creating the bundle

tujoworker:
My research so far is:

There is a difference in the output. But its minimal. Basically, every file (over 400) have some more imports such as this:

```js
import "core-js/modules/web.dom-collections.iterator.js";
```

Also, we may add a list of the supported browsers in a human readable table/list. Here is a PR that would make that possible: https://github.com/dnbexperience/browserslist-config

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/usage.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/usage.mdx
@@ -36,8 +36,8 @@ Read more in the [First Steps](/uilib/usage/first-steps/) section.
 
 ## Supported Browsers and Platforms
 
-Eufemia uses the following config for the bundle output, defined in `.browserslistrc`:
+Eufemia uses the [@dnb/browserslist-config](https://github.com/dnbexperience/browserslist-config) configuration for the bundle output, defined in `.browserslistrc`:
 
-- `defaults and supports es6-module`
+- `extends @dnb/browserslist-config`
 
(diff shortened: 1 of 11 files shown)


langz (approved):
🫡🥳

tujoworker:
Here's how it looks locally:

Screenshot 2025-03-09 at 17 06 22

and:

Screenshot 2025-03-09 at 17 06 28

packages/dnb-design-system-portal/src/docs/uilib/usage.mdx
```diff
@@ -36,8 +40,16 @@ Read more in the [First Steps](/uilib/usage/first-steps/) section.
...
 
-- `defaults and supports es6-module`
+### Configuration and Browserslist
 
-To see exactly which browsers this config supports, take a look [here](https://browsersl.ist/#q=defaults+and+supports+es6-module).
+Eufemia uses the [@dnb/browserslist-config](https://github.com/dnbexperience/browserslist-config) configuration for the bundle output, defined in `.browserslistrc`:
+
+- `extends @dnb/browserslist-config`
+
+It only effects the JavaScript bundle output, not the CSS bundle output.
```
langz, line 51:
```suggestion
It only affects the JavaScript bundle output, not the CSS bundle output.
```

Not important, I was just unsure if it's effects or affects, I think it's affects.

langz (approved):

PR #4705, opened 2025-03-10
fix(Forms): render HelpButton after label and before labelDescription

### LINKED ISSUE #4693: Field subtitle/sub-label/description on new line
## 🚀 Feature Proposal

`Field` has `labelDescription` for adding some extra information "between" the label and the form inputs.  
Sometimes I think it would be nice to get the label description on a new line.  
Also, when using the `help` prop, I think the button for showing the help text is better suited to follow the label, rather than the label description, though perhaps one could argue that sometimes the description should have a help text.

## Motivation
We would like to give the user information on the requirements for the field they are using.
I think readability would be better if the description was on a new line below the label, and the label help icon was following the label text, and not the description.
Image

## Pitch

I think it would be a tiny improvement to the `Field` component

tujoworker:
Fixes #4693

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/FieldBlock/FieldBlockDocs.ts
+++ b/packages/dnb-eufemia/src/extensions/forms/FieldBlock/FieldBlockDocs.ts
@@ -11,6 +11,11 @@ export const FieldBlockSharedProperties: PropertiesTableProps = {
     type: 'string',
     status: 'optional',
   },
+  labelDescriptionNextLine: {
+    doc: 'If true, the labelDescription will be displayed on the next line.',
+    type: 'boolean',
+    status: 'optional',
+  },
   labelSrOnly: {
     doc: 'Use `true` to make the label only readable by screen readers.',
     type: 'boolean',
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/create-component/FieldBlock/Examples.tsx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/create-component/FieldBlock/Examples.tsx
@@ -328,7 +328,8 @@ export const InlineHelpButtonVerticalLabelDescription = () => {
       <Form.Card>
         <Field.String
           label="Ønsket lånebeløp"
-          labelDescription={'\\nDescription'}
+          labelDescription="Description"
+          labelDescriptionNextLine
           help={{
             title: 'Hva betyr lånebeløp?',
             content: 'Dette er hvor mye du har tenkt å låne totalt.',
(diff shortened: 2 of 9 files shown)


langz:
@tujoworker, is this something you aim to include in release of v10.67.1 #4715 ?

packages/dnb-eufemia/src/extensions/forms/FieldBlock/FieldBlockDocs.ts
```diff
@@ -11,6 +11,11 @@ export const FieldBlockSharedProperties: PropertiesTableProps = {
     type: 'string',
     status: 'optional',
   },
+  labelDescriptionNextLine: {
```
langz, line 14: Seen in the light of the commit message/title _fix(Forms): render HelpButton after label and before labelDescription_, I do think the property name `labelDescriptionNextLine` is a bit strange? 🤔 
Shouldn't the property name be something about positioning the help button, or perhaps something in the help property like `help={{ title: 'Help title', content: 'Help content', position: 'something' }}`?

tujoworker: We have decided to call it for `labelDescriptionSameLine` – it will only effect the labelDescription, regardless if you have a help button or not.

packages/dnb-eufemia/src/extensions/forms/FieldBlock/FieldBlockDocs.ts
```diff
@@ -11,6 +11,11 @@ export const FieldBlockSharedProperties: PropertiesTableProps = {
     type: 'string',
     status: 'optional',
   },
+  labelDescriptionSameLine: {
+    doc: 'If true, the labelDescription will be displayed on the next line.',
```
langz, line 15:
```suggestion
    doc: 'If true, the labelDescription will be displayed on the same line as the label.',
```

langz (approved):
Looks great 💯 

Often when we introduce new properties, we mention them in the commit message/title, and it's probably not always necessary. But based on the commit message/title: `feat(Forms): render HelpButton after label and before labelDescription`, there's not easy to understand that there's a new available prop, and what it does.


Further, is the property name `labelDescriptionNextLine`descriptive enough? 🤔 
When I initially read it, without looking at any code examples, etc. I was thinking label description same line as what? But it's probably given that it's the same line as the label.

tujoworker:
What about calling that property `labelDescriptionInline`?

langz (approved):

packages/dnb-eufemia/src/extensions/forms/FieldBlock/FieldBlockDocs.ts
langz: I unresolved this again 🙏

PR #4706, opened 2025-03-10
feat(DatePicker): add support for swedish translations

joakbjerk:
- [x] Add Swedish `date-fns` locale 
- [x] Document supported `date-fns` locales
- [x] Add tests
- [x] Update story

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/components/date-picker/info.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/components/date-picker/info.mdx
@@ -17,3 +17,7 @@ The DatePicker component should be used whenever there is to enter a single date
 The DatePicker operates with a default JavaScript Date instance as well as string (ISO 8601) like `start_date="2019-05-05"` (yyyy-MM-dd).
 
 For more _DatePicker_ examples, have a look at [this CodeSandbox](https://codesandbox.io/embed/q88v3r8w6).
+
+### Translations and date formatting
+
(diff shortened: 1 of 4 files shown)


langz:
Good good, great start 🚀 

I think we need:
- docs on how to use this, with good examples with imports from date-fns, etc. Let's use swedish as an example, just to help out the developer who have requested this feature.
- docs on which date-fns version we use, meaning they(consumers) will most likely have to use the same version right?
- Maybe also an actual example, which will use swedish? Is that possible? Perhaps make it with `range` or so and for a month that we know is spelled different than norwegian?

langz:
And we need to rebase with main branch, looks like https://github.com/dnbexperience/eufemia/pull/4698 caused a few conflicts 🙏

packages/dnb-design-system-portal/src/docs/uilib/components/date-picker/Examples.tsx
```diff
@@ -181,14 +182,18 @@ export const DatePickerStatusMessage = () => (
...
-  <ComponentBox>
-    <DatePicker
-      label="DatePicker"
-      date={new Date()}
-      showInput
-      suffix={<HelpButton title="Modal Title">Modal content</HelpButton>}
-    />
-  </ComponentBox>
+  // To prevent `context.getTranslation(...).HelpButton is undefined` error on demo page,
+  // caused by setting the provider locale to be `sv-SE` in DatePickerWithCustomDateFormatting`
```
langz, line 186:
```suggestion
  // caused by setting the provider locale to be `sv-SE` in DatePickerWithCustomDateFormatting
```

langz:
Looking great 👏 
I'll take a demo/test after the deploy preview is ready 🤌

langz (approved):
❤️🎸

tujoworker (approved):

PR #4707, opened 2025-03-10
chore(Provider): add extensive locale test for `sv-SE`

langz:
Not sure if this extensive test is redundant or does provide value.
Anyhow, I want to see it go green once 😅

### DIFF
--- a/packages/dnb-eufemia/src/shared/__tests__/Provider.test.tsx
+++ b/packages/dnb-eufemia/src/shared/__tests__/Provider.test.tsx
@@ -561,7 +561,6 @@ describe('Provider', () => {
                   month: 'm',
                   year: 'å',
                 },
-                // TODO: add firstDay
               },
               Anchor: {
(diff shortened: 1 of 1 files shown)


packages/dnb-eufemia/src/shared/__tests__/Provider.test.tsx
```diff
@@ -480,5 +512,292 @@ describe('Provider', () => {
...
+                <Td align="right">
+                  Row 3 with <b>medium text</b>
+                </Td>
+              </Tr>
+            </tbody>
+          </Table>
+          <Upload acceptedFileTypes={['jpg', 'png']} />
+        </Provider>
+      )
+      expect(true).toBe(true)
```
langz, line 800: I'll have to find some other type of expect statement than this :)

langz:
I don't think "we need this", but it could be helpful, as it tests many of our components.

PR #4713, opened 2025-03-10
fix(Input): remove value attribute (⚠️ Use element.value instead in e.g. tests)

tujoworker:
Removing the `value` attribute from the HTML output, as best practice nowdays is to avoid updating it dynamically.  
Getting the `value` via the attribute was **never documented**, so this should be safe to remove.


langz (approved):
Looks good to me :)
Bye bye `value`attribute 👋 

You can decide if you want to include this in v10.67.1 https://github.com/dnbexperience/eufemia/pull/4715

PR #4726, opened 2025-03-11
fix(Forms): ensure `minDate`/`maxDate` does not interrupt first form submit

tujoworker:


tujoworker:
> Is it safe to remove this?
> One would think the code removed had some functionality?

Yes, talked with @joakbjerk about it. It was to show errors initially, when provided from outside. But we do not do this in all other fields. Its the onBlur event or the form submit event that will show errors. So we needed to add `validateInitially` to the tests that then failed.

PR #4727, opened 2025-03-11
fix(Value.SummaryList): styling `Value.*` components with help button

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1741697990566819?thread_ts=1740042283.488439&cid=CMXABCHEY
CSB of the problem: https://codesandbox.io/p/devbox/value-padding-summary-list-forked-5cxffw?workspaceId=ws_LbiDwJdEBdbwxH2AkyAyVE

Thoughts and description based on the newly added screenshot test:

valuesummarylist-have-to-match-with-help-help-button snap

The reporter also has some uncertainties regarding: _The help text has some padding and rounded corners when inside a SummaryList . Might be intentional, might not be? 🤔_ 

- [x]  Screenshot tests and example
- [x] The actual fix
- [x] Update screenshot tests to reflect fix


packages/dnb-eufemia/src/extensions/forms/Value/SummaryList/__tests__/SummaryList.screenshot.test.ts
```diff
@@ -36,6 +36,14 @@ describe('Value.SummaryList', () => {
     expect(screenshot).toMatchImageSnapshot()
   })
 
+  it('have to match with help help button', async () => {
```
langz, line 39:
```suggestion
  it('have to match with help button', async () => {
```

tujoworker (approved):
There are several significant changes. I would like to add more integrations tests when I get time. But I would like of you take a look over relevant examples.

tujoworker:
Added the possibility to have own widths for values. We may adjust them later.

packages/dnb-eufemia/src/extensions/forms/ValueBlock/style/dnb-value-block.scss
```diff
@@ -74,27 +82,39 @@
...
+        margin-right: 0.3em;
+      }
+    }
   }
 
   // Composition block
   &__composition--horizontal &__content {
     display: flex;
     flex-flow: row;
+    flex-wrap: wrap;
```
tujoworker, line 117: Makes it possible for --next-line to wrap on a new line. `flex-basis: 100%;` ensures that.

tujoworker (approved):

packages/dnb-eufemia/src/extensions/forms/FieldBlock/FieldBlockDocs.ts
```diff
@@ -78,7 +78,7 @@ export const FieldBlockProperties: PropertiesTableProps = {
   },
   composition: {
     doc: 'Use `true` for when you have more than one field wrapped.',
-    type: 'string',
+    type: 'boolean',
```
langz, line 81:
```suggestion
    type: 'true',
```
Not sure if it's correct to say type  `type: 'true'` or not, but TS complains if I do:
`<FieldBlock composition={false}>`:
`Type 'false' is not assignable to type 'true'.ts(2322)`

langz (commented):
Wow, great work @tujoworker, thanks for taking on such a complex task ⭐ 

The only thing I see, which I'm unsure of in the screenshot tests is the placement of the help button for Value's without labels, like Value streetNr in the street composition:
```
<Value.Composition
  label="Street"
  help={{
    open: true,
    title: 'Help title',
    content: 'Help content',
  }}
>
  <Value.String path="/streetName" />
  <Value.Number
    path="/streetNr"
    help={{
      open: true,
      title: 'Help title',
      content: 'Help content',
    }}
  />
</Value.Composition>
```

Should the help button be placed before(as it is) or after the value, when there's no label provided.

Screenshots, the value to look for is 12 that's the Value streetNr:
https://github.com/dnbexperience/eufemia/pull/4727/files#diff-005ce2d5c2762e2e3ab3ee800950078094ecb34729ccd37a84a7c857c4a37368

image

packages/dnb-eufemia/src/extensions/forms/Value/Composition/CompositionDocs.ts
```diff
@@ -1,10 +1,11 @@
 import { PropertiesTableProps } from '../../../../shared/types'
 import { ValueProperties } from '../ValueDocs'
 
-const { label, transformLabel } = ValueProperties
+const { label, transformLabel, help } = ValueProperties
 
 export const CompositionProperties: PropertiesTableProps = {
   label,
+  help,
```
langz, line 8: Nice catch👏

langz (commented):
In the given screenshot test, 
https://github.com/dnbexperience/eufemia/pull/4727/files#diff-f912aeea850e0147e5bd531dddae198be1e6096a18be371eccb78c7507bff642

image

Is it correct that we wrap all the labels so early? From a visual standpoint, it feels like they could use more of their available width before breaking into a new line 🤔 The help buttons takes the full width.
Or perhaps we/I need to add a property to make sure that the labels can use the full available width?

tujoworker:
> The help buttons takes the full width.

The content of the help button has also a max width, but it is larger than the field/value width of `large`. We can and probably should adjust it. But in another PR.

tujoworker:
> Should the help button be placed before(as it is) or after the value, when there's no label provided for a field in a Field.Composition?

Great question. I don't know. But I gave it a try with flex.

packages/dnb-eufemia/src/extensions/forms/ValueBlock/style/dnb-value-block.scss
```diff
@@ -50,11 +51,20 @@
...
+    // Make it possible to omit labels in a SummaryList with a horizontal/grid layout that has a label
+    margin-bottom: 0;
+  }
 
   &__help {
     display: flex;
     flex-flow: column;
 
+    .dnb-card & {
+      // Make it possible to that &__help--next-line wraps to a new line
```
langz, line 64:
```suggestion
      // Make it possible so that &__help--next-line wraps to a new line
```

PR #4728, opened 2025-03-11
docs(Value): add help button to example

langz:
Reason for adding is that it's nice to have a help button example incorporated in the [value docs](https://eufemia.dnb.no/uilib/extensions/forms/Value/) as well.


tujoworker (approved):
Nice 👍

PR #4731, opened 2025-03-12
fix(Wizard.Step): change text to step contains error

langz:


joakbjerk (approved):
🪜

tujoworker:
The only reason why I wrote and still think "form" is more descriptive, is that it can be mysterious for someone to read "step", because its not obvious for many people, that they are on a form "step".

PR #4733, opened 2025-03-12
fix(Field.BankAccountNumber): `0000 00 00000` should be invalid

langz:


joakbjerk (approved):

packages/dnb-eufemia/src/extensions/forms/Field/BankAccountNumber/BankAccountNumber.tsx
```diff
@@ -108,6 +108,9 @@ function BankAccountNumber(props: Props) {
 }
 
 function isValidAccountNumber(digits: string) {
+  if (digits === '00000000000') {
```
tujoworker, line 111:
```suggestion
  if (parseFloat(digits) === 0) {
```

Would be better in terms of supporting other countries.

tujoworker (approved):

PR #4737, opened 2025-03-13
fix(Forms): use `max-width` instead of `width` when defining `width`

tujoworker:


tujoworker:
In certain cases, especial with Field.Composition, we get unwanted results when using max-width. While there are ways to work around this, it looks like it needs a much larger rewrite.

Screenshot 2025-03-13 at 14 11 58

PR #4745, opened 2025-03-14
feat(Forms): adjust the "widths" of `Value.*` components

tujoworker:
More [info](https://dnb-it.slack.com/archives/C07JEJVSHJR/p1741942407932319).


packages/dnb-eufemia/src/extensions/forms/ValueBlock/style/dnb-value-block.scss
```diff
@@ -1,5 +1,5 @@
 @import '../../../../style/core/utilities.scss';
-@import '../../Field/style/value-sizes.scss';
+@import '../../Value/style/value-sizes.scss';
```
langz, line 2: 🫡

langz (approved):
Wow, thanks looking very very good. 

Should we add any docs to the Value docs about how it now looks, why it takes full width, etc, and what the reasoning behind it is?

Or is that not necessary for users to know?

PR #4760, opened 2025-03-16
fix(NumberFormat): render properly when `sv-SE` locale is given

### LINKED ISSUE #4753: sv-SE translation is not working for NumberFormat component
## 🐛 Bug Report

Change components language to Svenska.
Try to visit this page, it breaks.

https://eufemia.dnb.no/uilib/components/number-format/

tujoworker:
Fixes #4753

### DIFF
--- a/packages/dnb-eufemia/src/components/number-format/NumberUtils.js
+++ b/packages/dnb-eufemia/src/components/number-format/NumberUtils.js
@@ -309,7 +309,7 @@ export const format = (
     }
 
     if (value === 'invalid') {
-      aria = locales[locale].NumberFormat.not_available || 'N/A'
+      aria = locales[locale]?.NumberFormat.not_available || 'N/A'
     }
 
     // return "locale" as well value,l, since we have to "auto" option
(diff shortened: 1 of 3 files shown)


langz (approved):

packages/dnb-eufemia/src/components/number-format/NumberUtils.js
```diff
@@ -309,7 +309,7 @@ export const format = (
     }
 
     if (value === 'invalid') {
-      aria = locales[locale].NumberFormat.not_available || 'N/A'
+      aria = locales[locale]?.NumberFormat.not_available || 'N/A'
```
langz, line 312: Looks like a decent fix 👏 
But as far as I understand this, then this translation is not in use https://github.com/dnbexperience/eufemia/blob/8d9e2624e0aeccc74f5b70860a39f1e6d2b18e15/packages/dnb-eufemia/src/shared/locales/sv-SE.ts#L96

as `locales from '../../shared/locales'` will never have the `sv-SE` translations, but only for `nb-NO` and `en-GB`, which makes the translation https://github.com/dnbexperience/eufemia/blob/8d9e2624e0aeccc74f5b70860a39f1e6d2b18e15/packages/dnb-eufemia/src/shared/locales/sv-SE.ts#L96 dead code? As it will never be used? 🤔

PR #4763, opened 2025-03-17
feat(NumberForma): use component's translation for invalid aria text

langz:


packages/dnb-eufemia/src/components/number-format/__tests__/NumberUtils.test.ts
```diff
@@ -133,6 +133,24 @@ describe('Decimals format', () => {
     })
   })
 
+  it('should render custom invalid text when given', () => {
```
langz, line 136:
```suggestion
  it('should render custom invalid ARIA text when given', () => {
```

tujoworker (approved):
Nice work 👍

PR #4764, opened 2025-03-17
feat(Table): add view transition support to inserted table rows (accordion)

tujoworker:
Closes #3590

This PR:

- adds a `style` attribute to the accordion examples, so content does not change the column width, when they are rendered.
- adds `ViewTransition` to the accordion rows, so when many new tr's are rendered, they will "kind of" animate/expand (except for the last tr). This is a side-effect of adding a "viewTransitionName: `row-${id}`" to each tr. This only works in non-Firefox browsers.
- Why `ViewTransition`? To be short; animating an tr in "height" is at the moment not possible with CSS.
- Here is an [example](https://eufemia-git-feat-table-transition-view-eufemia.vercel.app/uilib/components/table/#expand-additional-rows).

https://github.com/user-attachments/assets/f1095a01-7568-4604-91d9-914c518ae703


langz (approved):
Nice nice, I don't know too much about the topic, but I think it looks good 💯 

I've tested  the [example](https://eufemia-git-feat-table-transition-view-eufemia.vercel.app/uilib/components/table/#expand-additional-rows) in the following browsers:

Safari: No animation, and feels a bit stuttering, 

Chrome: Looks great and behaves smooth.



Safari and Chrome on my iPhone:


https://github.com/user-attachments/assets/226f3fc0-d62b-4cfc-8d22-584c47ad6c52

langz:
Thanks for the improvements. I have the example one more test on my iPhone:
https://github.com/user-attachments/assets/59c1dd91-20c2-4437-b9af-3b223b2ce1d9

tujoworker:
Yeah, let's deal with that later 😄

langz (approved):

PR #4765, opened 2025-03-17
feat(InfoCard): add `stretch` property and don't stretch by default

tujoworker:
This PR changes:

- the card to only stretch to the max content width.
- to get the old behavior, use the `stretch` property.
- a card with centered content, will get margin with auto, in order to center itself. But not when the `stretch` property was given.


langz (approved):
I just tested the preview, on both desktop and mobile, and I think it looks good🥳

PR #4766, opened 2025-03-17
feat(DatePicker): replace locale based `date-fns` formatting with `Intl`

joakbjerk:
Relies on https://github.com/dnbexperience/eufemia/pull/4882

TODO:

- [x] refactor unused and undocumented `titleFormat` and `dayOfWeekFormat`props
- [x] remove `date-fns` locales in favor of `Provider` `locale`
- [x] replace `date-fns` format used in `formatSelectedDateTitle` inside `DatePicker` with `Intl`
- [x] Replace `quickFormat` with `formatDate`
- [x] Add tests for `selectedDateTitle` `aria-title`
- [x] Make swedish translations tests work


tujoworker (approved):
Nice 😍

tujoworker:
I think we can remove the docs (own commit) about the supported locales. Just because the situation is now the same like for all other components.

PR #4776, opened 2025-03-19
docs(DatePicker): add rationale behind why `correctInvalidDate` is not on by default

joakbjerk:


packages/dnb-design-system-portal/src/docs/uilib/components/date-picker/demos.mdx
```diff
@@ -84,6 +84,9 @@ import enUS from '@dnb/eufemia/shared/locales/en-US'
 
 ### Min and Max date correction
 
+The `correctInvalidDate` property is not on by default, as it can be confusing to screen reader users, when the value they type in changes, for seemingly no reasong to them.
```
langz, line 87:
```suggestion
The `correctInvalidDate` property is not on by default, as it can be confusing to screen reader users, when the value they type in changes, for seemingly no reasoning to them.
```

langz, line 87:
```suggestion
The `correctInvalidDate` property is not on by default, as it can be confusing to screen reader users, when the value they type in changes, for seemingly no reason to them.
```

langz (approved):

PR #4778, opened 2025-03-19
fix(DatePicker): correct invalid dates if  `correctInvalidDate` is `true` and `minDate` or `maxDate` is defined

joakbjerk:
### TODO

- [x] Move `invalidDates` to own hook
- [x] Add `invalidDates` to date correction function
- [x] Add test

`TODO` comments added to the code is meant for a later pr🙏

### DIFF
--- a/packages/dnb-eufemia/src/components/date-picker/DatePicker.tsx
+++ b/packages/dnb-eufemia/src/components/date-picker/DatePicker.tsx
@@ -118,6 +118,7 @@ export type DatePickerProps = {
   maxDate?: DateType
   /**
    * Corrects the input date value to be the same as either `minDate` or `maxDate`, when the user types in a date that is either before or after one of these. Defaults to `false`.
+   * Deprecate in v11 in favor of built in validation through Field.Date
    */
   correctInvalidDate?: boolean
   /**
--- a/packages/dnb-eufemia/src/components/date-picker/DatePickerInput.tsx
+++ b/packages/dnb-eufemia/src/components/date-picker/DatePickerInput.tsx
@@ -39,6 +39,7 @@ import { DatePickerEventAttributes } from './DatePicker'
 import { useTranslation } from '../../shared'
 import { DatePickerInputDates } from './hooks/useDates'
 import usePartialDates from './hooks/usePartialDates'
+import useInvalidDates from './hooks/useInvalidDates'
 
 export type DatePickerInputProps = Omit<
   React.HTMLProps<HTMLInputElement>,
@@ -104,12 +105,6 @@ export type DatePickerInputProps = Omit<
   ) => void
 }
 
-export type InvalidDates = {
-  invalidDate?: string
-  invalidStartDate?: string
-  invalidEndDate?: string
-}
-
 const defaultProps: DatePickerInputProps = {
   maskOrder: 'dd/mm/yyyy',
   maskPlaceholder: 'dd/mm/åååå',
@@ -150,11 +145,8 @@ function DatePickerInput(externalProps: DatePickerInputProps) {
   const [focusState, setFocusState] = useState<string>('virgin')
 
   const { partialDatesRef, setPartialDates } = usePartialDates()
+  const { invalidDatesRef, setInvalidDates } = useInvalidDates()
 
-  const invalidDatesRef = useRef<InvalidDates>({
-    invalidStartDate: null,
-    invalidEndDate: null,
-  })
   const isDateFullyFilledOutRef = useRef(false)
 
   const {
@@ -343,7 +335,7 @@ function DatePickerInput(externalProps: DatePickerInputProps) {
         }
       )
     },
-    [updateDates, callOnChangeHandler, hasHadValidDate]
+    [updateDates, callOnChangeHandler, hasHadValidDate, invalidDatesRef]
   )
 
   const callOnChange = useCallback(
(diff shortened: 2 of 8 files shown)


packages/dnb-eufemia/src/components/date-picker/DatePicker.tsx
```diff
@@ -118,6 +118,7 @@ export type DatePickerProps = {
   maxDate?: DateType
   /**
    * Corrects the input date value to be the same as either `minDate` or `maxDate`, when the user types in a date that is either before or after one of these. Defaults to `false`.
+   * Deprecate in v11 in favor of built in validation through Field.Date
```
langz, line 121: Is it deprecated? Or is it to be deprecated?

I don't fully understand what I should use instead of `correctInvalidDate` from Field.Date to still get the same behaviour 🤔 

```suggestion
   * @deprecated in v11, use validation through Field.Date instead.
```

langz:
I gave the [deploy preview example ](https://eufemia-git-fix-datepicker-correct-invalid-date-14ed6e-eufemia.vercel.app/uilib/components/date-picker/demos/#hidden-nav) a go, and I feel like it's correcting me when I'm entering a valid date 🤔 


https://github.com/user-attachments/assets/e16303ed-dd54-4660-9728-af5ee9b2d4c7

In the video, I enter 02052022

langz:
Do you have the link to the Slack thread again?

I'm just really curious about the motivation for this, as I personally find it very intrusive and annoying that the date I input will be overwritten, even if the date is not "valid". I personally, would rather want to see a descriptive error message, then correct my mistake, instead of the date being changed to something I never typed in.

Let's say if I type in the wrong year for a field with `minDate="2022-11-01"` (01.11.2021).
I type in 03.02.2021 instead of 03.02.2022 which was what I actually wanted to do. Then the date automatically get's set to 01.11.2021 (minDate), and I have to type in the whole date that I wanted (03.02.2022) again. Instead of just changing 2021 in the date I wrongly entered (03.02.2021) to 2022 if I would have gotten an error message.

joakbjerk:
> Do you have the link to the Slack thread again?
> 
> I'm just really curious about the motivation for this, as I personally find it very intrusive and annoying that the date I input will be overwritten, even if the date is not "valid". I personally, would rather want to see a descriptive error message, then correct my mistake, instead of the date being changed to something I never typed in.
> 
> Let's say if I type in the wrong year for a field with `minDate="2022-11-01"` (01.11.2021).
> I type in 03.02.2021 instead of 03.02.2022 which was what I actually wanted to do. Then the date automatically get's set to 01.11.2021 (minDate), and I have to type in the whole date that I wanted (03.02.2022) again. Instead of just changing 2021 in the date I wrongly entered (03.02.2021) to 2022 if I would have gotten an error message.

[Slack](https://dnb.enterprise.slack.com/archives/CMXABCHEY/p1742365744513439?thread_ts=1742365744.513439&cid=CMXABCHEY)

Totally agree. And just adds more unnecessary complexity to the code, and its not a great user experience. That is also the reason I wish we deprecate the correctInvalidDate property in v11

langz:
> > Do you have the link to the Slack thread again?
> > I'm just really curious about the motivation for this, as I personally find it very intrusive and annoying that the date I input will be overwritten, even if the date is not "valid". I personally, would rather want to see a descriptive error message, then correct my mistake, instead of the date being changed to something I never typed in.
> > Let's say if I type in the wrong year for a field with `minDate="2022-11-01"` (01.11.2021).
> > I type in 03.02.2021 instead of 03.02.2022 which was what I actually wanted to do. Then the date automatically get's set to 01.11.2021 (minDate), and I have to type in the whole date that I wanted (03.02.2022) again. Instead of just changing 2021 in the date I wrongly entered (03.02.2021) to 2022 if I would have gotten an error message.
> 
> [Slack](https://dnb.enterprise.slack.com/archives/CMXABCHEY/p1742365744513439?thread_ts=1742365744.513439&cid=CMXABCHEY)
> 
> Totally agree. And just adds more unnecessary complexity to the code, and its not a great user experience. That is also the reason I wish we deprecate the correctInvalidDate property in v11

Thanks for the link.
Yeah, we can/should discuss it in the team, what to do with it, and if we deprecate/remove it we should add some good words about it in `v11-info.mdx`.
Thank god the `correctInvalidDate` is opt-in 😅

langz:
> I gave the [deploy preview example ](https://eufemia-git-fix-datepicker-correct-invalid-date-14ed6e-eufemia.vercel.app/uilib/components/date-picker/demos/#hidden-nav) a go, and I feel like it's correcting me when I'm entering a valid date 🤔
> 
>  Screen.Recording.2025-03-25.at.21.02.35.mov 
> In the video, I enter 0-2-0-5-2-0-2-2 (02.05.2022) which I assume should be a valid date, then it corrects it to be 01.05.2022 it does so when I enter 2 as part of 2022.

Other than this experience, I think it works as expected and described in our existing docs 👍

joakbjerk:
> I gave the [deploy preview example ](https://eufemia-git-fix-datepicker-correct-invalid-date-14ed6e-eufemia.vercel.app/uilib/components/date-picker/demos/#hidden-nav) a go, and I feel like it's correcting me when I'm entering a valid date 🤔 
> 
> 
> https://github.com/user-attachments/assets/e16303ed-dd54-4660-9728-af5ee9b2d4c7
> 
> In the video, I enter 0-2-0-5-2-0-2-2 (02.05.2022) which I assume should be a valid date, then it corrects it to be 01.05.2022 it does so when I enter 2 as part of 2022.


Looks like it's caused when the month is set to 00 for a breif moment and it then fires the correction with the outdated date value when you type 05 🤔 

Again, this prop is just adding more complexity in return of worse UX, so I vote  for closing this pr

joakbjerk:
Closed because this functionality only adds complexity to the code, and leads to worse and confusing UX.

PR #4782, opened 2025-03-19
feat(Field.SelectCountry): enhance value type

langz:
Or perhaps this is "too breaking", possibly requiring our users to import and use the type `CountryISO`?


tujoworker (approved):
I think it should not break. Because, when a country code is given, that is not a part of the data set, it would not be valid anyway.

Issue #4771, opened 2025-03-18
Type issues with FieldsProps and FieldBlock


Sodnarts:
## 💥 Regression Report

We found two separate issues when upgrading from 10.62.4 to 10.66.0. 

1. When passing in the `error` from `useFieldProps` to `FieldBlock` we get a _Type 'unknown' is not assignable to type 'string'_ error.

2. When passing in `props: FieldProps<string, string>` to `useFieldProps(props)` we get a _Types of property 'onFocus' are incompatible._ error which points to _'ProvideAdditionalEventArgs' is not assignable to type 'ReceiveAdditionalEventArgs<string, DefaultErrorMessages>'_

Both issues seems to present when using the latest version of Eufemia as well.

## Last working version

Worked up to version: 10.62.4

Stopped working in version: 10.63.0 (for issue 1). 10.65.0 (for issue 2).

## To Reproduce

1. Pass `error` from `useFieldProps` to `FieldBlock` 
2. Pass props with type `FieldProps<string, string>` to `useFieldProps(props)`.

Here's a [CSB](https://codesandbox.io/p/sandbox/eloquent-tharp-z9pck9) that has reproduced the issue. 

## Expected behavior

Expected to not throw the errors described above.

## Link to repo / fork / branch (highly encouraged)

Please provide an example for how this feature would be used.

## Run `npx envinfo`

Paste the results here:

```bash

```

PR #4785, opened 2025-03-19
fix(Forms): enhance typing for `FieldBlock` and `useFieldProps`

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1742375173085389
Fixes https://github.com/dnbexperience/eufemia/issues/4771

CSB of the issue: https://codesandbox.io/p/sandbox/eloquent-tharp-z9pck9

For now, I've just added a stor, but not sure if that's too helpful as I don't believe the TypeScript settings in this repo is strict enough to catch/display the same errors as in the [CSB of the issue](https://codesandbox.io/p/sandbox/eloquent-tharp-z9pck9).

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/create-component.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/create-component.mdx
@@ -184,6 +184,7 @@ import {
   Form,
   FieldBlock,
   useFieldProps,
+  FieldProps,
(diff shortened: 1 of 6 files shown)


tujoworker (approved):
To test the changes, we need to enable strict mode. 
Also, I look into if its possible to enable strict mode for some files only. But that is really not possible in a good way – as of now. But this would be an important step towards safer types, when a consumer uses strict mode, which is often the case.

PR #4790, opened 2025-03-20
docs: add `xx-small` to the spacing table

tujoworker:
This is a draft PR. Looks like we support it in JavaScript, but not in CSS. So we would need to change the setup in CSS to make it actually work:

Screenshot 2025-03-20 at 14 17 55


tujoworker:
Closing this draft because the original documentation change is already covered on main: `xx-small` is documented and works as a standalone 0.25rem spacing value.

Combining it with another size, for example `small xx-small`, still does not produce 1.25rem. The JavaScript creates both CSS classes, but the CSS cascade resolves the margin to 1rem. Removing the current documentation note would therefore describe behavior we do not support.

If quarter-step combinations are needed in the future, we should handle that as a focused Space behavior change with CSS support and a test of the rendered margin.

PR #4799, opened 2025-03-21
docs: update `onInput` example (prevent typing of invalid characters)

tujoworker:
I tested the added example and found an issue when selecting the text. The updated example takes care of when selecting an area in the input field before typing.


langz (approved):
If it's important. I think it's safe to deploy latest main in our portal. As it's just docs changes and minor chores on main now.

tujoworker:
no problem. no rush.

PR #4801, opened 2025-03-24
chore: Cannot decrease to heading level 1

langz:
Improves the warning:

```
Eufemia Cannot decrease to heading level 1! Had before 2 - The new level is 2 
NB: This warning was triggered by:  #Description
```


tujoworker (approved):
Nice catch here as well 💰

PR #4819, opened 2025-03-26
feat(GlobalError): deprecate `status` in favor of `statusCode`

langz:
I like the word `statusCode` better than `status`.
It also seems that other source do as well:
- https://en.wikipedia.org/wiki/List_of_HTTP_status_codes
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status

Changing from `status` to `statusCode` also helps with reducing the confusion with the property `code`, see docs https://eufemia.dnb.no/uilib/components/global-error/properties/

But I'll also open a PR to change that property name as well, to reduce the confusion.

Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1742973159203409


packages/dnb-eufemia/src/components/global-error/__tests__/GlobalError.test.tsx
```diff
@@ -68,6 +68,73 @@ describe('GlobalError', () => {
     )
   })
 
+  it('should not render status code when statusCode is empty', () => {
```
langz, line 71: I added this test, to make sure that this existing functionality should stay the same now after I've deprecated status  property in favor of statusCode.

tujoworker (approved):

PR #4832, opened 2025-03-27
chore(Forms): refactor object to use maps

tujoworker:
We use maps a couple of places in Eufemia Forms. Maps in general can be faster, but are mostly less memory hungry. So I think some of our internals can and should be rewritten.


langz (approved):
Wow, we are so fast now:

[image](https://media2.giphy.com/media/iegKxpndVUjrARrJM5/200w.gif?cid=6c09b952zqfli70gxiagvi4su2pnnnapq04q6zg6abfya253&ep=v1_gifs_search&rid=200w.gif&ct=g)

PR #4834, opened 2025-03-27
fix(Field.Upload): async removal of file with same file name as other file

langz:
When async removing a file that has the same file name as an other file, both files displays as loading/removing, and only 1 file is removed. 

The expected result when clicking remove for file x is that only file x is loading/removing and only file x is removed.

CSB of the problem: https://codesandbox.io/p/devbox/exciting-pike-td395f

Video of the problem:

https://github.com/user-attachments/assets/549bccac-d856-4b48-a924-6104d592e666

CSB with the fix in this PR: https://codesandbox.io/p/devbox/sad-antonelli-dkdmtp?workspaceId=ws_LbiDwJdEBdbwxH2AkyAyVE

Sidenote: As of today, it's allowed with files of the same file name, that's not in our code classified as an "existing file".
These are the conditions that needs to be met for it to be an "existing file": https://github.com/dnbexperience/eufemia/blob/efbf145a25ef7c8274bd06df7b3c2dc695cbb4cd/packages/dnb-eufemia/src/components/upload/useUpload.ts#L49-L51

### DIFF
--- a/packages/dnb-eufemia/src/components/upload/UploadFileList.tsx
+++ b/packages/dnb-eufemia/src/components/upload/UploadFileList.tsx
@@ -45,11 +45,11 @@ function UploadFileList() {
       const idIsSame =
         fileListElement.id && fileListElement.id === fileToBeUpdated.id
 
-      const fileNameIsSame =
-        fileListElement?.file?.name &&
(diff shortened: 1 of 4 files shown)


tujoworker (approved):
Nice work 👏

PR #4851, opened 2025-03-31
chore(Avatar): refactor properties docs

langz:


packages/dnb-eufemia/src/components/avatar/AvatarDocs.ts
```diff
@@ -0,0 +1,97 @@
...
+  hasLabel: {
+    doc: 'If aria-hidden is set to `true` or if a label is given, typical inside a table or dl (definition list), then you can disable Avatar.Group as a dependent of Avatar. Use `true` to omit the `Avatar group required:` warning.',
+    type: 'boolean',
+    status: 'optional',
+  },
+  className: {
+    doc: 'Custom className for the component root.',
+    type: 'string',
+    status: 'optional',
+  },
```
tujoworker, line 43:
```suggestion
```

I don't think we need to document such a common prop. Also because we don't do it on other components.

packages/dnb-eufemia/src/components/avatar/AvatarDocs.ts
```diff
@@ -0,0 +1,97 @@
...
+  children: {
+    doc: 'The Avatars to group.',
+    type: 'React.ReactNode',
+    status: 'optional',
+  },
+  className: {
+    doc: 'Custom className for the component root.',
+    type: 'string',
+    status: 'optional',
+  },
```
tujoworker, line 86:
```suggestion
```

I don't think we need to document such a common prop. Also because we don't do it on other components.

tujoworker (approved):

PR #4867, opened 2025-04-01
chore(Forms): refactor `onSubmit` event to reuse existing logic

tujoworker:


langz (approved):
Looks good, I've not tested this 🔨 
But if there would be a change in actual functionality, one would assume our existing tests to catch this and fail 🤞

PR #4881, opened 2025-04-02
chore(Field.Date): improve localization in docs

langz:
This PR now just adds the `ChangeLocale` dropdown in the `Field.Date` docs.

TODO:
- [x]  Actually fix that localization does not work for `Field.Date` in our docs. Is it a real problem with the component or just for the docs in our portal? It seems like it updates the label based on the locale, but not the formatting of the date.

Here's a video from main branch when changing locale in DatePicker (expected behavior):

https://github.com/user-attachments/assets/c9fa9398-3a3f-4da6-abf6-889ab43dd72a

Here's a video from main branch when changing locale in Field.Date (unexpected behavior):

https://github.com/user-attachments/assets/7f1ddabd-4269-4aa0-b3b1-9cf40480c554


langz:
Thanks @joakbjerk, works great now 🚀

PR #4895, opened 2025-04-03
fix(Upload): compare existing properties to determine duplicate

langz:


packages/dnb-eufemia/src/components/upload/useUpload.ts
```diff
@@ -44,11 +44,23 @@ function useUpload(id: UploadProps['id']): useUploadReturn {
 
   const getExistingFile = useCallback(
     (file: File, fileItems: Array<UploadFile> = files) => {
+      const compareExistingProperty = function (
+        a: File,
+        b: File,
+        property: string
+      ) {
+        return (
+          Object.prototype.hasOwnProperty.call(a, property) &&
```
tujoworker, line 53:
```suggestion
          property in a &&
```

packages/dnb-eufemia/src/components/upload/UploadFileListCell.tsx
```diff
@@ -26,7 +26,7 @@ import { UploadFile, UploadFileNative } from './types'
 
 // Shared
 import { getPreviousSibling } from '../../shared/component-helper'
-import useUpload from './useUpload'
+import useUpload, { isFilesDuplicates } from './useUpload'
```
tujoworker, line 29:
```suggestion
import useUpload, { hasFilesDuplicates } from './useUpload'
```

I'm not sure. But "has" sounds more correct in that case.

tujoworker (approved):

packages/dnb-eufemia/src/components/upload/UploadFileListCell.tsx
langz: Agree, will rename tomorrow :)

langz: I tried improving it by renaming to `isFileEqual`, hope that makes it better 🤔

packages/dnb-eufemia/src/components/upload/useUpload.ts
```diff
@@ -64,4 +60,27 @@ function useUpload(id: UploadProps['id']): useUploadReturn {
...
+    a: File,
+    b: File,
+    property: string
+  ) {
+    return (
+      a &&
+      property in a &&
+      b &&
+      property in b &&
+      (a[property] === 0 || // If value is 0, which is default when not provided, we can't say whether the file is a duplicate or not, so we assume they are.
```
langz, line 74:
```suggestion
      (a[property] === 0 || // If value is 0, which is default when not provided, we can't say whether the file is a equal or not, so we assume they are.
```

packages/dnb-eufemia/src/components/upload/useUpload.ts
```diff
@@ -64,4 +60,27 @@ function useUpload(id: UploadProps['id']): useUploadReturn {
...
+        b[property] === 0 || // If value is 0, which is default when not provided, we can't say whether the file is a duplicate or not, so we assume they are.
```
langz, line 75:
```suggestion
        b[property] === 0 || // If value is 0, which is default when not provided, we can't say whether the file is equal or not, so we assume they are.
```

PR #4902, opened 2025-04-04
fix(Avatar.Group): forward `size`, `variant`, `color`, `backgroundColor` to single child

langz:
Fixes so setting `size`, `variant`, `color`, `backgroundColor` when single Avatar as child should work


langz:
https://github.com/dnbexperience/eufemia/pull/4903

PR #4904, opened 2025-04-04
chore: share `getColor` function

langz:
Not entirely sure if this is the best place to put it 🤔 

For now I moved it to `packages/dnb-eufemia/src/shared/helpers.js`


joakbjerk (approved):
`helpers.js` makes sense to me at least 🤷‍♂️

tujoworker (approved):

PR #4914, opened 2025-04-08
fix(Heading): warning when increasing heading to be `h2` when previous sibling is `h1`

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1743769418111539
CSB: https://codesandbox.io/p/devbox/value-padding-forked-5lrp2k

I do not think we should warn about this, as I believe:
```
<Heading>heading</Heading>
<Heading increase>subheading</Heading>
``` 
should result in h1 and h2 which it does, but without warnings.

I do think the reason why we log the warning today, is because we automatically change the second heading to be a h2 in the snippet:
```
<Heading>heading</Heading>
<Heading>subheading</Heading>
``` 
So hence, our code tries/thinks the following code will increase the heading level to h3 of the second heading which is already made into a h3:

```
<Heading>heading</Heading>
<Heading increase>subheading</Heading>
``` 

I think the code perhaps could be simplified or made a tiny bit easier to read.

### DIFF
--- a/packages/dnb-eufemia/src/components/heading/HeadingHelpers.ts
+++ b/packages/dnb-eufemia/src/components/heading/HeadingHelpers.ts
@@ -85,11 +85,16 @@ export const correctInternalHeadingLevel = ({
       }
     }
 
+    const counterWasCorrectedToLevel2 =
+      globalHeadingCounter?.current?.hasCorrection() &&
+      globalSyncCounter?.current?.level === 1 &&
+      counter?.getLevel() === 2
+
     if (level >= 1) {
       counter.setLevel(level)
     } else if (decrease) {
       counter.decrement()
-    } else if (increase) {
+    } else if (increase && !counterWasCorrectedToLevel2) {
       counter.increment()
     }
 
--- a/packages/dnb-eufemia/src/components/heading/HeadingHelpers.ts
+++ b/packages/dnb-eufemia/src/components/heading/HeadingHelpers.ts
@@ -67,6 +67,14 @@ export const correctInternalHeadingLevel = ({
     counter.enableBypassChecks()
   }
 
+  const hasCounterWasCorrectedToLevel2 = () => {
+    return (
+      globalSyncCounter?.current?.level === 1 &&
+      globalHeadingCounter?.current?.hasCorrection() &&
+      counter?.getLevel() === 2
+    )
+  }
+
   const update = (level: InternalHeadingLevel) => {
     if (!isRerender) {
       counter.makeMeReady({ level })
@@ -85,16 +93,11 @@ export const correctInternalHeadingLevel = ({
       }
     }
 
-    const counterWasCorrectedToLevel2 =
-      globalHeadingCounter?.current?.hasCorrection() &&
-      globalSyncCounter?.current?.level === 1 &&
-      counter?.getLevel() === 2
-
     if (level >= 1) {
       counter.setLevel(level)
     } else if (decrease) {
       counter.decrement()
-    } else if (increase && !counterWasCorrectedToLevel2) {
+    } else if (increase && !hasCounterWasCorrectedToLevel2()) {
       counter.increment()
     }
 
--- a/packages/dnb-eufemia/src/components/heading/__tests__/Heading.test.tsx
+++ b/packages/dnb-eufemia/src/components/heading/__tests__/Heading.test.tsx
@@ -317,6 +317,26 @@ describe('Heading component', () => {
     )
   })
 
+  it('have to not warn when setting h2 by increasing Header', () => {
+    const debugWarning = jest.fn()
+    render(
+      <>
+        <Heading debug={debugWarning}>Heading #1</Heading>
+        <Heading debug={debugWarning} increase>
+          Heading #2
+        </Heading>
+      </>
+    )
+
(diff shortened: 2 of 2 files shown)


andlbrei:
I see that the documentation might be improved to explain this.
https://eufemia.dnb.no/uilib/components/heading/

First there is this example which actually shows the "auto h2", but also implies you need to use `increase` to get to the next level for the h3, which is probably why I used it that way.
```html
<Heading.Level reset={1}>
  <Heading>h1</Heading>
  <Heading>h2</Heading> #I didn't notice this
  <Heading increase>h3</Heading>
  <Heading>still h3</Heading>
  <Heading increase>h4</Heading>
  <Heading increase>h5</Heading>
  <Heading decrease>h4</Heading>
  <Heading level={2}>back to h2</Heading>
  <Heading increase>h3</Heading>
</Heading.Level>
```

Then there is the text
```
Heading level core-concept
- A heading will inherit the level from its previous sibling.
```

Then later it is clarified:
```
Heading level rules and corrections

The correction will ensure that:

    a heading will start with a level 1.
    the second level will get corrected be level 2.
```
But I didn't read that 😅 

So I guess there is the choice between letting the `increase` on a "auto h2" become a h3, in which case the warning is correct, or remove the warning if using `increase` on a "auto h2" heading will not have any effect? 🤔

packages/dnb-eufemia/src/components/heading/__tests__/Heading.test.tsx
```diff
@@ -317,6 +317,26 @@ describe('Heading component', () => {
     )
   })
 
+  it('have to not warn when setting h2 by increasing Header', () => {
+    const debugWarning = jest.fn()
```
langz, line 321:
```suggestion
```

packages/dnb-eufemia/src/components/heading/__tests__/Heading.test.tsx
```diff
@@ -317,6 +317,26 @@ describe('Heading component', () => {
...
+    render(
+      <>
+        <Heading debug={debugWarning}>Heading #1</Heading>
```
langz, line 324:
```suggestion
        <Heading debug={warn}>Heading #1</Heading>
```

packages/dnb-eufemia/src/components/heading/__tests__/Heading.test.tsx
```diff
@@ -317,6 +317,26 @@ describe('Heading component', () => {
...
+        <Heading debug={debugWarning} increase>
```
langz, line 325:
```suggestion
        <Heading debug={warn} increase>
```

packages/dnb-eufemia/src/components/heading/__tests__/Heading.test.tsx
```diff
@@ -317,6 +317,26 @@ describe('Heading component', () => {
...
+      </>
+    )
+
+    expect(document.querySelectorAll('.dnb-heading')[0].textContent).toBe(
+      '[h1] Heading #1'
+    )
+    expect(document.querySelectorAll('.dnb-heading')[1].textContent).toBe(
+      '[h2] Heading #2'
+    )
+    expect(debugWarning).not.toHaveBeenCalled()
```
langz, line 337:
```suggestion
    expect(warn).not.toHaveBeenCalled()
```

langz:
> I see that the documentation might be improved to explain this. https://eufemia.dnb.no/uilib/components/heading/
> 
> First there is this example which actually shows the "auto h2", but also implies you need to use `increase` to get to the next level for the h3, which is probably why I used it that way.
> 
> ```
> <Heading.Level reset={1}>
>   <Heading>h1</Heading>
>   <Heading>h2</Heading> #I didn't notice this
>   <Heading increase>h3</Heading>
>   <Heading>still h3</Heading>
>   <Heading increase>h4</Heading>
>   <Heading increase>h5</Heading>
>   <Heading decrease>h4</Heading>
>   <Heading level={2}>back to h2</Heading>
>   <Heading increase>h3</Heading>
> </Heading.Level>
> ```
> 
> Then there is the text
> 
> ```
> Heading level core-concept
> - A heading will inherit the level from its previous sibling.
> ```
> 
> Then later it is clarified:
> 
> ```
> Heading level rules and corrections
> 
> The correction will ensure that:
> 
>     a heading will start with a level 1.
>     the second level will get corrected be level 2.
> ```
> 
> But I didn't read that 😅
> 
> So I guess there is the choice between letting the `increase` on a "auto h2" become a h3, in which case the warning is correct, or remove the warning if using `increase` on a "auto h2" heading will not have any effect? 🤔

Thanks for the valuable feedback @andlbrei.
@tujoworker, do you have any thoughts on what could be the preferable improvement in this scenario?

packages/dnb-eufemia/src/components/heading/HeadingHelpers.ts
```diff
@@ -85,11 +85,16 @@ export const correctInternalHeadingLevel = ({
       }
     }
 
+    const counterWasCorrectedToLevel2 =
```
tujoworker, line 88: As talked in the meeting, we should move this logic into a function / or different logic, so it not runs when not necessary. Also, the `globalSyncCounter?.current?.level === 1`  I would run before the functions.

tujoworker (approved):

PR #4916, opened 2025-04-09
chore(DatePicker): deprecate `correctInvalidDate`

joakbjerk:
is this a chore or a feat?


langz:
I would think it's a feat; in terms of it showing up in the release log. We should also make a post about it in the slack channel.

tujoworker (approved):

PR #4918, opened 2025-04-09
fix(InputMasked): allow 0 as value when `disallowLeadingZeroes` is set

langz:


tujoworker:
I have added a fix and aligned the tests.

Now – it's almost a question if the `Field.Currency` should have `disallowLeadingZeroes={true}` by default.

langz:
I made a CSB to test this, but I'm allowed to enter 01, 002, 00003, etc, when `disallowLeadingZeroes`.
https://codesandbox.io/p/devbox/eloquent-wildflower-m6rgns

langz (commented):
I think it looks good 🥇 
Can't remember correctly, but feel like we found a potential improvement or so on yesterdays 🚀 Eufemia Catchup meeting.

langz:
Also when changing from negative to positive value, it's possible to add a leading zero. But probably edge-case?


https://github.com/user-attachments/assets/d9c84fdf-4eb2-4058-bca8-c711e8be8f4d

tujoworker:
The more I dive into this, the more I realize that its almost impossible to "always" disallow a leading zero. Because there are so many edge cases, especially when we want to allow temporarily leading zeros during typing (100 => 00 => 200). So in order to make a "really safe" version, we would need to remove the leading zeros during the blur event. But sill, a user could experience this as a not so nice feature (100 => 00 => blur => undefined).

snorrekim (changes requested):
Im still seeing some issues, I'm going to take a closer look and add a commit. It's a bit too complex to explain right now, will need a bit of time on it.

tujoworker:
> Only potential improvement I can find is when using negative numbers

This is taken care of now.

tujoworker (approved):

PR #4935, opened 2025-04-10
feat(Upload): add `description` to fileItem

langz:
Can be demoed here: https://eufemia-git-feat-add-description-to-file-eufemia.vercel.app/uilib/components/upload/demos/#upload-with-file-description

Now looking like:

image

TODO: Test and see if this works in Field.Upload or that we have to add changes for it to work there as well


tujoworker (approved):
Looks good. Also, having an example + one test for the `Field.Upload` would be nice. Should be easy to document. It works already. See screenshot below.

Screenshot 2025-04-10 at 14 38 50

packages/dnb-eufemia/src/components/upload/UploadFileListCell.tsx
```diff
@@ -168,10 +170,17 @@ const UploadFileListCell = ({
...
         />
+        {getDescription()}
       </div>
     )
   }
 
+  function getDescription() {
+    return description ? (
+      <P className="dnb-upload__text">{description}</P>
+    ) : null
```
tujoworker, line 181:
```suggestion
    if(!description){
      return null
    }
    return <P className="dnb-upload__text">{description}</P>
```
I tend to prefer writing the logic like this. Just because its more "isolated" and the if statement can be extended without changing "what we expect" from the function/component. But no big deal.

langz:
> Looks good. Also, having an example + one test for the `Field.Upload` would be nice. Should be easy to document. It works already. See screenshot below.
> 
> 
> 
> Screenshot 2025-04-10 at 14 38 50
> 
> 
> 
> 

Thanks, I'll add a PR with an example in Field.Upload

packages/dnb-eufemia/src/components/upload/UploadFileListCell.tsx
langz: This caused a lint issue. Will fix and improve this later this evening.

PR #4937, opened 2025-04-10
fix(DatePicker): display correct dates when props are updated with `useLayoutEffect`

joakbjerk:
Fixes [this issue](https://dnb-it.slack.com/archives/CMXABCHEY/p1744105593951939)

### TODO

- [x] Replace input dates `useEffect` with `useLayoutEffect`
- [x] Add tests

### DIFF
--- a/packages/dnb-eufemia/src/components/date-picker/hooks/useDates.ts
+++ b/packages/dnb-eufemia/src/components/date-picker/hooks/useDates.ts
@@ -1,9 +1,13 @@
-import { useCallback, useLayoutEffect, useMemo, useState } from 'react'
+import React, { useCallback, useMemo, useRef, useState } from 'react'
 import { convertStringToDate, isDisabled } from '../DatePickerCalc'
 import isValid from 'date-fns/isValid'
 import format from 'date-fns/format'
 import { addMonths, isSameDay } from 'date-fns'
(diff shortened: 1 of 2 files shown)


packages/dnb-eufemia/src/components/date-picker/hooks/useDates.ts
```diff
@@ -145,7 +151,7 @@ export default function useDates(
 
   // Updated input dates based on start and end dates, move to DatePickerInput
   // TODO: Move to DatePickerInput
-  useEffect(() => {
+  useLayoutEffect(() => {
```
tujoworker, line 154: We should import it like all other places. Search for this hook and you will find it.

tujoworker (approved):

langz (approved):
Looking good🚀
Should we look/test for the same issue in other components that we have? Can we add this test of yours to other components where this can/could happen?

joakbjerk:
> Looking good🚀 Should we look/test for the same issue in other components that we have? Can we add this test of yours to other components where this can/could happen?
> 
> Have we actually tested that this fixes the issue highlighted in the CSB from @tujoworker?

Double checked the CSB, and the fix does not work in StrictMode

tujoworker:
Sounds we should add a test with StrictMode. At least, it works now in "prod".

PR #4938, opened 2025-04-10
chore(DatePicker): move `hasHadValidDate` to `DatePickerInput`

joakbjerk:
Moves the value `hasHadValidDate` variable to `DatePickerInput` since it's the only component that actually uses it. No need for it to live in the context and `useDates` hook

### DIFF
--- a/packages/dnb-eufemia/src/components/date-picker/DatePickerInput.tsx
+++ b/packages/dnb-eufemia/src/components/date-picker/DatePickerInput.tsx
@@ -179,10 +179,7 @@ function DatePickerInput(externalProps: DatePickerInputProps) {
 
   const translation = useTranslation().DatePicker
 
-  const hasHadValidDate = useMemo(
-    () => isValid(startDate) || isValid(endDate),
-    [startDate, endDate]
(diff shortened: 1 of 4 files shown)


langz (approved):

packages/dnb-eufemia/src/components/date-picker/DatePickerInput.tsx
```diff
@@ -180,6 +179,11 @@ function DatePickerInput(externalProps: DatePickerInputProps) {
 
   const translation = useTranslation().DatePicker
 
+  const hasHadValidDate = useMemo(
```
tujoworker, line 182: Do we need to memo it?

tujoworker (approved):

packages/dnb-eufemia/src/components/date-picker/DatePickerInput.tsx
joakbjerk: I do not think so, sometimes lint wants you to do it if its used as an dependency for another hook method (memo, callback)🤔

PR #4939, opened 2025-04-10
fix(InputMasked): ensure correct cursor position when suffix changes during typing

tujoworker:
This issue only occurred when the suffix changed while typing. In that case, the cursor was placed at the end because the lengths no longer matched the initially provided value.
We now send in the current/latest suffix, and the cursor position is set correctly.

In this video; the issue is shown all the way at the end of the video, where the cursor gets placed at the end.

https://github.com/user-attachments/assets/aee4df3c-3394-4b40-a653-0083b13e1ca1


langz (approved):
🥳 🍾

PR #4947, opened 2025-04-11
chore: remove SvgComparison.js

langz:


tujoworker (approved):
No, we don't use it. So we can remove it.

langz:
> No, we don't use it. So we can remove it.

Good, I've updated this PR to now remove the file `packages/dnb-eufemia/src/components/icon/lib/SvgComparison.js`

Issue #4931, opened 2025-04-10
bug(Tooltip): Tooltip alignment not working anymore


NathanBrodin:
## 💥 Regression Report

As discussed in [Slack](https://dnb-it.slack.com/archives/CMXABCHEY/p1744274312734339), the tooltip position is not correct, the code always acts as if  arrow="center" when we set the position to `right`.

## Last working version

Worked up to version: v9

Stopped working in version: v10

## To Reproduce

![Image](https://github.com/user-attachments/assets/c65b9570-e883-4cec-a766-40e9fb5f4dad)

In our app, here's how it was working in v9:

![Image](https://github.com/user-attachments/assets/c31e8242-c674-4878-90a5-df89f28edf9f)

And here's how it's working now in v10:

![Image](https://github.com/user-attachments/assets/48f555c1-ba05-4b27-a9af-bf90c5a77189)

PR #4951, opened 2025-04-11
fix(Tooltip): Tooltip alignment not working anymore

NathanBrodin:
Hey 👋 Fix for #4931 

- The cases where we would align top right were not handled so I introduced a function to correctly calculate the tooltip position for every tooltip alignments.
- I refactored the stories to actually show all the different uses cases with alignments, positions... as well with other properties and stuff. 

This does not break the use of the tooltip by other components (blank Anchor, buttons and such).

Have a nice weekend!

Fixes https://github.com/dnbexperience/eufemia/issues/4931

### DIFF
--- a/packages/dnb-eufemia/src/components/tooltip/TooltipContainer.tsx
+++ b/packages/dnb-eufemia/src/components/tooltip/TooltipContainer.tsx
@@ -127,8 +127,6 @@ export default function TooltipContainer(
       return // stop here
     }
 
-    let alignOffset = 0
-
     const elementWidth = element.offsetWidth
     const elementHeight = element.offsetHeight
     const rect = target.getBoundingClientRect()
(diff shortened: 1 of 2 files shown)


langz:
Thanks for the great work @NathanBrodin :) 
I've opened a temporary PR https://github.com/dnbexperience/eufemia/pull/4965 just to highlight what will visually change in the tooltip component with your PR.
I've simply taken the code changes from `packages/dnb-eufemia/src/components/tooltip/TooltipContainer.tsx` and updated the visual screenshot tests based on it.

I like the improvements of your contribution, but because of my lack of understanding of the Tooltip component in v9 and now in v10, I find it a bit difficult to say what the expected result is for all affected scenarios:
- [Take for instance when just setting align="right", should the arrow be positioned at the end of the component or should the content of the tooltip be placed at the end of the component?](https://github.com/dnbexperience/eufemia/pull/4965/files#diff-29610c4000f55e7b588d6dfa01e3702510368e80727af9549cde77003dec5aeb) The docs for align property says _defines the offset position to the target element the arrow appears_, so I would the arrow should be positioned at the end of the component?
-[ I also have the same question as above when align="left" and arrow="left"](https://github.com/dnbexperience/eufemia/pull/4965/files#diff-68b0181564848047d203464b9b5df7ade6c0c6db6ad38f06cb815a3504d65e73)

langz:
Closing this, as we decided to go with a different approach implemented in https://github.com/dnbexperience/eufemia/pull/4968

Thanks for the great work ❤️

PR #4993, opened 2025-04-23
fix(Value.SelectCurrency): handle invalid iso code values

langz:
When providing a not valid/supported iso code as `value`, the provided value will be displayed instead of the component breaking/failing.

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Value/SelectCurrency/info.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Value/SelectCurrency/info.mdx
@@ -4,7 +4,7 @@ showTabs: true
 
 ## Description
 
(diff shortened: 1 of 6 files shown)


packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Value/SelectCurrency/info.mdx
```diff
@@ -4,7 +4,7 @@ showTabs: true
 
 ## Description
 
-`Value.SelectCurrency` will render the selected currency.
+`Value.SelectCurrency` will render the selected currency display name by the `value`'s ISO code ([ISO 4217 code](https://en.wikipedia.org/wiki/ISO_4217)). It displays the currency name in the current locale, together with the currency ISO code, like `Norwegian krone (NOK)`. If the value is not a valid/supported ISO code, it renders the value provided.
```
langz, line 7:
```suggestion
`Value.SelectCurrency` will render the selected currency display name by the `value`'s ISO code ([ISO 4217 code](https://en.wikipedia.org/wiki/ISO_4217)). It displays the currency name in the current locale, together with the currency ISO code, like `Norwegian krone (NOK)`. If the value provided is not a valid/supported ISO code, it renders the value.
```

tujoworker (approved):
Nice 👏👏

PR #5000, opened 2025-04-24
feat(Field.Currency): defaults to disallowing leading zeroes

langz:
Related to https://github.com/dnbexperience/eufemia/pull/4918

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/Currency/__tests__/Currency.test.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/Currency/__tests__/Currency.test.tsx
@@ -50,14 +50,24 @@ describe('Field.Currency', () => {
     expect(input).toHaveAttribute('inputmode', 'decimal')
   })
 
-  it('should default to not allow leading zeroes when ', async () => {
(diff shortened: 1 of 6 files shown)


packages/dnb-eufemia/src/extensions/forms/Field/Currency/__tests__/Currency.test.tsx
```diff
@@ -50,6 +50,16 @@ describe('Field.Currency', () => {
     expect(input).toHaveAttribute('inputmode', 'decimal')
   })
 
+  it('should default to not allow leading zeroes ', async () => {
```
langz, line 53: Test should be updated at some point, in relation to the changes in https://github.com/dnbexperience/eufemia/pull/4918

langz:
I think this is ready for review now :) 
@tujoworker

tujoworker (approved):
Nice! But I think we should first merge #4918 and then test this one again, properly.

tujoworker (changes requested):
Following today’s meeting, we decided to postpone setting this as the default for this field for some time. Just to get a better understanding of how the changed made in #4918 will work out for projects using it.

PR #5012, opened 2025-04-25
fix(Forms): ensure Field.Selection with the autocomplete variant receives showIndicator in the first onType parameter object

tujoworker:
More [info](https://dnb-it.slack.com/archives/CMXABCHEY/p1745592615795799).


langz (approved):
🤾

PR #5019, opened 2025-04-28
feat(Field.DateOfBirth): new component for selecting date of birth

langz:


tujoworker:
> Day input field, which values are valid? Should just 1 be valid, or should we enforce 01?

It now auto correct during blur.

>     Add support for returnFormat, and dateFormat?

`dateFormat` is added. It does both. I don't think we should separate this for fields to avoid complexity.

>     Should onChange only return a value when a date is correctly entered, sort of the same as a [Field.Date](https://eufemia.dnb.no/uilib/extensions/forms/feature-fields/Date/)? But how will this affect the validator?

I think its good as it is for now.

>     https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete#bday-month must support autocomplete. Handle browser autofill/autocomplete for the month value which is in an autocomplete.

I could not find anyway to test this. We have at least added these.

>     How should the validator work? I find the logic challenging complex here compared to a regular input field. Now we have 3 that sort of should "work like one".

Seems to work fine to me now.

tujoworker (approved):
Nice work 👏

PR #5021, opened 2025-04-28
fix(DatePicker): cancel and reset functionality

joakbjerk:
Relies on https://github.com/dnbexperience/eufemia/pull/5046

Changes added based on notes from meeting

### Cancel button
Should reset to the previous date/state/value so when the user has made a choice it should use the previous value and if there is no value from the backend and the user has selected a date and then presses the button it should clear/empty the value
        
### Reset button
Should go back to the value of the date/startDate/endDate prop, should work the same way cancel does now


packages/dnb-eufemia/src/components/date-picker/hooks/useSubmittedCalendarDates.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export default function useSubmittedCalendarDates() {
+  const submittedCalendarDatesRef = useRef<SubmittedCalendarDates>({})
+
+  const setSubmittedCalendarDates = useCallback(
+    (dates: SubmittedCalendarDates) => {
+      submittedCalendarDatesRef.current = {
+        ...submittedCalendarDatesRef.current,
+        ...dates,
+      }
```
tujoworker, line 17:
```suggestion
      submittedCalendarDatesRef.current = dates
```

I don't think we need to merge them. I think we simply can overwrite the existing dates.

tujoworker (approved):
I think it looks good. Just check by comment above.

packages/dnb-eufemia/src/components/date-picker/hooks/useSubmittedCalendarDates.ts
joakbjerk: Agreed 🙏

packages/dnb-eufemia/src/components/date-picker/hooks/useSubmittedCalendarDates.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export type SubmittedCalendarDates = Pick<
+  DatePickerDates,
+  'startDate' | 'endDate'
+>
+
+export default function useSubmittedCalendarDates() {
+  const submittedCalendarDatesRef = useRef<SubmittedCalendarDates>({})
+
+  const setSubmittedCalendarDates = useCallback(
```
tujoworker, line 12:
```suggestion
  const setSubmittedDates = useCallback(
```

Do we need to have `Calendar` in the name? Does it give any extra value? 🤔 Just thinking out loud.

joakbjerk: I think not, based on the context. It's a mouthfull 😄

langz (approved):
Looking great 😊 

Is there any changes or so we have to do in Field.Date based on these this PR? Probably not, but sometimes those two components are very intertwined.

joakbjerk:
> Looking great 😊
> 
> Is there any changes or so we have to do in Field.Date based on these this PR? Probably not, but sometimes those two components are very intertwined.

Looking at the tests there should not be, since its just a change of the internal functionality, and not a new prop or brand new logic for the `DatePicker` 🫡

PR #5022, opened 2025-04-28
chore(Field.Selection): support setting `autoComplete` for autocomplete variant

langz:
To be used in https://github.com/dnbexperience/eufemia/pull/5019

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/Selection/Selection.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/Selection/Selection.tsx
@@ -26,7 +26,9 @@ import type { DropdownAllProps } from '../../../../components/Dropdown'
 import { HelpProps } from '../../../../components/help-button/HelpButtonInline'
 import { DrawerListProps } from '../../../../fragments/DrawerList'
 import {
+  AssertNoMissing,
   convertCamelCasePropsToSnakeCase,
+  KeysWithUnderscore,
   ToCamelCase,
 } from '../../../../shared/helpers/withCamelCaseProps'
 import useDataValue from '../../hooks/useDataValue'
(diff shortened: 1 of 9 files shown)


packages/dnb-eufemia/src/extensions/forms/Field/Selection/Selection.tsx
```diff
@@ -311,6 +311,7 @@ function Selection(props: Props) {
                     Object.freeze(autocompleteProps)
                   ) as AutocompleteAllProps)
                 : null)}
+              autoComplete={autocompleteProps?.autoComplete}
```
langz, line 314: I have to do this, even though I would like to just pass the `autoComplete` property through the line above, `(convertCamelCaseProps( Object.freeze(autocompleteProps) ) as AutocompleteAllProps) : null)}`, but that's not possible. As it would change `autoComplete` into `auto-complete` which `Autocomplete` component will not recognize. And I think it's a bad idea to manually add the `auto-complete` property to `Autocomplete` component, as it's default/part of HTML.

This finding also indicates that all other HTML properties which is written in camelCase in React, will not be supported in `Field.Selection`'s and `autocompleteProps` and `dropdownProps` as they will be converted to snake_case.

tujoworker: I' have added a type-safe solution where don't need to add this extra property. Check it out.

tujoworker (approved):

langz (commented):
Nice, thanks 👏 

Looking good, though I hope this is not something we'll eventually use everywhere in our codebase, as I do find it a bit difficult to understand. But as an exception for the code we have today, it looks good 👍 

With this solution, which I think is good enough, If I understand it correctly we'll still have the problem when passing in a native react camelCase property which the component has its own snake_case variant of. Like for instance `onChange` where the Dropdown component has a `on_change`. When using the Dropdown component by itself, one could set both `on_change` and `onChange`, but when using Field.Selection `dropdownProps:{{onChange: somefn}}` it would replace the AutoComplete's `on_change` and not set the `onChange` event.

Which I think is fine, and probably not possible to fix, hence why I don't want this functionality to be used many many places. Converting camelCase props to snake_case by nature can and will potentially lead to issues no matter how it is solved/implemented, I think. As there might exist component properties that has the name of a react html property/even just written in snake_case, potentially. Not saying we have many occurrences of it today.

PR #5033, opened 2025-04-29
fix(Field.Upload): support width

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1745934779449369


packages/dnb-eufemia/src/extensions/forms/Field/Upload/__tests__/__image_snapshots__/upload-have-to-match-width-small.snap.png
langz, line 1: Perhaps we should do some improvements to this width? Doesn't look really good.

packages/dnb-eufemia/src/components/upload/style/dnb-upload.scss
```diff
@@ -20,6 +20,7 @@
 
   position: relative;
   display: flex;
+  flex-basis: 100%;
```
langz, line 23: not sure if `flex-basis` or `width` is best.

langz:
> I wonder if its better to capture only one snapshot for all sizes. We do that in a couple of other fields. Just because then you see very clearly in the snapshot the sizes and what may be off in relation to the others.

Sure, I can refactor it to being so.

So one, example and one snapshot for all the sizes of a given component?

tujoworker:
For readability we set a max width of 60ch to content that should not get too wide.

Like the help button content or statuses, if they are stretched:

Screenshot 2025-04-30 at 11 58 18

So also the Field.Upload component has this width today.

But maybe that's how it should work? Like having a remove button all the way to the right?

Screenshot 2025-04-30 at 11 50 22

langz:
> For readability we set a max width of 60ch to content that should not get too wide.
> 
> Like the help button content or statuses, if they are stretched:
> 
> Screenshot 2025-04-30 at 11 58 18
> So also the Field.Upload component has this width today.
> 
> But maybe that's how it should work? Like having a remove button all the way to the right?
> 
> Screenshot 2025-04-30 at 11 50 22

I see, we can continue to somehow have this, if we could set the inner width, the content inside of the brackets to follow the max width of 60ch or so?

langz:
After rebasing with https://github.com/dnbexperience/eufemia/pull/5039, I do think I'm happy with the result.
Do we think the default width of the Field.Upload should be stretch or large? Many other fields has default width as large.

tujoworker (approved):

langz:
> After rebasing with #5039, and updating screenshot tests [27938d7](https://github.com/dnbexperience/eufemia/commit/27938d7640df8fafea7208f1251ffe0043141339) I do think I'm happy with the result, as it's now possible for it to stretch. Do we think the default width of the Field.Upload should be `stretch` or `large`? Many other fields has default width as large. As of now the default width of Field.Upload is `stretch`.
> 
> We could also consider setting a "internal max-width", as discussed and pictured here [#5033 (comment)](https://github.com/dnbexperience/eufemia/pull/5033#issuecomment-2841512980), but not 100% sure.
> 
> @tujoworker

Never mind, I must have done something strange (some local changes or so), as stretch is not working 😅 
Have not added a fix for it yet, so it's understandable 😄

PR #5037, opened 2025-04-30
fix(Field.Expiry): support width

langz:
Just opened this PR and added these tests as I saw this is the only field that does not support any width prop values like `small`, `medium`, `large` and `stretch`, even though the docs says so https://eufemia.dnb.no/uilib/extensions/forms/feature-fields/Expiry/properties/

Not sure if/how this field can support different widths?
How would it look? 🤔 

TODO:
- [ ] The actual fix


langz:
Replaced with https://github.com/dnbexperience/eufemia/pull/5057 :)

PR #5041, opened 2025-04-30
feat(Forms): always validate fields in all Wizard steps (without keepInDOM)

tujoworker:
More [info](https://dnb-it.slack.com/archives/CMXABCHEY/p1745994632904719).


langz (approved):
A bit difficult for me to reproduce and understand.
Reach out if you want to share the specific use case and/or CSB with me, then I can assist with some testing, etc.

PR #5048, opened 2025-05-02
chore(Portal): bump mdx version

tujoworker:
Here are the release notes to v3 (no breaking changes for use case it looks like): https://github.com/mdx-js/mdx/releases/tag/3.0.0

But we need to look over a good amount of pages to confirm that noting breaks.


langz (approved):
I spent 3 minutes clicking through multiple pages, and during those minutes I did not find any issues or concerns.

PR #5049, opened 2025-05-02
fix(Forms): ensure fields do not overflow in width on a certain screen size

tujoworker:
On a certain screen width, a large FieldBlock will overflow. It "can" and probably will cut over Field.String in the default width.

With the issue:
Screenshot 2025-04-30 at 15 31 44

With this PR, the large size will adjust it to a lower width, so it will not overflow:
Screenshot 2025-04-30 at 16 13 16


langz (approved):
👏

PR #5051, opened 2025-05-02
fix(Forms): ensue the edit container for Iterate and Section shows red border on errors

tujoworker:


joakbjerk (approved):

langz (approved):
We could consider if we would want a screenshot test as well. Just give it a quick evaluation 🙏

tujoworker:
> We could consider if we would want a screenshot test as well. Just give it a quick evaluation 🙏

I added a visual test. 👍

Issue #4886, opened 2025-04-02
DatePicker - not possible to select the input value and paste in a new value


langz:
## 🐛 Bug Report

Motivation: https://dnb-it.slack.com/archives/C02H0BM875H/p1743247188881739

I'm not able to CTRL+A to select values in the DatePicker field, and paste inn a new value, i.e `15.04.2025`.

## To Reproduce

Try to do so in the demo/example: https://eufemia.dnb.no/uilib/components/date-picker/demos/#default-datepicker-with-input

PR #5053, opened 2025-05-02
feat(DatePicker): copy whole formatted date when coping from input field

tujoworker:
Relates to #4886

You can test it here: https://eufemia-git-feat-date-picker-copy-eufemia.vercel.app/uilib/extensions/forms/feature-fields/Date/ by copy&pasting from one field to another.

### DIFF
--- a/packages/dnb-eufemia/src/components/date-picker/DatePickerCalc.ts
+++ b/packages/dnb-eufemia/src/components/date-picker/DatePickerCalc.ts
@@ -251,6 +251,20 @@ export function correctV1Format(date: string) {
   return date
 }
 
+function parseHumanDate(
+  input: string,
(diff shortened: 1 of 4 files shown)


joakbjerk (approved):
Looks good! 👍

Tested in Firefox on desktop, and Brave, Chrome and Edge on android.

Copy works as expected when selecting the field, and the pressing the paste prompt, but the quick paste prompt does not work 🤔

PICTURE WILL COME HERE

tujoworker:
> but the quick paste prompt does not work

We use HTML standards. I don't think we can do anything about that.

joakbjerk:
> > but the quick paste prompt does not work
> 
> We use HTML standards. I don't think we can do anything about that.

Looks good on android then 😄

langz (approved):
🥳

PR #5059, opened 2025-05-05
feat(Icons): add `digital_advise`, `aino`, `lightningstrike`

tujoworker:


langz:
Do we have to remove the aino tag from the chatbot icon?
https://eufemia-git-icon-aino-eufemia.vercel.app/icons/#chatbot

Or doesn't that matter?

tujoworker:
The tags are only for documentation. We get it from Figma. So should be changed there if it matters. 😅

Issue #3241, opened 2024-01-18
ToggleButton: refactor to use correct semantic elements to enhance screen reader UX


tujoworker:
Our ToggleButtons can be used in various variants. And the screen reader support is fine as well. 

- Its read out as a toggle button by e.g. VO (by using `aria-pressed`).
- Also the "state" is read out correctly (by using `aria-checked`).
- And they are accessible with the tab key. Which is debatable if that is wanted or not.

But there are some issues that can be solved with a refactor:

- Semantically, we can't use e.g. `aria-required`, which is only allowed on type="input" elements. So axe tests do fail. Even e.g. VO does read them out correctly.
- When [only one ToggleButton](https://eufemia.dnb.no/uilib/components/toggle-button/demos/#togglebutton-with-status-messages-and-a-group-variant-as-radio) can be selected at a time `type="radio"`, maybe, it should also behave like a [radio button group](https://eufemia.dnb.no/uilib/components/radio/demos/)? So a arrow key would be used to change the value, instead focus and enter/space.

I think, the last point is highly debatable, because, we should probably not change the user input behavior for when the component "looks" the same.

Also, it looks more like a button, so I would argue, that a user wants to use the tab key to access these buttons, instead of a arrow key.

A quick solution, to make `aria-required` be acceptable, is to change:

- `aria-pressed` to `aria-checked`
- and add these three attributes to the button component:

```jsx
<Button
  role="checkbox"
  type="checkbox"
  element="span"
  ...
/>
```

This makes `aria-required` to be accepted by axe. But now VO reads the "toggle button" out as a checkbox. Which is in my opinion not a good move. But again, debatable.

PR #5064, opened 2025-05-06
fix(Forms): support `required` in Field.Selection with `button` variant

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1746520226276669

The fix does so [Field.Selection variant button](https://eufemia.dnb.no/uilib/extensions/forms/base-fields/Selection/demos/#toggle-buttons):
- ToggleButton has `role="radio"`
- ToggleButton.Group has`role="radiogroup"`
- ToggleButton component when role radio sets `aria-checked` instead of `aria-pressed`

I think role radio is correct as one can only select one of these toggle buttons / items, read https://www.w3.org/TR/wai-aria/#radio

Fixes https://github.com/dnbexperience/eufemia/issues/3241

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/Selection/__tests__/Selection.test.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/Selection/__tests__/Selection.test.tsx
@@ -775,8 +775,28 @@ describe('variants', () => {
 
       const buttons = document.querySelectorAll('button')
       expect(buttons.length).toEqual(2)
-      expect(buttons[0].getAttribute('aria-pressed')).toBe('false')
-      expect(buttons[1].getAttribute('aria-pressed')).toBe('false')
+      expect(buttons[0].getAttribute('aria-checked')).toBe('false')
+      expect(buttons[1].getAttribute('aria-checked')).toBe('false')
+    })
+
+    it('has radio roles', () => {
+      render(
+        <Field.Selection variant="button">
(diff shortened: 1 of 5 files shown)


packages/dnb-eufemia/src/extensions/forms/Field/Selection/__tests__/Selection.test.tsx
```diff
@@ -2135,7 +2148,6 @@ describe('variants', () => {
           await axeComponent(result, {
             rules: {
               // Because of aria-controls and aria-required is not allowed on buttons – but VO still reads it
-              'aria-allowed-attr': { enabled: false },
               'aria-valid-attr-value': { enabled: false },
```
langz, line 2151: I think we can improve this in a separate PR

langz: https://github.com/dnbexperience/eufemia/pull/5065

langz:
> I think we can go for it. This closes #3241. We probably need to do the same on `Field.ArraySelection`.
> 
> For VoiceOver and NVDA, the read roles for ToggleButton and Radio buttons are the same now. But how they control the elements are different. We could "potentially" add a `aria-describedby` with a hidden description of it.

Sounds good 🚀 
For `Field.ArraySelection`, we'll add that in https://github.com/dnbexperience/eufemia/pull/5074

PR #5072, opened 2025-05-07
feat(Pagination): add support for horizontal `paginationBarLayout`

tujoworker:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1746527457334889


packages/dnb-eufemia/src/components/pagination/Pagination.js
```diff
@@ -191,6 +193,7 @@ class PaginationInstance extends React.PureComponent {
       children,
       className,
       barSpace,
+      paginationLayout, // eslint-disable-line
```
langz, line 196:
```suggestion
      paginationLayout,
```

packages/dnb-eufemia/src/components/pagination/PaginationDocs.ts
```diff
@@ -6,6 +6,11 @@ export const PaginationProperties: PropertiesTableProps = {
     type: ['pagination', 'infinity'],
     status: 'optional',
   },
+  paginationLayout: {
```
langz, line 9:
```suggestion
  paginationBarLayout: {
```

Should we name it `paginationLayout` or `paginationBarLayout`?
I see that it only changes the layout of the pagination bar, so perhaps we should rename `paginationLayout` to `paginationBarLayout`?

Worth a consideration 📿

langz (approved):

packages/dnb-eufemia/src/components/pagination/PaginationDocs.ts
tujoworker: Yes, that is better – I think too.

PR #5086, opened 2025-05-08
fix(Field.PhoneNumber): select country  code input value on single click

tujoworker:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1743066020148259

https://github.com/user-attachments/assets/5a0d896c-9d64-4acf-8aa7-52203d1a0df2


langz (approved):
I like it :)

Is this only needed for `Field.PhoneNumber`?
Or should this be the default functionality for:
- `Field.SelectCurrency`
- `Field.SelectCountry`
- `Field.Selection` variant `autocomplete`

tujoworker:
Yeah, I think it should be the same there as well.

PR #5087, opened 2025-05-08
fix(Upload): remove file by id or file equality

langz:
Motivation is that when transforming where one make a new File() the files will not be equal, hence the file will not be removed...


packages/dnb-eufemia/src/extensions/forms/Field/Upload/__tests__/Upload.test.tsx
```diff
@@ -1813,6 +1814,87 @@ describe('Field.Upload', () => {
     })
   })
 
+  it('should remove correct file when transformIn and transformOut is used to change the file', async () => {
```
langz, line 1817: I'll just have to double check this test, before merging

tujoworker (approved):

packages/dnb-eufemia/src/extensions/forms/Field/Upload/__tests__/Upload.test.tsx
langz: Yes, looking good now :)

PR #5091, opened 2025-05-08
fix(Value): help button should not wrap alone

langz:
Not completed, I still miss a css attribute, but not sure exactly which one.
Somehow if I change the element of the help button from a `button` to a `span`, it works as expected 🤔


langz:
I've updated the screenshot tests, and it looks to be working fine, except for when Values with help is used in a Value.SummaryList:

image

langz:
I think it looks good :) 
Only concern is this:
`packages/dnb-eufemia/src/extensions/forms/ValueBlock/__tests__/__image_snapshots__/valueblock-for-ui-have-to-match-wrapping.snap.png`:
image

Zooming in:

image

But it's most likely a edge case where one will have such long words? 😄

tujoworker (approved):

PR #5005, opened 2025-04-25
feat(Forms): add `requireCommit` property to Iterate.PushContainer to show error during submit or next step after user input

tujoworker:
- Fixes #4985 and #4072
- PR [Preview](https://eufemia-git-feat-forms-push-container-require-ch-5525ad-eufemia.vercel.app/uilib/extensions/forms/Iterate/PushContainer/demos/#require-commit)

A red border is shown:

Screenshot 2025-04-25 at 08 23 50

also for the filled variant:

Screenshot 2025-04-25 at 08 23 02

PR #5102, opened 2025-05-10
feat(Forms): add `preventUncommittedChanges` in Form.Isolation to show error during submit after user input

tujoworker:
Same functionality as #5005 – so a lot of code were moved.
Builds on top of #5127

Link to deploy preview: https://eufemia-git-feat-forms-isolation-requirecommit-eufemia.vercel.app/uilib/extensions/forms/Form/Isolation/demos/#require-commit


packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Isolation/info.mdx
```diff
@@ -162,3 +163,44 @@ function MyForm() {
...
+        <Field.String path="/isolated" />
+        <Form.Isolation.CommitButton resetAfterCommit />
+        <Form.Isolation.ResetButton showWhenErrorIsPresent />
+      </Form.Isolation>
+    </Form.Handler>
+  )
+}
+```
+
+The `showWhenErrorIsPresent` property ensures that the reset button is displayed only when the "commit required" error is visible. This helps prevent users from resetting the form unnecessarily.
```
langz, line 206: If we think it could come up more alternatives to when one would like to show this button, one could also potentially name the property `showWhen`, and then have different values, like `showWhen="errorIsPresent"`.
Or just `visibleWhen`, as we use this in other places in forms.

packages/dnb-eufemia/src/extensions/forms/Form/Isolation/Isolation.tsx
```diff
@@ -49,6 +52,10 @@ export type IsolationProviderProps<Data extends JsonObject> = {
    * Prevent the form from being submitted when there are fields with errors inside the Form.Isolation.
    */
   bubbleValidation?: boolean
```
langz, line 54: Can the `requireCommit` property have any correlation or collisions with the `bubbleValidation` property, if both are set? Or ca a specific combination of the two lead to some strange UX? 🤔

langz (approved):

packages/dnb-eufemia/src/extensions/forms/Form/Isolation/Isolation.tsx
tujoworker: No, I don't think so. Because `requireCommit` already tells the outer form to not precede. That's the nature of this prop. And `bubbleValidation` only effects each single field, and tells them, that errors should reach the outer form, instead of the isolated one. While the isolated date still stays isolated.

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Isolation/info.mdx
tujoworker: Yeah, I think that's a good idea. Because this prop is ugly long.

langz: Does it exist any other occurrences of naming like this in our codebase? If so, I would vote for that option🤣👍

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Isolation/info.mdx
```diff
@@ -182,11 +183,11 @@ function MyForm() {
 }
 ```
 
-### Define the reset data
+### Define your own data reference (`dataReference`)
 
-Technically, when you use `resetAfterCommit`, the `Form.Isolation` will reset its data context based on the "initial" internal data set.
+Technically, when you use `requireCommit` or `resetAfterCommit`, the `Form.Isolation` will use its "initial" internal data set to crate a reference. This reference is used to either compare if there is a change or to reset the data context after a commit.
```
langz, line 188:
```suggestion
Technically, when you use `requireCommit` or `resetAfterCommit`, the `Form.Isolation` will use its "initial" internal data set to create a reference. This reference is used to either compare if there is a change or to reset the data context after a commit.
```

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Isolation/info.mdx
```diff
@@ -182,11 +183,11 @@ function MyForm() {
...
 
-But in some situations, you may need a different data set than the internal data set given at the initial render.
+But in some situations, you may need a different data set than the initial data set given at the initial render.
```
langz, line 190: Perhaps we could provide an example of a situation like this? I don't fully see it 😅

langz:
`requireCommit`, to me it could sound like one is required to make a commit, or so, which is not the case. 
I feel like the prop name could be more descriptive in regards to preventing dirty/unsaved changes.
Also using the word require makes me think about the html required attribute/property.

Potential alternatives:
- `preventUnsavedChanges`
- `requireChangesToBeCommited`
- 

Not sure if I want to change it, as we already use it in Iterate.PushContainer, etc.
We don't need to change this as part of this PR or this release, but could be worth thinking about.

langz:
`resetAfterCommit`, what does it reset? Should we state that it resets the values of the isolated form?

Potential alternatives:
- `resetValuesAfterCommit`
- `resetInputsAfterCommit`
- `resetFieldsAfterCommit`

packages/dnb-eufemia/src/extensions/forms/Form/Isolation/Isolation.tsx
```diff
@@ -56,6 +57,10 @@ export type IsolationProviderProps<Data extends JsonObject> = {
...
+  /**
+   * If the container should be committed before the form is submitted.
```
langz, line 61:
```suggestion
   * If the isolation containing changes should be committed before the form is submitted.
```

packages/dnb-eufemia/src/extensions/forms/Form/Isolation/IsolationDocs.ts
```diff
@@ -25,6 +25,11 @@ export const IsolationProperties: PropertiesTableProps = {
     type: 'boolean',
     status: 'optional',
   },
+  requireCommit: {
+    doc: 'If the container should be committed before the form is submitted.',
```
langz, line 29:
```suggestion
    doc: 'If the isolation containing changes should be committed before the form is submitted.',
```

langz (approved):
Functionality-wise, I think this works great :) 

One possible improvement I see is https://github.com/dnbexperience/eufemia/pull/5102#issuecomment-2886063784

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Isolation/demos.mdx
```diff
@@ -23,8 +23,16 @@ import * as Examples from './Examples'
...
 
-When no `defaultValue` is set on the Form.Isolation (inner context), the the default value from Form.Handler (outer context) is used for the initial value.
+When no `defaultValue` is set on the Form.Isolation (inner context), the default value from Form.Handler (outer context) is used for the initial value.
 
-When pressing "Legg til" the default value from Form.Isolation is inserted again, because `resetDataAfterCommit` is set to `true`.
+When pressing the "Legg til / Add"-button, the default value from Form.Isolation is inserted again, because `resetDataAfterCommit` is set to `true`.
 
 <Examples.InsideSection />
+
+### Require commit
```
langz, line 32:
```suggestion
### Prevent uncommited changes
```

langz:
> `resetAfterCommit`
> Also, could it be that one would only want to reset a given field of the `Form.Isolation`? How would one do that? Is this something we should support?

What do you think @tujoworker ?

tujoworker:
> Could it be that one would only want to reset a given field of the Form.Isolation?

Yes, but then it probably is easier to do it imperative/manually. 

> How would one do that? 

I think using the [snapshot functionality](https://eufemia.dnb.no/uilib/extensions/forms/Form/useSnapshot/) could be interesting to use here.

Is this something we should support?

Yes, but we may add such a feature at a later stage. 

> Or should we only support resetting all fields/children of the Form.Isolation?

I would argue, that this is the most common way a form would work.

langz (approved):
I gave it a round of testing now, and it seems to work great 😊 👏

PR #5106, opened 2025-05-12
fix(Forms): support `required` in Field.Selection with `dropdown` variant

langz:
TODO:
- [x] Extensive testing using VoiceOver and NVDA

### DIFF
--- a/packages/dnb-eufemia/src/components/dropdown/Dropdown.js
+++ b/packages/dnb-eufemia/src/components/dropdown/Dropdown.js
@@ -241,7 +241,7 @@ export default class Dropdown extends React.PureComponent {
         {...this.props}
         id={id}
         data={data || children}
-        opened={null}
+        opened={false}
         tagName="dnb-dropdown"
         ignore_events={false}
         prevent_selection={
@@ -641,6 +641,8 @@ class DropdownInstance extends React.PureComponent {
                       </span>
                     </>
                   }
+                  role="combobox"
+                  title={title}
                   {...triggerParams}
                 />
               )}
--- a/packages/dnb-eufemia/src/components/dropdown/__tests__/Dropdown.test.tsx
+++ b/packages/dnb-eufemia/src/components/dropdown/__tests__/Dropdown.test.tsx
@@ -75,6 +75,12 @@ describe('Dropdown component', () => {
     ).toBe(
       (mockData[props.value] as DrawerListDataArrayObject).selected_value
     )
+    expect(
+      document.querySelector('.dnb-dropdown__trigger')
+    ).toHaveAttribute(
+      'title',
+      (mockData[props.value] as DrawerListDataArrayObject).selected_value
+    )
 
     keydown(32) // space
 
@@ -98,6 +104,13 @@ describe('Dropdown component', () => {
     expect(
       document.querySelector('.dnb-dropdown__text__inner').textContent
     ).toBe(
+      (mockData[(props.value as number) + 1] as DrawerListDataArrayObject)
+        .selected_value
+    )
+    expect(
+      document.querySelector('.dnb-dropdown__trigger')
+    ).toHaveAttribute(
+      'title',
       (mockData[(props.value as number) + 1] as DrawerListDataArrayObject)
         .selected_value
     )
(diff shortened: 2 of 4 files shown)


packages/dnb-eufemia/src/components/dropdown/Dropdown.js
```diff
@@ -241,7 +241,7 @@ export default class Dropdown extends React.PureComponent {
         {...this.props}
         id={id}
         data={data || children}
-        opened={null}
+        opened={false}
```
langz, line 244: Setting this to false instead of null will lead to `aria-expanded="false"` on initial render, instead of how it is today, where the `aria-expanded` attribute is not rendered.

packages/dnb-eufemia/src/components/dropdown/Dropdown.js
```diff
@@ -639,6 +639,8 @@ class DropdownInstance extends React.PureComponent {
                       </span>
                     </>
                   }
+                  role="combobox"
+                  title={title}
```
langz, line 643: After enabling the ARIA tests again, I would get this error, hence adding title to the dropdown trigger button:
```
    "Buttons must have discernible text (button-name)"

    Fix any of the following:
      Element does not have inner text that is visible to screen readers
      aria-label attribute does not exist or is empty
      aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty
      Element has no title attribute
      Element's default semantics were not overridden with role="none" or role="presentation"
```

tujoworker:
We can't use `role="combobox"` on a Dropdown, because e.g. VoiceOver reads out that the user can "type". So it can be used for an "Autocomplete".

Screenshot 2025-05-14 at 10 38 28

langz:
> We can't use `role="combobox"` on a Dropdown, because e.g. VoiceOver reads out that the user can "type". So it can be used for an "Autocomplete".
> 
> Screenshot 2025-05-14 at 10 38 28

Okok, one can sort of type a bit in a Dropdown, when dropdown is active one can type one letter and the hover/selected item will change.

I see that there's different libraries and examples where dropdowns use the combobox role.

https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/ and https://mui.com/material-ui/react-select/

tujoworker:
We do already use `role="listbox"` in the list that opens.

`combobox` describes that it should be combined with an `input` and a list to select something from. We should not use it in the Dropdown.

If aria-required is the problem, we can also think of making an exception, and not set it there. But rather add an `aria-describedby="error-msg"`, where the error-msg describes the "requirement".

langz:
> We do already use `role="listbox"` in the list that opens.
> 
> `combobox` describes that it should be combined with an `input` and a list to select something from. We should not use it in the Dropdown.
> 
> If aria-required is the problem, we can also think of making an exception, and not set it there. But rather add an `aria-describedby="error-msg"`, where the error-msg describes the "requirement".

_The combobox input field can either be a single-line text field that supports editing and typing, similar to a HTML [<input>](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input) with a [<datalist>](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/datalist), or an element that only displays the current value of the combobox._ - https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/combobox_role#description

tujoworker:
Right. https://www.w3.org/TR/wai-aria-1.2/#combobox also says something similar:

> or an element with role="presentation" that allows the user to select a value from a collection of possible values.

And her we have a coded example: https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/

So I guess, even VoiceOver does read out loud that the user can type, we can go for it.

tujoworker (approved):

langz:
> Right. https://www.w3.org/TR/wai-aria-1.2/#combobox also says something similar:
> 
> > or an element with role="presentation" that allows the user to select a value from a collection of possible values.
> 
> And her we have a coded example: https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/
> 
> So I guess, even VoiceOver does read out loud that the user can type, we can go for it.

Great, thanks for taking the time to test 🔉 and look into the docs 📖

PR #5109, opened 2025-05-12
fix(Forms): ensure support for `labelSrOnly` in Field.Toggle

langz:
As Field.Boolean uses Field.Toggle under the hood, this fixes it in Both Field.Toggle and Field.Boolean


tujoworker (approved):
🤩

PR #5120, opened 2025-05-14
fix(Field.Selection): apply correct `width` to prevent label wrapping

tujoworker:
Motivation: https://dnb-it.slack.com/archives/C07JEJVSHJR/p1747208014972039?thread_ts=1747140976.337829&cid=C07JEJVSHJR


langz (approved):
Looking good 🥇

PR #5123, opened 2025-05-14
fix(Drawer): containerPlacement on small screen

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1747044916958009


joakbjerk (approved):
The solution and screen shot tests looks good to me 🙏

PR #5124, opened 2025-05-14
fix(Dropdown): fix `React.ReactNode` support for `title` prop

joakbjerk:
fixes this [issue](https://dnb-it.slack.com/archives/CMXABCHEY/p1747217419048679)

### DIFF
--- a/packages/dnb-eufemia/src/components/dropdown/Dropdown.js
+++ b/packages/dnb-eufemia/src/components/dropdown/Dropdown.js
@@ -483,7 +483,10 @@ class DropdownInstance extends React.PureComponent {
 
     const handleAsMenu =
       isTrue(action_menu) || isTrue(more_menu) || isTrue(prevent_selection)
-    const isPopupMenu = isTrue(more_menu) || !(_title && _title.length > 0)
+
+    const title = this.getTitle(_title)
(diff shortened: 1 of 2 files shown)


packages/dnb-eufemia/src/components/dropdown/Dropdown.js
```diff
@@ -483,7 +483,10 @@ class DropdownInstance extends React.PureComponent {
 
     const handleAsMenu =
       isTrue(action_menu) || isTrue(more_menu) || isTrue(prevent_selection)
-    const isPopupMenu = isTrue(more_menu) || !(_title && _title.length > 0)
```
langz, line 486: Hmm, makes me think why one would want/need to check the length of the title as well ? 🤔

langz (approved):

packages/dnb-eufemia/src/components/dropdown/Dropdown.js
joakbjerk: The only scenario I can think of, would be to see if it's an empty string or not, for when the title is of string type. But the `!title` snippet would achieve the same result since and empty string is falsy 🤔

tujoworker (approved):

PR #5127, opened 2025-05-14
feat(Forms): add `resetAfterCommit` to Form.Isolation

tujoworker:
Builds on top of #5126


langz:
Is this new property only useable and available on Form.Isolation.CommitButton?
Or on other commitbuttons as well?
Maybe the only commitbutton is Form.Isolation.CommitButton? I've not checked. Hehe

langz (approved):

langz:
I ran it locally, as the deploy preview did not work, and it doesn't seem to be working when submitting/pressing enter. It does not reset.

https://github.com/user-attachments/assets/fdea61fd-a9ac-4686-a8c1-3b12f96ff36f

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Isolation/info.mdx
```diff
@@ -162,3 +162,22 @@ function MyForm() {
   )
 }
 ```
+
+## Clear data after commit
+
+You can clear the isolation the user committed by using `resetAfterCommit`:
```
langz, line 168:
```suggestion
You can reset the isolation the user committed by using `resetAfterCommit`:
```

langz:
When trying out the default data example, I notice if I manually clear/empty the input field, then click the commit button, the value is not commited which is understandable as it's empty and its required to have a value, however the input of the isolated field is reset to the default data. I don't feel like this behavior match the name of the prop resetAfterCommit, as I would perhaps expect the input to only be reset after a "successful commit" - not everytime pressing the commit button 🤔 


https://github.com/user-attachments/assets/e6b9eafb-1d8c-4ffc-af22-c6e438bc5079



Also in this scenario, the animation/error message is a bit strange, as it almost animates in before it gets removed. This is is not important, just a tiny edge-case.

tujoworker:
I moved the property away from the button and up to the `Form.Isolation`. Also, pressing enter "should" work now. I also added a way to refresh the initial snapshot of the data set, which is used when resetting the isolated data context.

langz:
> When trying out the default data example, I notice if I manually clear/empty the input field, then click the commit button, the value is not commited which is understandable as it's empty and its required to have a value, however the input of the isolated field is reset to the default data. I don't feel like this behavior match the name of the prop resetAfterCommit, as I would perhaps expect the input to only be reset after a "successful commit" - not everytime pressing the commit button 🤔 
> 
> 
> 
> 
> 
> https://github.com/user-attachments/assets/e6b9eafb-1d8c-4ffc-af22-c6e438bc5079
> 
> 
> 
> 
> 
> 
> 
> Also in this scenario, the animation/error message is a bit strange, as it almost animates in before it gets removed. This is is not important, just a tiny edge-case.

I still experience this:

https://github.com/user-attachments/assets/19797f53-2776-4d83-8bc2-3b1f5d2e959b

tujoworker:
> When pressing the "Legg til"-button in this scenario, I think I would expect to see the error message that it's required and that the input would still stay empty and not being reset to the default value.

This example uses `defaultData`. It is used to "reset" after commit. I will try to write a better message. Or should we leave it empty maybe? Then we get what you described.

Screenshot 2025-05-15 at 16 24 35

**Update:** I updated the text of the example, because it also makes sense like is today, when thinking of the default value handling:


> This example has a `defaultValue` both on the Form.Handler and the Form.Isolation.
>
> When no `defaultValue` is set on the Form.Isolation (inner context), the the default value from Form.Handler (outer context) is used for the initial value.
>
> When pressing "Legg til" the default value from Form.Isolation is inserted again, because `resetAfterCommit` is set to `true`.
 
 Does that work for you?

langz:
> > > When pressing the "Legg til"-button in this scenario, I think I would expect to see the error message that it's required and that the input would still stay empty and not being reset to the default value.
> > 
> > 
> > 
> > This example uses `defaultData`. It is used to "reset" after commit. I will try to write a better message. Or should we leave it empty maybe? Then we get what you described.
> > 
> > 
> > 
> > Screenshot 2025-05-15 at 16 24 35
> > 
> > 
> 
> Okay, I'm not sure if I get my message across clearly. 
> 
> But I do find the interaction strange, based on the naming of the prop resetAfterCommit. In the scenario I've recorded in two previous messages, why does it reset? My input (empty) was not committed. I would not expect the input field to be reset from empty to default data when there was no change committed. I would just expect to see a required error message or so. 
> 
> Try for yourself:
> Remove all data in the "Isolated"-field, then press "legg til". 
> 
> https://github.com/user-attachments/assets/ff098002-b7ad-4f8d-959c-7c54b8dbc683
> 
> https://eufemia-git-feat-forms-isolation-commit-reset-af-d56531-eufemia.vercel.app/uilib/extensions/forms/Form/Isolation/demos/#inside-a-section
> 
> Do you find this logical? That my empty value gets replaced with the default value, while my empty value is not committed?

But we can improve on this in the feature as well, and release what we have now. Perhaps it's easier to show what I mean in person.

tujoworker:
> My input (empty) was not committed. I would not expect the input field to be reset from empty to default data when there was no change committed. I would just expect to see a required error message or so.

Now I did understand it. Thank you 🤩

I have added a test&fix and think that it should work as expected now.

langz (approved):

PR #5130, opened 2025-05-15
chore(Form.useSnapshot): add safeguard when using hook

tujoworker:
It looks like #5129 did introduce a issue during portal build.

Screenshot 2025-05-15 at 10 56 38


langz (approved):
Great, thanks :) 

Just noticed it on main yeah 👍

PR #5140, opened 2025-05-16
chore(Form.Isolation): errorIsPresent -> uncommitedChangePrevented

langz:
My motivation is that when reading the following code:
```
    <Form.Handler>
      <Form.Isolation preventUncommitedChanges resetDataAfterCommit>
        <Field.String path="/isolated" />
        <Form.Isolation.CommitButton />
        <Form.Isolation.ResetButton showWhen="errorIsPresent" />
      </Form.Isolation>
    </Form.Handler>
```

I don't fully understand that there's a relation between`<Form.Isolation.ResetButton showWhen="errorIsPresent" />` and  `<Form.Isolation preventUncommitedChanges>`. Further, I feel like ` <Form.Isolation.ResetButton showWhen="errorIsPresent" />` is very generic. If the reset button should be showed when error is present, why does it not show if the required error is present, but only when the uncommited changes error is displayed?

From:
`<Form.Isolation.ResetButton showWhen="errorIsPresent" />`

To:
`<Form.Isolation.ResetButton showWhen="uncommitedChangePrevented" />`

Now looking like:
```
    <Form.Handler>
      <Form.Isolation preventUncommitedChanges resetDataAfterCommit>
        <Field.String path="/isolated" />
        <Form.Isolation.CommitButton />
        <Form.Isolation.ResetButton showWhen="uncommitedChangePrevented" />
      </Form.Isolation>
    </Form.Handler>
```

I'm open to other naming suggestions that could help solve the issues described above.

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Isolation/Examples.tsx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Isolation/Examples.tsx
@@ -253,7 +253,7 @@ export const PreventUncommitedChanges = () => {
 
               <Flex.Horizontal>
                 <Form.Isolation.CommitButton />
-                <Form.Isolation.ResetButton showWhen="errorIsPresent" />
+                <Form.Isolation.ResetButton showWhen="uncommitedChangePrevented" />
               </Flex.Horizontal>
             </Flex.Stack>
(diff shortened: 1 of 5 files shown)


tujoworker (approved):
I very much prefer `uncommittedChangeDetected`.

Note: "uncommitted" is written with two tt.

PR #5148, opened 2025-05-19
fix(useTranslation): ensure correct TypeScript typing support

tujoworker:


joakbjerk (approved):
Looks good to me 🔥

packages/dnb-eufemia/src/shared/useTranslation.tsx
```diff
@@ -31,7 +32,9 @@ export default function useTranslation<T = Translation>(
   const { locale, translation } = useContext(Context)
   const { assignUtils } = useAdditionalUtils()
 
-  return useMemo(() => {
+  return useMemo<
+    TranslationFlatToObject<T> & AdditionalReturnUtils
+  >(() => {
     const id = typeof messages === 'string' ? messages : undefined
     if (id) {
       return formatMessage(id, args, translation)
```
andlbrei, line 40: @tujoworker just wanted to let you know that using like `useTranslation("My.ID")`, with an ID as input, results in a bad TS return type. Probably because `formatMessage` has `any` as return type, and the explicit return type for `useMemo` here does not include `string`.  
We are migrating whatever usage we have of this to the other way of using `useTranslation()`, since we already do that for our Forms components, so no need to fix it for our sake.  
Should this be deprecated perhaps?

PR #5149, opened 2025-05-19
chore(DatePicker): use translations as default values for props

joakbjerk:
Makes the following props use the translation values as default values, instead of having the values be hard coded in several different places throughout the `DatePicker`
- `maskOrder`
- `maskPlaceholder`
- `dateFormat`
- `returnFormat`
- `submitButtonText`
- `cancelButtonText`
- `resetButtonText`
- `firstDay`

Relies on https://github.com/dnbexperience/eufemia/pull/5148 to prevent type errors in `DatePickerInput`


langz (approved):
Good good 👍 

One could argue that it's a bit more difficult to figure out what the default values for props in DatePicker. As some properties is listed in defaultProps with its default value, and other props has its default values listed else where in the code.

tujoworker (approved):

PR #5152, opened 2025-05-19
chore(DatePicker): rename `onChange` for `DatePickerRange`  to `onPickerChange`

joakbjerk:
To prevent confusion since it's not the same as the `DatePicker` `onChange`


langz (approved):
Looking good 👍 

I assume this is only internal.
If not, I would expect changes to tests, deprecating existing `onChange` prop, etc, etc 😅

PR #5155, opened 2025-05-19
fix(Field.Indeterminate): add support for `required` property

langz:
Deploy preview: https://eufemia-git-chore-add-axe-tests-to-fields-eufemia.vercel.app/uilib/extensions/forms/base-fields/Indeterminate/demos/

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/Indeterminate/__tests__/Indeterminate.test.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/Indeterminate/__tests__/Indeterminate.test.tsx
@@ -831,4 +831,88 @@ describe('Indeterminate', () => {
       )
     })
   })
+
+  it('should support required property', async () => {
+    const onSubmit = jest.fn()
+
+    render(
+      <Form.Handler onSubmit={onSubmit}>
+        <Field.Indeterminate
+          dependencePaths={['/child1', '/child2']}
+          propagateIndeterminateState="auto"
+          valueOn="what-ever"
+          valueOff="you-name-it"
+          path="/parent"
+          required
+        />
+        <Field.Boolean path="/child1" />
+        <Field.Toggle
+          path="/child2"
+          valueOn="custom-on"
+          valueOff="custom-off"
+        />
+      </Form.Handler>
+    )
+
+    const form = document.querySelector('form')
+    const [parent, child1] = Array.from(document.querySelectorAll('input'))
+
+    fireEvent.submit(form)
+
+    expect(onSubmit).toHaveBeenCalledTimes(0)
+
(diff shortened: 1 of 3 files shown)


packages/dnb-eufemia/src/extensions/forms/Field/Indeterminate/__tests__/Indeterminate.test.tsx
```diff
@@ -831,4 +832,99 @@ describe('Indeterminate', () => {
...
+      )
+
+      const input = document.querySelector('input')
+      expect(input).toHaveAttribute('aria-required', 'true')
+    })
+
+    it('should have aria-invalid', () => {
+      render(
+        <Form.Handler>
+          <Field.Indeterminate
```
langz, line 901: Reason for this test failing is something with Field.Indeterminate passing `value={false}` to `Field.Toggle`.
Not sure why though...

langz:
The problem/challenge here is that the following code will not display a required error message:
```
        <Field.Indeterminate
          dependencePaths={['/checkbox1', '/checkbox2', '/checkbox3']}
          label="Indeterminate"
          required
          validateInitially
        />
```

Reason for this is because the value of a Field.Indeterminate is always set to `valueOn` (defaults to `true`) or `valueOff` (defaults to `false`).
See here for the logic: https://github.com/dnbexperience/eufemia/blob/91947b110a0b3a55580a5c1339cb5907b6aa0d7a/packages/dnb-eufemia/src/extensions/forms/Field/Indeterminate/Indeterminate.tsx#L43-L48

And since the value is either `true` or `false`, it indirectly means that the user has provided data/input.
The required error message is only displayed if the user has not provided data/input, meaning the value is `undefined`. But as stated above, the value of `Field.Indeterminate` can never be `undefined`, hence never being able to display/show the required error message.

Not sure how/if we should fix this 🤔 

@tujoworker

packages/dnb-eufemia/src/extensions/forms/Field/Indeterminate/__tests__/Indeterminate.test.tsx
```diff
@@ -831,4 +831,88 @@ describe('Indeterminate', () => {
...
+
+    expect(onSubmit).toHaveBeenCalledTimes(0)
+
+    await userEvent.click(parent)
+    fireEvent.submit(form)
+
+    expect(onSubmit).toHaveBeenCalledTimes(1)
+  })
+
+  it('should change internal required state when indeterminate state changes', async () => {
```
tujoworker, line 877: In this we change the state from "outside". So we only modify it on the children. But since the state of the "Field.Indeterminate" then changes, it resolves the required validation. I think that this is correct. Therefore this test. Could it be named better?

PR #5157, opened 2025-05-19
feat(Card, Form.Card, Wizard, StepIndicator): controlling card outset

snorrekim:
* Fix Wizard step container background overflowing on small screens
* Add `<Card.Provider>` that can emulate a `<Card>` parent context


packages/dnb-eufemia/src/extensions/forms/Form/Card/__tests__/Card.test.tsx
```diff
@@ -70,4 +73,80 @@ describe('Form.Card', () => {
       )
     }
   })
+  describe('CardContext', () => {
```
tujoworker, line 76:
```suggestion
  })
  
  describe('CardContext', () => {
```

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Wizard/Container/demos.mdx
```diff
@@ -49,3 +49,15 @@ const onSubmitRequest: OnSubmitRequest = ({ getErrors }) => {
...
+
+### Outset and layout (Card.Provider)
+
+Like `<Form.Card>`, the container for the steps will be "outset" from the regular padding unless the wizard is
+inside a `<Card>` (of `<Form.Card>`) component.
+
+This can be disabled by wrapping the area in a `<Card.Provider hasParentCard>` (or `<Form.Card.Provider hasParentCard>`) to tell all children
+to behave as if they are inside a `<Card>`. Or by setting `outset={false}`.
+
+See more details in the [Form.Card "Outset" example](/uilib/extensions/forms/Form/Card/#outset) .
```
tujoworker, line 61:
```suggestion
The wizard navigation area ([StepIndicator](/uilib/components/step-indicator/)) will "outset" (break out) from the layout using negative margins.

If the `Wizard.Container` is placed inside a parent `<Card>` or another wrapper that messes up with the layout, you can turn it off in two ways:

– Wrap the Wizard.Container in `<Form.Card.Provider hasParentCard>` to make all nested cards act like they’re inside a `<Card>`.
– Or set `outset={false}` on each card.

See the [Form.Card "Outset" example](/uilib/extensions/forms/Form/Card/#outset) for more details.
```

packages/dnb-eufemia/src/components/card/Card.tsx
```diff
@@ -118,11 +118,37 @@ function Card(props: Props) {
...
+   *
+   * Defaults to existing context value.
+   */
+  hasParentCard,
+  children,
+}: {
+  hasParentCard?: boolean
+  children: React.ReactNode
+}) {
+  const context = useContext(CardContext) ?? {}
```
tujoworker, line 145:
```suggestion
  const context = useContext(CardContext) || {}
```

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Card/info.mdx
```diff
@@ -7,5 +7,7 @@ hideInMenu: true
 
 `Form.Card` is a wrapper for the [Card](/uilib/components/card/) component to make it easier to use inside a form.
 
-- It will set `outset` to `true` by default.
-- It will set `stack` to `true` by default.
+- It will set `outset` to `true` by default (instead of `false`). [See example below.](#outset)
+- It will set `stack` to `true` by default (instead of `false`).
+
+But it otherwise identical to the `<Card>` component.
```
tujoworker, line 13:
```suggestion
But in all other ways, it works just like the [Card](/uilib/components/card/) component.
```

packages/dnb-eufemia/src/extensions/forms/Form/Card/CardDocs.ts
```diff
@@ -1,14 +1,15 @@
...
-    ...CardProperties.outset,
-    doc: 'Same as `outset` in [Card](/uilib/components/card/properties). Defaults to `true`.',
+    doc: 'Whether or not to break out (using negative margins) on larger screens. Same as `outset` in [Card](/uilib/components/card/properties). But defaults to `true`',
+    type: 'boolean',
+    status: 'optional',
   },
   stack: {
-    ...CardProperties.stack,
-    doc: 'Same as `stack` in [Card](/uilib/components/card/properties). Defaults to `true`.',
+    doc: 'True to stack the sub components with lines between. The `spacing` will default to `medium`. Same as `stack` in [Card](/uilib/components/card/properties). But defaults to `true`',
```
tujoworker, line 10:
```suggestion
    doc: 'True to stack the sub components with space between. The `spacing` will default to `medium`. Same as `stack` in [Card](/uilib/components/card/properties). But defaults to `true`',
```

packages/dnb-eufemia/src/components/card/Card.tsx
```diff
@@ -118,11 +118,37 @@ function Card(props: Props) {
...
+ * Can be used to tell children to act as if they are in a Card, even if they are not.
+ */
+Card.Provider = function CardProvider({
+  /**
+   * Defines if Card components should act as if it is nested inside another Card.
+   * Used to control side-margins outset behaviour.
+   *
+   * Defaults to existing context value.
+   */
+  hasParentCard,
```
tujoworker, line 139:
```suggestion
  suppressOutsetOfNestedCards,
```

I'd like to find a declarative name, so that a developer reading the prop can roughly understand what it does to the layout.

Any suggestions are welcome.

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Card/demos.mdx
```diff
@@ -9,6 +9,16 @@ import * as Examples from './Examples'
 
 <Examples.BasicUsage />
 
+## Outset
+
+`<Form.Card>` sets `outset` by default to perserve content alignement in forms.
+If for some reason this breaks layout `outset={false}` can be used to position
+the `<Form.Card>` as a regular `<Card>` component.
+
+Alternatively, the wrapper `<Form.Card.Provider hasParentCard>` can be used to have children behave as if they are inside a `<Card>` component.
```
langz, line 18: I don't 100% understand the use case for this 🤔 
Why would I want to use _the wrapper `<Form.Card.Provider hasParentCard>` to have children behave as if they are inside a `<Card>` component_? Couldn't I rather just write the code instead and get the same result?:
```
  <Card>
    <Form.Card>
      <P>
        Told to behave as if inside a parent card. Will stay inside the
        content area of the parent even on small screens.
      </P>
    </Form.Card>
  </Card>
```

tujoworker: You could. But then the first card draws a outline. Which in such cases was provided by a "fake" (non-Eufemia) card or wrapper of some sort that gets in the way "visually", because of the negative margin we add when we `outset`.

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Wizard/Container/demos.mdx
snorrekim: Fixed, but with a small change, instead of 

"If the `Wizard.Container` is placed inside a parent `<Card>` or another wrapper that messes up with the layout, you can turn it off in two ways:"

it's:

"This outset is turned off if the `Wizard.Container` is placed inside a `<Card>`, but if placed in a different wrapper that messes with the layout, you can manually turn it off in two ways:"

packages/dnb-eufemia/src/components/card/Card.tsx
snorrekim: I agree. The issue is that we aren't just disabling outset, we are also disabling responsive breakout. So it might be a bit misleading. You won't get the same behaviour as just setting `outset={false}` you also need `responsive={false}` 

Suggestions:

- forceCardsToRespectPadding
- disableOutsetAndBreakoutOfCards
- containCardsToContentArea

tujoworker: The `responsive={false}` again will ensure it does not "break" out. So they both are within a mechanism that "breaks-out" from the layout. So could we maybe then use the term "breakout" instead of outset?

- `suppressBreakoutOfNestedCards`

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Card/demos.mdx
langz: Okay, so you are saying that the` <Form.Card.Provider hasParentCard>` does more than _to have children behave as if they are inside a `<Card>` component_ (which is what the docs says it does)?

langz:
Looking good, we could consider adding an entry for this in the changelog before releasing. 

https://github.com/dnbexperience/eufemia/pull/5165

packages/dnb-eufemia/src/extensions/forms/Form/Card/__tests__/Card.test.tsx
```diff
@@ -70,4 +73,81 @@ describe('Form.Card', () => {
       )
     }
   })
+
+  describe('CardContext', () => {
```
tujoworker, line 77:
```suggestion
  describe('Card.Provider', () => {
```

packages/dnb-eufemia/src/extensions/forms/Form/Card/__tests__/Card.test.tsx
```diff
@@ -70,4 +73,81 @@ describe('Form.Card', () => {
...
+    const contextRef: React.MutableRefObject<CardContextState> =
+      React.createRef()
+
+    const ContextConsumer = () => {
+      contextRef.current = useContext(CardContext)
+      return null
+    }
```
tujoworker, line 84:
```suggestion
    }

```

PR #5158, opened 2025-05-19
feat(Iterate.RemoveButton): add support for `itemNo`

tujoworker:
```tsx
<Iterate.Array value={['foo', 'bar']}>
  <RemoveButton text="Remove item {itemNo}" />
</Iterate.Array>
```


langz (approved):
🫡

PR #5162, opened 2025-05-20
fix(Forms): ensure Iterate.Array with `required` prop removes error in Wizard step

### LINKED ISSUE #5159: "Sticky" validation error from `Iterate.PushContainer`
## 🐛 Bug Report

Validation error message is shown after navigating to new page after having triggered a validation error in a `Iterate.PushContainer` with `preventUncommittedChanges`.  
It seems as if the validation error message is stuck after filling all the missing fields. Ajv does not report any validation errors from the component.  

## To Reproduce

https://codesandbox.io/p/sandbox/gracious-cookies-324nps

## Expected behavior

I expect no validation error message to be shown when the content is valid.

## Eufemia Version

10.74.0

tujoworker:
Fixes #5159


langz (approved):
🚀

PR #5166, opened 2025-05-20
fix(DatePicker): show month label when `onlyMonth` is `true`

joakbjerk:


langz:
Nice look 😎🕶️
Is this a feature or a fix?
I can see that the property is already documented, etc.

tujoworker (approved):

packages/dnb-eufemia/src/components/date-picker/DatePickerCalendar.tsx
```diff
@@ -491,26 +491,51 @@ function DatePickerCalendar(restOfProps: DatePickerCalendarProps) {
...
-            id={id}
-            nr={nr}
-            date={month}
-            locale={locale}
-          />
-          {yearNavigation && (
+      <div
+        className={classnames(
+          'dnb-date-picker__header',
+          hideNav && 'dnb-date-picker__header--only-month'
```
langz, line 497: If I read the code correctly. 

This class which is named something only month is the parent of possible children who is not displaying only month, perhaps we should find a different name for the class?

langz, line 497: I also don't understand why this class is appended when hideNav, when the DatePickerCalender of potential type="month" is only displayed when !hideNav and it additionally have logic for deciding if type is month. Shouldn't we use the same logic both for the classname and the condition of which the datepickercalender with type month is rendered?

langz (approved):

packages/dnb-eufemia/src/components/date-picker/DatePickerCalendar.tsx
joakbjerk: Refactored the display logic, and renamed the css classname, hope this makes more sense 😄

PR #5167, opened 2025-05-21
feat(DateFormat): add date formatting component

joakbjerk:
### TODO

- [x] Add docs
- [x] Screenshot tests
- [x] Add tests
- [x] Add story
- [x] Add provider support
- [x] Add `Space` support
- [x] Add `skeleton` support
- [x] Make dates screen reader friendly
- [x] Add aria tests
- [x] Add base functionality
- [x] Move `formatDate` to `DateFormatUtils`
- [x] Add inline demo example
- [x] Remove all non `dateStyle` examples
- [x] Remove non `dateStyle` props
- [x] Add invalid date safeguard
- [x] Update invalid date safeguard to show given invalid value

### Features to add later

- Relative time
- time support (hour, second minute)
- Range


packages/dnb-eufemia/src/components/date-format/DateFormatUtils.ts
```diff
@@ -0,0 +1,50 @@
+import { AnyLocale } from '../../shared/Context'
+import { LOCALE as defaultLocale } from '../../shared/defaults'
+import { convertStringToDate } from '../date-picker/DatePickerCalc'
+import { DateType } from '../date-picker/DatePickerContext'
+
+export type FormatDateOptions = {
+  locale?: AnyLocale
+  formatOptions?: Intl.DateTimeFormatOptions
+  /**
+   * @deprecated use `options={{dateStyle}}` instead.
```
langz, line 10: How come a new component has deprecated props? 🤔

langz:
Cool cool, I'll give this a test tomorrow (Friday).
Hopefully the deploy preview is building/running correctly by then 🤞 
If not, perhaps we can just add a dummy commit, so that it hopefully rebuilds.

Is there any places internally in our codebase where we should use this component as well? 🤔

joakbjerk:
> Is there any places internally in our codebase where we should use this new `DateFormat` component? 🤔

Maybe `Value.Date`, swap out the `formatDate` function with the actual new component? 🤔

packages/dnb-eufemia/src/components/date-format/DateFormatDocs.ts
```diff
@@ -0,0 +1,29 @@
+import { PropertiesTableProps } from '../../shared/types'
+
+export const DateFormatProperties: PropertiesTableProps = {
+  date: {
```
langz, line 4: Should we call it `date` or `value`?
In the following components we call it `value`:
- https://eufemia.dnb.no/uilib/components/number-format/properties/
- https://eufemia.dnb.no/uilib/components/list-format/properties/

packages/dnb-eufemia/src/components/date-format/DateFormatDocs.ts
```diff
@@ -0,0 +1,29 @@
...
+    type: ['Date', 'string'],
+    status: 'optional',
+  },
+  dateStyle: {
+    doc: 'Defines the formatting  used for dates (weekday, day, month year). Cannot be used together with `weekday`, `day`, `month` and `year`. Defaults to `long`.',
+    type: ['long', 'medium', 'short', 'full'],
+    status: 'optional',
+  },
+  locale: {
+    doc: 'Locale used for formatting. Defaults to `nb-NO`',
```
langz, line 15:
```suggestion
    doc: 'Locale used for formatting. Defaults to `nb-NO`.',
```

langz:
It would be awesome if we could "safeguard" the component from "non valid" values, instead of breaking/errors, like these:
Screenshot 2025-05-30 at 10 29 57
Screenshot 2025-05-30 at 10 29 43
Screenshot 2025-05-30 at 10 29 27
Screenshot 2025-05-30 at 10 29 20
Screenshot 2025-05-30 at 10 29 13
Screenshot 2025-05-30 at 10 28 58

langz:
Do we support a "duration string"? Or perhaps that's not necessary, or expected 🤔 
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/time

langz:
Should we document and expose the `formatDate` function as well?
Could there be use cases where one would want to use that only?

Both in NumberFormat and ListFormat, the "underlying" functions are exposed:
- https://eufemia.dnb.no/uilib/components/number-format/#formatting-only-interceptor
- https://eufemia.dnb.no/uilib/components/list-format/#formatting-only

joakbjerk:
> Do we support a "duration string"? Or perhaps that's not necessary, or expected 🤔 I just saw that it's supported in the time element/tag. https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/time

We were thinking of adding it later as it's own feature, to keep this pr mvp 👍

joakbjerk:
> Should we document and expose the `formatDate` function as well? Could there be use cases where one would want to use that only?
> 
> Both in NumberFormat and ListFormat, the "underlying" functions are exposed:
> 
>     * https://eufemia.dnb.no/uilib/components/number-format/#formatting-only-interceptor
> 
>     * https://eufemia.dnb.no/uilib/components/list-format/#formatting-only

Yeah we could do that. I know of at least one instance of a developer using the `formatDate` function

packages/dnb-eufemia/src/components/date-format/DateFormatDocs.ts
```diff
@@ -0,0 +1,29 @@
...
+    type: ['Date', 'string'],
+    status: 'optional',
+  },
+  dateStyle: {
+    doc: 'Defines the formatting  used for dates (weekday, day, month year). Cannot be used together with `weekday`, `day`, `month` and `year`. Defaults to `long`.',
+    type: ['long', 'medium', 'short', 'full'],
+    status: 'optional',
+  },
+  locale: {
+    doc: 'Locale used for formatting. Defaults to `nb-NO`.',
```
langz, line 15: Could potentially be nice to provide a link to all locales supported.
Assuming it's some kind of web standard, probably this? https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/lang

packages/dnb-eufemia/src/components/date-format/stories/__tests__/DateFormat.test.tsx
```diff
@@ -0,0 +1,177 @@
...
+    rerender(<DateFormat value="2025-08-01" dateStyle="long" />)
+    expect(dateFormat).toHaveTextContent('1. august 2025')
+
+    rerender(<DateFormat value="2025-08-01" dateStyle="medium" />)
+    expect(dateFormat).toHaveTextContent('1. aug. 2025')
+
+    rerender(<DateFormat value="2025-08-01" dateStyle="short" />)
+    expect(dateFormat).toHaveTextContent('01.08.2025')
+  })
+  describe('date value formats', () => {
```
tujoworker, line 26:
```suggestion
  })

  describe('date value formats', () => {
```

tujoworker (approved):
Godt jobbet 👏

langz (approved):
Great great 👍🥳
Thanks for looking into my comments. Have a great weekend!

packages/dnb-design-system-portal/src/docs/uilib/components/date-format.mdx
```diff
@@ -0,0 +1,18 @@
+---
+title: 'DateFormat'
+description: 'A ready to use DNB date formatter.'
+showTabs: true
```
langz, line 4:
```suggestion
showTabs: true
status: 'new'
```

So that we'll get the new icon:

Screenshot 2025-06-02 at 10 48 47

langz:
Before merging, let's just quickly discuss with @tujoworker if we should release v10.76.0 or v10.75.3 first 🙏

langz:
> It would be awesome if we could "safeguard" the component from "non valid" values as `children` or as `value`, instead of breaking/errors, like these: Screenshot 2025-05-30 at 10 29 57 Screenshot 2025-05-30 at 10 29 43 Screenshot 2025-05-30 at 10 29 27 Screenshot 2025-05-30 at 10 29 20 Screenshot 2025-05-30 at 10 29 13 Screenshot 2025-05-30 at 10 28 58
> 
> I guess we should just either return what was sent in, or return an empty string or so? It's safest to return what was sent in I believe.

In regards to this, is it best/correct to return invalid date when the date is invalid, or to return the value the user has passed in?

image

joakbjerk:
> > It would be awesome if we could "safeguard" the component from "non valid" values as `children` or as `value`, instead of breaking/errors, like these: Screenshot 2025-05-30 at 10 29 57 Screenshot 2025-05-30 at 10 29 43 Screenshot 2025-05-30 at 10 29 27 Screenshot 2025-05-30 at 10 29 20 Screenshot 2025-05-30 at 10 29 13 Screenshot 2025-05-30 at 10 28 58
> > I guess we should just either return what was sent in, or return an empty string or so? It's safest to return what was sent in I believe.
> 
> In regards to this, is it best/correct to return `invalid date` when the date is invalid, or to return the value the user has passed in? I would perhaps think it's best to just return the value the user has passed in 🤔 My reason/thought is just that we now change the value from x to` Invalid date` which doesn't make any sense for the end-user. I guess for him it would provide more meaning to see `x`, even though it's not a valid date?
> image

The main issue with doing that is that the component will return nothing if the value is not a valid `ReactNode`, so it would be hard to pickup on there being an issue, I think perhaps 😄 Not sure what the best way to go ahead to solve this woulde be 🤔 
Screenshot 2025-06-04 at 11 01 18

Issue #5052, opened 2025-05-02
Support Icon in `selected_value` in a Dropdown/DrawerList


tujoworker:
<img width="398" alt="Image" src="https://github.com/user-attachments/assets/6b91bfe7-551f-44b7-9f00-89b76daa2748" />

This results in: Invalid hook call. Hooks can only be called inside of the body of a function component. 

CSB reprod: https://codesandbox.io/p/sandbox/late-voice-4d72cz?file=/src/App.tsx:21,47

PR #5172, opened 2025-05-21
fix(Dropdown): support Icon in `selected_value` 

langz:
fixes https://github.com/dnbexperience/eufemia/issues/5052


packages/dnb-eufemia/src/fragments/drawer-list/DrawerListHelpers.ts
```diff
@@ -252,6 +253,7 @@ export function parseContentTitle(
         convertJsxToString(dataItem.selected_value, separator, (word) => {
           const nestedChildren =
             !word.props.children &&
+            word?.type !== Icon &&
```
langz, line 256: This only fixes it for Icon.

So if passing in any other components or custom components that use hooks, the same error as reported in the issue will occur. 

Might exist a better fix/solution that will work for all components that use hooks here?🫠

PR #5212, opened 2025-05-28
fix: ensure `line-height-lead` has correct value when consumed via JavaScirpt in DNB Eiendom

tujoworker:


tujoworker:
No, it's not a change on the line-height, but only the auto generated file, that gets its variables from SCSS.

PR #5224, opened 2025-06-02
fix(Accordion): fix Accordion toggle behavior when not used inside an `Accordion.Group`

joakbjerk:
Fixes [this issue](https://dnb-asa.atlassian.net/jira/software/c/projects/EDS/boards/734?assignee=712020%3A11a274b8-3b2a-4c7f-9968-07cf788ff7dd&selectedIssue=EDS-789) by consolidating the `expandBehaviour` and `expandBehavior` and adding a default value to the `expandBehavior` used in context, for grouped accordions used outside of an `Accordion.Group`

## Testing

[Deploy Preview](https://eufemia-git-fix-accordion-unable-to-collapse-group-bug-eufemia.vercel.app/uilib/components/accordion/#customized-accordion)

[Current Docs](https://eufemia.dnb.no/uilib/components/accordion/#customized-accordion)

### DIFF
--- a/packages/dnb-eufemia/src/components/accordion/Accordion.tsx
+++ b/packages/dnb-eufemia/src/components/accordion/Accordion.tsx
@@ -461,7 +461,8 @@ export type GroupProps = AccordionProps & {
 }
 
 const Group = ({
-  expandBehaviour = undefined,
+  // expandBehaviour can be removed in v11
+  expandBehaviour,
   expandBehavior = 'single',
   ...props
 }: GroupProps) => {
@@ -532,6 +533,7 @@ const Group = ({
       onInit={onInit}
       {...props}
       group={group}
+      // expandBehaviour can be removed in v11
       expandBehavior={expandBehaviour || expandBehavior}
       expanded_id={expandedId || props.expanded_id}
     />
--- a/packages/dnb-eufemia/src/components/accordion/Accordion.tsx
+++ b/packages/dnb-eufemia/src/components/accordion/Accordion.tsx
@@ -461,7 +461,7 @@ export type GroupProps = AccordionProps & {
 }
 
 const Group = ({
-  expandBehaviour = 'single',
+  expandBehaviour = undefined,
   expandBehavior = 'single',
   ...props
 }: GroupProps) => {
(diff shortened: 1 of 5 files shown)


packages/dnb-eufemia/src/components/accordion/Accordion.tsx
```diff
@@ -461,7 +461,7 @@ export type GroupProps = AccordionProps & {
 }
 
 const Group = ({
-  expandBehaviour = 'single',
+  expandBehaviour = undefined,
   expandBehavior = 'single',
```
langz, line 465: Do we need to set this to `undefined` as well? Or is it only for the old deprecated `expandBehaviour` prop?

packages/dnb-eufemia/src/components/accordion/AccordionProviderContext.ts
```diff
@@ -17,14 +17,14 @@ type AccordionGroupContextProps = {
   onInit?: (...args: any[]) => any
   collapseAccordionCallbacks?: React.MutableRefObject<(() => void)[]>
   collapseAllHandleRef?: React.MutableRefObject<() => void>
-  /**
-   * @deprecated – Replaced with expandBehavior, expandBehaviour can be removed in v11
-   */
-  expandBehaviour?: AccordionGroupProps['expandBehaviour']
```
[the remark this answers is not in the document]
joakbjerk: Yeah, I think you are correct 🤔

langz (approved):

packages/dnb-eufemia/src/components/accordion/Accordion.tsx
joakbjerk: Only the old `expandBehaviour`, it should be enoughtto just remove `expandBehaviour` in v11, but I'll double check if can actually remove the default prop in this pr. I don't think that should break any functionality tied to the old `expandBehaviour`

joakbjerk: Seems like we have to have the `expandBehaviour` destructured at top of the function, to prevent the prop from being added as a HTML attribute 😅 but removing the `undefined` default value worked. It's also a little redundant to assign `undedfined` as default value, since it's automatically `undefined` when not defined 😝

PR #5232, opened 2025-06-04
fix(Field.Selection & Field.ArraySelection): remove error border when info or warning message text

langz:
This fix improves the following scenarios:

image
image


packages/dnb-eufemia/src/extensions/forms/hooks/useFieldProps.ts
```diff
@@ -2683,6 +2683,7 @@ export function checkForError(
   >
 ) {
   return potentialErrors.some((error) => {
+    console.log('checkForError', error)
```
langz, line 2686: Added this log, to be able to look at what's causing the error from the portal, or for others as well

langz: Removed again now

tujoworker (approved):

PR #5233, opened 2025-06-05
chore: align versions of dependencies across the workspace

tujoworker:
Motivation: We should always strive to align the same version across the two packages.


langz (approved):
🫶

PR #5234, opened 2025-06-05
chore: add `Eufemia.versions` support

tujoworker:
Also, it should with that be possible to get a short sha. We may document that as a feature if we see that it works.

Here is a test setup: https://stackblitz.com/edit/dsuwq5hy

Screenshot 2025-06-05 at 14 00 40


langz (approved):
🥬

PR #5236, opened 2025-06-05
chore(StepIndicator): remove duplicated StepIndicator tests

tujoworker:
From what I can see, is that these tests do not test a different look. They look like leftovers from v1.


tujoworker:
Alright 🫡

PR #5237, opened 2025-06-05
fix(Autocomplete): empty string `input_value` should clear the input value

langz:


packages/dnb-eufemia/src/components/autocomplete/Autocomplete.js
```diff
@@ -404,7 +404,7 @@ class AutocompleteInstance extends React.PureComponent {
 
       if (
         props.input_value !== 'initval' &&
-        props.input_value?.length > 0
+        (props.input_value?.length > 0 || props.input_value === '')
```
tujoworker, line 407: Do we even need a check now?

tujoworker (approved):

PR #5245, opened 2025-06-10
fix(Field.Selection): hover state for radio-list variant

langz:
Not sure if we need the z-indexes set?
Let's see if the screenshot tests fail 😅


tujoworker:
Also, tested it quickly with touch. And as expected – its fine too.

PR #5253, opened 2025-06-12
feat(Breadcrumbs): add `onToggle` property

joakbjerk:
Solves [this ticket](https://dnb-asa.atlassian.net/jira/software/c/projects/EDS/boards/734?assignee=712020%3A11a274b8-3b2a-4c7f-9968-07cf788ff7dd&selectedIssue=EDS-737)

### TODO

- [x] Add `onToggle` prop
- [x] Add tests
- [x] Add `isCollapsed` property to callback

### DIFF
--- a/packages/dnb-eufemia/src/components/breadcrumb/BreadcrumbDocs.ts
+++ b/packages/dnb-eufemia/src/components/breadcrumb/BreadcrumbDocs.ts
@@ -85,7 +85,7 @@ export const BreadcrumbEvents: PropertiesTableProps = {
     status: 'optional',
   },
   onToggleCollapse: {
-    doc: 'Will be called when breadcrumb collapse is toggled.',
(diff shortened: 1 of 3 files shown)


packages/dnb-eufemia/src/components/breadcrumb/BreadcrumbDocs.ts
```diff
@@ -84,6 +84,11 @@ export const BreadcrumbEvents: PropertiesTableProps = {
     type: 'function',
     status: 'optional',
   },
+  onToggleCollapse: {
```
langz, line 87: Does it exist a better word, that means both toggle and collapse? Or should we rather make one event for onToggleCollapse and one for onToggleExpand?

As when just naming it onToggleCollapse, Inwould not expect it to fire on expand, only on collapse.

langz (approved):

PR #5260, opened 2025-06-16
feat(Field.Upload): support `children`

langz:
Builds on top of https://github.com/dnbexperience/eufemia/pull/5259


tujoworker (approved):
Twice as Nice! But it makes we wonder; do we need a visual tests for "children" usage? (just because they are time-consuming to run on the CI).

PR #5266, opened 2025-06-17
fix(Field.PhoneNumber): truncate to 8 digits when changing country code to NO

langz:


packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber/PhoneNumber.tsx
```diff
@@ -93,6 +93,11 @@ const defaultMask = [
   /\d/,
 ]
 
+const numberOfDigitsInDefaultMask = defaultMask.filter(
```
langz, line 96: I could perhaps just use the value `8`, instead of all of this code.
But it makes it a bit more dynamic, so that one potentially could change the `[defaultMask](https://github.com/dnbexperience/eufemia/pull/5266/files#diff-38c29b6761161bce42ae3132ad81661b225c7b82739dcb4b447dc12b5e37d422L82)` (not that I assume we ever will).

tujoworker (approved):

PR #5267, opened 2025-06-18
feat: add style isolation support

tujoworker:
This PR provides a way to isolate its styles, so you can use different Eufemia versions along side each other on the same page.

**Links**
- The docs: https://eufemia-git-feat-style-isolation-custom-scope-eufemia.vercel.app/uilib/usage/customisation/styling/style-isolation/
- Test case with two Eufemia versions with one bundler: https://stackblitz.com/edit/dsuwq5hy-miib48ne
- Similar test case where we set a "custom scope" for both Eufemia versions and apps: https://stackblitz.com/edit/dsuwq5hy-zfetefrp
- Similar test case where we nested styles in each other: https://stackblitz.com/edit/dsuwq5hy-nzftxjwh
- CSB: https://codesandbox.io/p/devbox/blissful-river-j8mtc8
- POC file: https://www.figma.com/board/gBy4hNcxXOKWxwj6DOQYOd/Style-Isolation

## How it works

The isolated styles are in files with the suffix `--isolated.min.css`. All selectors inside these files are scoped with a prefix e.g. `eufemia-scope--1_2_3`, which in practice means that the original selector `.dnb-button` will be `.eufemia-scope--1_2_3 .dnb-button`.

```html
<div class="eufemia-scope--1_2_3">
  <!-- App 1 using Eufemia v1.2.3 -->
  <button class="dnb-button">Eufemia 1.2.3</button>
</div>

<div class="eufemia-scope--2_1_0">
  <!-- App 2 using Eufemia v2.1.0 -->
  <button class="dnb-button">Eufemia 2.1.0</button>
</div>
```

In order to apply the styles, you need to wrap your app with the `IsolatedStyleScope` component.

```tsx
import '@dnb/eufemia/style/isolated'
import { IsolatedStyleScope } from '@dnb/eufemia/shared'

function MyApp() {
  return <IsolatedStyleScope>Your app content</IsolatedStyleScope>
}
```

Optionally, but recommended – add the PostCSS plugin to your build tool.


packages/dnb-design-system-portal/src/docs/uilib/usage/customisation/styling/style-isolation.mdx
```diff
@@ -0,0 +1,279 @@
...
+  return <IsolatedStyleScope>Your app content</IsolatedStyleScope>
+}
+```
+
+Optionally (but recommended), add the PostCSS plugin to your build tool. This ensures your custom styles have the correct CSS specificity when overriding or extending Eufemia styles.
+
+### Good to know
+
+- CSS `:root` selectors are rewritten to target the scoped container (e.g., .eufemia-scope--1_2_3).
+- While a `html` selector are left untouched and continues to target the global `<html>` element.
```
langz, line 40: I did not quite understand this sentence, and doesn't feel like it "fit in"/match the one above.

packages/dnb-design-system-portal/src/docs/uilib/usage/customisation/styling/style-isolation.mdx
```diff
@@ -0,0 +1,279 @@
...
+
+**Notes:**
+
+- Remove existing `.dnb-core-style` classes.
+- Font files are loaded from the CDN (Read more [about hosted fonts](/uilib/typography/#hosted-fonts-cdn)), so they are shared between Eufemia versions.
+
+### 3. Add the PostCSS plugin
+
+1. Find and add a [PostCSS extension or loader](https://github.com/postcss/postcss#usage) for your build tool.
+2. Then Create a `postcss.config.js` file in your project root:
```
langz, line 78:
```suggestion
2. Then create a `postcss.config.js` file in your project root:
```

packages/dnb-design-system-portal/src/docs/uilib/usage/customisation/styling/style-isolation.mdx
```diff
@@ -0,0 +1,279 @@
...
+```
+
+### Get the current scope hash
+
+So you can use `getStyleScopeHash` to get the scope hash for the current Eufemia version.
+
+Via the `IsolatedStyleScope` component:
+
+```tsx
+import { getStyleScopeHash } from '@dnb/eufemia/shared/IsolatedStyleScope.js'
```
langz, line 200:
```suggestion
import { getStyleScopeHash } from '@dnb/eufemia/shared/IsolatedStyleScope'
```

Or perhaps it's correct to use the extension as well?

packages/dnb-eufemia/package.json
```diff
@@ -25,6 +25,7 @@
     "build:copy": "./scripts/postbuild/copy-build-artifacts.sh",
     "build:es": "./scripts/postbuild/babel-es.sh",
     "build:esm": "./scripts/postbuild/babel-esm.sh",
+    "build:lebab": "babel-node --extensions .ts ./src/plugins/postcss-isolated-style-scope/scripts/transformToESM",
```
langz, line 28: lebab using babel ? 🤔  👍 
Aren't those opposites? 🤔

langz (approved):

langz:
What happens if one nest isolated styles and `IsolatedStyleScope`? I assume that will work as expected? 🤔

packages/dnb-eufemia/package.json
tujoworker: yes 👍 – but in that case the transform script (transformToESM), which uses "lebab", is written in TypeScript – that's why we need something other than Node.js v18 to run it. Would we use the newest Node.js, we could run it without a transformer.

tujoworker:
> I'm not sure how the apps will do it, but there's probably a possibility that they will nest as well🤣😅

If they are from two different Eufemia versions, they do not share the same context. Then they will be nested:

Screenshot 2025-06-20 at 11 26 09

If they nest within the same version, then I also think we should add support for it too. Added a new commit for that now.

packages/dnb-design-system-portal/src/docs/uilib/usage/customisation/styling/style-isolation.mdx
```diff
@@ -0,0 +1,307 @@
...
+}
+
+render(
+  <IsolatedStyleScope>
+    <Component />
+  </IsolatedStyleScope>,
+)
+```
+
+Optionally, you can provide a different scope hash to the hook if you need to retrieve an element from a nested scope:
```
kimroen, line 147: This is the first time we mention "scope hash" in the documentation (besides `scopeHash` in the configuration example).

Can we define what that is? What is hashed about it?

tujoworker: Thank you Kim for the feedback. I agree and will update the description.

PR #5283, opened 2025-06-24
chore(DateFormat): remove logs from test

langz:


joakbjerk:
I'm probably missing misunderstanding or missing something here 😄 but it seems like the console.log is added? 🤔

PR #5300, opened 2025-06-25
fix(Field.PostalCodeAndCity): `postalCode` and `city` should override existing props

langz:


packages/dnb-eufemia/src/extensions/forms/Field/PostalCodeAndCity/PostalCodeAndCity.tsx
```diff
@@ -148,11 +147,11 @@ function PostalCodeAndCity(props: Props) {
         inputMode="numeric"
         autoComplete="postal-code"
         data-country-code={country || countryCode}
+        {...postalCode}
```
langz, line 150: This might be breaking, I've not analysed the code too much, let's see what the test says.

PR #5319, opened 2025-07-01
docs: add GitHub and Figma links to component docs

langz:


packages/dnb-design-system-portal/src/shared/tags/TabBar.tsx
```diff
@@ -109,6 +113,8 @@ export default function TabBar({
           {title}
         </AutoLinkHeader>
       )}
+      {figmaUrl && <Link href={figmaUrl}>Figma</Link>}
+      {gitHubUrl && <Link href={gitHubUrl}>GitHub</Link>}
```
langz, line 117: TODO: Use Anchor instead of Link

packages/dnb-design-system-portal/src/shared/tags/TabBar.tsx
```diff
@@ -109,6 +113,8 @@ export default function TabBar({
           {title}
         </AutoLinkHeader>
       )}
+      {figmaUrl && <Link href={figmaUrl}>Figma</Link>}
```
langz, line 116: Add margin right.

langz:
Looking great🤗

I have one thought, as we say in the "relevant urls" say docs on GitHub, I also think we should write code on GitHub. But I feel like that's a bit unnecessarily long. 

Perhaps the section could be named something else? Or just the actual links 🤔

PR #5326, opened 2025-07-15
feat(DateFormat): add support for relative time (auto updated)

joakbjerk:
👉 PR Preview: https://feat-dateformat-relative-tim.eufemia-e25.pages.dev/uilib/components/date-format/demos/

Originally it was speculated that we could use a [duration string](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/time#valid_datetime_values) value for the `datetime` attribute. But from testing it seems like VoiceOver does not pick up on, or reads the duration string at all.

I've tried setting it up like in [this example](https://www.jasha.eu/blogposts/2019/05/voiceover-safari-time-element.html), but the VoiceOver ignores the duration string, and reads the sentence as "It takes 47 meters to travel to the Hauge" in a broke norwegian accent (the system language i tested) instead of "It takes 47 minutter to travel to the Hauge" as the example claims it should.

But I feel like this is a non issue, since the [Intl.RelativeTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat) API can take a date object and turn it into a readable sentence for screen readers anyways. So i propose dropping the duration string support for the purpose of this feature.

And which relative time units should we support? As of now I've added support for all of them 

- seconds
- minutes
- hours
- days
- weeks
- months
- years

I feel like weeks might not be needed, as 27 days ago, would be easier to read and comprehend, than 3 weeks ago 🤔

## TODO

- [x] Decide which units to support
- [x] FIx days to month calculation, to take the different month lengths into consideration.

### DIFF
--- a/packages/dnb-design-system-portal/src/core/ChangeLocale.tsx
+++ b/packages/dnb-design-system-portal/src/core/ChangeLocale.tsx
@@ -26,6 +26,12 @@ export default function ChangeLocale({ listUSLocale = null, ...props }) {
         .filter((key) => {
           return !listUSLocale && key === 'en-US' ? false : true
         })
+        .sort((a, b) => {
+          // Sort based on the order defined in languageDisplayNames
+          const aIndex = Object.keys(languageDisplayNames).indexOf(a)
+          const bIndex = Object.keys(languageDisplayNames).indexOf(b)
+          return aIndex - bIndex
+        })
         .map((key) => {
           const title = languageDisplayNames[key].label
           return <Field.Option key={key} value={key} title={title} />
--- a/packages/dnb-design-system-portal/src/docs/uilib/components/date-format/Examples.tsx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/components/date-format/Examples.tsx
@@ -57,9 +57,21 @@ export const RelativeTime = () => {
   return (
     <ComponentBox>
       <P>
+        <strong>Basic relative time:</strong>
+        <br />
         Skrevet{' '}
         <DateFormat
           value={new Date(new Date().getTime() - 30 * 1000)}
+          relativeTime
+        />
+        <br />
+        <DateFormat
+          value={new Date(new Date().getTime() - 2 * 60 * 1000)}
+          relativeTime
+        />
+        <br />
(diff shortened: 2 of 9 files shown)


tujoworker (approved):

langz:
How do we know that the users of relative time (devs) want's to write "for x siden":
image

I assume it could be phrased so many different ways:
`Sist innlogging: 35 sekunder siden.`
`For 30 sekunder siden logget du inn.` In this example the F is upper-cased, component would today only return lowercase f.
`30 sekunder siden forrige innlogging.`

langz:
A sidenote, should the `DateFormat` support a time duration? Like in the example [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/time#try_it).
See [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/time#valid_datetime_values) under _Valid duration string_.

langz (approved):

langz:
Which other props should be supported when using `relativeTime`?
The following props did not work in combination with relativeTime:
- space props like `top`, `bottom`, etc.
- `dateStyle`
image

tujoworker:
> How do we know that the users of relative time (devs) want's to write "for x siden": image
> 
> I assume it could be phrased so many different ways:
> 
>     * `Sist innlogging: 35 sekunder siden.`
> 
>     * `30 sekunder siden forrige innlogging.`
> 
>     * `For 30 sekunder siden logget du inn.` In this example the `F` is upper-cased, component would today only return a lowercased `f.`

Intl.RelativeTimeFormat is designed to always provide complete, grammatically correct phrases in the target language. It doesn't allow you to get just the time value without the contextual words.

tujoworker:
> Which other props should be supported when using `relativeTime`? The following props did not work in combination with `relativeTime`:
> 
>     * space props like `top`, `bottom`, etc.
> 
>     * `dateStyle`

They should work now.

langz (approved):
Great work 📅📆

tujoworker:
Could be. But we add such features later when needed. ☺️👌

PR #5333, opened 2025-07-28
fix(StyleIsolation): change font base path

langz:


tujoworker:
No, we still move these files as before. 
The change only affects when the plugin is used. 

But the tests need to be adjusted. You can test it locally as well.

tujoworker:
Also, check if the skeleton and Sbanken fonts are hosted.

tolgaz:
We are only hosting fonts listed in https://eufemia.dnb.no/uilib/typography/#hosted-fonts-cdn in our bucket. Should we host all fonts in https://github.com/dnbexperience/eufemia/tree/main/packages/dnb-eufemia/assets/fonts ?

tujoworker:
> Should we host all fonts in https://github.com/dnbexperience/eufemia/tree/main/packages/dnb-eufemia/assets/fonts ?

Yes, that would be good 🙏

thebreiflabb:
Given the latest official browser support, would it be enough to contain this to only the woff2 formats, as long as all the different fonts are included?
https://caniuse.com/woff2

packages/dnb-eufemia/src/plugins/postcss-font-url-rewrite/__tests__/font-url-rewrite-plugin.test.ts
```diff
@@ -254,14 +254,6 @@ describe('font-url-rewrite-plugin', () => {
         new RegExp(`^${basePathRegex}dnb/DNB-Bold\\.(woff2|woff)$`)
       )
     }
-
-    // Check CDN accessibility for each URL and fail if any are not accessible
-    for (const fontUrl of urls) {
```
langz, line 259: I removed this test, as does it make sense for us to test "external" urls, that we can't control?
As of right now, it did fail when visiting: https://assets.dnb.no/shared/fonts/1.0.0/dnb/DNB-Bold.woff

tujoworker: I would like to keep these tests, as this is a way to check if the link is actually valid and working.

tujoworker: We can remove checking the `woff` file. Now the failed test should go green.

PR #5334, opened 2025-07-28
fix: skip-link styling

langz:
fixes: https://dnb-it.slack.com/archives/CMXABCHEY/p1753708180948019

I'm not entirely sure how/what the fix here should be, seems like the [following change](https://github.com/dnbexperience/eufemia/pull/5306/files#diff-ef3774a725c2e6dad3648a609dfe7ba96fad8c9d1d36e362575ff55f24a8eebc) in this [PR](https://github.com/dnbexperience/eufemia/pull/5306) makes it fail in release `v10.77.1`.

Could there be a bigger issue with the `[skip-isolation]` strategy?

To experience the issue, I made a CSB using [v10.77.1](https://codesandbox.io/p/devbox/keen-thunder-kt4f4l?workspaceId=ws_LbiDwJdEBdbwxH2AkyAyVE) and one using [v10.77.0](https://codesandbox.io/p/devbox/inspiring-hooks-24vw4t?workspaceId=ws_LbiDwJdEBdbwxH2AkyAyVE).

@tujoworker

### DIFF
--- a/packages/dnb-eufemia/src/style/core/helper-classes/skip-link.scss
+++ b/packages/dnb-eufemia/src/style/core/helper-classes/skip-link.scss
@@ -94,17 +94,17 @@
 // How to use the "skip-link"
 // 1. Have this on the top of your app: <a class="dnb-skip-link dnb-button" href="#dnb-app-content">Skip to content</a>
 // 2. Place this id="dnb-app-content" on your main content container
-[skip-isolation] .dnb-skip-link {
+.dnb-skip-link {
   position: absolute;
   z-index: 0;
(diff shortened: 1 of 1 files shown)


langz:
> Omg, we should go back to :not(#isolated) 😱 by rewriting the plugin. Will look into it asap.
> 
> Can you add a visual test? Though we had that already.

I can look into it.
It's however not failing in the portal, https://eufemia.dnb.no/uilib/usage/accessibility/focus/#skip-link

tujoworker:
Yeah, true. It does not fail because it has the plugin 😄

PR #5343, opened 2025-07-29
fix(Table): allow expanding by chevron when text is selected

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1753777131358149

By clicking the "button" which is not a button, but just an icon in the most left cell, we should always be able to expand or close the accordion.

### DIFF
--- a/packages/dnb-eufemia/src/shared/helpers.js
+++ b/packages/dnb-eufemia/src/shared/helpers.js
@@ -266,6 +266,17 @@ export function getSelectedText() {
   return ''
 }
 
+export function emptySelectedText() {
+  try {
+    if (window.getSelection && window.getSelection().empty) {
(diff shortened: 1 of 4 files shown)


packages/dnb-eufemia/src/shared/helpers.js
```diff
@@ -266,6 +266,17 @@ export function getSelectedText() {
...
 
+export function emptySelectedText() {
+  try {
+    if (window.getSelection && window.getSelection().empty) {
+      window.getSelection().empty()
+    }
+  } catch (e) {
+    //
+  }
+  return ''
```
tujoworker, line 277:
```suggestion
```

tujoworker (approved):

PR #5345, opened 2025-07-29
fix(GlobalStatus): color contrasts of icon and background in success state

langz:
Motivation for this PR is that I believe the color contrast between the background color (green) and the color (green) of the information icon is not WCAG compliant.

Not sure if we want to use the check icon or not.
That's my initial best solution for improving the color contrast in the globalstatus with state success.

Visual change:
image

I could also change to using the  regular [information](https://eufemia.dnb.no/icons/#information) icon?
What do we think?


tujoworker:
Decorative content is exempt from contrast requirements. 

WCAG 2.1 (SC 1.4.11 – Non‑text Contrast): This criterion requires a minimum 3:1 contrast ratio for meaningful graphical UI components and graphical objects required for understanding. However, purely decorative graphics (like presentation/decorative icons) that carry no information are explicitly exempt.

PR #5351, opened 2025-07-29
fix(NumberFormat): display dash `-` for values `null`, `undefined`, and `''`

langz:
Motivation: https://dnb.enterprise.slack.com/archives/CMXABCHEY/p1753791730435249


tujoworker (approved):
👏

Issue #5347, opened 2025-07-29
`PushContainer` open after uploading document


andlbrei:
## 🐛 Bug Report

**This is an issue in production**

A weird one, but one of several bugs we have where `PushContainers` are opened by themselves.  
After uploading a document on one page of the form, a `PushContainer` in a previous page opens, and there are validation errors because of it.  

(This issue has previously been DM'ed to @tujoworker )

## To Reproduce

[![Demo](https://dnb.enterprise.slack.com/files/U05UFPH6258/F0904SEL0SE/screen_recording_2025-06-05_at_10.18.49.mov)](https://dnb.enterprise.slack.com/files/U05UFPH6258/F0904SEL0SE/screen_recording_2025-06-05_at_10.18.49.mov)

<video src="https://dnb.enterprise.slack.com/files/U05UFPH6258/F0904SEL0SE/screen_recording_2025-06-05_at_10.18.49.mov" width="320" height="240" controls></video>

https://stackblitz.com/edit/eufemia-starter-ppmo9om1?file=src%2FApp.tsx

## Expected behavior
`PushContainers` are only open after being opened by a user, and do not open by themselves.  

## Eufemia Version
10.77.4

PR #5363, opened 2025-08-05
fix(Wizard): should not report a change when comparing files/objects

langz:
Closes https://github.com/dnbexperience/eufemia/issues/5347

[Original reprod from issue.](https://stackblitz.com/edit/eufemia-starter-ppmo9om1?file=src%2FApp.tsx)
[Reprod using this PR.](https://stackblitz.com/edit/eufemia-starter-8gqukxux?file=package.json)

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Form/Isolation/useHasContentChanged.ts
+++ b/packages/dnb-eufemia/src/extensions/forms/Form/Isolation/useHasContentChanged.ts
@@ -50,7 +50,7 @@ function hasValueChanged(a, b) {
       return true
     }
 
-    // Check fo deep-equality of the values
+    // Check for deep-equality of the values
     return keysA.some((k) => hasValueChanged(a[k], b[k]))
   }
 


packages/dnb-eufemia/src/extensions/forms/Form/Isolation/useHasContentChanged.ts
```diff
@@ -34,3 +37,15 @@ export default function useHasContentChanged({
...
 }
+
+function isValuesDifferent(value1, value2) {
+  if (
+    typeof value1 === 'object' &&
+    value1 !== null &&
+    typeof value2 === 'object' &&
+    value2 !== null
+  ) {
+    return JSON.stringify(value1) !== JSON.stringify(value2)
```
tujoworker, line 48: Good to read today's (yesterday) news that stringify will be twice as fast in V8 starting with version 13.8 (Chrome 138).

https://v8.dev/blog/json-stringify

tujoworker (commented):
Nice findings. 👏

I would like to confirm it, so need to use my Mac for doing so during the next days.

tujoworker (approved):
Nice work 👏

I think we still should not use JSON.stringify for comparing objects. Also, the order of the keys may change, but not the value. Which would not work with stringify. So I added a more robust version.

Issue #5332, opened 2025-07-24
`Iterate.PushContainer` is not closable if inside a `Wizard.Step` with an unrelated validation error in it.


boyum:
## 🐛 Bug Report

In certain cases, `Iterate.PushContainer` is not closable if inside a `Wizard.Step` with an unrelated validation error in it. It soft locks the step and users need to reload the page to continue.

## To Reproduce

1. Open CSB: https://codesandbox.io/p/sandbox/still-https-s8t28p
2. Navigate from step 1 to step 2
3. Navigate from step 2 to step 3 (thus triggering validation and get a validation error on the address input)
4. Navigate back from step 2 to step 1
5. Navigate from step 1 to step 2
6. See that the `Iterate.PushContainer` is now permanently in the open state

## Expected behavior

`Iterate.PushContainer` should not be affected by unrelated fields and should not soft lock.

## Run `npx envinfo`

Paste the results here:

```bash
  System:
    OS: macOS 15.5
    CPU: (12) arm64 Apple M3 Pro
    Memory: 82.25 MB / 18.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.17.0 - ~/.nvm/versions/node/v20.17.0/bin/node
    Yarn: 4.6.0 - ~/.nvm/versions/node/v20.17.0/bin/yarn
    npm: 8.19.4 - ~/dev/dnb/eufemia/node_modules/.bin/npm
  Managers:
    Homebrew: 4.5.9 - /opt/homebrew/bin/brew
    Maven: 3.9.9 - /opt/homebrew/bin/mvn
    pip3: 21.2.4 - /usr/bin/pip3
    RubyGems: 3.0.3.1 - /usr/bin/gem
  Utilities:
    Make: 3.81 - /usr/bin/make
    GCC: 17.0.0 - /usr/bin/gcc
    Git: 2.39.5 - /usr/bin/git
    Clang: 17.0.0 - /usr/bin/clang
    Curl: 8.7.1 - /usr/bin/curl
  Servers:
    Apache: 2.4.62 - /usr/sbin/apachectl
  IDEs:
    IntelliJ: 2024.3.5
    VSCode: 1.102.1 - /usr/local/bin/code
    Vim: 9.1 - /usr/bin/vim
    Xcode: /undefined - /usr/bin/xcodebuild
  Languages:
    Bash: 3.2.57 - /bin/bash
    Go: 1.23.5 - /opt/homebrew/bin/go
[...]

PR #5364, opened 2025-08-05
fix(PushContainer): should not open when it exists error in other fields in step

langz:
First draft of fixing https://github.com/dnbexperience/eufemia/issues/5332

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Iterate/PushContainer/PushContainer.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Iterate/PushContainer/PushContainer.tsx
@@ -304,6 +304,7 @@ function PushContainer(props: AllProps) {
             preventUncommittedChanges={preventUncommittedChanges}
             showResetButton={showResetButton}
             outerContext={outerContext}
+            required={required}
             {...rest}
           >
             {children}
@@ -325,6 +326,7 @@ function NewContainer({
   rerenderPushContainer,
   preventUncommittedChanges,
   outerContext,
+  required,
   children,
   ...rest
 }) {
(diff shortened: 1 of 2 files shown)


packages/dnb-eufemia/src/extensions/forms/Iterate/PushContainer/PushContainer.tsx
```diff
@@ -350,8 +352,11 @@ function NewContainer({
     visibilityContext?.isVisible === false
       ? false
       : Boolean(
-          !showOpenButton || containerMode === 'edit' || showCommitStatus
+          !showOpenButton ||
+            containerMode === 'edit' ||
+            ((required || hasContentChanged) && showCommitStatus)
```
langz, line 357: So now it will only open if there's a commitstatus and the pushcontainer is either required or had content changed.
Not sure if this is all-encompassing for all scenarios that may arise where it should be open.

packages/dnb-eufemia/src/extensions/forms/Iterate/PushContainer/PushContainer.tsx
```diff
@@ -325,20 +326,20 @@ function NewContainer({
...
   outerContext,
+  required,
   children,
   ...rest
 }) {
   const { containerMode, switchContainerMode } =
     useContext(IterateItemContext) || {}
   containerModeRef.current = containerMode
 
-  const { hasContentChanged, showStatus: showCommitStatus } =
```
langz, line 335: Renamed this back to rather being showStatus, as I initially thought showCommitStatus would only be if the pushcontainer or the step was attempted commited/submitted, but since it's actually showStatus it can potentially be true for so many other scenarios, like if exists any error, etc.

tujoworker (approved):
Very nice work 👏

> I think it should always be possible to close the pushcontainer from the user's perspective.
So if we in the future open it wrongly programmatically, like we did in https://github.com/dnbexperience/eufemia/issues/5332, then the user is allowed to close it, and can continue...

I agree on this. We should look into it.

Issue #5356, opened 2025-07-30
Cannot delete from array when initial data is incomplete (sometimes)


andlbrei:
## 🐛 Bug Report

**This is an issue in production**

My suspicion is that when the data on an itemPath is incomplete/invalid the element cannot be deleted, in certain cases.  
When these cases are, is not clear; setting the section in the first step of the wizard will make deletion work fine.  
Also, it seems like the data is partially removed from the state, but not completely.  

Video demonstrating the issue:

[](https://github.com/user-attachments/assets/8aedc145-d8c7-40bd-bd74-16b8b9f5d2f3)

## To Reproduce

https://stackblitz.com/edit/eufemia-starter-9wvvsxtb

- Navigate to the next step and try to delete the last element in the array.  

If you modify the reprod and remove the first step then deletion works fine, for some reason.  

## Expected behavior

I expect an element to be deleted from an array when the delete button is pressed, regardless if the data is valid or not.

## Eufemia Version

10.77.4

PR #5365, opened 2025-08-06
fix(Iterate.Array): cannot delete from array when initial data is incomplete

langz:
Closes https://github.com/dnbexperience/eufemia/issues/5356

[Original reproduction from issue](https://stackblitz.com/edit/eufemia-starter-9wvvsxtb)
[Reproduction from this PR](https://stackblitz.com/edit/eufemia-starter-jxegwqg9?file=package.json)

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Iterate/Array/__tests__/Array.test.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Iterate/Array/__tests__/Array.test.tsx
@@ -2328,7 +2328,7 @@ describe('Iterate.Array', () => {
                   ownershipAttributes: [],
                 },
                 {
-                  name: 'Has ownershipAttributes  as null',
+                  name: 'Has ownershipAttributes as null',
                   ownershipAttributes: null,
                 },
                 {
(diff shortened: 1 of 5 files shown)


packages/dnb-eufemia/src/extensions/forms/Iterate/Array/__tests__/Array.test.tsx
```diff
@@ -2303,6 +2303,111 @@ describe('Iterate.Array', () => {
...
+                },
+                {
+                  name: 'Has not specified ownershipAttributes',
+                },
+                {
+                  name: 'Has empty array',
+                  ownershipAttributes: [],
+                },
+                {
+                  name: 'Has ownershipAttributes  as null',
```
langz, line 2331:
```suggestion
                  name: 'Has ownershipAttributes as null',
```

tujoworker:
Until now, we temporarily replaced the array data when an item was removed so it could be animated. However, we also removed the item immediately from the data context. When a field’s value was undefined, it would then add the item back into the data context during the animation. Without animation it did work fine.

So now we remove the item from the data context after the animation has finished.

PR #5376, opened 2025-08-11
fix(PushContainer): styling of variant filled in error state on small screen

langz:


packages/dnb-eufemia/src/extensions/forms/Form/Section/style/dnb-form-section.scss
```diff
@@ -37,7 +35,6 @@
 
     &--variant-filled &__inner {
       --space: var(--spacing-small);
-      background-color: var(--color-lavender);
```
tujoworker, line 38: This one looks to me like it should not be there.

tujoworker (approved):
I think its fine now. Great work 👏

PR #5383, opened 2025-08-13
chore: remove new badge from AriaLive & Section components

langz:
Components were created approximately 2years ago


tujoworker (approved):
Fine for me. The section is much older. But got a big facelift, which was the reason we had it marked as "new".

Issue #4194, opened 2024-10-30
Showing a previously removed value results in undefined error


andlbrei:
## 🐛 Bug Report

When working with `Visibility`, toggling between removing a value using `useData().remove()` when hiding and then showing the fields related to to removed value causes an error.

I notice that after removing a value, it's not actually removed, but is set to `undefined`.
In another case, which is not relevant any more, I used `set()` to get the value properly removed from the data, but I don't think that is a very elegant solution.

Could/should `remove()` remove the key as well as the value from the data?
Or are there better ways to handle removing values when hiding them to avoid them being part of submitted data?

```
TypeError
Cannot use 'in' operator to search for '1' in undefined
    at Object.set (https://swf6zr.csb.app/node_modules/
dnb/eufemia/extensions/forms/utils/json-pointer/json-pointer.js:45:16
    at eval (https://swf6zr.csb.app/node_modules/
dnb/eufemia/extensions/forms/DataContext/Provider/Provider.js:426:28
    at eval (https://swf6zr.csb.app/node_modules/
dnb/eufemia/extensions/forms/hooks/useFieldProps.js:1011:95
    at eval (https://swf6zr.csb.app/node_modules/
dnb/eufemia/extensions/forms/hooks/useFieldProps.js:1039:7
commitHookEffectListMount
https://swf6zr.csb.app/node_modules/react-dom/cjs/react-dom.development.js:23150:26
commitLayoutEffectOnFiber
https://swf6zr.csb.app/node_modules/react-dom/cjs/react-dom.development.js:23268:17
commitLayoutMountEffects_complete
https://swf6zr.csb.app/node_modules/react-dom/cjs/react-dom.development.js:24688:9
commitLayoutEffects_begin
https://swf6zr.csb.app/node_modules/react-dom/cjs/react-dom.development.js:24674:7
commitLayoutEffects
https://swf6zr.csb.app/node_modules/react-dom/cjs/react-dom.development.js:24612:3
commitRootImpl
https://swf6zr.csb.app/node_modules/react-dom/cjs/react-dom.development.js:26823:5
commitRoot
https://swf6zr.csb.app/node_modules/react-dom/cjs/react-dom.development.js:26682:5
performSyncWorkOnRoot
https://swf6zr.csb.app/node_modules/react-dom/cjs/react-dom.development.js:26117:3
flushSyncCallbacks
https://swf6zr.csb.app/node_modules/react-dom/cjs/react-dom.development.js:12042:22
eval
[...]

PR #5398, opened 2025-08-15
fix(Forms): ensure labels from `Value.*` components can be served from a `Value.*` component itself

tujoworker:
In regards to #4194 and [this comment](https://github.com/dnbexperience/eufemia/issues/4194#issuecomment-3188760960).


langz (approved):
Looking great. Will this also work in combination with inheritLabel?

And will it work using a Value as label for both a Field and a Value?

tujoworker:
> Looking great. Will this also work in combination with inheritLabel?
> 
> And will it work using a Value as label for both a Field and a Value?

I looks like so as the added test runs green.

PR #5407, opened 2025-08-21
chore: create country translation files for en-GB and nb-NO

tujoworker:
Related to #5405


langz (approved):
Wow, so nice, so fast 🐎

PR #5411, opened 2025-08-26
Use tsdown to bundle eufemia

asmadsen:
We've had issues with using eufemia in vitest, and SSR mode with Vite. Our issues has mainly been caused by the fact that eufemia only does file to file transpilation, however proper ESM modules requires fully qualified paths during imports, meaning one has to specify each file as there is no implicit extension (`some-folder/some-file.js` instead of `some-folder/some-file`) and no implicit index resolution when pointing to a folder (`some-folder/index.js` instead of `some-folder`) something babel doesn't do. Using a bundler it will do reference resolving, and update the imports to be valid.

# Backwards compatibility
As we don't want to require a breaking change for this change, I've configured tsdown to bundle everything to follow the same structure and output as the current setup. The only change should be the rewritten imports to be compliant with the ESM format.

To validate the changes it has been tested against all apps written for the new online bank, the main issues has been related to the [custom cjs mapping required by jest](https://eufemia.dnb.no/uilib/usage/first-steps/module-formats/#jest-and-esm-node-testing-environments), and the places that rely on implicit directory index imports as they need to be manually declared in the exports field.

# Missing parts:
- [ ] Exports handling all directory imports
  -  As mentioned when validating it with apps, when an app import a index file by specifying the directory it needs to be manually defined in the exports field. The entries currently added is the ones required to get all apps building, so there may be apps that fail because of this. We should discuss how to handle this, should everything be defined or is it acceptable to require apps to update some imports.
- [ ] Stripping of prop-types
- [ ] Testing of ESM and UMD bundles


tujoworker:
Thank you so much for your contribution 🙏  

I fully agree that this is an issue we need to address. It's been a known problem for a long time, but as always, prioritization and limited capacity have delayed progress.  

That's why a contribution like this makes a big difference — it helps us move forward and gives us a reason to prioritize it.  

---

## Some background

- About four years ago I created a Babel plugin (unpaid) intended to solve the extension issue:  
  👉 [babel-plugin-fully-specified](https://www.npmjs.com/package/babel-plugin-fully-specified)  

Unfortunately, it was never prioritized further, and the issue remained.  

- Later, Kim and I worked on a first reproduction of using **Vitest with Eufemia** in a second (intermediate) package:  
  👉 [eufemia-vitest PR #1](https://github.com/tujoworker/eufemia-vitest/pull/1)  

---

## Plan going forward

- [ ] 1. Move out everything unrelated to the bundler migration into separate PRs.  
- [ ] 2. Align the config and ensure all tests pass.  
- [ ] 3. Discuss potential breaking issues.  
- [ ] 4. Discuss usage of **alpha** and **beta** versions.  
- [ ] 5. Automate manually defined export paths.  
- [ ] 6. Add tests/contracts to verify the changes and new setup.  

---

Looking forward to collaborating with you to make this happen 🚀

tujoworker (commented):
Thank you for your work so far!

I did a deep dive into the output, and it looks very promising.
We’re getting both an `/es` and a `/cjs` directory, which is great. In this PR we also end up with an `/esm` directory — but that one we can skip.

I’ve also left a few comments here and there on things I think are worth taking a closer look at.

PR #5412, opened 2025-08-26
fix(Field.Date): properly handle required error

tujoworker:


langz (approved):
Thanks, I've been testing it a few times in the deploy preview, and I've not found any anomalies, works great 👏

PR #5424, opened 2025-09-01
fix(Input): ensure error, focus, and hover states work with autofilled content

### LINKED ISSUE #5217: Error state style issue in input fields when filling with the browsers autocomplete feature (autofill)
When getting an error when verifying postal code, https://eufemia-git-feat-forms-api-connectors-bring-addr-1a098a-eufemia.vercel.app/uilib/extensions/forms/Connectors/Bring/demos/#verify-a-postal-code

I only see the red error border for postnr. when hovering over it.
Shouldn't it be visible when the error gets displayed?

https://github.com/user-attachments/assets/bf718204-69aa-4122-a549-86459e62e273

_Originally posted by @langz in https://github.com/dnbexperience/eufemia/issues/4575#issuecomment-2900822565_

tujoworker:
Fixes #5217

I also tried adding a visual test for these styles. There is a way to do this via Chromium and its DevTools, but it would require a significant refactoring.
Another option would be to use a "simulation" class, but I think this approach is too fragile from a maintenance perspective and could negatively impact CSS specificity in certain cases.

### DIFF
--- a/packages/dnb-eufemia/src/components/input/style/dnb-input.scss
+++ b/packages/dnb-eufemia/src/components/input/style/dnb-input.scss
@@ -535,10 +535,6 @@
     caret-color: var(--color-white);
   }
 
-  &[data-has-content='false'] &__input:focus:autofill {
-    --autofill-inset: none;
-  }
-
   &[data-input-state='focus'] &__placeholder {
     display: none;
   }
--- a/packages/dnb-eufemia/src/components/input/style/dnb-input.scss
+++ b/packages/dnb-eufemia/src/components/input/style/dnb-input.scss
@@ -141,6 +141,15 @@
     --b-width: var(--input-border-width--hover);
     --b-inset: var(--input-border-inset--hover);
   }
+
+  &__border:has(#{&}__input:autofill) {
+    --b-color: var(--input-border-color);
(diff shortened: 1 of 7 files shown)


langz (approved):
Great, thanks for considering adding a visual test 🤝

packages/dnb-eufemia/src/components/input/style/dnb-input.scss
```diff
@@ -227,13 +237,12 @@
...
-  // change the autocomplete appearance once filled out
+  // Set the background color to the default background color.
+  // Some browsers change the background color so users can see the difference between "autofill" and "user input".
+  // Which can be a bit jarring in DNBs design. We may consider to change this to a custom color in the future.
+  // NB: There are more than this ":autofill" in the theme file.
   &__input:autofill {
-    // set the border and the background
-    box-shadow:
-      0 0 0 var(--input-border-width) var(--input-border-color)
-        var(--autofill-inset, inset),
```
langz, line 235: Should we also remove https://github.com/dnbexperience/eufemia/blob/f609a1c3cf48b0f0e2a509c6ef42a7ac38772a48/packages/dnb-eufemia/src/components/input/style/dnb-input.scss#L530 ?

PR #5435, opened 2025-09-03
fix(DrawerList): swap aria-selected and aria-current

snorrekim:
Reverse of `aria-current` and `aria-selected` to be correct with accessibility rules.

- `aria-selected` is `true` for the selected option (the input value) and `false` for the other options.
- `aria-current` is true for the focused option (when navigating), and is removed from other option.

The use of `aria-current` is not needed by WAI-ARIA specs, but added by us for screen reader compatibility.


tujoworker (approved):
Works fine on NVDA ✅

PR #5442, opened 2025-09-04
fix(CopyOnClick): show Tooltip after copy

tujoworker:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1756972697635629


packages/dnb-eufemia/src/components/copy-on-click/__tests__/__image_snapshots__/copyonclick-have-to-match-default.snap.png
langz, line 1: I believe the left one (deleted is correct), as it has 80px width.

langz (approved):
🚀

PR #5444, opened 2025-09-04
fix(Dropdown, Autocomplete): keyboard navigation fixes

snorrekim:
Commit message:

--- 

Changes:
- Keyboard navigation no longer focuses the container `ul` between loops.
- `aria-activedescendant` now controlled by the `DrawerList` context.
- If an item is selected, it receives focus when the list is opened, instead of the previously focused element.

Fixes:
- Keyboard navigation no longer loops to the selected item.
- `Autocomplete` now has a `aria-activedescendant` set when first opened so screen readers can navigate the list
- `Autocomplete` now loops in both directions, not just from the bottom.
- Pressing any key no longer resets focus unless it's a search with a result. 

--- 

Fixes for some of the various issues reported here: https://github.com/dnbexperience/eufemia/issues/5528

I'll quote it here and update this list as I fix each one:

> - **BUG: Loops to selected:** 
>   - Should loop to the last item when looping "up", not to the selected item.

Fixed: no longer navigates to the item before or after the selected item when looping.

> - **IMPROVEMENT: List not focused (Dropdown):** 
>   - Should focus the list between loops when looping in both directions, not just "up"

Fixed/Changed: now only focuses the ul when the list first opens. Once user starts navigating only the items get focused.

> - **BUG/IMPROVEMENT: Doesn't loop (Autocomplete):**
>   - Should loop in both directions, not just "down"

Fixed: Autocomplete now loops in both directions

> - **IMPROVEMENT: Selected is not focused:** 
>   - Should focus selected item when re-opening the list. (Focus priority should be as follows: selected item > last focused item > the list.)

Implemented: if an item is selected, the focus is set to that item when the list is re-opened.

> - **BUG: Search resets focus:** 
>   - Should only reset focus on search if there is a result.

Fixed: now search only changes focus if it finds a result to focus on

> - **IMPROVEMENT: we should probably not focus the ul list to be accessibility compliant**
>   - https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#kbd_focus_activedescendant states that when the controlling element receives focus the aria-activedescendant should point to the item that is active. So there should always be an active item. Though the idea is that the controlling element always has DOM focus, and the items are just referenced and visually styled. So we sohuld perhaps implement that properly.  
>   - Today we even fake it for screen readers by pretending the activedecendant is the first or selected element. If not it can't start navigating.

Fixed-ish: First, we make sure that there `aria-activedecendant` always points to a valid item. Autocomplete was missing this. But we still start with no item focused. In those cases the aria-activedescendant defaults to the first item. This is a compromise between design and accessibility, screen readers require an "active" item to be able to properly navigate, but when you open a list we don't want to already have an item focused unless it is the selected item.

### DIFF
--- a/packages/dnb-eufemia/src/fragments/drawer-list/DrawerListProvider.tsx
+++ b/packages/dnb-eufemia/src/fragments/drawer-list/DrawerListProvider.tsx
@@ -649,7 +649,7 @@ export default class DrawerListProvider extends React.PureComponent<
    * @param {number} active_item The item to set as active
    * @param {object} param1
    * @property {boolean} fireSelectEvent Whether the onSelect event should get emitted
-   * @property {boolean} scrollTo Whether the list should scroll to the new active item nor not
+   * @property {boolean} scrollTo Whether the list should animate the scroll to the new active item or not
    * @property {event} event The event object to forward to the emitted events
    */
   setActiveItemAndScrollToIt = (
@@ -1211,20 +1211,19 @@ export default class DrawerListProvider extends React.PureComponent<
       }
 
       const { selected_item, active_item } = this.state
+      const newActiveItem =
+        parseFloat(selected_item as string) > -1
+          ? selected_item
+          : active_item
       dispatchCustomElementEvent(this.state, 'on_show', {
         ...args,
-        data: getEventData(
-          parseFloat(selected_item as string) > -1
-            ? selected_item
-            : active_item,
-          this.state.data
-        ),
+        data: getEventData(newActiveItem, this.state.data),
         attributes: this.attributes,
         ulElement: this._refUl.current,
       })
 
       this.setActiveItemAndScrollToIt(
-        parseFloat(active_item as string) > -1 ? active_item : -1,
+        parseFloat(newActiveItem as string) > -1 ? newActiveItem : -1,
         { scrollTo: false }
       )
     }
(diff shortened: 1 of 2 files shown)


snorrekim:
Conclusions after meeting: (Directions are reversed if the list is above the input.)

- Bug: when moving up the list loops to the selected element, not to the bottom element. It should loop to the bottom. Moving down loops as normal. 
  - When navigating past the first item focus is placed on the list element, this is a sort of navigation reset and therefore behaves as if the selected element is focused.
  - Fix: navigating should continue looping as normal even with a selected element.
- Behaviour (Dropdown): When navigating to past the first element the list is selected, before continuing to the last element. This does not happen when you navigate past the last element.
  - The reason for selecting the list is unclear, but we'll keep it as it is assumed to help screen readers.
  - Autocomplete keeps focus on the input field, and therefore has no need to select the list.
  - Fix: The list should be selected when looping in both directions.
- Bug (Autocomplete): When navigating past the first item it does not loop to the last. When navigating past the last is does loop.
  - Fix: It should loop in both direction.
- Behaviour: When closing and opening the list, the focus is kept on the last focused element
  - In a large list this means that if the focused element is different from the selected element it can be impossible to find the selected element again, and it's not obvious what is currently selected.
  - Fix: If an item is selected, the focus should be on the selected item when re-opening the list. If no item is selected we should still remember the last focused item.
- Bug: clicking any key that doesn't have a behaviour resets focus
  - The fallback default key-click behaviour is to do a search, but if no results are found the focus is reset.
  - Fix: only change focus if a result is found.
  
  Summary of fixes:

- When looping "up", go to the "last" item, not to the selected item.
- Dropdown: focus the list between loops when looping in both directions, not just "up"
- Autocomplete: loop in both directions, not just "down"
- When opening the list, focus priority should be as follows: selected item > last focused item > the list.
- Only reset focus on search if there is a result.

snorrekim:
One more minor issue. in Dropdown, when searching with keyboard, it remembers the last search after arrow navigation. So it goes to the next result, instead of the first.

If press "B" and go to the "Bar" item, and then navigate up to another item, if I press "B" again I should go back to the "Bar" item, not the next item that starts with "B".

PR #5445, opened 2025-09-05
chore(CopyOnClick): screenshot test when in Drawer

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1756972697635629

Not sure about the actual fix yet...


packages/dnb-design-system-portal/src/docs/uilib/components/copy-on-click/screenshot-test.mdx
langz, line 1: Because the screenshot test is of a drawer that has `openState="opened"` I've moved this screenshot test to an own page, so that it does not interfere with the others, as the drawer will always be open.

tujoworker (approved):

PR #5449, opened 2025-09-08
fix(CopyOnClick): use original Tooltip component to show tooltip for better support

tujoworker:
Motivation: https://dnb-it.slack.com/archives/C04NL73S2Q5/p1757066513223379


langz (approved):
Looking great, 
I can see from the screenshots that the new tooltips has more space between the tooltip and the content. But this distance/space is perhaps inline with the other tooltips that we have now?

tujoworker:
I think so: 
Screenshot 2025-09-08 at 08 24 50

PR #5463, opened 2025-09-10
fix(Forms): ensure Form.Iterate and Form.Section runs validation when cancel button is pressed

### LINKED ISSUE #5431: Canceling in a `Form.Section` Edit mode with invalid data is allowed before step validation
## 🐛 Bug Report

**Production bug, but not critical**

When the form has initial data which is invalid, the form knows about this and opens the `Form.Section` where the `Field` for the data resides in `Edit`mode, instead of `View` mode.  
However, the user is allowed to cancel `Edit` mode even though the `Field` is invalid, possibly because the `Field` is also unchanged.  

Also, there is something weird going on where two `Form.Sections` representing the same data appears when trying to go to the next step of the form.  

## To Reproduce

1. Have a validation schema, and initial data which does not validate for the form.  
1. Have a field for the data in a `Form.Section` with both `View` and `Edit` containers.  
1. The `Form.Section` should open in `Edit` mode
1. You should be able to cancel and go to `View` mode

https://codesandbox.io/p/devbox/remove-data-on-hide-forked-yhjv89

https://github.com/user-attachments/assets/2e586909-6deb-47de-acf7-328564ab9830

## Expected behavior

I expect the user to not be allowed to cancel editing when they are in an invalid state.  

### Browser – JS

10.79.0

tujoworker:
Fixes #5431

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/DataContext/FieldBoundary/FieldBoundaryContext.ts
+++ b/packages/dnb-eufemia/src/extensions/forms/DataContext/FieldBoundary/FieldBoundaryContext.ts
@@ -44,6 +44,12 @@ export interface FieldBoundaryContextState {
    * To set the local visible error state.
    */
   revealError?: (path: Path, hasError: boolean) => void
+
+  /**
+   * To manually verify if there are any field errors.
+   * Use it for when "hasError" is one tick behind.
+   */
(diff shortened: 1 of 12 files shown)


langz (approved):
🙏 📿

PR #5471, opened 2025-09-11
feat(Forms): add Zod support and use it internally

tujoworker:
Should be 100% backwards compatible. 
Makes it possible to use Eufemia Forms without CSP unsafe-eval.


langz:
👏

PR #5484, opened 2025-09-12
fix(Modal): prevent scrolling content behind modal on iOS

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1757666908255409

Not entirely sure if this is the best solution, and what the side-effects could be.


tujoworker:
I will look at this after my holiday next week.

PR #5486, opened 2025-09-12
fix(Dialog): make scrollable content full height

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1757666908255409


tujoworker:
I will look at this after my holiday next week.

PR #5489, opened 2025-09-16
fix(Field.Selection, Field.ArraySelection, Field.Toggle, Field.Boolean): enhance accessibility for radio and button variants

tujoworker:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1756978409797769 and https://dnb-it.slack.com/archives/C04NL73S2Q5/p1758018609062269

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/Boolean/__tests__/Boolean.test.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/Boolean/__tests__/Boolean.test.tsx
@@ -1334,6 +1334,49 @@ describe('Field.Boolean', () => {
         expect(second).toHaveAttribute('aria-invalid', 'true')
       })
     })
+
+    it('should have aria-labelledby and role="radiogroup" on fieldset when label is given', () => {
+      render(<Field.Boolean variant="buttons" label="Legend" />)
+
+      const fieldset = document.querySelector('fieldset')
+      const legend = document.querySelector('legend')
+
+      expect(fieldset).toHaveAttribute('aria-labelledby', legend.id)
+      expect(fieldset).toHaveAttribute('role', 'radiogroup')
+      expect(legend).toHaveAttribute('id')
+    })
+
+    it('should not have fieldset when no label is given', () => {
+      render(<Field.Boolean variant="buttons" />)
+
+      const fieldset = document.querySelector('fieldset')
+      expect(fieldset).not.toBeInTheDocument()
+
+      // Toggle buttons don't have individual labels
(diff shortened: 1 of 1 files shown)


packages/dnb-eufemia/src/components/radio/RadioGroup.js
```diff
@@ -237,7 +237,11 @@ export default class RadioGroup extends React.PureComponent {
       <RadioGroupContext.Provider value={context}>
         <div className={classes}>
           <AlignmentHelper />
-          <Fieldset className="dnb-toggle-button-group__fieldset">
+          <Fieldset
+            className="dnb-toggle-button-group__fieldset"
+            aria-labelledby={label ? id + '-label' : undefined}
```
snorrekim, line 242: should use a `const legendId = id + '-label'` since the id is used multiple places

packages/dnb-eufemia/src/extensions/forms/FieldBlock/FieldBlock.tsx
```diff
@@ -141,6 +141,8 @@ export type Props<Value = unknown> = SharedFieldBlockProps &
     disableStatusSummary?: boolean
     /** For internal use only */
     required?: boolean
+    /** Role for the fieldset element when using fieldset */
+    fieldsetRole?: string
```
snorrekim, line 145:
```suggestion
    fieldsetRole?: 'group' | 'radiogroup'
```

Technically fieldset can also have role `"presentation"` but I don't think that's needed here.

packages/dnb-eufemia/src/components/form-row/FormRow.js
```diff
@@ -235,7 +235,16 @@ export default class FormRow extends React.PureComponent {
...
     return (
       <Context.Provider value={providerContext}>
-        <Fieldset useFieldset={useFieldset}>
+        <Fieldset
+          useFieldset={useFieldset}
+          aria-labelledby={
+            useFieldset && label
+              ? label_id
+                ? label_id
+                : id + '-label'
```
snorrekim, line 244: Same as others: Since we are now using this id multiple places, perhaps create a `const labelId = label_id || id + '-label'`

packages/dnb-eufemia/src/components/radio/RadioGroup.js
snorrekim: `id + '-label'` is used 3 times in this file.

tujoworker: I see 👏

langz (approved):

PR #5491, opened 2025-09-18
fix(Dialog, Drawer): enhance screen reader support and focus management

tujoworker:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1757332666406779 and I think even https://dnb-it.slack.com/archives/CMXABCHEY/p1758017383453619

This PR:
- Ensures to set focus, even when no close button and no title was found.
- Ensures `aria-hidden` is not present in parent elements, so ctr+opt+array-key navigation works properly in VoiceOver.


langz (approved):
Looking good, seems like there's a few relevant tests failing, that's worth looking into.

PR #5493, opened 2025-09-19
release of v10.82.0

tujoworker:


langz (approved):
Great job 😃 
Do you want an own release with this Zod feature only? Makes sense to me.

Issue #3124, opened 2023-12-20
bug(Field.Currency): strange behaviour when entering big/large numbers


langz:
## 🐛 Bug Report

When entering values in the number field, it suddenly changes values.

All I do is entering the value `8`, many times.

Video:

https://github.com/dnbexperience/eufemia/assets/1359205/9ec77cf7-a442-4778-bc90-64f6c16150d0

## To Reproduce

Go to the [following example in our docs](https://eufemia.dnb.no/uilib/extensions/forms/feature-fields/Currency/#empty):
- Enter `8` many many times
- Then just observe what happens after entering `8` for the 17th time(input changes from `88888888888888888` to `88888888888888900`) and 22th time pressing 8(input changes to `8,888888888888` and onChange returns `8.88888888888889e+21`).

## Expected behavior

I expect it to behave as a regular input field, see the following [CSB](https://codesandbox.io/p/sandbox/boring-sky-kvqgxv?file=%2Fsrc%2FApp.tsx), and video(left is input type `string` and right is input type `number`):

https://github.com/dnbexperience/eufemia/assets/1359205/52788f63-a4f1-44e6-82c0-d6f2788cfd2d

snorrekim:
9007199254740991 is Number.MAX_SAFE_INTEGER. Any higher, and it loses precision. Even google's Material UI (https://mui.com/base-ui/react-number-input/) limit their number input to that. So I think that's what we should do as well.

PR #5498, opened 2025-09-22
fix(Field.Number): show error immediately when the input exceeds the maximum possible number

tujoworker:
Reported by @andlbrei

This is a follow up to this [issue](https://github.com/dnbexperience/eufemia/issues/3124). 

To reproduce:

1. Visit https://fix-forms-number-large-numbe.eufemia-e25.pages.dev/uilib/extensions/forms/base-fields/Number/demos/
2. Enter a large number like `123123123123123123` or `9007199254740992`
3. Now, the error should up. While before, it will first show on blur.

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/Number/Number.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/Number/Number.tsx
@@ -211,8 +211,6 @@ function NumberComponent(props: Props) {
       if (value === '') {
         return props.emptyValue
       }
-      validateContinuouslyRef.current =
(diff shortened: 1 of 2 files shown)


langz (approved):
Worked great when testing it 🙏 
However, does it work when pasting?
When pasting `9 007 199 254 740 992`, should I expect to see/get an error displayed immediately? Or only after "changing it"(adding an additional value) or by blurring?

Also, if I paste in `900719925474099` then add the value `2` to the end of it, so that I get `9 007 199 254 740 992` which should be larger than the maximum value displayed in the error message (`9007199254740991`), I would expect to see an error message? But don't get one.

PR #5501, opened 2025-09-22
chore(Field.Expiry): improve autocomplete docs

langz:


packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/feature-fields/Expiry/info.mdx
```diff
@@ -11,7 +11,7 @@ import { Field } from '@dnb/eufemia/extensions/forms'
 render(<Field.Expiry />)
 ```
 
-It supports the HTML [autocomplete](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete) attribute (`cc-exp-month` and `cc-exp-year`) if no value is given.
+It supports the HTML [autocomplete](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete) attribute, and by default set it to `cc-exp-month` for the month field, and to `cc-exp-year` for the year field.
```
tujoworker, line 14:
```suggestion
It supports the HTML [autocomplete](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete) attribute, which is by default set to `cc-exp-month` for the month field and `cc-exp-year` for the year field.
```

tujoworker (approved):

PR #5516, opened 2025-09-24
chore: ensure no ViewTransition is enabled during visual tests

tujoworker:
It may be the cause of such wired snapshots:
Screenshot 2025-09-24 at 14 40 04

Like when ViewTransition runs – even it theoretical should not:
Screenshot 2025-09-24 at 15 09 45

Or maybe a transition to fullscreen. Even I can't find it to happen.
Screenshot 2025-09-24 at 14 59 37

### DIFF
--- a/packages/dnb-design-system-portal/gatsby-node.js
+++ b/packages/dnb-design-system-portal/gatsby-node.js
@@ -271,6 +271,11 @@ exports.onCreateWebpackConfig = ({
       // Webpack 4 to 5 migration
       plugins.provide({ process: 'process/browser' }),
 
+      // Setup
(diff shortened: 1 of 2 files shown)


langz (approved):
🎉

tujoworker:
Kind of ironic, that the tests [do fail](https://github.com/dnbexperience/eufemia/actions/runs/17978024829/job/51136331309?pr=5516) with the same 😄
So definitively not fixed.

Screenshot 2025-09-24 at 15 50 27

tujoworker:
Suddenly it fails with something like that? 🤔

image

PR #5521, opened 2025-09-25
chore(Tabs): rename files to ts/tsx

tujoworker:
This way we can better track a ts rewrite.

- Rename of tiles to ts or tsx.
- Move type definitions from .d.ts to their new ts/tsx files.

We may use parts of #3065 later.


langz (approved):
Thanks, this is great 🥇 

My only concern is if we still provide the same types, using the same names, exported from the same places as before, so that it won't possibly "break" (complain) in other apps.

PR #5523, opened 2025-09-25
chore(Dropdown): `icon_position` tests

langz:


langz:
> Actually, when we force it the left aligned, then we mess up with the existing feature. 🤔
> 
> Screenshot 2025-09-25 at 09 47 20

Yeah, I fixed it by adding `!align_dropdown`, so now:
```
    if (isTrue(independent_width) && !icon_position && !align_dropdown) {
      icon_position = 'left'
    }
```

langz:
This PR does now only apply missing icon_position tests.

PR #5550, opened 2025-10-01
fix(GlobalStatus): scroll into view on Firefox

langz:
Motivation: https://dnb-it.slack.com/archives/C07JEJVSHJR/p1759232347266619

### DIFF
--- a/packages/dnb-eufemia/src/components/global-status/GlobalStatus.js
+++ b/packages/dnb-eufemia/src/components/global-status/GlobalStatus.js
@@ -354,6 +354,8 @@ export default class GlobalStatus extends React.PureComponent {
       this._scrollToStatusTimeout = isElementVisible(element, isDone)
       if (element && typeof element.scrollIntoView === 'function') {
         // wait a tick, to make sure that the element is visible, as firefox needs that
+        // or else it reports:
+        // scroll anchoring was disabled in a scroll container because of too many consecutive adjustments
         await wait(1)
         element.scrollIntoView({
           block: 'center',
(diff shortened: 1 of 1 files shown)


packages/dnb-eufemia/src/components/global-status/GlobalStatus.js
```diff
@@ -739,4 +743,6 @@ const isElementVisible = (elem, callback, delayFallback = 1e3) => {
   return null
 }
 
+const wait = (duration) => new Promise((r) => setTimeout(r, duration))
```
tujoworker, line 746:
```suggestion
const wait = () => new Promise((r) => requestAnimationFrame(r))
```

langz:
> > > I think, if requestAnimationFrame does work here, then we should use it over timeout.
> > 
> > 
> > Did you try it out?
> 
> I'll give it a go tomorrow 🤝🫡

Will try it out here https://github.com/dnbexperience/eufemia/pull/5553

langz:
> > > > I think, if requestAnimationFrame does work here, then we should use it over timeout.
> > > 
> > > 
> > > Did you try it out?
> > 
> > 
> > I'll give it a go tomorrow 🤝🫡
> 
> Will try it out here #5553

Did not work as expected, https://github.com/dnbexperience/eufemia/pull/5553#issuecomment-3359520162

packages/dnb-eufemia/src/components/global-status/GlobalStatus.js
langz: https://github.com/dnbexperience/eufemia/pull/5553#issuecomment-3359520162

PR #5563, opened 2025-10-06
fix(ChildrenWithAge): add translations for locale `da-DK` & `sv-SE`

langz:
Motivation:
It actually crashes/fails when using the Eufemia Portal with one of the locales that's missing translations.

https://github.com/user-attachments/assets/0a039364-c51d-4a30-b442-aa3fbcc9f42c

Todo in the future, could also be nice that the block/component does not actually break/fail if one tries to use a locale which does not have translations for it.


tujoworker:
> Todo in the future, could also be nice that the block/component does not actually break/fail if one tries to use a locale which does not have translations for it.

I'll look into it.

PR #5573, opened 2025-10-07
chore(Portal): add custom <Details> component

snorrekim:
Lets us add collapsed content in our `.mdx` files:

```
<Details>
  <Summary>Want to learn more?</Summary>

  This is more information that is relevant, but not so much that we should dedicate screenspace to it.

</Details>
```

portal details

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/components/table/demos.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/components/table/demos.mdx
@@ -86,12 +86,10 @@ A sticky table header with `sticky="css-position"` and `max-height` on the `Tabl
 
 ### Several tables in one container
 
-<details>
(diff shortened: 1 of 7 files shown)


snorrekim:
For some reason the code in the preview does not match the code in the preview.

snorrekim:
For some reason, `visibility` is magically transitionable in one direction. So when `visibility` goes from `hidden` to `visible` it is instant, but from `visible` to `hidden` it actually waits until the end of the transition. So we can use that to prevent keyboard navigation without cutting the animation short.

snorrekim:
I changed the animation curve for the height transition. Since we are using max-height it times it as if it is animating to a 100rem height container, regardless of actual size. So by slowing down the start of the animation it will not be so fast in the beginning.

PR #5575, opened 2025-10-07
chore(Forms): add Zod schema translations support (part #2)

tujoworker:
I messed up. Here is some more I wanted to get in with #5572


langz (approved):
Aha, I see, thanks for the explanation🫡

PR #5609, opened 2025-10-15
chore(Field.PhoneNumber): improve transformer example

langz:


packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber/PhoneNumber.tsx
```diff
@@ -33,7 +33,7 @@ import { DrawerListDataArrayItem } from '../../../../fragments/DrawerList'
 export type AdditionalArgs = {
   phoneNumber: string
   countryCode: string
-  iso: string
+  iso?: string
```
langz: If this would not be optional, then the updated example transformIn example which does not return iso, would result in a TS error as then the iso property would be required.

PR #5616, opened 2025-10-15
break(FormRow, FormSet)!: remove FormRow & FormSet

langz:


tujoworker:
> We should probably add docs for how to replace:
> <FormRow label="Legend:">

We have an example here: https://eufemia.dnb.no/uilib/layout/form-row/#deprecation

PR #5624, opened 2025-10-15
break(Section)!: remove `spacing` property

langz:


langz:
We need to look into all updated/changed screenshot tests.

PR #5658, opened 2025-10-16
chore(Switch): remove deprecated leftovers

langz:
Just in case we forget to remove the deprecated types


langz:
Done in https://github.com/dnbexperience/eufemia/pull/5650

PR #5667, opened 2025-10-16
fix(Autocomplete): skip unnecessary render of AriaLive

tujoworker:
Motivation: request by @vron to make Autocomplete more performant.

This ensures a more performant first render.


langz (approved):
Great 👏 

Seems like our SelectCountry and SelectCurrency components needs to get their tests re-aligned, as they now seems to fail 📽️

tujoworker:
Jepp, I'm looking into it. It needs a more correct way of defining when it was actually in use.

PR #5687, opened 2025-10-16
break(Card)!: remove styling as we have outset

langz:


tujoworker (approved):

tujoworker (changes requested):
Not sure why its marked as deprecated. Or if we need to adjust something in the props instead.

PR #5695, opened 2025-10-16
break(StepIndicator)!: remove `on_render` property

langz:


langz:
Followed-up in https://github.com/dnbexperience/eufemia/pull/5697

PR #5702, opened 2025-10-17
feat(renderWithFormatting): add new text formatter (simple markdown like) to shared tools

tujoworker:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1760612792849759
Docs preview: https://feat-render-with-formatting.eufemia-e25.pages.dev/uilib/usage/customisation/localization/#formatted-messages

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/usage/customisation/localization.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/usage/customisation/localization.mdx
@@ -271,6 +271,37 @@ function MyApp() {
 }
 ```
 
+#### Use without translations
+
+You can also use `renderWithFormatting` directly, without the translation context. This is handy for static copy or small strings you build at runtime.
+
+```tsx
+import { renderWithFormatting } from '@dnb/eufemia/shared'
+
+const text =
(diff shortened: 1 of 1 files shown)


langz (approved):
Wow, cool feature 👏 

Have we considered any other ways of doing it? Or is that not necessary?

I like this approach though, as it's like markdown. Does it support all MD features or only a subset? Do we manually do this, or use som kind of md plugin/package that helps us with it?

tujoworker:
> Have we considered any other ways of doing it? Or is that not necessary?

Technically, there aren’t many other good ways to achieve this (e.g. BBCode).

- `'Dette er [b]viktig[/b] og [i]betont[/i].'`

We could allow limited HTML, but that quickly becomes messy to read and maintain.

- `'Dette er <strong>viktig</strong> og <em>betont</em>.'`

Since practically all platforms already support a simple subset of Markdown, it feels like the right and most readable choice here as well.

- `'Dette er **viktig** og _betont_.'` => GitHub: Dette er **viktig** og _betont_.

langz:
Just a thought:
It seems a bit cumbersome to first change a translation for a given key, then remember to use renderWithFormatting for the given translation key. As one might change the translation key in a different file than where one must add the renderWithFormatting.

PR #5713, opened 2025-10-17
chore: remove `@ungap/structured-clone`

langz:
Worth a discussion, could potentially be a bit early to remove now, or?

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Form/Isolation/Isolation.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Form/Isolation/Isolation.tsx
@@ -34,12 +34,6 @@ import IsolatedContainer, { isolationError } from './IsolatedContainer'
 import IsolationContext from './IsolationContext'
 import type { OnCommit, Path } from '../../types'
 
-/**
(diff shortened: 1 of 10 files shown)


packages/dnb-eufemia/src/extensions/forms/Form/Isolation/Isolation.tsx
```diff
@@ -233,7 +227,10 @@ function IsolationProvider<Data extends JsonObject>(
       let isolatedData = structuredClone(mountedData)
 
       if (typeof transformOnCommitProp === 'function') {
-        isolatedData = transformOnCommitProp(isolatedData, outerData)
+        isolatedData = transformOnCommitProp(
+          isolatedData,
+          outerData
+        ) as Data
```
langz: Not sure if this is the best/correct thing to do.

tujoworker (commented):
I really wish we could remove it. But sadly first in iOS 15.4 [this got available](https://developer.mozilla.org/en-US/docs/Web/API/Window/structuredClone) it looks like. 

But currently we promise [support for iOS 13.1](https://eufemia.dnb.no/uilib/usage/).

PR #5725, opened 2025-10-20
fix(Form.Iterate): preserve JSX when replacing `{itemNo}` in labels

tujoworker:
This ensures `{itemNo}` works when labels are React nodes (e.g. formatted or wrapped), aligning with e.g. Iterate containers and Field.Toggle behavior.

This makes the `{itemNo}` work with `renderWithFormatting`.

**NB:** This will create a conflict when rebasing `main` on top of `v11`.


langz (approved):
Ahh, nice 👏

PR #5727, opened 2025-10-20
feat(Translation): enhance support for typing via generic `<Translation<SupportedTypes> id={(t) => t.my.string} />`

tujoworker:
Motivation: This way a dev does not need to create a wrapper component.


langz (approved):
🦭

PR #5782, opened 2025-10-20
break(DrawerListPortal): remove `include_owner_width` property

langz:
Doesn't seem like it's used?

I should rather rename this one, and not remove.


langz:
Follow-up for renaming https://github.com/dnbexperience/eufemia/pull/5791

PR #5788, opened 2025-10-20
break(Dropdown, Autocomplete, DrawerList)!: remove events

langz:


langz:
> Ok, so all the bugs seems to be cause mainly by the `validateDOMAttributes` that used to remove. things like `on_change` but does not remove `onChange`.

Should we extend it to remove onChange, etc, or is there an other approach? Just not passing those props down to the rendered html element?

snorrekim:
So, turns out we can't just remove all camel case events, because some of the code is dependant on "some" camel case events to pass through. Autocomplete needs `onMouseDown` to be passed in this case. It's messy. Still looking at it.

langz:
How's it going with this @snorrekim?
Do you have some work locally on your machine that's not commited? 
Or could/should I look at this again? 👀

snorrekim:
I structured the regexes in `component-helper.ts` as constants. (With JSDocs `*private` because we might want to export them for tests, but they shouldn't be used by our users)

I didn't write the tests for now, but I think this is how we should structure all our regexes, so they have names and are easily testable.

So I did:
```
/** @private */
const startsWithCamelCaseRegex = /(^[a-z]{1,}[A-Z]{1})/
...
if(!startsWithCamelCaseRegex.test(i))
```

instead of just 
```
if(!/(^[a-z]{1,}[A-Z]{1})/.test(i))
```

PR #5885, opened 2025-10-31
feat(Typography): add support for `proseMaxWidth` (for default width) to P and H2 etc.

tujoworker:
As an addition to #5857


langz (approved):
Smart, that looks great 👍 

Should it be possible to not specify a value for `proseMaxWidth`, just for simplicity:
```
        <P proseMaxWidth>
          This paragraph uses proseMaxWidth="auto".
        </P>
```

Or does one have to specify `"auto"`?

Perhaps we should consider calling the value `default` instead of `auto`?
As `60ch` is sort of the default value, not the auto value, or? 🤔

langz (approved):

PR #5888, opened 2025-10-31
feat(Field.PhoneNumber): add `numberLabel` and forward `label` to the field block (legend/fieldset)

tujoworker:
Motivation: Pedro 🕺

This change is potentially "breaking" – but I think not many, probably nobody did use a `label`, because we provide a default `label` already. And if someone have defined a label, then it will still show up. Just in a different place.


langz (approved):
Looking good. Property name is fitting, and matching the other property names 👏🫡

langz:
> Looking good. Property name is fitting, and matching the other property names 👏🫡

Should we also make a new translation, that has a matching name? Then deprecate the "old one", PhoneNumber.label?

PR #5898, opened 2025-11-03
fix(Field.Date): should reset value when pressing reset button

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1762173393020519

TODO:
- [x] The actual fix. Not sure if we should just in `Field.Date` clear/empty the date on reset "manually", or if we should try to reuse the logic of clearing/emptying from `DatePicker`


langz:
Wow, great job 👍 
Seems to work great 🥇 ❤️

PR #5939, opened 2025-11-12
fix(Form.useData): should set `data` to `defaultData` when `data` is not provided

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1761828345372899

[Here's a CSB describing and showcasing the issue](https://codesandbox.io/p/devbox/gifted-boyd-8jly2c?workspaceId=ws_LbiDwJdEBdbwxH2AkyAyVE)


langz (commented):
Is there more work that needs to be done in this PR?

PR #5966, opened 2025-11-21
feat(Popover): add new component (for internal use only)

tujoworker:
Here we use the code form Tooltip and enhance it with some extra props and features.

**Motivation:** To be used inside Tooltip (#5968), DatePicker (#5969), WordDefinition (#5970), and maybe later DrawerList with Dropdown and Autocomplete.

Slack link: https://dnb-it.slack.com/archives/C081G91GCT0/p1763379660378859?thread_ts=1762864839.379969&cid=C081G91GCT0

Docs Preview: https://feat-popover.eufemia-e25.pages.dev/uilib/components/popover/

PR #5968, opened 2025-11-21
feat(Tooltip): implement Popover

tujoworker:
Based on #5966

Docs Preview: https://feat-popover-tooltip.eufemia-e25.pages.dev/uilib/components/tooltip/

I wonder if we should move most of the Tooltip visual tests over to Popover? We don't have arrow and position tests there. But in Tooltip, we have many. I don't want them twice in the code base.


langz:
I've only done a quick comparison of the tooltip component on [main](https://eufemia.dnb.no/uilib/components/tooltip/demos/) vs. the [deploy preview for this PR](https://feat-popover-tooltip.eufemia-e25.pages.dev/uilib/components/tooltip/demos/), and these are the changes/discrepancies I've found:


Screenshot 2025-11-28 at 12 31 14


Hover over Top in the following example on [main](https://eufemia.dnb.no/uilib/components/tooltip/demos/#tooltip-with-delay) and in [deploy preview](https://eufemia.dnb.no/uilib/components/tooltip/demos/#tooltip-with-delay).

Other than that minor observation, I think it looks great :)

Further inspections is  that this is caused by the change of the `size` property I believe, and that the `size="large"` has gotten more spacing/padding, but I'm fine with that :) Just wanted to highlight the difference.

langz (approved):
I think this is good-to-go:

Image

PR #5947, opened 2025-11-13
fix(DatePicker): `alignPicker="auto"` should align based on the available space in the current viewport

langz:
I think we can release it as this for now, and then based on feedback from the users, we can decide if we should set this as our default functionality. I believe this should be our default functionality, as without it it can actually make it so that users can't find/see the calendar and will not be able to input the date as they wish. 

Motivation: 
- https://dnb-it.slack.com/archives/CMXABCHEY/p1760350493482459
- https://dnb-it.slack.com/archives/C0597DPEXS7/p1756713305297779

Deploy preview demo: https://feat-datepicker-alignment-au.eufemia-e25.pages.dev/uilib/components/date-picker/visual-tests/
Stackblitz: https://stackblitz.com/edit/vx7tceub?file=src%2FApp.tsx
Stackblitz Demo: https://vx7tceub.stackblitz.io
CSB: https://codesandbox.io/p/devbox/dawn-tdd-v7z6c4
CSB demo: https://v7z6c4.csb.app/

TODO:
- [ ] Consider making this into a more reusable component, but that can also be done at a later point in time. 

Video as demo of how the datepicker now will reposition itself:

https://github.com/user-attachments/assets/2eb581ef-ef4d-42db-862d-a6fb5dc19397

# Repositioning inside of a ScrollView (dialog, modal, table):

This PR also fixes issues with datepickers in modals, like the following issue:

https://github.com/user-attachments/assets/8b350976-21d2-4043-bbb6-b3b2f4522be1

It will now reposition itself:

https://github.com/user-attachments/assets/1473d3c3-2e56-49c3-b207-62237b83ba30

# Initial positioning when `opened`:

This PR also fixes issues with initial positioning of the portal:

https://github.com/user-attachments/assets/dfb292ca-77c3-48fb-bf67-eefc26187cff

It will now position itself correct:

https://github.com/user-attachments/assets/2de955fa-94eb-41c7-a6ed-064008b8035a

PR #5969, opened 2025-11-21
feat(DatePicker): auto align position if needed (implement Popover)

tujoworker:
Based on https://github.com/dnbexperience/eufemia/pull/5966

Motivation: 
- https://dnb-it.slack.com/archives/CMXABCHEY/p1760350493482459
- https://dnb-it.slack.com/archives/C0597DPEXS7/p1756713305297779

Docs Preview: https://feat-popover-datepicker.eufemia-e25.pages.dev/uilib/components/date-picker/

### DIFF
--- a/packages/dnb-eufemia/src/components/popover/PopoverContainer.tsx
+++ b/packages/dnb-eufemia/src/components/popover/PopoverContainer.tsx
@@ -540,15 +540,41 @@ function PopoverContainer(props: PopoverContainerProps) {
         placementKey = 'top'
         nextLeft = topPlacement.left
         nextTop = topPlacement.top
-      } else if (placementKey === 'top' && !fitsTop) {
-        if (fitsBottom) {
+      } else if (placementKey === 'top' && !fitsTop && fitsBottom) {
+        placementKey = 'bottom'
+        nextLeft = bottomPlacement.left
+        nextTop = bottomPlacement.top
+      } else if (!fitsTop && !fitsBottom) {
+        // If neither fits in the viewport, we pick the one with more visible area.
+        // However, we must ensure that 'top' placement doesn't go off the top of the document (negative coordinate),
+        // as that content would be unreachable.
+        const topInvalid = topPlacement.top < 0
+
+        if (topInvalid) {
           placementKey = 'bottom'
           nextLeft = bottomPlacement.left
           nextTop = bottomPlacement.top
         } else {
-          placementKey = 'bottom'
-          nextLeft = bottomPlacement.left
-          nextTop = bottomPlacement.top
+          const getVisibleHeight = (y: number) => {
+            const top = y
+            const bottom = y + elementHeight
+            const visibleTop = Math.max(top, viewportTopEdge)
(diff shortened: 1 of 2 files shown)


tujoworker:
Issue Screenshot 2025-11-24 at 10 25 13

langz:
[In the following example](https://t7s67waa.stackblitz.io), open it in fullscreen mode, set window size to be 666px, then try to press the 4 different datepickers in the 4 different corners. They are all visible, but the positioning and the arrow is a bit off:

Screenshot 2025-11-27 at 22 31 55
Screenshot 2025-11-27 at 22 31 47
Screenshot 2025-11-27 at 22 31 39
Screenshot 2025-11-27 at 22 31 32

langz:
Should it position itself correctly in the [following example](https://stackblitz.com/edit/t7s67waa-2w3sri9a?file=src%2FApp.tsx), or is it some kind of strange edge-case?
I used this setup to test the positioning I was doing in https://github.com/dnbexperience/eufemia/pull/5947, so that it could position itself correctly when scrolling both horizontally and vertically.

tujoworker:
> [In the following example](https://t7s67waa.stackblitz.io), open it in fullscreen mode, set window size to be 666px, then try to press the 4 different datepickers in the 4 different corners. They are all visible, but the positioning and the arrow is a bit off:

I think that's because we have a fixed margin of 16px to the browser edge. At the moment I don't think its a problem, because we want this space.

tujoworker:
> Should it position itself correctly in the [following example](https://stackblitz.com/edit/t7s67waa-2w3sri9a?file=src%2FApp.tsx), or is it some kind of strange edge-case? I used this setup to test the positioning I was doing in #5947, so that it could position itself correctly when scrolling both horizontally and vertically.

There is something wrong. When I use 90vw, it works fine. But above it does not anymore. But I don't know if this is a issue we need to care?

tujoworker:
> Wouldn’t we want the calendar to open top here? Or is it fine to have it open in the bottom, so the user barely can see it opens and have to scroll down.
> ScreenRecording_11-28-2025.09-53-58_1.mov

The reason is that it is too large. The user would not be able to navigate as the navigation is behind the browser window at the top.

langz:
> > Wouldn’t we want the calendar to open top here? Or is it fine to have it open in the bottom, so the user barely can see it opens and have to scroll down.
> > ScreenRecording_11-28-2025.09-53-58_1.mov
> 
> The reason is that it is too large. The user would not be able to navigate as the navigation is behind the browser window at the top.

Is this true in all scenarios? I don't fully understand this.
In the following scenario, there's plenty of space available above the input field, but it still opens below.
I don't see why _the user would not be able to navigate as the navigation is behind the browser window at the top._, as there's many 1000px available on the top.

Example: https://share.icloud.com/photos/06aknWzpevF6n6yaA1Am1S_TA
Had to post a link to icloud, as file is above 9mb which is the upper limit that GitHub supports

matthiasdoerstel:
Hi 💯 looking good. Commenting for future reference, so you have my name.

PR #5970, opened 2025-11-21
feat(TermDefinition): add new component

tujoworker:
Motivation: https://dnb-it.slack.com/archives/C081G91GCT0/p1762867695896529?thread_ts=1762864839.379969&cid=C081G91GCT0

Based on https://github.com/dnbexperience/eufemia/pull/5966

Docs Preview: https://feat-word-definition.eufemia-e25.pages.dev/uilib/components/term-definition/

### DIFF
--- a/packages/dnb-eufemia/src/components/term-definition/TermDefinitionDocs.ts
+++ b/packages/dnb-eufemia/src/components/term-definition/TermDefinitionDocs.ts
@@ -18,8 +18,8 @@ export const TermDefinitionProperties: PropertiesTableProps = {
     status: 'optional',
   },
   '[Space](/uilib/layout/space/properties)': {
-    doc: 'Supports spacing props such as `top`, `right`, `bottom`, `left`, or `space`.',
-    type: ['string', 'number', 'boolean', 'object'],
+    doc: 'Spacing properties like `left` or `right` are supported.',
+    type: ['string', 'object'],
     status: 'optional',
   },
 }
--- a/packages/dnb-eufemia/src/components/anchor/AnchorDocs.ts
+++ b/packages/dnb-eufemia/src/components/anchor/AnchorDocs.ts
@@ -77,7 +77,7 @@ export const AnchorProperties: PropertiesTableProps = {
     status: 'optional',
   },
   '[Space](/uilib/layout/space/properties)': {
-    doc: 'Spacing properties like `top` or `bottom` are supported.',
+    doc: 'Spacing properties like `left` or `right` are supported.',
     type: ['string', 'object'],
     status: 'optional',
   },


langz:
Should the header and close button of the WordDefinition be vertically centered?
This is a screenshot from the deploy preview as of today:

Screenshot 2025-11-24 at 10 12 54

langz:
Is it intentional that the active state does not have the dotted underline?


https://github.com/user-attachments/assets/679194ef-8d0a-4ce7-aacf-4f12ee10c784

tujoworker:
> When tabbing to the word definition anchor, then clicking enter to open the word definition, if I so tab backwards (shift+tab) what should the active element be? I would think it would be the word definition anchor, which was the last active element before pressing enter?

Can you test this again?

tujoworker:
> Would it be a suitable alternative to name it `TermDefinition`, rather than `WordDefinition`?

yes.

tujoworker:
> Is it intentional that the active state does not have the dotted underline?
> ScreenRecording_11-26-2025.22-01-50_1.mp4

yeah, as the moment it is. @snorrekim thought it would be nice to keep the hover state when open. What do you think?

tujoworker:
> In VG, when tabbing into the "word definition"-component, it automatically opens. Not sure if this is good or bad UX for screenreaders 🤔
> Screen.Recording.2025-11-24.at.10.46.27.mov
> 
> This is just an observation

The one from VG is bad for screen readers as it always reads out the whole content. Also, opening it with focus, is against the purpose of using tab keys as navigation mechanism and space/enter to make actions on it.

tujoworker:
> Getting the following issue when visiting https://feat-word-definition.eufemia-e25.pages.dev/uilib/components/word-definition/ now:

its renamed to term-definition. Here it is: https://feat-word-definition.eufemia-e25.pages.dev/uilib/components/term-definition/

langz:
> > Is it intentional that the active state does not have the dotted underline?
> 
> > ScreenRecording_11-26-2025.22-01-50_1.mp4
> 
> 
> 
> yeah, as the moment it is. @snorrekim thought it would be nice to keep the hover state when open. What do you think?

It's fine, I don't have a too opinionated opinion about it. My first thought was that I did not understand why the dotted underline was removed. But I think the existing behavior now is fine, as it matches how a anchor's active state is as well where the underline is removed, so it's consistent at least.

langz:
One personal thought in regards to the design (I'm no designer) is that I feel it’s a bit much space underneath/between the term and the dotted underline, and perhaps between each dott as well? Here’s a screenshot from reading it on my phone, and a screenshot of the regular text decoration style dotted:

IMG_0074
IMG_0075

packages/dnb-eufemia/src/components/term-definition/TermDefinitionDocs.ts
```diff
@@ -0,0 +1,25 @@
...
+    status: 'required',
+  },
+  placement: {
+    doc: 'Defines the preferred popover placement relative to the trigger.',
+    type: ['top', 'right', 'bottom', 'left'],
+    defaultValue: 'bottom',
+    status: 'optional',
+  },
+  '[Space](/uilib/layout/space/properties)': {
+    doc: 'Supports spacing props such as `top`, `right`, `bottom`, `left`, or `space`.',
```
langz: I tried top, and bottom, and nothing happened when I used those. So not sure if we support it?

PR #5985, opened 2025-11-25
fix(P): reset nesting inside Dialog, Drawer and Modal (render p instead of span)

tujoworker:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1763993859777309?thread_ts=1763739844.450909&cid=CMXABCHEY


langz (approved):
Great 👍

PR #5993, opened 2025-11-25
fix(Field.PhoneNumber): should set `undefined` in Form.Handler when phoneNumber is not set

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1764074316592609


langz:
I've added a change in https://github.com/dnbexperience/eufemia/pull/5993/commits/335565f269cecb403e64c591557234a3e8089663 that we could review @tujoworker

PR #6025, opened 2025-12-02
fix(Field.Upload): should not display error removing file using async `fileHandler`

langz:
Fixes an issue introduced in https://github.com/dnbexperience/eufemia/pull/5998, as we now have to handle `undefined` value as well. Previously it was only an empty list `[]`.

[Can be demoed here, by uploading a file, then pressing remove](https://eufemia.dnb.no/uilib/extensions/forms/feature-fields/more-fields/Upload/demos/#with-asynchronous-file-handler)

image


langz (commented):
I should rather make a variable for `files` instead of using optional chaining so much.

PR #6026, opened 2025-12-03
fix(Tooltip): improve performance by disabling keepInDOM

### LINKED ISSUE #6011: Is it correct that the portal should render 500+ react portals?
## 🐛 Bug Report

Is it correct that the portal should render 500+ React portals?

Image

Can be seen by visiting https://eufemia.dnb.no/icons/, open the dev console, and search for `eufemia-scope--portal`

tujoworker:
Fixes #6011


langz (approved):
I've tested the change in a few components using tooltip, and I've seen no difference in its behaviour, which is good :)

PR #6027, opened 2025-12-03
fix(NumberFormat): show Tooltip on user copy action

tujoworker:
In this PR we replace the current implementation by using the original Tooltip component.


langz:
Looks to be placing the tooltip in a bit strange place:


https://github.com/user-attachments/assets/c85f6d43-0167-4255-adc7-fd0180541f24

PR #6030, opened 2025-12-03
fix(Popover): make VoiceOver focus handling reliable

tujoworker:


langz (approved):
🤝 
I've not tested/verified it, but I guess you have 🥇

PR #6032, opened 2025-12-03
fix(Wizard): prevent navigation when there's an ongoing async `fileHandler` in Field.Upload

langz:
Based on #6044 and #6045.


tujoworker (approved):

langz:
Thanks so much, this is working great 🎆

langz:
Here's a [scenario](https://codesandbox.io/p/devbox/youthful-feather-98k858?workspaceId=ws_LbiDwJdEBdbwxH2AkyAyVE) where it does not work as expected, when wrapped in an `Iterate.Array` 🤔 

Here's a CSB with the expected behavior:
https://codesandbox.io/p/devbox/lively-sun-nr2yft?workspaceId=ws_LbiDwJdEBdbwxH2AkyAyVE

PR #6041, opened 2025-12-04
fix(NumberFormat): omit select & copy helper on touch devices

### LINKED ISSUE #6033: NumberFormat selection on iOS
Not related to the changes in this PR, but related to Numberformat on iOS. It’s difficult/impossible to select the text and copy it, so I will never be able to see the tooltip on user copy action. Usually when text is selected on iOS Safari and Chrome, I can do a long press and release, and then I'llget these options, but I'm not able to do so using a NumberFormat:

image

Here's a video of me trying to get the options above presented when using the NumberFormat, not really interesting video. 

https://share.icloud.com/photos/0059Hr5oYIf5E69VxNGlAfoSw

So while it's helpful on desktop fo selecting and copying the number, it's actually more difficult on iOS

_Originally posted by @langz in https://github.com/dnbexperience/eufemia/pull/6027#issuecomment-3606770934_

tujoworker:
Fixes #6033

### DIFF
--- a/packages/dnb-eufemia/src/components/number-format/NumberFormat.tsx
+++ b/packages/dnb-eufemia/src/components/number-format/NumberFormat.tsx
@@ -20,6 +20,7 @@ import {
   extendPropsWithContextInClassComponent,
   extendDeep,
   detectOutsideClick,
+  isTouchDevice,
 } from '../../shared/component-helper'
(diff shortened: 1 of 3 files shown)


langz (approved):
Tested on my phone, and on my laptop, looking good :) 

On phone when selecting the whole number, and pressing copy, there's no tooltip shown, but I guess that's fine and expected?

PR #6042, opened 2025-12-04
break(Logo)!: fix brand svg import color

snorrekim:
- Add color info to svg instead of using css.
- Add title text to svg instead of using props.

BREAKING CHANGES:
- Remove props `brand` and `variant`.
- Instead, each logo svg must be separately imported.
- Use the `svg` prop to give the component the logo svg.
- Dynamic logo change based on theme is possible using a callback in the `svg` prop.


tujoworker (approved):
Very good work 🤩 (would [this](https://github.com/dnbexperience/eufemia/pull/6042#discussion_r2626013949) improve the overflow issue?)

packages/dnb-eufemia/src/components/logo/__tests__/__image_snapshots__/logo-for-eiendom-aligns-with-text.snap.png
tujoworker: Image

Changing the viewBox to `369` would fix the overflow issue I think.

```diff
- 0 0 368 80
+ 0 0 369 80
```

snorrekim: I looked at it. Assuming the smallest logo we support is 16px height, then we would need to add 2px to the eiendom logo, but it differens from logo to logo, and it's not a simple percentage, because what 0.5px width is on 16px height is different for each logo. And each logo has a different scale for their viewbox as well. Eiendom has 380 width, but carnegie has 641. And some logos even get cut on the bottom because they are more rectangular, like the carnegie logo.

But it would be doable, I just think it should be a separate PR. Because we should also code it in a way that is easily reversible so as not to loose the "source" version.

snorrekim: Did some tests, these are not precise:
- Eiendom needs +2px width. 
- DNB needs +1.5px width. 
- Carnegie needs +10.8px width and +10.5px height

You know, I'm just gonna say, we set `overflow: visible` for logos. Not all svgs, just logos. Why would you ever want to cut the logos?

PR #6044, opened 2025-12-05
fix(Forms): prevent Wizard step navigation during pending validation

tujoworker:
- Add hasFieldState check before calling onSubmit
- Prevent navigation to next step when fields have pending validation state
- Add hasFieldState to WizardContainer dependencies
- To make #6032 work.
- Based on #6045.


langz:
Not introduced in this PR, but found this challenge/issue when testing this PR.
https://github.com/dnbexperience/eufemia/issues/6048

PR #6045, opened 2025-12-05
fix(Forms): improve async validator detection and behavior

tujoworker:
- Add handleAsAsync flag to FieldInternalsValue type
- Update hasFieldWithAsyncValidator to check handleAsAsync flag
- Fix async behavior logic when skipErrorCheck is true
- Add hasFieldWithAsyncValidator to ContextState interface
- Add missing dependency to useFieldProps hook

To make #6032 work.

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/hooks/useFieldProps.ts
+++ b/packages/dnb-eufemia/src/extensions/forms/hooks/useFieldProps.ts
@@ -2028,6 +2028,7 @@ export default function useFieldProps<Value, EmptyValue, Props>(
       additionalArgs,
       asyncBehaviorIsEnabled,
       defineAsyncProcess,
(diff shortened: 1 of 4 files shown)


packages/dnb-eufemia/src/extensions/forms/hooks/useFieldProps.ts
```diff
@@ -2028,6 +2028,7 @@ export default function useFieldProps<Value, EmptyValue, Props>(
       additionalArgs,
       asyncBehaviorIsEnabled,
       defineAsyncProcess,
+      executeOnChangeRegardlessOfUnchangedValue,
```
langz: 👍

langz (approved):
Looking good :)

PR #6050, opened 2025-12-05
fix(Iterate.Array): set data context value to `undefined` when no value was given

tujoworker:
In relation to #6049 

To make JSON (Ajv) and Zod schemas work with dependent schema sections, they need a value — for example an empty array.

This PR adds an empty array as the default value, similar to how fields add undefined (emptyValue) as their default.

At this point, neither me nor @andlbrei see this causing issues, although there may of course be unforeseen scenarios we haven’t thought of yet.

Preview: https://fix-forms-array-default-valu.eufemia-e25.pages.dev/uilib/extensions/forms/Form/schema-validation/


langz (approved):

tujoworker:
I check if "undefined" could be an option to use. So do not merge.

tujoworker:
`undefined` is no option.

tujoworker:
Had one more talk with @andlbrei and we still need to take another look into setting undefined instead of []

PR #6069, opened 2025-12-10
fix(Icon): render large icons with correct stroke width

tujoworker:
Deploy docs: https://fix-icon-sizes.eufemia-e25.pages.dev/uilib/components/icon/#icon-sizes


packages/dnb-design-system-portal/src/docs/uilib/usage/accessibility/assets/beach.js
langz: 👋

PR #6076, opened 2025-12-10
feat(Forms): add `onStatusChange` event to all Field.* components

tujoworker:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1765277593649449?thread_ts=1765269047.731149&cid=CMXABCHEY

Deploy preview: https://feat-forms-on-status-change.eufemia-e25.pages.dev/

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/Expiry/Expiry.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/Expiry/Expiry.tsx
@@ -47,11 +47,24 @@ function Expiry(props: ExpiryProps = {}) {
 
   const {
     onBlurValidator = expiryValidator,
+    onChangeValidator: onChangeValidatorProp,
     errorMessages: propErrorMessages,
     validateInitially: validateInitiallyProp,
+    validateContinuously,
     value: valueProp,
     transformIn: transformInProp,
+    onStatusChange,
   } = props
+
+  // When validateContinuously is enabled, also validate on change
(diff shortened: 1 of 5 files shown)


langz:
Nice nice nice, looking good 🎸

langz:
As `onStatusChange` in the following scenario is only executed on blur, as I guess we have an `onBlurValidator` (which I guess is mostly common in Eufemia forms) then I would not be able to know that the field is successful (not having an error) before actually blurring the field.

In the scenario where I want to disable a custom submit button under this field, and only enable it when there's no errors then I will have to tab out (blur out) of the field before the submit button would be enabled. I think that's a bit strange, and would expect the submit button to become enabled when I've successfully entered a valid expiry date.

But I guess I as a developer could perhaps fix this, by rather using an `onChangeValidator`? 🤔 

Video just shows that `onStatusChange` fires on blur:


https://github.com/user-attachments/assets/589d3309-6d6a-4f54-b4c3-b61fa909bfc9

langz:
> > But I guess I as a developer could perhaps fix this, by rather using an `onChangeValidator`? 🤔
> 
> They should be able to use `validateContinuously`.

Yeah, that seems like a good combination. Only issue is that I don't really get `validateContinuously` to be working as expected. And that's nothing introduced in this PR. I also tested with Field.OrganizationNumber, Field.NationalIdentityNumber, etc.

https://github.com/user-attachments/assets/9038e547-283a-4b70-b03e-39d32b67ea49

langz (approved):

PR #6101, opened 2025-12-12
fix(Field.PhoneNumber): support `labelDescriptionInline` property

langz:


tujoworker (approved):
So its actually also a TypeScript "fix"?

Issue #6103, opened 2025-12-13
`PostalCodeAndCity.City` not flexible in horizontal layout


andlbrei:
## 🐛 Bug Report

The container for `PostalCodeAndCity` takes up extra space, pushing the content to the right of it out even when setting `width` for `City`.  
When the content uses minimal space the container also leaves a wider gap than what the parent container assigns, so the gaps in a row of `Fields` will be inconsistent.

<img width="1300" height="1183" alt="Image" src="https://github.com/user-attachments/assets/0f62e644-ffde-4a5e-80d0-6a1334fb299c" />

<img width="1299" height="112" alt="Image" src="https://github.com/user-attachments/assets/a2e5e198-057c-4847-a248-56901286feef" />

## To Reproduce

Use "Open preview in new tab" for best results
https://stackblitz.com/edit/eufemia-starter-fxtjgsco

## Expected behavior

I expect the gap between `PostalCodeAndCity` and other `Field` components to be consistent, and the flex behaviour to be consistent.  

## Eufemia Version

10.92.0

PR #6104, opened 2025-12-15
fix(Field.PostalCodeAndCity): width in horizontal layout

langz:
Too many `_supportsSpacingProps` causes issues.
As `Field.PostalCodeAndCity` is already rendering a `Field.Composition` which has `_supportsSpacingProps`, it's not needed in `Field.PostalCodeAndCity` and causes problems in `Flex.Horizontal` layout.

Fixes https://github.com/dnbexperience/eufemia/issues/6103

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/PostalCodeAndCity/PostalCodeAndCity.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/PostalCodeAndCity/PostalCodeAndCity.tsx
@@ -186,5 +186,5 @@ function PostalCodeAndCity(props: Props) {
   )
 }
 
-PostalCodeAndCity._supportsSpacingProps = false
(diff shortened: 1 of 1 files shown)


packages/dnb-eufemia/src/extensions/forms/Field/PostalCodeAndCity/__tests__/PostalCodeAndCity.test.tsx
```diff
@@ -533,4 +533,8 @@ describe('Field.PostalCodeAndCity', () => {
       expect(city).toHaveAttribute('aria-invalid', 'true')
     })
   })
+
+  it('should have constant of _supportsSpacingProps=false', () => {
```
langz:
```suggestion
  it('should have constant of _supportsSpacingProps=undefined', () => {
```

PR #6123, opened 2025-12-17
feat(Tabs): add support for Badge notification inside a tab

tujoworker:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1765879948897889

As @snorrekim mentioned in the Slack thread, the position is not as given in Figma. So we may make some CSS adjustments in Tabs – are my firth thoughts. So we may actually changing this PR. We will have a discussion about this tomorrow.

Preview: https://chore-tabs-badge-example.eufemia-e25.pages.dev/uilib/components/tabs/demos/#tabs-with-badge-notification


langz:
Would it be too much flexibility to provide some kind of offset property, so that the devs could re-position the badge with minor adjustments? 🤔

tujoworker:
> Would it be too much flexibility to provide some kind of offset property, so that the devs could re-position the badge with minor adjustments? 🤔

I would say, the goal is that the tab has the same width as before, also when it comes to hover and focus. So we almost need to use position: absolute to pop it out of the flow.

langz (approved):

PR #6130, opened 2025-12-17
fix(Forms): improve type safety for custom validators for `Field.Name`

tujoworker:
Based on #6127

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/Name/Name.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/Name/Name.tsx
@@ -11,7 +11,7 @@ export type NameValidator = ValidatorWithCustomValidators<
   }
 >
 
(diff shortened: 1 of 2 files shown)


packages/dnb-eufemia/src/extensions/forms/Field/Name/Name.tsx
```diff
@@ -1,11 +1,25 @@
...
 import { FormError } from '../../utils'
 
+export type NameValidator = ValidatorWithCustomValidators<
+  string,
+  {
+    nameValidator: Validator<string>
+  }
+>
+
+export type NameCompanyValidator = ValidatorWithCustomValidators<
```
langz: I know the field name is Name.Company, but sort of like to read CompanyNameValidator instead of NameCompanyValidator. But I guess both alternatives will work 👍

PR #6146, opened 2025-12-18
fix(Form.InfoOverlay): should work correctly with `reduceToVisibleFields`

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1766050348926099


packages/dnb-eufemia/src/extensions/forms/Form/InfoOverlay/__tests__/InfoOverlay.test.tsx
```diff
@@ -487,4 +487,52 @@ describe('Form.InfoOverlay', () => {
     })
     expect(outerContext.data).toEqual(undefined)
   })
+
+  it('should work together with onSubmit and reduceToVisibleFields', () => {
```
langz: I checked that this test fails on main branch without the changes in this PR 🙏

Issue #6152, opened 2025-12-19
Wizard with async fileHandler in Field.Upload in Iterate.Array


langz:
Here's a [scenario](https://codesandbox.io/p/devbox/youthful-feather-98k858?workspaceId=ws_LbiDwJdEBdbwxH2AkyAyVE) where it does not work as expected, when wrapped in an `Iterate.Array` 🤔 

Video (Next button is not disabled and loading after the click, and there's some issues with the Field.Upload's as only the last file uploaded is actually "uploaded" / displayed in the list of files):

https://github.com/user-attachments/assets/fcccccd4-b9bc-4119-b434-05a7ef296a74

Here's a CSB of the original issue that we improved, just to demo that the next button is disabled and loading/spinning, etc).
https://codesandbox.io/p/devbox/lively-sun-nr2yft?workspaceId=ws_LbiDwJdEBdbwxH2AkyAyVE

_Originally posted by @langz in https://github.com/dnbexperience/eufemia/issues/6032#issuecomment-3674288071_

langz:
This is now fixed, could not reproduce :)

PR #6159, opened 2025-12-22
fix(Wizard): prevent navigation when ongoing async `fileHandler` in Field.Upload in a Iterate.Array

langz:
It should do the same as in https://github.com/dnbexperience/eufemia/pull/6032, just that it's wrapped in an `Iterate.Array`.

Will be the 2nd and hopefully last step to be fixing https://github.com/dnbexperience/eufemia/issues/6152

Related to https://github.com/dnbexperience/eufemia/pull/6156

For now, I've only added a test.

TODO:
- [x] The actual fix.


tujoworker (approved):

langz (commented):
Looks great, thanks 👨‍🦯

PR #6165, opened 2026-01-02
chore(MultiInputMask): wrap event handlers in useCallback

langz:


joakbjerk (approved):

tujoworker (changes requested):
We should not merge this, as it will cause conflicts with v11 and my ongoing work.

PR #6174, opened 2026-01-02
chore: replace substr as it's deprecated

langz:
Motivation: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr


joakbjerk (approved):
🔧

PR #6177, opened 2026-01-02
chore(CopyOnClick, NumberFormat): iOS v13 workaround is no longer needed

langz:
iOS v13 workaround is no longer needed (iOS 13 was released in 2019).


joakbjerk (approved):
🔥

langz:
Could look into doing this also, https://github.com/dnbexperience/eufemia/pull/6581
When the time is right.

PR #6233, opened 2026-01-07
chore: upgrade @playwright/test to v1.57.0

langz:
New version should have some performance gains:
- Faster test execution
- Improved parallel test running
- Better resource utilization

I've updated screenshot tests locally using firefox 144.0.2


langz:
There's a lot of changes in the visual tests/screenshots which does not seem accurate.
I think this will require quite some more work for us to align, and is probably best to do after v11 to avoid huge conflicts between v10 and v11 for now.

@tujoworker

PR #6239, opened 2026-01-07
chore: remove draft from .mdx files

langz:
Not entirely sure what `draft: true` means, but to me these .mdx files are not "drafts" as they are actually used 🤔


tujoworker (changes requested):
We can maybe rename it? But we can't remove it, as these pages are hidden in menu and other places where we "list" components.

Image

langz:
> We can maybe rename it? But we can't remove it, as these pages are hidden in menu and other places where we "list" components.
> 
> Image

Aha, so that's what it's used for. Thanks. We also have a similar key, showInMenu or hideInMenu or so( but that's probably for an other menu. 

I'll take a look at it 😄

PR #6247, opened 2026-01-08
chore: convert all type-only imports to use the import type

langz:
Could be worth looking into: https://typescript-eslint.io/blog/consistent-type-imports-and-exports-why-and-how/#benefits-of-enforcing-type-only-importsexports


langz:
> Nice! I wish we can do that in v11 instead. Just to save us from a lot of extra work.

Good, it's now a PR open towards v11 for this now 🙏 
https://github.com/dnbexperience/eufemia/pull/6249

PR #6288, opened 2026-01-13
chore: remove lockfile-lint

langz:
Does not look like we are using this?


tujoworker (changes requested):
I think we rather should enable it. Here's a [PR](https://github.com/dnbexperience/eufemia/pull/6298).

PR #5413, opened 2025-08-26
feat: upgrade date-fns to latest version (v4) – ⚠️ It may effect your direct usage of date-fns

asmadsen:
- Update the way it's functions are imported

PR #6291, opened 2026-01-13
chore: optimize date-fns imports

langz:
Optimization: Use direct imports for better tree-shaking


tujoworker (changes requested):
We just did the opposite in this [PR](https://github.com/dnbexperience/eufemia/pull/5413)  because of the tsdown work. I don't think it has any impact on tree shaking when I look how they export the functions.

Many years ago, [date-fns](https://date-fns.org/) did recommend the slash import. But that is not needed anymore it looks.

langz:
> We just did the opposite in this [PR](https://github.com/dnbexperience/eufemia/pull/5413) because of the tsdown work. I don't think it has any impact on tree shaking when I look how they export the functions.
> 
> Many years ago, [date-fns](https://date-fns.org/) did recommend the slash import. But that is not needed anymore it looks.

Okok 👍

PR #6301, opened 2026-01-14
chore: upgrade lint-staged to v15

langz:
Key improvements in v15:
- Better performance with faster file processing
- Improved ESM support
- Better handling of glob patterns
- More reliable stashing mechanism
- Better error messages and debugging


tujoworker (approved):
I don't think anyone is using it. So we may remove it at some point of time. But we may keep it as of now. ;)

PR #6304, opened 2026-01-14
chore: remove globby

langz:
Benefits:
- Removed external dependency (globby is only transitive now)
- Using native Node.js API
- Reduced bundle size
- Better performance with native implementation


langz:
I don't think it's worth to remove this

PR #6372, opened 2026-01-20
break(Dropdown)!: replace `preventSelection`, `actionMenu`, and `moreMenu` with `mode`

langz:
To me, this setup makes sense as its mutually exclusive, so that one cannot accidentally mix conflicting modes like preventSelection, moreMenu and actionMenu. 
This is just a suggestion. Need to discuss this in the team.

The `mode` property accepts the following values (just my initial suggestion):
- `default`
- `prevent`
- `action-menu`
- `more-menu`


packages/dnb-eufemia/src/components/dropdown/MIGRATION_GUIDE.md
langz: This is AI generated, and should not be included.
But could extract the essence of this and move it into v11 docs.

tujoworker (commented):
I think we should keep preventSelection. Because it could be used with the menu modes.

So with that I think mode would not be the best name. Could be something menu related.

langz:
Solved in https://github.com/dnbexperience/eufemia/pull/7380

PR #6523, opened 2026-01-29
break(DatePicker)!: replace `alignPicker` with `alignCalendar`

langz:
Not 100% sure about this one. But today on [main](https://eufemia.dnb.no/uilib/components/date-picker/properties/#properties), the prop docs says: 

Use right to change the preferred `calendar` alignment direction. Defaults to left. If the DatePicker is close to the edge of the screen, the alignment of the `calendar` will change automatically to fit in the viewport.


tujoworker (changes requested):
I assume we don't need the prop anymore now that we use Popover. Did not double check it yet.

langz:
> I assume we don't need the prop anymore now that we use Popover. Did not double check it yet.

Yeah, we could of course remove it, as it with `Popover` will reposition itself correctly if needed.
But I think the property still makes sense, so that users can decide the alignment of the calendar/picker.
I assume we have this functionality that one can align the Popover in other components as well.

`alignPicker="right"` looks like:

Screenshot 2026-04-07 at 12 20 07



`alignPicker="left"` / default / not any value set looks like:

Screenshot 2026-04-07 at 12 20 21

PR #6525, opened 2026-01-30
break(Modal, Dialog, Drawer)!: replace `alignContent` value `centered` with `center-both`

langz:
Just a suggestion.
I never understood the difference between center and centered based on just reading the values of alignContent, so this is just a suggestion to make the distinction clearer.


tujoworker (commented):
I don't think "both" does help make it better. But dont have a suggestion yet.

PR #6544, opened 2026-02-05
fix(Field.Email): support unicode in local part and domain

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1770281489118729

This automatically convert internationalized domain names to their ASCII Punycode representation before validation, providing seamless SMTPUTF8 support

We should either implement https://github.com/dnbexperience/eufemia/pull/6543 or https://github.com/dnbexperience/eufemia/pull/6544

Other thoughts, should we support displaying punicode values, like `user@xn--mller-kva.com` as unicode to the user in both `Field.Email` and `Value.Email`? So that the user would see `user@müller.com` instead of the punycode version `user@xn--mller-kva.com`?


tujoworker:
I think this PR #6543 uses less code and is in line what we have today.

PR #6697, opened 2026-02-18
docs(Dropdown): document `on_show_focus` and `on_hide_focus`

langz:


tujoworker (changes requested):
Can we do this in v11 instead?

PR #6718, opened 2026-02-20
chore: prettier trailing comma

langz:
With "all", trailing commas are now added everywhere valid — including function parameters and arguments — which is the Prettier 3+ default and reduces diff noise when adding new items.

We should wait with merging this until end of v11, as it will cause heavy conflicts.


tujoworker (approved):
I guess this means a lot of files needs a small change? Then we should merge this right before release, and not now, in order to prevent conflicts.

PR #6771, opened 2026-02-24
chore: remove bypassActWarning

langz:
Now that we are using React v19, one would think that we do not longer need this.
TODO: Need to fix/improve the tests.

Prompt to use:

Prompt: Fix React act() warnings in Eufemia test suite

After removing the bypassActWarning() suppression from [setupJest.js](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), approximately 1,169 "not wrapped in act" console warnings surface across the test suite. All tests still pass — these are warnings only, not failures. The goal is to eliminate these warnings by fixing the root causes in the test files.

Context:

React 19.2.3, @testing-library/react 16.3.0, @testing-library/user-event 14.6.1
The codebase already uses await userEvent in 92 files — this is the established pattern
All 344 test suites (8,749 tests) pass with the suppression removed
What to fix and how:

fireEvent calls (~2,021 call sites across 127 files): These are the primary source of warnings. For each fireEvent call that triggers a state update, either:

Preferred: Replace with the equivalent await userEvent call (e.g., fireEvent.click(el) → await userEvent.click(el), fireEvent.change(input, { target: { value: 'x' } }) → await userEvent.type(input, 'x') or await userEvent.clear(input) + await userEvent.type(input, 'x')). The test function must be async.
Fallback (when userEvent has no equivalent): Wrap in act(): act(() => { fireEvent.scroll(el) }). Import act from @testing-library/react.
rerender() calls (~1,060 call sites across 152 files): Wrap in act():

Timer advances (~10 call sites across 7 files): Wrap jest.advanceTimersByTime() and jest.runAllTimers() in act():

Files: DateFormat.test.tsx, DateFormatUtils.test.ts, ModalContent.test.tsx, NumberFormat.test.tsx, TableStickyHeader.test.tsx, createMinimumAgeValidator.test.tsx, createMinimumAgeVerifier.test.ts

Suspense-related warnings (3 occurrences): Wrap render in await act(async () => { ... }).

Execution approach:

Work through files one at a time
After fixing each file, run its test with npx jest --testPathPatterns='<path>' --no-coverage to verify it passes AND produces zero "not wrapped in act" warnings (pipe output through grep -c 'not wrapped in act' to verify count is 0)
Do NOT change test assertions or behavior — only wrap/replace the event/render calls to satisfy act()
Follow existing code style (Prettier/ESLint config, camelCase, etc.)
Prefer [document.querySelector](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) over screen from @testing-library/react (per project conventions)
Find affected files by running:

To get a per-file breakdown of which test files produce warnings, you can run individual test suites and check.

Start with the smallest/easiest batch (timer files), then move to high-fireEvent components, then bulk rerender fixes.

### DIFF
--- a/packages/dnb-eufemia/src/components/autocomplete/__tests__/Autocomplete.test.tsx
+++ b/packages/dnb-eufemia/src/components/autocomplete/__tests__/Autocomplete.test.tsx
@@ -756,7 +756,9 @@ describe('Autocomplete component', () => {
       })
     })
 
-    await wait(2)
+    await act(async () => {
+      await wait(2)
+    })
 
     const nodes = document.querySelectorAll('.dnb-sr-only')
 
@@ -772,7 +774,9 @@ describe('Autocomplete component', () => {
       })
     })
 
-    await wait(2)
+    await act(async () => {
+      await wait(2)
+    })
 
     const nodes1 = document.querySelectorAll('.dnb-sr-only')
     expect(nodes1[nodes1.length - 1].textContent).toBe('2 alternativer')
@@ -788,7 +792,9 @@ describe('Autocomplete component', () => {
       })
     })
 
-    await wait(2)
+    await act(async () => {
+      await wait(2)
+    })
 
     const nodes2 = document.querySelectorAll('.dnb-sr-only')
     expect(nodes2[nodes2.length - 1].textContent).toBe('3 alternativer')
@@ -803,7 +809,9 @@ describe('Autocomplete component', () => {
       })
     })
 
-    await wait(2)
+    await act(async () => {
+      await wait(2)
+    })
 
     const nodes3 = document.querySelectorAll('.dnb-sr-only')
     expect(nodes3[nodes3.length - 1].textContent).toBe(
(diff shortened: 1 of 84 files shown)


packages/dnb-eufemia/src/components/autocomplete/__tests__/Autocomplete.test.tsx
```diff
@@ -797,18 +859,22 @@ describe('Autocomplete component', () => {
...
-      target: { value: 'cc bb more' },
+    await act(async () => {
+      fireEvent.change(document.querySelector('.dnb-input__input'), {
+        target: { value: 'cc bb more' },
+      })
     })
     expect(
       document.querySelectorAll('li.dnb-drawer-list__option')[0].outerHTML
     ).toBe(
-      '<li class="first-of-type first-item closest-to-top closest-to-bottom dnb-drawer-list__option" role="option" tabindex="-1" aria-selected="false" data-item="3" id="option-autocomplete-id-3"><span class="dnb-drawer-list__option__inner"><span class="dnb-drawer-list__option__item"><span>item <span class="dnb-drawer-list__option__item--highlight">cc</span> second</span></span></span></li>'
```
langz: TODO, look into this change

PR #6831, opened 2026-03-01
break!: remove `scrollbarAppearance`

langz:


tujoworker (changes requested):
I don't think we can and should remove it. It was an important feature requested by UX for years. Here is Firefox before/after.

Image
Image

PR #6872, opened 2026-03-02
refactor(Button): convert from class to function component

langz:


langz:
> huh, why close it?

Not sure if it's possible to remove all class components as part of v11, as I'm unsure if we have time. 

I therefore closed this, and rather focused on tasks that I know will provide value as of v11, such as removing prop-types and converting to TS. 

We can always re-open this, and many more PRs related to migrating away from class components.

PR #6896, opened 2026-03-03
break!: remove what-input dependency

langz:
Deploy preview: https://chore-remove-what-input.eufemia-e25.pages.dev/uilib/

Replace the what-input library (~3.5KB) with a lightweight
inputTracker utility (~35 lines) that sets data-whatinput on <html>.

CSS changes:
- Replace html[data-whatinput] selectors with :focus-visible for
  buttons, checkboxes, radios, switches, tabs, sliders, and other
  non-text-input components
- Replace html[data-whatintent] hover guards with @media (hover: hover)
- Keep html[data-whatinput='keyboard'] gating for text inputs, textareas,
  and drawer-list where :focus-visible cannot distinguish keyboard from
  mouse focus per CSS spec
- Remove @supports not selector(:focus-visible) fallbacks
- Remove deprecated whatInput()/whatInputNot() SCSS mixin wrappers

JS changes:
- Add src/shared/helpers/inputTracker.ts as a minimal what-input
  replacement (tracks pointer/keyboard via pointerdown and keydown)
- Remove what-input import and Tabs whatInput integration
- Update component-helper to initialize inputTracker on load

BREAKING CHANGE: Removes what-input npm dependency. Consumer code relying
on html[data-whatinput] or html[data-whatintent] attributes set by
what-input will need to provide its own input tracking.

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/about-the-lib/releases/eufemia/v11-info.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/about-the-lib/releases/eufemia/v11-info.mdx
@@ -17,6 +17,7 @@ v11 of @dnb/eufemia contains _breaking changes_. As a migration process, you can
 1. Replace `innerRef` with `ref` on all Eufemia components.
 2. Replace `<Context.Provider>` with `<Context>` when using Eufemia contexts directly (React 19 deprecation).
 3. If you import Eufemia SCSS source files with `@import`, replace with `@use` and namespace your mixin/variable calls (see [SCSS: @import → @use](#scss-import--use)).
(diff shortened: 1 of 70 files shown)


tujoworker:
Exiting! But its also a breaking change. Because teams do most probably use the HTML tag it sets. So we need to announce it in v11-info.mdx 🧸

tujoworker:
1. visit the preview: https://chore-remove-what-input.eufemia-e25.pages.dev/uilib/
2. Clicking in the area with all the text, shows the focus outline. Before we have checked whether its a keyboard focus event or mouse/touch etc. And in this case, we listened for the keyboard to focus.

PR #6949, opened 2026-03-09
refactor: move postcss-selector-parser to devDependencies

langz:


tujoworker (changes requested):
Actually, it needs to be a part of deps, and not devDeps, because its used by the PostCSS plugin: https://eufemia.dnb.no/uilib/usage/customisation/styling/style-isolation/#3-add-the-postcss-plugin

PR #7164, opened 2026-03-23
refactor(Breadcrumb): rename 'spacing' prop to 'innerSpace' for consistency

langz:


tujoworker (approved):
Is that case not handled in the other Section PR?

innerSpace!=100%spacing

PR #7193, opened 2026-03-23
break(Flex)!: rename sizeCount to columns, --has-size to --has-span

langz:


tujoworker (changes requested):
Not sure if that makes sense. Because I think they can and will wrap. If so, it's debatable if we can call it column.

PR #7194, opened 2026-03-23
feat: add use client directives for RSC compatibility

langz:
Todo:
- review if we need use directive in all these components
- Does it exist a linting rule or such that could determine if a component needs use directive or not?


tujoworker (changes requested):
We have a Babel plugin that ensures that for us automatically.

PR #7195, opened 2026-03-23
break(Section)!: remove deprecated spacing prop

langz:


tujoworker (approved):
innerSpace is a bit different. I
It also adds left and right padding. So there comes the inline&block naming in handy. 

Also, is there a v11-info entry?

PR #7233, opened 2026-03-26
feat(List): add emptyState prop for empty list rendering

langz:
Add emptyState prop to List.Container that renders custom content when the list has no children. This avoids rendering an empty ul element and enables user-friendly messaging like 'No items found'.

TODO: Discuss if this is something we should have or not.

### DIFF
--- a/packages/dnb-eufemia/src/components/list/Container.tsx
+++ b/packages/dnb-eufemia/src/components/list/Container.tsx
@@ -10,6 +10,7 @@ export type ListContainerProps = {
   separated?: boolean
   skeleton?: SkeletonShow
   disabled?: boolean
+  emptyState?: React.ReactNode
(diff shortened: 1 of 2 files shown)


tujoworker (changes requested):
Eufemia Form uses emptyValue. 


I don't think a property is the right direction. And it's very case by case how they want or need to deal with "that issue".


I could see something like this:

<List.Item.Placeholder>

It could provide its ovn/different look. Or inherit the basis look.

Alternatively a good example of how to use <List.Item.Basic> for a placeholder could work too.

tujoworker:
I'm not sure, should we add such an example? I guess it will be handled different for case to case. Also when UX is involved.

Screenshot 2026-04-07 at 08 21 19

PR #7331, opened 2026-04-07
chore: upgrade to Storybook v10

tujoworker:
On v11 Storybook crashes currently.
Should be tested on other machines before merging.

PR #7336, opened 2026-04-07
chore(Forms): use `interface` over `type` for performance in docgen

tujoworker:
Fixes an issue cause by this change #6778
It causes docgen (used by Storybook) to use a lot of memory, so it may crash. And if it not crashes, it slows down the performance.

More details:
The issue is that `type = { ... } & Base` forces react-docgen-typescript to structurally expand intersection types recursively, while `interface extends` uses nominal/lazy resolution. The fix: keep the exported name as `type` (satisfying your ESLint rule) but use an internal `interface` for the actual type body, so react-docgen gets the efficient resolution path.

Related PR #7331

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/types.ts
+++ b/packages/dnb-eufemia/src/extensions/forms/types.ts
@@ -352,13 +352,15 @@ export type ErrorProp<Value> = MessageProp<
   | Array<string | React.ReactElement | Error | FormError>
 >
 
-export type UseFieldProps<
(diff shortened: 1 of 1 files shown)


langz (approved):
❤️

I was not aware that this could happen.
Any way we could "safe guard" us from stuff like this happening again? Any tests or checks that we could add or enforce for this? 🤔

tujoworker:
No, not really. It depend very much on the tool etc.

PR #7337, opened 2026-04-07
feat(List): add `to`, `element`and `elementProps` to List.Item.Action

ahmetacer5:
## Summary
- Add `element`, `to`, and `elementProps` props to `List.Item.Action`, enabling React Router client-side navigation
- Follows the established pattern used by `Anchor`, `Button`, and `BreadcrumbItem`
- Without this, using `href` always causes a full page reload
- `elementProps` is fully typed — it infers allowed props from the `element` component
- Add `a:visited { color: inherit }` to the action href anchor CSS reset to prevent browser default visited-link styling

### Usage
```tsx
import { Link } from 'react-router-dom'

<List.Item.Action
  element={Link}
  to="/my-route"
  title="Go somewhere"
  elementProps={{ replace: true, state: { from: 'list' } }}
/>
```

## Test plan
- [x] All 57 existing + new ItemAction tests pass
- [x] Verify client-side navigation works with React Router in a consumer app
- [x] Verify `href` (without `element`/`to`) still works as before
- [x] Verify disabled/pending states prevent navigation with `to`
- [ ] Verify `elementProps` provides correct type inference for the given `element`

### DIFF
--- a/packages/dnb-eufemia/src/components/list/ListDocs.ts
+++ b/packages/dnb-eufemia/src/components/list/ListDocs.ts
@@ -331,6 +331,16 @@ export const ItemActionProperties: PropertiesTableProps = {
     type: 'string',
     status: 'optional',
   },
+  to: {
+    doc: 'Use this prop only if you are using a router Link component as the `element` that uses the `to` property to declare the navigation url.',
+    type: 'string',
+    status: 'optional',
+  },
+  element: {
+    doc: 'Define what HTML or React element should be used for the link (e.g. `element={Link}` for React Router). Only applicable when `href` or `to` is set.',
+    type: 'React.Element',
+    status: 'optional',
+  },
   target: {
     doc: 'Link target (e.g. `_blank` for new tab). Only applicable when `href` is set.',
     type: 'string',
--- a/packages/dnb-eufemia/src/components/list/ListDocs.ts
+++ b/packages/dnb-eufemia/src/components/list/ListDocs.ts
@@ -342,7 +342,7 @@ export const ItemActionProperties: PropertiesTableProps = {
     status: 'optional',
   },
   elementProps: {
-    doc: 'Additional props forwarded to the `element` component (e.g. `{ replace: true, state: { from: "list" } }` for React Router Link). Only applicable when `element` is set.',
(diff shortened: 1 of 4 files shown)


packages/dnb-eufemia/src/components/list/style/dnb-list.scss
```diff
@@ -294,6 +294,10 @@
         grid-row: 1 / -1; // Center vertically
         text-decoration: none;
         color: inherit;
+
+        &:visited {
+          color: inherit;
```
tujoworker: I think we should address this globally, but in a following up PR. So fine for now 🫶

tujoworker (approved):
Thank you 👏

packages/dnb-eufemia/src/components/list/style/dnb-list.scss
tujoworker: Actually, it cause a [style lint error](https://github.com/dnbexperience/eufemia/actions/runs/24124495919/job/70385755128?pr=7337) how its positioned. So I think we should remove it in this PR.

ahmetacer5: Thanks for the review :)

packages/dnb-eufemia/src/components/list/ListDocs.ts
```diff
@@ -331,6 +331,21 @@ export const ItemActionProperties: PropertiesTableProps = {
...
+    type: 'string',
+    status: 'optional',
+  },
+  element: {
+    doc: 'Define what HTML or React element should be used for the link (e.g. `element={Link}` for React Router). Only applicable when `href` or `to` is set.',
+    type: 'React.Element',
+    status: 'optional',
+  },
+  elementProps: {
+    doc: 'Additional props forwarded to the `element` component (e.g. `{ replace: true, state: { from: "list" } }` for React Router Link). Only applicable when `element` is set.',
```
tujoworker: > Only applicable when `element` is set.

Is that true? I think it will forward them to the existing Anchor/Button also.

tujoworker:
```suggestion
    doc: 'Additional props forwarded to the `element` component (e.g. `{ replace: true, state: { from: "list" } }` for React Router Link).',
```

PR #7347, opened 2026-04-08
refactor: replace Context.Consumer with useContext hook

langz:
useContext is considered the Modern, preferred method

### DIFF
--- a/packages/dnb-eufemia/src/extensions/payment-card/PaymentCard.tsx
+++ b/packages/dnb-eufemia/src/extensions/payment-card/PaymentCard.tsx
@@ -21,6 +21,7 @@ import {
 
 import { formatCardNumber } from './components/CardNumber'
 import CardFigure from './components/CardFigure'
+import type { CardData } from './components/CardFigure'
 
 import { ProductType, CardType, BankAxeptType } from './utils/Types'
 import Designs, { defaultDesign } from './utils/CardDesigns'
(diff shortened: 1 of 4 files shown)


tujoworker (approved):

packages/dnb-eufemia/src/extensions/payment-card/PaymentCard.tsx
```diff
@@ -187,43 +188,75 @@ function PaymentCard(props: PaymentCardProps) {
...
+  params: Record<string, unknown>
+  cardData: CardData
+  cardStatus: PaymentCardCardStatus
+  cardNumber: string
+  digits: string | number
+  id: string
+  skeleton: SkeletonShow
+  variant: PaymentCardVariant
+}) {
+  const { translation } = React.useContext(Context)
```
tujoworker:
```suggestion
  const { translation } = useContext(Context)
```

langz: Sure, I'm fine with that, but I do find it strange that we would like the same file to use both `React.useContext `and `useContext`, https://github.com/dnbexperience/eufemia/blob/ca61bcecb7a48aa2cc8f74f39d19ce3d09f3abdc/packages/dnb-eufemia/src/extensions/payment-card/PaymentCard.tsx#L139

PR #7354, opened 2026-04-08
chore(Theme): allow resetting `surface` with value `"default"` and remove setting class for `surface`

snorrekim:


packages/dnb-eufemia/src/shared/__tests__/Theme.test.tsx
```diff
@@ -118,6 +118,54 @@ describe('Theme', () => {
     )
   })
 
+  it('provides surface nest through the theme context', () => {
```
langz: I found the it statement a bit difficult to read and understand, here's a suggestion:
```suggestion
  it('resets surface to default when nested inside a dark surface theme', () => {
```

Not something we have to change.

packages/dnb-eufemia/src/shared/__tests__/Theme.test.tsx
```diff
@@ -118,6 +118,54 @@ describe('Theme', () => {
...
+    )
+
+    expect(receivedTheme).toEqual(
+      expect.objectContaining({
+        surface: 'default',
+      })
+    )
+  })
+
+  it('provides surface nest nest through the theme context', () => {
```
langz: I found the it statement a bit difficult to read and understand especially `nest nest` , here's a suggestion:

```suggestion
  it('provides surface through the theme context', () => {
```

Or

```suggestion
  it('inherits parent surface when no surface prop is given', () => {
```

Not something we have to change.

packages/dnb-eufemia/src/shared/Theme.tsx
```diff
@@ -130,7 +133,9 @@ export function getThemeClasses(theme: ThemeProps, className = null) {
     'eufemia-theme',
     name && `eufemia-theme__${name}`,
     name && variant && `eufemia-theme__${name}--${variant}`,
-    surface && `eufemia-theme__surface--${surface}`,
+    surface &&
+      surface !== 'default' &&
+      `eufemia-theme__surface--${surface}`,
```
langz: Seems like this, that `<Theme surface="default">` does not produce class `eufemia-theme__surface--default`, is the main change here. So we could potentially add a test asserting that this class is not added when surface="default".

langz (approved):
Looks good to me, though I don't 100% understand the problem and why this is needed 😅 👍

tujoworker (approved):

packages/dnb-eufemia/src/shared/__tests__/Theme.test.tsx
snorrekim: I just forgot to fix the names. I just added some random words. I'll update it.

packages/dnb-eufemia/src/shared/Theme.tsx
snorrekim: I'll add the test.

snorrekim:
I just realized that we should maybe make some changes to the defaults now that it is expected that we use `Theme` multiple times on a page for surface.

Right now, it creates a wrapper element by default, but that is only needed for the first element. And the wrapper element should not add classes for surface, because that isn't a page wide thing. We have `Theme.Context` but that's an easy mistake to make.

I believe we should default to no wrapper if it is a child context.

packages/dnb-eufemia/src/shared/__tests__/Theme.test.tsx
langz: Thank you @snorrekim 🙇

packages/dnb-eufemia/src/shared/Theme.tsx
langz: Thank you @snorrekim 🙇

packages/dnb-eufemia/src/shared/Theme.tsx
```diff
@@ -17,7 +17,10 @@ export type ThemeSizes = 'basis'
 export type PropMapping = string
 export type ContrastMode = boolean
 export type DarkMode = boolean
-export type ThemeSurface = 'dark'
+/**
+ * Changes component style based on background. Defaults to `undefined`.
```
langz: Not sure if this comment adds to my understanding of the topic.
The ThemeSurface of values 'default' | 'dark' defaults to undefined?

langz (approved):

tujoworker:
> I just realized that we should maybe make some changes to the defaults now that it is expected that we use `Theme` multiple times on a page for surface.
> 
> Right now, it creates a wrapper element by default, but that is only needed for the first element. And the wrapper element should not add classes for surface, because that isn't a page wide thing. We have `Theme.Context` but that's an easy mistake to make.
> 
> I believe we should default to no wrapper if it is a child context.

I see actually the value of having wrappers with classes for surface. Will show you on monay.

tujoworker:
> I just realized that we should maybe make some changes to the defaults now that it is expected that we use `Theme` multiple times on a page for surface.
> 
> Right now, it creates a wrapper element by default, but that is only needed for the first element. And the wrapper element should not add classes for surface, because that isn't a page wide thing. We have `Theme.Context` but that's an easy mistake to make.
> 
> I believe we should default to no wrapper if it is a child context.

Did remove the class in a new commit.

tujoworker:
Will be when #7424 is merged.

packages/dnb-design-system-portal/src/docs/uilib/components/anchor/Examples.tsx
```diff
@@ -198,56 +190,76 @@ export const AnchorExampleDisabled = () => (
...
       <Example>
-        <DarkSurface data-visual-test="anchor-contrast-no-hover">
+        <Section
+          surface="dark"
+          innerSpace
+          data-visual-test="anchor-contrast-no-hover"
+        >
           <Anchor
             href="/uilib/components/anchor"
             className="dnb-anchor--surface-dark dnb-anchor--no-hover"
```
langz, line 202: Should we manually set this class `dnb-anchor--surface-dark` when wrapped in a Section with surface="dark"?

This is a recurring pattern in this Examples.tsx file.

packages/dnb-eufemia/src/components/anchor/Anchor.tsx
```diff
@@ -82,10 +83,10 @@ export type AnchorProps = {
    */
   disabled?: boolean
   /**
-   * Changes component style based on background.
-   * Default: `undefined`
+   * Used to override the `surface` value from `Theme.Context`. Changes component style based on background.
+   * Default: `"default"` if there is no `Theme.Context`.
```
langz, line 87:
```suggestion
   * Default: `undefined` if there is no `Theme.Context`.
```

Is it actually the default value `"default"`?
I couldn't find any code in Anchor which sets that value as default, as we have in button.

https://github.com/dnbexperience/eufemia/blob/34448a0fee5f72354b9f70123e81a0b798d070f1/packages/dnb-eufemia/src/components/button/Button.tsx#L295

packages/dnb-eufemia/src/shared/ThemeDocs.ts
```diff
@@ -32,8 +32,8 @@ export const ThemeProperties: PropertiesTableProps = {
     status: 'optional',
   },
   surface: {
-    doc: 'Defines the surface the themed content is placed on. Use `dark` to apply styles intended for dark surfaces.',
-    type: ['"dark"'],
+    doc: 'Adjusts component appearance for use on dark backgrounds. Use `dark` when content is placed on a dark surface. Default is `undefined`.',
```
langz, line 35: Default is `undefined`, but it has a value "default".
I guess it's fine, but it makes it a bit difficult to understand for me whenever there's a value "default" which is not actually the default value 😅

packages/dnb-eufemia/src/shared/ThemeDocs.ts
```diff
@@ -32,8 +32,8 @@ export const ThemeProperties: PropertiesTableProps = {
...
+    type: ['"dark"', '"default"'],
```
langz, line 36: Is "default" the opposite value of "dark"?
If so, perhaps "light" could be an option?

"reset"? that has an explicit intent.

I don't fully grasp all of these changes, so ignore this suggestion if it's not in line with what's already discussed.

langz:
I guess the latest comments I provided was not worth looking into @tujoworker? I'm all fine with that.

packages/dnb-eufemia/src/shared/ThemeDocs.ts
tujoworker: Yes, the naming can be discussed more. 
@snorrekim came with the name suggestion. It sounded right in the meetings we had back then.

packages/dnb-eufemia/src/components/anchor/Anchor.tsx
tujoworker: Yeah, we should remove the default value actually.

packages/dnb-design-system-portal/src/docs/uilib/components/anchor/Examples.tsx
tujoworker: Looks not right. 😅

PR #7355, opened 2026-04-08
chore: migrate hardcoded spacing values to --spacing-* tokens

langz:


langz:
We discussed in a meeting that we rather want to go through all manually coded rem values used in margins and passing, and evaluate each occurrence if it’s suited to use the spacing tokens or not. Therefore we closed this.

PR #7356, opened 2026-04-08
refactor: centralize z-index scale with CSS custom properties

langz:
Add src/style/core/z-index.scss with a centralized z-index token scale:
  --z-index-section:  1
  --z-index-dropdown: 100
  --z-index-popover:  1000
  --z-index-tooltip:  1100
  --z-index-modal:    3000

Component-level custom properties (--modal-z-index, --popover-z-index, --tooltip-z-index, --section-z-index) now reference the centralized tokens via var() instead of hardcoded numeric values.

This makes the z-index layering system discoverable and adjustable from a single place while preserving backward compatibility for consumers who override the component-level custom properties.

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/usage/customisation/styling.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/usage/customisation/styling.mdx
@@ -252,6 +252,45 @@ This allows usage like:
 
 The `tokens-tailwind.css` files are available for all themes: `ui`, `sbanken`, and `carnegie`.
 
+## Z-Index Scale \{#z-index\}
+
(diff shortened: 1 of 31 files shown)


tujoworker (approved):
Would be nice to have documented. Maybe in the styling docs or so. So when we search a-index it will come up. Also we then can reference the docs internally.

langz:
> Would be nice to have documented. Maybe in the styling docs or so. So when we search a-index it will come up. Also we then can reference the docs internally.

Great, I've made an attempt to add this in https://github.com/dnbexperience/eufemia/pull/7356/changes/3ab179d5b205f8611ad737d8abac59d82ff3fd5f

PR #7365, opened 2026-04-09
feat(Portal): add carnegie theme to SidebarMenu

langz:


packages/dnb-design-system-portal/src/shared/menu/SidebarMenu.module.scss
```diff
@@ -24,6 +24,16 @@
...
     }
 
+    :global(.eufemia-theme__carnegie) & {
+      --is-inside-bg: var(--ca-color-pale-green);
+      --is-inside-active-path-color: var(--ca-color-ocean-green);
+      --is-inside-active-path-bg: var(--ca-color-pale-green);
+      --is-inside-active-path-bg-l-3: var(--ca-color-pale-green);
+      --dnb-sidebar-menu-text-color: var(--ca-color-ocean-green);
+      --dnb-sidebar-menu-hover-bg: var(--ca-color-pale-green);
+      --dnb-sidebar-menu-focus-bg: var(--ca-color-ocean-green);
```
langz: This does not look correct 😓

tujoworker (changes requested):
That did not turn out good. Can't see whats selected.
I have also tried this a couple of weeks ago. Basically used tokens for most of the portal styles. But gave up, because its so hard to get right. Its better UX does that job. Its easier to do in Figma.

packages/dnb-design-system-portal/src/shared/menu/SidebarMenu.module.scss
langz: Will look into finding appropriate values for the carnegie theme :)

langz:
> That did not turn out good. Can't see whats selected. I have also tried this a couple of weeks ago. Basically used tokens for most of the portal styles. But gave up, because its so hard to get right. Its better UX does that job. Its easier to do in Figma.

Hey, thanks for taking a look at my draft.
I just wanted to take a stab at improving the experience when user selects the carnegie theme :)

PR #7368, opened 2026-04-09
fix(Portal): replace hardcoded white with --color-white token in SearchBar

langz:


packages/dnb-design-system-portal/src/shared/menu/SearchBar.module.scss
```diff
@@ -20,7 +20,7 @@
     }
 
     .dnb-form-label {
-      color: white; /* only for axe-core */
+      color: var(--color-white); /* visually hidden label - needs contrast for axe-core */
```
langz:
```suggestion
      color: var(--color-white); /* only for axe-core */
```

PR #7378, opened 2026-04-09
refactor(Tooltip): use design tokens

tujoworker:
- Based on #7381
- Preview: https://chore-popover-tokens.eufemia-e25.pages.dev/uilib/components/tooltip/

PR #7381, opened 2026-04-09
chore(Popover): use design tokens and rename theme to `surface`

tujoworker:


tujoworker:
Its not possible to make this change, without first chaning a ton of tooltip visual snapshots and later back again. So this will will be a part of PR #7378

PR #7386, opened 2026-04-10
refactor(Switch): use design tokens

joakbjerk:
Adding the correct color tokens to this component will also require to change the size of the switch button so it fits inside the background, as the color of the button is meant to be in contrast to the background color. 

The switch button is kind of difficult to see here
Screenshot 2026-04-10 at 21 25 39

But easy to see here, as shown with the intended design in figma
Screenshot 2026-04-10 at 21 26 39

The figma sketches is also missing tokens for the different action states (hover, active etc.)


packages/dnb-eufemia/src/components/switch/__tests__/Switch.screenshot.test.ts
langz: Add  'carnegie' to this describe.each as well 🙏 

```suggestion
describe.each(['ui', 'sbanken', 'carnegie'])('Checked Switch for %s', (themeName) => {
```

tujoworker (approved):
Nice, looks good.

PR #7389, opened 2026-04-10
break(Forms)!: make ajv an optional peer dependency

langz:
BREAKING CHANGE: ajv and ajv-errors are now optional peer dependencies.

Motivation: Reduce bundle size for apps that don't use JSON Schema validation. In Eufemia Forms v11, Ajv is no longer included as a direct dependency — this PR extends that by making ajv and ajv-errors optional peer dependencies instead.


tujoworker:
What is the motivation behind this, and what does it improve?

At the moment, I mostly see downsides. 🤔
- We still have a lot of tests and some logic that rely on a specific Ajv version.
- If a consumer uses a different version, it may cause problems.
- It could make future breaking changes harder to handle.
- It seems like it may make Ajv harder to use than it is today.

That said, I do like the idea of not having Ajv as a dependency. 🤷‍♂️
I'll take a closer look into it.

langz:
> 
> 
> What is the motivation behind this, and what does it improve?
> 
> 
> 
> At the moment, I mostly see downsides. 🤔
> 
> - We still have a lot of tests and some logic that rely on a specific Ajv version.
> 
> - If a consumer uses a different version, it may cause problems.
> 
> - It could make future breaking changes harder to handle.
> 
> - It seems like it may make Ajv harder to use than it is today.
> 
> 
> 
> That said, I do like the idea of not having Ajv as a dependency. 🤷‍♂️
> 
> I'll take a closer look into it.

Motivation is just to not have ajv as a dependency, as Ajv is no longer included as a dependency in Eufemia Forms v11. Which reduces bundle size for applications that don't use JSON Schema validation.

So I was thinking the same for the rest of the ajv dependencies. But I might be wrong. 
I'm fine with us having a subset of ajv dependencies.

I now added "motivation" to the PR description as well.

tujoworker:
> Which reduces bundle size for applications that don't use JSON Schema validation.

No, it should not be a part of bundles when not used as it is now.

langz:
Too me it makes sense that the users who provide ajv and its version also provide ajv-errors and its version. 

Or else we definitely need to document which version of ajv that we support which matches our internal ajv error version, and thus can be provided from users. We might have it already ajv as peer dep. 

Will look more into it.

langz:
Closing this, as I feel that Eufemia is tightly coupled to specific ajv internals, and will still need to maintain specific versions of ajv and ajv related packages itself.

tujoworker:
I double-checked the bundling behavior and can confirm that `ajv` and `ajv-errors` is not included when it is not used.

tujoworker:
How ever, our own `ajvErrorsToFormErrors` and `ajvErrorsToOneFormError` functions are still a part of the bundle. Also in this PR (double checked it as well).

Screenshot 2026-04-14 at 07 28 58

PR #7398, opened 2026-04-13
chore(Theme): replace darkMode with `colorScheme` prop

tujoworker:
Add colorScheme prop with values auto, light, dark, and inherit. 

It was not documented, so no breaking change.

### DIFF
--- a/packages/dnb-eufemia/src/shared/__tests__/Theme.test.tsx
+++ b/packages/dnb-eufemia/src/shared/__tests__/Theme.test.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { render } from '@testing-library/react'
+import { act, render } from '@testing-library/react'
 import Context from '../Context'
 import type { ThemeAllProps } from '../Theme'
 import Theme from '../Theme'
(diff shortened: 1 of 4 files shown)


packages/dnb-eufemia/src/shared/__tests__/Theme.test.tsx
```diff
@@ -86,16 +86,134 @@ describe('Theme', () => {
...
+      </Theme>
+    )
+
+    const parent = document.querySelector('#parent')
+    const child = document.querySelector('#child')
+    expect(Array.from(parent.classList)).toContain(
+      'eufemia-theme__color-scheme--dark'
+    )
+    expect(Array.from(child.classList)).toContain(
+      'eufemia-theme__color-scheme--inherit'
```
langz: If it inherits colorScheme from parent, shouldn't this be `--dark` then, instead of `--inherit`?

langz (approved):

langz:
Probably just an edge-case, but which CSS class does one get if rendering a single `<Theme colorScheme="inherit" />` ?

tujoworker:
I think we should remove it actually. From a CSS point of view, it makes no sense.

PR #7424, opened 2026-04-14
feat(Section): add surface prop and provide it via Theme provider

tujoworker:
The default token is defined by UX.


langz (approved):
Looks great to me, my only potential improvement could be to add a test verifying that an explicit surface prop on `<Section>` takes precedence over an inherited `<Theme surface="dark">`. For example:

```
<Theme surface="dark">
  <Section surface={undefined}>content</Section>  // should NOT get dark classes?
</Theme>
```

PR #7425, opened 2026-04-14
chore: stop AI agents from auto-updating changelog files

langz:
As we most likely do not want to have entries in our changelog files, based on previous PR experiences.


tujoworker (approved):
Yeah, I agree. Even though we were explicit about “new” content, everything still ended up in there all the time.

PR #7427, opened 2026-04-14
chore(Theme): rename default surface to `light`

tujoworker:
Follow up of #7354 and new review comments.


packages/dnb-eufemia/src/components/button/Button.tsx
```diff
@@ -110,7 +110,7 @@ export type ButtonProps = {
    */
   variant?: ButtonVariant
   /**
-   * Used to override the `surface` value from `Theme.Context`. Changes component style based on background. Defaults to `"default"` if there is no `Theme.Context`.
+   * Used to override the `surface` value from `Theme.Context`. Changes component style based on background. Defaults to `"light"` if there is no `Theme.Context`.
```
langz:
```suggestion
   * Used to override the `surface` value from `Theme.Context`. Changes component style based on background. Defaults to `undefined` if there is no `Theme.Context`.
```

langz (approved):
Thanks, looks great :) 

Maybe @snorrekim should add his thoughts in regards to default vs light naming

PR #7429, opened 2026-04-14
chore(Portal): update CodeBlock component with new toolbar

tujoworker:
Preview: https://chore-code-block.eufemia-e25.pages.dev/uilib/components/anchor/demos/

### DIFF
--- a/packages/dnb-design-system-portal/src/shared/tags/CodeBlock.tsx
+++ b/packages/dnb-design-system-portal/src/shared/tags/CodeBlock.tsx
@@ -3,19 +3,30 @@
  *
  */
 
-import React from 'react'
+import React, {
+  useCallback,
+  useContext,
+  useMemo,
+  useRef,
+  useState,
+} from 'react'
 import clsx from 'clsx'
 import { Highlight, Prism } from 'prism-react-renderer'
 import Tag from './Tag'
-import { Button, Space } from '@dnb/eufemia/src/components'
+import { Checkbox, Space, ToggleButton } from '@dnb/eufemia/src/components'
+import Theme from '@dnb/eufemia/src/shared/Theme'
+import type {
+  ThemeColorScheme,
+  ThemeSurface,
+} from '@dnb/eufemia/src/shared/Theme'
 import { makeUniqueId } from '@dnb/eufemia/src/shared/component-helper'
 import { Context } from '@dnb/eufemia/src/shared'
 import { createSkeletonClass } from '@dnb/eufemia/src/components/skeleton/SkeletonHelper'
 import {
   liveCodeEditorStyle,
-  toolbarStyle,
+  exampleBoxStyle,
   codeBlockStyle,
-  plainBackgroundStyle,
+  toolbarStyle,
 } from './CodeBlock.module.scss'
 import {
   LiveProvider,
@@ -33,6 +44,10 @@ export type CodeSectionProps = {
   hideToolbar?: boolean
   hideCode?: boolean
   hidePreview?: boolean
+  /**
+   * use surface="default" to just show the button to toggle surface, without forcing it to be dark by default
+   */
+  surface?: ThemeSurface
   reactLive?: boolean
   language?: string
   className?: string
@@ -49,7 +64,7 @@ const CodeBlock = ({
   reactLive: isReactLive,
   ...props
 }: CodeSectionProps) => {
-  const context = React.useContext(Context)
+  const context = useContext(Context)
 
   if (!language) {
     language = ((String(props && props.className).match(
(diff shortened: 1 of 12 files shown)


langz (approved):
Looks great 🚀 

Not sure if it's possible to make the bottom border of the live code block with a radius as well, as we have when "collapsed" in the toolbar

Image

How it looks when only displaying toolbar, where there's a border radius:

Image

packages/dnb-design-system-portal/src/shared/tags/CodeBlock.tsx
```diff
@@ -183,102 +191,140 @@ function LiveCode(props: LiveCodeProps) {
...
+            <ToggleButton
+              checked={!hidePreview}
+              onChange={({ checked }) => setHidePreview(!checked)}
+              size="medium"
+            >
+              Preview
+            </ToggleButton>
+          )}
+
+          {process.env.NODE_ENV === 'development' && (
```
langz: Will this only be displayed locally when doing "development"?
Would be nice to have it in deploy previews etc as well.

packages/dnb-design-system-portal/src/shared/tags/CodeBlock.tsx
```diff
@@ -183,102 +191,140 @@ function LiveCode(props: LiveCodeProps) {
...
-        {!global.IS_TEST && !hideCode && (
-          <Space
-            className={clsx(
-              'dnb-live-editor',
-              createSkeletonClass('code', context.skeleton)
-            )}
-            top={omitWrapper}
-            ref={editorElementRef}
-          >
-            <span className="dnb-sr-only">Code Editor</span>
```
langz, line 227: I see we now have      ` <Space title="Code Editor"` instead, does that need the className="dnb-sr-only as well, or fine as it is now?

langz:
Would be nice that the example code is not reset when clicking on other features/toggles like "Dark mode" in the toolbar:

https://github.com/user-attachments/assets/0fa3a84f-b9ec-4c5b-8ff8-25114cb690d5



I guess this happens because: _LivePreviewComponent, ToolbarComponent, and LiveEditorComponent are declared as arrow functions inside LiveCode and used as JSX elements (<LivePreviewComponent />). React compares component type by reference — since these get a new identity on every render, React will unmount and remount them on every state change_

langz:
A future feature request for the toolbar could be a reset button, to reset back to the original example.

packages/dnb-design-system-portal/src/shared/tags/CodeBlock.tsx
tujoworker: Its fine as it is when using title/aria-label.

langz:
I've not looked into the actual code used when toggling surface and dark mode, etc. but it's very important that it uses the same mechanics as the actual code and/or app users will do themselves. So that it can be used for testing with high accuracy.

tujoworker:
Did commit updates covering most of the feedback, expect of:
> A future feature request for the toolbar could be a reset button, to reset back to the original example.

langz:
> Looks great 🚀
> 
> Not sure if it's possible to make the bottom border of the live code block with a radius as well, as we have when "collapsed" in the toolbar
> 
> Image
> How it looks when only displaying toolbar, where there's a border radius:
> 
> Image

[I checked now](https://chore-code-block.eufemia-e25.pages.dev/uilib/components/anchor/demos/#anchor-in-headings), and it looks like great, I like the design, thanks :) 

Not too important, but I feel like if the codebox/block would have a border radius in the bottom, then it would feel even more like the whole "section" (example, toolbar, codesection) is one uniformed thing. As it would have the same border radius in the top and bottom of the "whole section".
Screenshot 2026-04-15 at 09 41 13


No need to spend time on this.

langz (approved):

packages/dnb-design-system-portal/src/shared/tags/CodeBlock.tsx
```diff
@@ -187,98 +194,115 @@ function LiveCode(props: LiveCodeProps) {
...
+              <Checkbox
+                checked={colorScheme !== 'dark' && surface === 'dark'}
+                onChange={({ checked }) =>
+                  setSurface(checked ? 'dark' : undefined)
+                }
+                size="medium"
+                label="Dark surface"
+              />
+            )}
+          </Space>
```
langz, line 283: I don't fully follow the logic, but when setting checked state based on other values, etc, does it exist a scenario where we would end up in a strange state after toggling dark mode on/off a few times and perhaps the surface as well.

tujoworker: Yes. The compatibility supporting both will vary from component to component and needs to be adjusted independently.

langz:
Looks very good now 🥇 

image

PR #7435, opened 2026-04-14
chore(Button): add missing surface prop documentation

langz:
The surface prop exists in Button's TS type with JSDoc but was missing from ButtonDocs.ts. Added it with the correct ThemeSurface union type.

### DIFF
--- a/packages/dnb-eufemia/src/components/button/ButtonDocs.ts
+++ b/packages/dnb-eufemia/src/components/button/ButtonDocs.ts
@@ -27,7 +27,7 @@ export const ButtonProperties: PropertiesTableProps = {
     status: 'optional',
   },
   surface: {
-    doc: 'Used to override the `surface` value from `Theme.Context`. Changes component style based on background. Defaults to `"default"` if there is no `Theme.Context`.',
+    doc: 'Used to override the `surface` value from `Theme.Context`. Changes component style based on background. Defaults to `undefined` if there is no `Theme.Context`.',
     type: ['"default"', '"dark"'],
     status: 'optional',
   },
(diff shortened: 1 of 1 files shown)


packages/dnb-eufemia/src/components/button/ButtonDocs.ts
```diff
@@ -26,6 +26,11 @@ export const ButtonProperties: PropertiesTableProps = {
     type: ['"primary"', '"secondary"', '"tertiary"', '"unstyled"'],
     status: 'optional',
   },
+  surface: {
+    doc: 'Used to override the `surface` value from `Theme.Context`. Changes component style based on background. Defaults to `"default"` if there is no `Theme.Context`.',
```
langz:
```suggestion
    doc: 'Used to override the `surface` value from `Theme.Context`. Changes component style based on background. Defaults to `undefined` if there is no `Theme.Context`.',
```

packages/dnb-eufemia/src/components/button/ButtonDocs.ts
```diff
@@ -26,6 +26,11 @@ export const ButtonProperties: PropertiesTableProps = {
     type: ['"primary"', '"secondary"', '"tertiary"', '"unstyled"'],
     status: 'optional',
   },
+  surface: {
+    doc: 'Used to override the `surface` value from `Theme.Context`. Changes component style based on background. Defaults to `undefined` if there is no `Theme.Context`.',
```
langz, line 30: Dependant on https://github.com/dnbexperience/eufemia/pull/7427, which states that it defaults to undefined in jsdoc

PR #7502, opened 2026-04-15
fix(GlobalStatus): prevent unintended visibility when show defaults to auto

langz:
Issue: https://v11.eufemia-e25.pages.dev/uilib/components/global-status/demos/
Deploy preview: https://fix-global-status-show-auto.eufemia-e25.pages.dev/uilib/components/global-status/demos/

Two issues from the class-to-functional conversion:

1. provider.init() received raw ownProps (show: undefined) instead of
   props with defaults (show: 'auto'). The provider add method sets
   show to true when undefined.

2. The onUpdate callback used truthy/falsy checks instead of strict
   boolean checks. The old class component used isTrue() which only
   returns true for literal true/1. Since 'auto' is truthy, it
   incorrectly triggered setIsActive(true) when any provider update
   occurred.

### DIFF
--- a/packages/dnb-eufemia/src/components/global-status/GlobalStatus.tsx
+++ b/packages/dnb-eufemia/src/components/global-status/GlobalStatus.tsx
@@ -337,7 +337,10 @@ function GlobalStatusComponent(ownProps: GlobalStatusProps) {
     useState<GlobalStatusResult>(() => {
       const provider = GlobalStatusProvider.create(ownProps.id)
       providerRef.current = provider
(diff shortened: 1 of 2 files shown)


langz:
I might have to consider if this is just a small issue, or if it's a sign of a bigger issue related to the refactoring from class component to functional component 🤔 

EDIT: After looking at it, I believe this was a one-off bug specific to GlobalStatus's unique show: 'auto' semantics, not a systemic issue with the class-to-functional migration.

tujoworker (approved):

PR #7504, opened 2026-04-15
break(NumberFormat)!: use detectCountryCode for E.164 phone formatting

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1776246147152709

- Integrate detectCountryCode into formatPhone so all phone formatting
  consumers (including NumberFormat.PhoneNumber) auto-detect country codes
  from spaceless E.164 (+) and 00-prefixed numbers for all countries.
- Move detectCountryCode utility from extensions/forms to shared so core
  NumberUtils can use it without depending on forms.
- Extract lightweight countryCodes list (CDC strings only) to avoid
  pulling the full countries data into the NumberFormat bundle.
- Remove duplicate country code detection from Value.PhoneNumber since
  formatPhone now handles it directly.
- Phone formatting no longer treats bare numbers starting with 47 as
  having a Norwegian country code. Use E.164 format (+47...) or 00
  prefix (0047...) to indicate a country code.
- Strip formatting from tel/sms link href for RFC 3966 compliance.
- Apply dashed CDC parenthesization consistently for all formatting paths.


langz:
This is now ready for review @tujoworker

tujoworker (approved):

PR #7513, opened 2026-04-16
feat(List): add `List.ShowMoreButton` to toggle items visibility

tujoworker:
Motivation: https://dnb-it.slack.com/archives/C081G91GCT0/p1776157948812889
Deploy preview: https://feat-list-show-more.eufemia-e25.pages.dev/uilib/components/list/demos/#with-showmorebutton-and-visiblecount

Introduces a decoupled show more/less button that links to `List.Container` via `useSharedState` and a shared `id` prop.

- New `List.ShowMoreButton` component (tertiary button with chevron icon)
- New `visibleCount` prop on `List.Container` to limit visible items
- Localized default labels (nb-NO, en-GB, sv-SE, da-DK)
- Demo example and tests


packages/dnb-eufemia/src/components/list/ListShowMoreButton.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  return (
+    <Button
+      variant="tertiary"
+      text={expanded ? resolvedShowLess : resolvedShowMore}
+      icon={expanded ? chevron_up : chevron_down}
+      iconPosition="right"
+      onClick={handleClick}
+      aria-expanded={expanded}
+      {...rest}
```
langz, line 53: Not too important, but I guess users now can overrride all the props listed above.
Could consider either spreading rest props first before the props we "must have", or add omit to the type or such.

Worth a consideration.

packages/dnb-eufemia/src/components/list/ListShowMoreButton.tsx
```diff
@@ -0,0 +1,62 @@
...
+  }, [expanded, update])
+
+  return (
+    <Button
+      variant="tertiary"
+      text={expanded ? resolvedShowLess : resolvedShowMore}
+      icon={expanded ? chevron_up : chevron_down}
+      iconPosition="right"
+      onClick={handleClick}
+      aria-expanded={expanded}
```
langz, line 51: Could there also be a use for `aria-controls`?

packages/dnb-eufemia/src/components/list/ListShowMoreButton.tsx
tujoworker: We do omit the types actually I think.

langz:
It looks very very nice, great feature 🐰

langz:
EDIT: Nothing to see here,  it actually works as expected, I'm not sure how I was able to experience it, can't reproduce it.

Could it "grow" in the opposite direction when pressing the "vis mer"-button.
When clicking, the screen sort of "jumps". I would hope that the "vis mer"-button would stay put in the same place, and then the list would "grow"/expand further down and potentially out of view.

Here's how it is today:


https://github.com/user-attachments/assets/d968f009-6485-4124-88c1-634a04fb79ff

langz:
Should we have some animation?

packages/dnb-eufemia/src/components/list/ListShowMoreButton.tsx
```diff
@@ -0,0 +1,62 @@
...
+  type SharedStateId,
+} from '../../shared/helpers/useSharedState'
+import useTranslation from '../../shared/useTranslation'
+
+export type ListShowMoreButtonSharedState = {
+  expanded: boolean
+}
+
+export type ListShowMoreButtonProps = {
+  id: SharedStateId
```
langz: I'm not 100% sure about naming, but could a user expect the `id` prop to be a regular HTML `id` prop that would render in the Button in the HTML? That's for instance how it works on `List.Container`, etc.

Not sure how we usually treat or do this in the codebase.

An alternative could be to introduce a `listId`, or something

langz (approved):

langz:
When used with Accordions, if one is open before clicking "vis mindre" or "vis mer", it now collapses. I feel like it should still be expanded:

https://github.com/user-attachments/assets/4e831aec-f242-4968-ac13-743813b3d516



https://github.com/user-attachments/assets/8900819c-e59a-4b9a-81c9-474ad357a031

langz:
Could it be named only `List.ShowMore`? I think `List.ShowMoreButton` is fine, but it's a bit longer than the other List.X components.

Or is it a `List.ShowMoreToggleButton`?

tujoworker:
> Should we have some animation?

Yes, it does now.

tujoworker:
> Could it be named only `List.ShowMore`? I think `List.ShowMoreButton` is fine, but it's a bit longer than the other List.X components.
> 
> Or is it a `List.ShowMoreToggleButton`?

We keep it as. Should work fine. It describes that it is just a button.

tujoworker:
> When used with Accordions, if one is open before clicking "vis mindre" or "vis mer", it now collapses. I feel like it should still be expanded:

To make that happen, we need to make sure the components get not unmounted. So we end up having them in the DOM now with a `hidden` HTML attribute.

PR #7525, opened 2026-04-16
fix(List): prevent space-between gaps in accordion items

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1776326485319869

Reprod of issue: https://stackblitz.com/edit/fy7efl3t-8wainmbw?file=src%2FApp.tsx
Reprod of the fix: https://stackblitz.com/edit/fy7efl3t-zhfjf9ha?file=package.json,src%2FApp.tsx


langz:
Seems like this fix is working @tujoworker 
I've tried a few, but this is the only one I've experienced working.
Do you think this has any potential downsides?

tujoworker (approved):
Should be fine I think. We could consider to enable/disable it base if it's open or closed.

PR #7539, opened 2026-04-17
chore(Code,Pre): add design token support

tujoworker:


packages/dnb-eufemia/src/elements/code/style/code-mixins.scss
```diff
@@ -52,6 +54,10 @@
   font-size: inherit;
   font-family: var(--font-family-monospace);
 
+  // Pre is using syntax highlighting to determine the text color, this property works as a fallback color in case of no highlighting
+  color: var(--token-color-background-positive-subtle);
+  background-color: var(--token-color-text-neutral);
```
langz: Are these correct? color using background token, and background-color using color token?
It might result in the correct colors, but semantically it looks strange.
Perhaps this is dark mode or surface, etc 😂 

It's also fine to keep it as is, but could add a comment explaining why.

packages/dnb-eufemia/src/elements/code/style/code-mixins.scss
```diff
@@ -19,9 +19,11 @@
   text-decoration: inherit;
   line-height: calc(var(--line-height-xx-small--em) / #{$code-scale});
 
-  color: inherit;
+  background-color: var(--token-color-background-positive-subtle);
+  box-shadow: inset 0 0 0 0.0625rem
+    var(--token-color-stroke-neutral-subtle);
 
-  border-radius: 0.1875em;
+  border-radius: 0.25rem;
```
langz: `em` to `rem`?
Thoughts about scaling with font size

langz (approved):

packages/dnb-eufemia/src/style/themes/ui/prism/dnb-prism-theme.ts
```diff
@@ -7,7 +7,7 @@ const italic = 'italic' as const
 const prismTheme = {
   plain: {
     color: 'var(--color-mint-green-12)',
-    backgroundColor: '#222',
+    backgroundColor: 'var(--token-color-text-neutral)',
```
langz: backgroundColor using color token?
It might result in the correct colors, but semantically it looks a bit off.
Perhaps this is dark mode or surface, etc 😂 

It's also fine to keep it as is, but could add a comment explaining why.

langz (approved):

packages/dnb-eufemia/src/style/themes/ui/prism/dnb-prism-theme.ts
tujoworker: Yeah, we can try using `--token-color-decorative-first-intense-static` instead.

tujoworker: no :D 
Screenshot 2026-04-17 at 10 22 37

langz: Reason for using these "semantically wrong" tokens is because we don't have tokens for this scenario I would guess.

Not sure if it's then best to not use tokens and rather just use colors, or just use tokens but add comments to the usage of them

tujoworker: Can't find a token. But I think we should go for a static color there anyway. So use the good old --color- then.

PR #7540, opened 2026-04-17
chore: improve design tokens filter

tujoworker:
When I search for design tokens, I don't find the base/advanced filter helpful. But rather want to quickly see e.g. all "On dark" tokens. Therefore, I think it makes more sense to add this kind of filter system.

Preview: https://chore-design-tokens-filter-i.eufemia-e25.pages.dev/uilib/usage/customisation/theming/design-tokens/tokens/#background

### DIFF
--- a/packages/dnb-design-system-portal/src/uilib/utils/designTokens.ts
+++ b/packages/dnb-design-system-portal/src/uilib/utils/designTokens.ts
@@ -40,7 +40,44 @@ export type ThemeName =
   | 'sbankenLight'
   | 'sbankenDark'
   | 'carnegie'
-export type TokenAudience = 'base' | 'state' | 'advanced'
+export type TokenModifier =
+  | 'hover'
+  | 'pressed'
+  | 'focus'
+  | 'disabled'
+  | 'inverse'
+  | 'ondark'
+  | 'onlight'
+  | 'onsubtle'
+  | 'subtle'
+  | 'bold'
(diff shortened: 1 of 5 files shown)


packages/dnb-design-system-portal/src/uilib/utils/designTokens.ts
```diff
@@ -40,7 +40,44 @@ export type ThemeName =
...
+  'ondark',
+  'onlight',
+  'onsubtle',
+  'subtle',
+  'bold',
+  'base',
+  'muted',
+  'intense',
+  'alternative',
+  'destructive',
```
langz, line 76: I think this can result in strange behavior as destructive is also a group?

Let's say filtering by "Hover", that works intuitively — it shows only tokens that have a hover variant (action-hover, destructive-hover, etc.), never the base token itself.

Filtering by "Destructive" catches the base destructive token too, because its group name happens to be in the modifier list. So it behaves like a group filter, not a modifier filter.

Worth considering

langz (approved):

packages/dnb-design-system-portal/src/uilib/utils/designTokens.ts
tujoworker: Yes, its AND based. But I think that's good. Because you can go strict, by just using one at a time. But you can also enhance from there.

PR #7543, opened 2026-04-17
break(Space)!: rewrite spacing system to use a dynamic solution

tujoworker:

PR #7557, opened 2026-04-19
break(Space)!: use CSS custom properties for responsive margins

tujoworker:
Second attempt after #7543
I think its a good move so far.

The only downside I can see, is the style prop gets overwritten. But that's the same issue with `innerSpace`. So for now we then show such a message for the `space` (could do the same for innerSpace too):

Screenshot 2026-04-19 at 13 43 36

---

**Update:** We do use the old system as long as not a responsive media query value is provided.


packages/dnb-eufemia/src/components/space/style/dnb-space.scss
```diff
@@ -5,132 +5,167 @@
...
+  //       var(--spacing-x-large)
+  //   );
+  // }
+  // // 10rem
+  // &--xx-large-x2#{&}--x-large {
+  //   margin-#{$direction}: calc(
+  //     var(--spacing-xx-large) + var(--spacing-xx-large) +
+  //       var(--spacing-x-large)
+  //   );
+  // }
```
langz: Do we need all of thse comments?

packages/dnb-eufemia/src/components/space/style/dnb-space.scss
```diff
@@ -5,132 +5,167 @@
...
+  &--x-small,
+  &--small,
+  &--medium,
+  &--large,
+  &--x-large,
+  &--xx-large,
+  &--xx-large-x2 {
+    &:not([style*='--margin-#{$short}-']) {
+      &::before {
+        content: 'Missing --margin-#{$short}-* custom property!';
```
langz: Can this end up in production?
Or interfere in anyway we would not like?

packages/dnb-eufemia/src/components/space/style/dnb-space.scss
tujoworker: I'll remove it for now. Could use it as a fallback. But I think we can go without it.

packages/dnb-eufemia/src/components/space/style/dnb-space.scss
tujoworker: As it is now, it will. Can think about it to not show it in prod.

packages/dnb-eufemia/src/components/space/SpacingUtils.ts
```diff
@@ -87,8 +89,150 @@ export const createSpacingProperties = (
...
+    // Nothing resolvable – skip work and return an empty style
+    if (Object.keys(merged).length === 0) {
+      return {}
+    }
+
+    mediaSpace = {
+      small: merged,
+      medium: merged,
+      large: merged,
+    }
```
langz: applySpacing always produces style even for non-responsive spacing — A simple top="large" generates --margin-t-s: 2rem; --margin-t-m: 2rem; --margin-t-l: 2rem; (identical values across all breakpoints)

packages/dnb-eufemia/src/components/global-error/GlobalError.tsx
```diff
@@ -131,16 +131,15 @@ export default function GlobalError(localProps: GlobalErrorAllProps) {
     children: text,
   }
 
-  const params = {
+  const params = applySpacing(attributes, {
+    ...attributes,
```
langz: Looks a bit suspicious.
If attributes contains spacing props they'll appear as both spacing input and leaked into the target before stripping.

langz:
Instead of overwriting style, could this be an option?

Only emit custom properties when spacing is actually responsive
This is the simplest high-impact fix. Non-responsive spacing (e.g., top="large") doesn't need custom properties at all — the CSS classes already handle it. Only emit --margin-* when the space prop contains media breakpoints:

```
// In computeMarginProperties, before the else branch:
} else {
  const merged = mergeSpacing(individualProps)
  if (Object.keys(merged).length === 0) {
    return {} // no spacing at all
  }

  // NEW: skip custom properties for non-responsive spacing.
  // The CSS classes alone are sufficient.
  return {}
}
```

This eliminates the style attribute for the vast majority of components (any using simple top="large" etc.), completely avoiding the overwrite concern for non-responsive use cases.

packages/dnb-eufemia/src/components/space/SpacingUtils.ts
tujoworker: Yeah, like `innerSpace` does today. We may look into improving this.

langz: `<Button top="large">` produces `style="--margin-t-s: 2rem; --margin-t-m: 2rem; --margin-t-l: 2rem;"`

langz (approved):

langz:
Thoughts regarding naming:

The innerSpace properties use --space-* while the new margin properties use --margin-*. This is semantically clear (space = padding, margin = margin), but it creates a naming asymmetry — the prop is called `space` but its custom properties are --margin-*, while the prop called `innerSpace` uses --space-*. 
A developer seeing --space-t-s could reasonably think it maps to the `space` prop, not `innerSpace`.

tujoworker:
> Thoughts regarding naming:
> 
> The innerSpace properties use --space-* while the new margin properties use --margin-_. This is semantically clear (space = padding, margin = margin), but it creates a naming asymmetry — the prop is called `space` but its custom properties are --margin-_, while the prop called `innerSpace` uses --space-*. A developer seeing --space-t-s could reasonably think it maps to the `space` prop, not `innerSpace`.

I, I also saw it and thought, we should rename --space to --padding. I guess that's the right move. But we can do (I will do) that in another PR.

tujoworker:
> Instead of overwriting style, could this be an option?
> 
> Only emit custom properties when spacing is actually responsive This is the simplest high-impact fix. Non-responsive spacing (e.g., top="large") doesn't need custom properties at all — the CSS classes already handle it. Only emit --margin-* when the space prop contains media breakpoints:
> 
> ```
> // In computeMarginProperties, before the else branch:
> } else {
>   const merged = mergeSpacing(individualProps)
>   if (Object.keys(merged).length === 0) {
>     return {} // no spacing at all
>   }
> 
>   // NEW: skip custom properties for non-responsive spacing.
>   // The CSS classes alone are sufficient.
>   return {}
> }
> ```
> 
> This eliminates the style attribute for the vast majority of components (any using simple top="large" etc.), completely avoiding the overwrite concern for non-responsive use cases.

yeah, but then we need to get back the static defined CSS for spacing. I also thought this could be the right way to go. But yeah, maybe we should un-comment them again and make that happen.

The only possible downside is that with the current approach, it would be possible to define theoretically every value. Including values above 10rem, which is not possible in v10. But I think that's fine.

packages/dnb-eufemia/src/components/global-error/GlobalError.tsx
tujoworker: [Improved](https://github.com/dnbexperience/eufemia/pull/7557/changes/ff3da3350c45d8dabcb40833bbd24884378b9e88).

PR #7577, opened 2026-04-20
chore(StyleLint): add rule for supporting scoped dark mode

tujoworker:
Based on #7576 so tests do not fail like so:

Screenshot 2026-04-20 at 09 39 51

### DIFF
--- a/packages/dnb-eufemia/scripts/stylelint/plugins/__tests__/token-name-policy.test.ts
+++ b/packages/dnb-eufemia/scripts/stylelint/plugins/__tests__/token-name-policy.test.ts
@@ -330,4 +330,66 @@ describe('token-name-policy stylelint rule', () => {
       temp.cleanup()
     }
   })
+
+  it('flags var(--token-...) references inside :root selector', async () => {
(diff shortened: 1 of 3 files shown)


packages/dnb-eufemia/scripts/stylelint/plugins/__tests__/token-name-policy.test.ts
```diff
@@ -330,4 +330,66 @@ describe('token-name-policy stylelint rule', () => {
...
+    expect(
+      errors.some((e) =>
+        e.text.includes('must not be used in :root, html, or body')
+      )
+    ).toBe(false)
+  })
+
+  it('allows var(--token-...) declarations in tokens.scss files', async () => {
+    const errors = await lintWithRule({
+      code: ':root { --token-color-background-neutral: var(--dnb-greyscale-0); }',
```
langz, line 385: For this test, should `var(--dnb-greyscale-0);` also be a token that matches `TOKEN_REFERENCE_REGEX`?

langz (approved):

PR #7582, opened 2026-04-20
refactor(pre): use design tokens

langz:
Replace var(--color-white) with var(--token-color-text-neutral-inverse) and var(--color-black-80) with var(--token-color-background-marketing) in the pre/code block styling.

### DIFF
--- a/packages/dnb-eufemia/src/elements/code/style/code-mixins.scss
+++ b/packages/dnb-eufemia/src/elements/code/style/code-mixins.scss
@@ -56,8 +56,8 @@
   font-family: var(--font-family-monospace);
 
   // Pre is using syntax highlighting to determine the text color, this property works as a fallback color in case of no highlighting
-  color: var(--color-white);
-  background-color: var(--color-black-80);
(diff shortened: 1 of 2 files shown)


tujoworker (approved):

tujoworker (approved):

packages/dnb-eufemia/src/elements/code/style/code-mixins.scss
```diff
@@ -56,8 +56,8 @@
   font-family: var(--font-family-monospace);
 
   // Pre is using syntax highlighting to determine the text color, this property works as a fallback color in case of no highlighting
-  color: var(--color-white);
-  background-color: var(--color-black-80);
+  color: var(--token-color-text-neutral-inverse);
```
tujoworker, line 59: I think we should double check that.

tujoworker:
I think we should close this for now. The changes where towards the `pre` tag. But they are not used right away. We do not have a "component" for it either. We have this comment above the colors:

> // Pre is using syntax highlighting to determine the text color, this property works as a fallback color in case of no highlighting

PR #7589, opened 2026-04-20
refactor(Value.Block): use design tokens

langz:
Replace var(--color-black-80) with var(--token-color-text-neutral-onlight) for placeholder text color.

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/ValueBlock/style/dnb-value-block.scss
+++ b/packages/dnb-eufemia/src/extensions/forms/ValueBlock/style/dnb-value-block.scss
@@ -81,7 +81,7 @@
   }
 
   &__placeholder {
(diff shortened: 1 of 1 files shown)


packages/dnb-eufemia/src/extensions/forms/ValueBlock/style/dnb-value-block.scss
```diff
@@ -81,7 +81,7 @@
   }
 
   &__placeholder {
-    color: var(--color-black-80);
+    color: var(--token-color-text-neutral-onlight);
```
langz:
```suggestion
    color: var(--token-color-text-neutral-alternative);
```

joakbjerk (approved):
Looks good to me 🫡

PR #7592, opened 2026-04-20
chore(Heading): use design tokens for debug style

langz:
Replace var(--color-black-20) with var(--token-color-text-neutral-subtle) and var(--color-black) with var(--token-color-text-neutral) in debug label styling.


tujoworker (approved):

langz (commented):
Fancy 💯

PR #7597, opened 2026-04-20
refactor(Stat): use design tokens

langz:
Replace --color-success-green with --token-color-text-positive, --color-fire-red with --token-color-text-destructive, --color-black-80/55/20 with --token-color-text-neutral-onlight/alternative/subtle, --color-mint-green-25 with --token-color-background-positive-subtle, and --color-fire-red-8 with --token-color-background-error-subtle.


tujoworker:
Done here #7638

PR #7598, opened 2026-04-20
refactor(VippsWalletButton): use design tokens

langz:
Replace var(--color-white) with var(--token-color-text-neutral-inverse) for button text color.


tujoworker:
White should be used.

PR #7600, opened 2026-04-20
chore(Field.Password): toggle visibility while input has focus

langz:
The show/hide password button did not work when clicked while the input had focus. The ToggleVisibilityButton was defined as a component via useCallback, causing React to unmount and remount the button DOM node when the Password component re-rendered on blur. This broke the click event since the mousedown target was removed before mouseup.

Changed to useMemo returning a JSX element instead, keeping a stable DOM node across re-renders.

Issue in v11: https://v11.eufemia-e25.pages.dev/uilib/extensions/forms/feature-fields/more-fields/Password/demos/
How it works in v10:
https://eufemia.dnb.no/uilib/extensions/forms/feature-fields/more-fields/Password/demos/
Deploy preview of the fix:
https://fix-password-toggle-visibili.eufemia-e25.pages.dev/uilib/extensions/forms/feature-fields/more-fields/Password/demos/

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/Password/Password.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/Password/Password.tsx
@@ -93,7 +93,15 @@ function Password({
     [props]
   )
 
-  const ToggleVisibilityButton = useCallback(() => {
+  const preventFocusChange = useCallback((event: React.MouseEvent) => {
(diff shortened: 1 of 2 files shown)


tujoworker (approved):
Nice 👍

langz:
This fix is not working yet :(

Works now

PR #7604, opened 2026-04-20
chore(GlobalStatus): restore autoScroll={false} behavior

langz:
The class-to-functional conversion dropped the autoScroll check in scrollToStatus(). Setting autoScroll={false} was silently ignored, causing the component to always scroll into view when opened.

This restores the original guard that returns early when autoScroll is false, using propsRef to access the current value inside the memoized callback.


tujoworker (approved):
Nice!

PR #7605, opened 2026-04-20
chore(Autocomplete): fix indicator label direction

langz:
The default labelDirection was changed to 'vertical', but existing usages with a label prop should retain horizontal layout to preserve previous appearance.

- Autocomplete: indicator ProgressIndicator now explicitly horizontal


langz:
TEsted, and it works as expected.

PR #7606, opened 2026-04-20
chore: add showDefaultLabel and horizontal labelDirection to random value examples

langz:
Added labelDirection="horizontal" to the circular random value example and added showDefaultLabel with labelDirection="horizontal" to the linear random value example.

Preview: https://fix-progress-indicator-examp.eufemia-e25.pages.dev/uilib/components/progress-indicator/#circular-progressindicator-with-a-label-in-a-horizontal-direction

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/components/progress-indicator/Examples.tsx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/components/progress-indicator/Examples.tsx
@@ -94,6 +94,7 @@ export const ProgressIndicatorCircularRandomExample = () => (
               type="circular"
               progress={value}
               showDefaultLabel
+              labelDirection="horizontal"
               noAnimation
             />
             <Button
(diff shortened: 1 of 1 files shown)


tujoworker (approved):

langz:
Not sure if this fixes it.
I think what I want to fix is actually related to migration of `<FormRow centered>` -> `<Flex.Horizontal align="centered">` which does not result in the same visual result

packages/dnb-design-system-portal/src/docs/uilib/components/progress-indicator/Examples.tsx
```diff
@@ -244,6 +245,8 @@ export const ProgressIndicatorLinearRandomExample = () => (
             <ProgressIndicator
               type="linear"
               progress={value}
+              showDefaultLabel
+              labelDirection="horizontal"
```
langz:
```suggestion
```

Issue #7615, opened 2026-04-20
v11: CountryFlag – "Eufemia Forms" demo changes value when opening/closing


langz:
Split from #7601.

### Description

The "Eufemia Forms" demo changes value when opening/closing (from Sweden to Norway).

- **v11:** https://v11.eufemia-e25.pages.dev/uilib/components/country-flag/demos/#eufemia-forms
- **prod:** https://eufemia.dnb.no/uilib/components/country-flag/demos/#eufemia-forms

https://github.com/user-attachments/assets/10582a53-c0be-481c-9146-e9bb86dd038b

PR #7624, opened 2026-04-20
fix(DrawerList): prevent value reset on blur for lazy-loaded data

langz:
Fixes: Fixes https://github.com/dnbexperience/eufemia/issues/7615

When components like SelectCountry, SelectCurrency, and PhoneNumber lazy-load their full data on focus, the data change caused selectedItem indices to become stale in prepareDerivedState. This resulted in the value resetting to the first item (e.g. Norway).

Recalculate selectedItem directly in the data-change block of prepareDerivedState, co-locating the fix with the cause.


tujoworker (approved):
👏

PR #7633, opened 2026-04-21
chore(PaginationProvider): use useIsomorphicLayoutEffect for onEnd disptch

langz:
The onEnd callback was dispatched via useEffect (after paint) while all other pending callbacks in the same file use useIsomorphicLayoutEffect (before paint) to match the class component's setState callback timing.

This inconsistency could cause a visible frame where the UI shows the ended state before the onEnd callback fires. Switching to useIsomorphicLayoutEffect aligns the timing with the stated design intent and the other callbacks.

### DIFF
--- a/packages/dnb-eufemia/src/components/pagination/PaginationProvider.tsx
+++ b/packages/dnb-eufemia/src/components/pagination/PaginationProvider.tsx
@@ -379,8 +379,10 @@ const PaginationProvider = (props: any) => {
     ]
   )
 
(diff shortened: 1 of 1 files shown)


tujoworker (approved):

tujoworker:
I renamed it to chore. My thoughts are – this is caused by e.g. React v19 or the class to fc refactor. So its an "internal issue" that caused it. No need to be mentioned in the release log.

PR #7658, opened 2026-04-21
fix(DrawerList): clip arrow to triangle shape to prevent diamond on overscroll

langz:
Motivation is to fix this:

https://github.com/user-attachments/assets/c950d1ce-6799-4851-882c-de1f9889f1a2

Deploy preview: https://fix-drawer-list-arrow-clip-p.eufemia-e25.pages.dev/uilib/components/autocomplete/demos/


langz:
This does not work as expected.
We only want to remove the "bottom" part of the triangle which actually is a diamond, so that it's only looking like a triange pointing upwards.

EDIT: should work as expected now.

tujoworker (approved):

langz:
Fixed in https://github.com/dnbexperience/eufemia/pull/7668

PR #7661, opened 2026-04-22
chore(List): replace forwardRef with ref prop in test mock

langz:
React 19 passes ref as a regular prop, making forwardRef unnecessary. Convert the MockLink test helper to a plain function component that accepts ref as a prop.


tujoworker (approved):

tujoworker:
Better to use "chore". Refactor I think will appear in the release log.

langz:
> Better to use "chore". Refactor I think will appear in the release log.

Great, I'll remember to do so.
I guess we'll probably have to manually edit the next release log as well, as I guess there will be a few fixes here and there 😅

PR #7662, opened 2026-04-22
chore: convert class components to function components in tests/stories

langz:
Convert React.PureComponent class components to function components with hooks (useState + useEffect) in:
- Dialog.stories.tsx (ModalRerenderExample)
- Modal.stories.tsx (ModalRerenderExample)
- Accordion.test.tsx (DidRender)

The withComponentMarkers.test.tsx class component is intentionally kept as it tests class component support.


tujoworker (approved):
Nice find!

PR #7681, opened 2026-04-22
refactor: replace default React imports with named imports

langz:
Replace `import React from 'react'` with specific named imports (e.g. `import { useState, useEffect } from 'react'`) and type-only imports (e.g. `import type { ReactNode } from 'react'`) across the codebase. This removes unnecessary default namespace imports while keeping the code functionally identical.


tujoworker:
Icon files are generated. We need to remove it in the generation script.

langz:
> Icon files are generated. We need to remove it in the generation script.

Thanks for the feedback @tujoworker, I added this here https://github.com/dnbexperience/eufemia/pull/7681/changes/7a31390111d208f84684596b785b2798229ca570#diff-7e475a7b7d44593192fb46446467c20ea24f99f66d2363721092dbaff376595f :)

langz:
@tujoworker, @snorrekim & @joakbjerk, this PR is now ready for review :) 

Personally I like this a lot, and believe it is "the way to go" with current React ecosystem conventions.

langz:
> @tujoworker, @snorrekim & @joakbjerk, this PR is now ready for review :)
> 
> Personally I like this a lot, and believe it is "the way to go" with current React ecosystem conventions.

@tujoworker, @snorrekim & @joakbjerk, I think this PR is ready for review again :)

tujoworker:
I think we can mark it as a feature. It needs to be mentioned in the release lot. 

⚠️ I'm not sure about the conferences for all the bundlers. That nees to be confirmed.

tujoworker (approved):

tujoworker:
@codex review

langz:
> I think we can mark it as a feature. It needs to be mentioned in the release lot.
> 
> ⚠️ I'm not sure about the conferences for all the bundlers. That nees to be confirmed.

I think the only theoretical risk is extremely old bundler configurations that can't resolve named imports from react's CJS entry, but that would already be broken with React 17+ JSX transform I guess?

But yeah, a manual sanity check of the built artifacts could perhaps provide full confidence.

PR #7686, opened 2026-04-22
chore: rename grid SCSS files to follow dnb- prefix convention

langz:
Renamed grid-container.scss → dnb-grid-container.scss and grid-item.scss → dnb-grid-item.scss to match the naming convention used by all other component SCSS files (e.g. dnb-button.scss, dnb-badge.scss).


tujoworker (changes requested):
That means they are processed as additional packages. All SCSS files starting with dnb- are treated as “packages”, and we already do this in many places.
I don’t think the rewrite gives us any additional benefits.

Docs: https://eufemia.dnb.no/contribute/getting-started/making-changes/#css-packages

langz:
> That means they are processed as additional packages. All SCSS files starting with dnb- are treated as “packages”, and we already do this in many places. I don’t think the rewrite gives us any additional benefits.
> 
> Docs: https://eufemia.dnb.no/contribute/getting-started/making-changes/#css-packages

Aha, thanks. 
I do agree that it does not make sense. I just saw a pattern, that this was named differently. But was not aware of these docs, thanks.

PR #7701, opened 2026-04-22
refactor(Skeleton): use design tokens

tujoworker:
Preview: https://refactor-skeleton-design-tok.eufemia-e25.pages.dev/uilib/components/skeleton/


langz (commented):
I assume the new colors is correct, but I find them very very difficult to see on a white background, almost impossible. So I hope no apps will use skeletons on a white background.
Would think it's some kind of WCAG color contrast issue?
Perhaps it's only intended to be used on "non-white backgrounds"?

image


https://refactor-skeleton-design-tok.eufemia-e25.pages.dev/uilib/components/skeleton/demos/

tujoworker:
I agree. I also think they are too light. Lets tackle this better.

langz (approved):
Thanks, looks better now.

PR #7711, opened 2026-04-23
chore(Checkbox): use stroke tokens for border instead of background tokens

langz:

Replace background-category tokens with stroke tokens for border properties:
- --checkbox-color-border-on: use stroke-selected instead of background-selected
- --checkbox-color-border-on--disabled: use stroke-action-disabled instead of background-action-disabled


joakbjerk:
The border should be the same color as the background when checked, and checked in disabled state. The Figma sketches does not have a border color tokens for the checked states, as the components do not have borders in sketches when checked. Setting the borders to transparent in the code makes the checkboxes slightly smaller than the should be when checked.

I feel like the background color tokens works here, since the intent is for the borders to be the same color as the background. We could of course change the `--checkbox-color-border-on` be `--token-color-stroke-selected` as they are the same color in all the themes for now, but setting `--checkbox-color-border-on--disabled` to '--token-color-stroke-action-disabled' does not work, as they are different colors 🤷

tujoworker:
@langz will add comment in another PR.

PR #7712, opened 2026-04-23
chore(Switch): use stroke tokens for border instead of background tokens

langz:

Replace background-category tokens with stroke tokens for border-color:
- checked state: use stroke-selected instead of background-selected
- disabled checked state: use stroke-action-disabled instead of background-action-disabled


joakbjerk:
Same situation as [here](https://github.com/dnbexperience/eufemia/pull/7711#issuecomment-4304443079) 🫡

tujoworker:
@langz will add comment in another PR.

PR #7719, opened 2026-04-23
chore: remove forwardRef from test

langz:


packages/dnb-eufemia/src/elements/Element.tsx
```diff
@@ -97,7 +97,7 @@ function Element(localProps: ElementAllProps) {
 
   const isFragment = Tag === React.Fragment
 
-  if (!isFragment && typeof Tag !== 'function' && ref) {
```
langz, line 100: I will look a bit more into this.

langz: We don’t have class components anymore, but could be that other apps could pass those in

packages/dnb-eufemia/src/elements/Element.tsx
```diff
@@ -97,7 +97,7 @@ function Element(localProps: ElementAllProps) {
...
+  if (!isFragment && ref) {
```
langz:
```suggestion
  if (!isFragment && typeof Tag !== 'function' && ref)) {
```

packages/dnb-eufemia/src/elements/Element.tsx
```diff
@@ -97,7 +97,7 @@ function Element(localProps: ElementAllProps) {
 
   const isFragment = Tag === React.Fragment
 
-  if (!isFragment && typeof Tag !== 'function' && ref) {
+  if (!isFragment && typeof Tag !== 'function' && ref)) {
```
langz:
```suggestion
  if (!isFragment && typeof Tag !== 'function' && ref) {
```

PR #7723, opened 2026-04-23
fix(List): improve a11y for `List.Item.Action`

tujoworker:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1776938215268739


langz (approved):
Would an existing axeComponent test have seen/found this issues, or perhaps that's not needed?

karini6:
> Would an existing axeComponent test have seen/found this issues, or perhaps that's not needed?

I noticed there is a test I would expect to catch this in List.test.tsx, I was surprised to see that it passed. Good reminder that axe tests don't catch as much as we'd like them to, I guess :D

tujoworker:
> Would an existing axeComponent test have seen/found this issues, or perhaps that's not needed?

Yes. Now we have one.

tujoworker:
> > Would an existing axeComponent test have seen/found this issues, or perhaps that's not needed?
> 
> I noticed there is a test I would expect to catch this in List.test.tsx, I was surprised to see that it passed. Good reminder that axe tests don't catch as much as we'd like them to, I guess :D

Hmm, yeah, you think about this one I guess:

https://github.com/dnbexperience/eufemia/blob/main/packages/dnb-eufemia/src/components/list/__tests__/List.test.tsx#L227

tujoworker:
Yes, we had a problem in our axe tests! Will either add the fix here in in a separate PR, depending on the outcome.

PR #7726, opened 2026-04-23
chore(Tag): remove dead variables

langz:

Replace icon-category token with semantically correct background token for --tag-icon-background-color in active/pressed state:
- use background-action-pressed instead of icon-action-pressed

The --tag-icon-background-color feeds into the SVG circle fill, making background the correct token category.


joakbjerk:
The variable did not actually do anything, so lets remove it all together 😝

PR #7736, opened 2026-04-23
chore(Portal): implement "Light mode" in CodeBlock when portal is in Dark mode

tujoworker:


packages/dnb-design-system-portal/src/shared/tags/CodeBlock.tsx
```diff
@@ -164,6 +164,8 @@ function LiveCode(props: LiveCodeProps) {
     props.surface
   )
 
+  console.log('colorScheme', colorScheme)
+
```
langz:
```suggestion
```

langz (approved):

packages/dnb-design-system-portal/src/shared/tags/CodeBlock.tsx
```diff
@@ -278,12 +281,20 @@ function LiveCode(props: LiveCodeProps) {
...
+                          ? inheritedDark
+                            ? 'light'
+                            : 'dark'
+                          : undefined
+                      )
+                    }}
                     size="medium"
-                    label="Dark mode"
+                    label={inheritedDark ? 'Light mode' : 'Dark mode'}
                   />
```
langz, line 296: Should we reset colorScheme when the inherited mode changes?

I think I was into it here https://github.com/dnbexperience/eufemia/pull/7737#pullrequestreview-4164604223

Does not really matter, as it's dev-only so the blast radius is small, but it's the kind of subtle desync that leads to confusing behavior when toggling dark mode at the portal level while also toggling it per-CodeBlock.

langz (approved):

packages/dnb-design-system-portal/src/shared/tags/CodeBlock.tsx
tujoworker: Yeah, let's have a try with this approach. I think it makes sense when everything falls into place.

PR #7737, opened 2026-04-23
chore(Portal): add dark mode toggle button

tujoworker:


packages/dnb-design-system-portal/src/shared/menu/PortalToolsMenu.tsx
```diff
@@ -5,12 +5,17 @@ import { H2, P, Anchor } from '@dnb/eufemia/src'
...
+import ToggleDarkMode from '../../core/ToggleDarkMode'
 import ChangeLocale from '../../core/ChangeLocale'
 import ChangeStyleTheme from '../../core/ChangeStyleTheme'
 import { buttonStyle } from './PortalToolsMenu.module.scss'
 import type { TooltipPlacement } from '@dnb/eufemia/src/components/tooltip/types'
 import type { ButtonProps } from '@dnb/eufemia/src/components/Button'
 
+const isDev =
+  process.env.NODE_ENV === 'development' ||
+  process.env.GATSBY_IS_PREVIEW === 'true'
```
langz, line 17: 👍 
Nice, I like having stuff in preview 🤗

langz (approved):
Great, I think it looks good :) 

One could consider looking into which "logic" one would like to have in terms of priority between dark mode set in "portal tools" vs dark mode set in "code block", but I guess that's just nitpicking, as users will easily see if it's actually darkmode or not regardless of what the checkboxes say:

https://github.com/user-attachments/assets/30a72bd3-e3e7-4f36-bf8f-3785b7eea7f3

PR #7740, opened 2026-04-23
refactor(Tabs): use design tokens

tujoworker:
Preview: https://refactor-tabs-design-tokens.eufemia-e25.pages.dev/uilib/components/tabs/


packages/dnb-eufemia/src/components/tabs/Tabs.tsx
```diff
@@ -1277,7 +1273,6 @@ Tip: Check out other solutions like <Tabs.Content id="unique">Your content, outs
               onMouseEnter={onMouseEnterHandler}
               onClick={onClickHandler}
               onKeyUp={onKeyDownHandler}
-              onMouseDown={onMouseDown}
```
langz, line 1280: Assuming this is correct and intentional 🙏

packages/dnb-eufemia/src/components/tabs/style/dnb-tabs.scss
```diff
@@ -239,26 +245,75 @@
...
+      color: var(--tab-title-color--selected);
+    }
+
+    &:not([disabled]).selected:active &__title {
+      color: var(--token-color-text-action-pressed);
     }
 
     // selected only
+    /* stylelint-disable-next-line no-duplicate-selectors */
     &:not([disabled]).selected &__title {
```
langz: Can/should this and line 304 be merged together into one?

langz (approved):

packages/dnb-eufemia/src/components/tabs/Tabs.tsx
tujoworker: Yeah, we don’t get a mouse-down effect with that change. No tests depend on it, but I’m pretty sure it was added intentionally at some point.

We should probably dig a bit deeper before removing it.

That said, even in production today, there is no visible active-state change, even though the CSS styles for it exist.

packages/dnb-eufemia/src/components/tabs/style/dnb-tabs.scss
tujoworker: The second one is the active state. But yeah, could be joined.

packages/dnb-eufemia/src/components/tabs/Tabs.tsx
tujoworker: Before v11 we had this comment. I'll dig deeper.

> // Once we press with mouse, the focus makes the scroll view move
> // In order to enhance UX, we prevent that and only allow a real click

tujoworker: I'm not able to reproduce any issue like that. I think its fine to remove.

packages/dnb-eufemia/src/components/tabs/style/dnb-tabs.scss
tujoworker: Done

PR #7746, opened 2026-04-24
fix(FieldBlock): render label as span when not associated with a form field

langz:
When FieldBlock is used without a forId (i.e. the content is not a form field), the label element was rendered as <label> without a for attribute. This caused Chrome's accessibility audit to flag it with `'a label is not associated with a form field'`. Now renders as <span> when `forId` is not provided, and `<label>` when it is.

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/FieldBlock/style/dnb-field-block.scss
+++ b/packages/dnb-eufemia/src/extensions/forms/FieldBlock/style/dnb-field-block.scss
@@ -37,9 +37,6 @@ fieldset.dnb-forms-field-block {
   &--width {
     &-stretch {
       flex-grow: 1;
-      label.dnb-form-label {
-        max-width: none;
-      }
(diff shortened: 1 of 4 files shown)


packages/dnb-eufemia/src/extensions/forms/FieldBlock/style/dnb-field-block.scss
```diff
@@ -70,6 +67,10 @@ fieldset.dnb-forms-field-block {
 
     margin-right: 0;
 
+    .dnb-forms-field-block--width-stretch & {
+      max-width: none;
+    }
+
```
langz, line 73: Added this so that this will apply to whether the underlying element is `<label>`, `<span>`, or `<legend>`.
So that it will still look as expected, no matter what underlying element it is used for the given label

tujoworker (approved):
👏

PR #7759, opened 2026-04-24
fix(Dropdown, Autocomplete): add outline (border) to DrawerList

tujoworker:
In order to support dark mode, the drawer list needs an outline. 
The motivation is to get the same look as the Popover. Except for the different drop shadow.


langz (approved):
Good good, assuming they use the same tokens as in popover then 🙏

packages/dnb-eufemia/src/fragments/drawer-list/style/themes/dnb-drawer-list-theme-ui.scss
```diff
@@ -208,6 +217,9 @@
 
   &__arrow::before {
     @include utilities.sharpDropShadow();
```
langz:
```suggestion
```

I guess this can be removed, as we manually set box-shadow on the line below

PR #7766, opened 2026-04-24
refactor(Timeline): use design tokens

joakbjerk:
Preview: https://feat-timeline-design-tokens.eufemia-e25.pages.dev/uilib/components/upload/

### DIFF
--- a/packages/dnb-eufemia/src/components/timeline/__tests__/__snapshots__/Timeline.test.tsx.snap
+++ b/packages/dnb-eufemia/src/components/timeline/__tests__/__snapshots__/Timeline.test.tsx.snap
@@ -108,7 +108,7 @@ exports[`Timeline scss should match style dependencies css 1`] = `
   padding: 0;
   color: var(--token-color-icon-neutral);
   background-color: var(--token-color-background-page-background);
-  --border-color: var(token-color-stroke-neutral);
+  --border-color: var(--token-color-stroke-neutral);
   --border-width: 0.0625rem;
   box-shadow: inset 0 0 0 var(--border-width) var(--border-color);
   border-color: transparent;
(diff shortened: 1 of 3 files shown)


packages/dnb-eufemia/src/components/timeline/__tests__/__snapshots__/Timeline.test.tsx.snap
```diff
@@ -128,20 +103,47 @@ exports[`Timeline scss should match style dependencies css 1`] = `
   justify-content: center;
   height: auto;
   padding: 0;
+  color: var(--token-color-icon-neutral);
+  background-color: var(--token-color-background-page-background);
+  --border-color: var(token-color-stroke-neutral);
```
langz:
```suggestion
  --border-color: var(--token-color-stroke-neutral);
```

packages/dnb-eufemia/src/components/timeline/style/themes/dnb-timeline-theme-ui.scss
langz, line 1: Should this file be removed? Or is that not the case for all conversions?
In quite a few design token conversetions this file gets removed, but it might not be so for all?

langz (approved):

packages/dnb-eufemia/src/components/timeline/style/themes/dnb-timeline-theme-ui.scss
joakbjerk: this file still contains font weights and sizes, which are not tokenized as far as I know

langz: Ahhh, there's so many levels to this. We will never actually be done supporting tokens for any components, as it will always be leftovers. Haha. 

Probably not that bad, but sort of feels like it never really is finished.

PR #7769, opened 2026-04-25
fix(MCP): make @modelcontextprotocol/sdk a devDependency — ⚠️ install it yourself to use the MCP server (beta)

tujoworker:
Move it to devDependencies and document that applications using the MCP server need to install it themselves. The reason is because of a vulnerability issue reported by some security scanners. Not the one we use in this repo.

This dependency was not used by apps at runtime, so it should not be a vulnerability issue when you did not use the MCP server.

As we still state the MCP server is in beta, I think this move is ok to do as a fix release.


langz (approved):

langz:
We should also remember to make a post about it in Slack, sp that users expect this to be failing, and will know to make adjustments.

langz:
We could also add a note in v11-info.mdx, perhaps AI agents would easier understand and fix the "breaking change"

PR #7779, opened 2026-04-27
feat(Table): add useTableKeyboardNavigation "plugin" hook

tujoworker:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1776687358381519
Deploy preview: https://feat-table-keyboard-navigati.eufemia-e25.pages.dev/uilib/components/table/info/#keyboard-navigation
Demo: https://feat-table-keyboard-navigati.eufemia-e25.pages.dev/uilib/components/table/demos/#table-with-keyboard-navigation

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/components/table/Examples.tsx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/components/table/Examples.tsx
@@ -1098,13 +1098,9 @@ export const KeyboardNavigation = () => (
                     <Field.String label="Name" labelSrOnly value="Ola" />
                   </Td>
                   <Td>
-                    <Field.String
-                      label="Email"
-                      labelSrOnly
-                      value="ola@example.no"
-                    />
+                    <Field.Email labelSrOnly value="ola@example.no" />
                   </Td>
-                  <Td>
+                  <Td verticalAlign="middle">
                     <Button
                       variant="tertiary"
                       icon={trashIcon}
@@ -1122,14 +1118,40 @@ export const KeyboardNavigation = () => (
                     />
                   </Td>
                   <Td>
-                    <Field.String
-                      label="Email"
+                    <Field.Email
                       labelSrOnly
                       value="kari@example.no"
                       multiline
                     />
                   </Td>
+                  <Td verticalAlign="middle">
+                    <Button
+                      variant="tertiary"
+                      icon={trashIcon}
+                      tooltip="Delete row"
+                    />
+                  </Td>
+                </tr>
+                <tr>
                   <Td>
+                    <Field.Number
+                      label="Quantity"
+                      labelSrOnly
+                      showStepControls
+                      width="small"
+                      value={1}
+                    />
+                  </Td>
+                  <Td>
+                    <Field.Number
+                      label="Quantity"
+                      labelSrOnly
+                      showStepControls
+                      width="small"
+                      value={2}
+                    />
+                  </Td>
+                  <Td verticalAlign="middle">
                     <Button
                       variant="tertiary"
                       icon={trashIcon}
--- a/packages/dnb-design-system-portal/src/docs/uilib/components/table/Examples.tsx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/components/table/Examples.tsx
@@ -38,12 +38,15 @@ import {
   Section,
   Stat,
 } from '@dnb/eufemia/src/components'
-import Table from '@dnb/eufemia/src/components/table/Table'
-import Th from '@dnb/eufemia/src/components/table/TableTh'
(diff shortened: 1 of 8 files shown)


langz:
Nice, great 👍 

I assume we'll release v11.0.2 before this, or is this feature urgent?

packages/dnb-eufemia/src/components/table/useTableKeyboardNavigation.tsx
```diff
@@ -0,0 +1,236 @@
...
+  enabled?: boolean
+}
+
+function getCells(table: HTMLTableElement) {
+  const rows = Array.from(table.querySelectorAll('tr'))
+  return rows.map((row) =>
+    Array.from(row.querySelectorAll<HTMLTableCellElement>('td, th'))
+  )
+}
+
```
langz, line 24: Do we support nested tables, how would this work in that scenario?
Like if a `td` contains a `table`, etc.

packages/dnb-design-system-portal/src/docs/uilib/components/table/Examples.tsx
```diff
@@ -1068,6 +1071,94 @@ export const AccordionRow = () => {
   )
 }
 
+export const KeyboardNavigation = () => (
```
langz, line 1074: Could we render/display this example?
I don't think we do in our docs/demos.mdx right now?

langz:
Should use role="grid"?
https://www.w3.org/WAI/ARIA/apg/patterns/grid/#:~:text=and%20row%20selection.-,Data%20Grids%20For%20Presenting%20Tabular%20Information,are%20interacting%20with%20the%20grid.

packages/dnb-eufemia/src/components/table/useTableKeyboardNavigation.tsx
tujoworker: no, Tables should not be nested because of a11y.

tujoworker:
> Should use role="grid"? https://www.w3.org/WAI/ARIA/apg/patterns/grid/#:~:text=and%20row%20selection.-,Data%20Grids%20For%20Presenting%20Tabular%20Information,are%20interacting%20with%20the%20grid.

not as long as we use semantic tables.

packages/dnb-design-system-portal/src/docs/uilib/components/table/Examples.tsx
tujoworker: Yes, the deploy preview: https://feat-table-keyboard-navigati.eufemia-e25.pages.dev/uilib/components/table/demos/#table-with-keyboard-navigation

langz: Great, it was not there when I first added the comment you see. 
Or perhaps I was just too blind to find it.

langz:
https://github.com/user-attachments/assets/6db6a627-801b-482f-a474-f99954bad0e0

tujoworker:
> Screen.Recording.2026-04-27.at.13.58.07.mov

This is fixed with #7804

langz:
Since ArrowUp/ArrowDown always navigates the grid, I assume it will be problematic when used with field that can change value on ArrowUp/ArrowDown? Like number fields and / or datepicker. But I guess this will collide either way, perhaps users using this hook should be informed of that.

As of right now, both happens, both navigation and changing value on one key press.


https://github.com/user-attachments/assets/34629d6e-17d9-4f8e-9972-c2cc4a76c427

langz:
One of the agents I used suggests _Bug: Memory leak in setupFocusCleanup setupFocusCleanup adds a focusout event listener to the table element but it's never cleaned up. The useEffect cleanup only removes the keydown listener_

packages/dnb-eufemia/src/components/table/useTableKeyboardNavigation.tsx
```diff
@@ -0,0 +1,236 @@
...
+      col = Math.min((grid[row]?.length ?? 1) - 1, col + 1)
+      break
+  }
+
+  const nextCell = grid[row]?.[col]
+  if (nextCell) {
+    event.preventDefault()
+    focusCell(nextCell)
+  }
+}
```
langz, line 190: When pressing ArrowLeft on the leftmost cell (or ArrowUp on the top row), row/col clamps to the same value, so nextCell is the current cell. preventDefault() still fires, blocking browser defaults for no reason.

```suggestion
const currentCell = grid[position.row]?.[position.col]
const nextCell = grid[row]?.[col]
if (nextCell && nextCell !== currentCell) {
  event.preventDefault()
  focusCell(nextCell)
}
```

langz (approved):
🪝

PR #7780, opened 2026-04-27
chore: use InlineImg component for font-size screenshot in accessibility intro

langz:
Motivation: Not able to see the image here https://main.eufemia-e25.pages.dev/uilib/intro/03-accessibility/#200-you-said

But I guess perhaps this could be based on some environment stuff, as I'm able to see it here https://eufemia.dnb.no/uilib/intro/03-accessibility/#200-you-said

🤔


langz:
Tested, and it works as expected

PR #7793, opened 2026-04-27
fix: read key from element instead of props for React 19 compatibility

langz:
In React 19, 'key' is no longer included in element.props. Read key directly from element.key instead of destructuring from element.props. Affects flex/utils.tsx and flex/Container.tsx.

Motiation: https://react.dev/warnings/special-props


joakbjerk (approved):
👍

PR #7794, opened 2026-04-27
refactor(Avatar): use design tokens

joakbjerk:
Preview: https://feat-avatar-design-tokens.eufemia-e25.pages.dev/uilib/components/avatar/


packages/dnb-eufemia/src/components/avatar/style/dnb-avatar.scss
```diff
@@ -6,6 +6,8 @@
 @use '../../../style/core/utilities.scss' as utilities;
 
 .dnb-avatar {
+  --color: var(--token-color-text-neutral-inverse);
+  --background-color: var(--token-color-decorative-first-bold);
```
langz: Not entirely sure if we want to set this both here and in --primary, seems redundant, as we also always set --primary class as default I believe.

langz:
Remove `require.resolve('../style/themes/dnb-avatar-theme-ui.scss')` from Avatar.test.tsx 🙏

langz (approved):

packages/dnb-eufemia/src/components/avatar/style/dnb-avatar.scss
joakbjerk: Looks like color has to be defined at the top there, so that both `primary` and `secondary` inherits the same text color, but the background color can be moved to primary 🙏

packages/dnb-eufemia/src/components/avatar/style/dnb-avatar.scss
```diff
@@ -6,6 +6,7 @@
 @use '../../../style/core/utilities.scss' as utilities;
 
 .dnb-avatar {
+  --color: var(--token-color-text-neutral-inverse);
```
tujoworker: I think we need more info in a var than just color. --text-color? --basis-color?

joakbjerk: text color 🙏

PR #7804, opened 2026-04-27
chore(Portal): fix CodeBlock color in dark mode

tujoworker:
Screenshot 2026-04-27 at 14 22 45

### DIFF
--- a/packages/dnb-eufemia/src/style/themes/ui/prism/dnb-prism-theme.ts
+++ b/packages/dnb-eufemia/src/style/themes/ui/prism/dnb-prism-theme.ts
@@ -7,7 +7,7 @@ const italic = 'italic' as const
 const prismTheme = {
   plain: {
     color: 'var(--token-color-decorative-third-muted-static)',
-    backgroundColor: 'var(--token-color-background-marketing)',
+    backgroundColor: 'transparent',
   },
   styles: [
     {


packages/dnb-eufemia/src/style/themes/ui/prism/dnb-prism-theme.ts
```diff
@@ -7,7 +7,7 @@ const italic = 'italic' as const
 const prismTheme = {
   plain: {
     color: 'var(--token-color-decorative-third-muted-static)',
-    backgroundColor: 'var(--token-color-background-marketing)',
+    backgroundColor: 'transparent',
```
tujoworker, line 10: We just use the background color of the pre element. I think that's just fine.

langz: What does it mean, that all CodeBlocks must then be in a pre element?
But perhaps all CodeBlocks already is in a pre element?

langz (approved):

packages/dnb-eufemia/src/style/themes/ui/prism/dnb-prism-theme.ts
tujoworker: Yes, it is.

PR #7808, opened 2026-04-28
style(Portal): enable dark mode ☀️🌑

tujoworker:


langz (approved):
Which decorater to use? (chore/feat)?
Does not really matter I guess, but using chore could make it so that we'll be able to release a bugfix release if needed.

tujoworker:
We can use style I think.

PR #7818, opened 2026-04-28
chore: replace deprecated KeyboardEvent.which with event.key

langz:
https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/which

KeyboardEvent.which is deprecated. Replace all usage with the modern event.key string-based API.

Changes:
- whatInput.ts: ignoreMap and specificMap now use key name strings instead of numeric keycodes, and reads event.key instead of event.which
- specificKeys() signature changed from number[] to string[]
- Updated all callers in Tabs.tsx, MenuRoot.tsx, component-helper.ts
- Updated whatInput tests to use key names


tujoworker (approved):
Nice!

PR #7826, opened 2026-04-28
chore(Avatar, Badge, Dropdown, Input, Pagination, Tag, Timeline, Tooltip, ScrollView): remove null entries from defaultProps objects

langz:
Remove redundant null entries from defaultProps. When using extendPropsWithContext, null defaults are equivalent to undefined (both allow context override), so only meaningful non-null defaults need to be kept.


tujoworker (approved):
Before, when we had class components, we used the "null" declarations to help during the props merge process. But I guess this works now because tests are fine with the change.

PR #7830, opened 2026-04-28
feat(Forms): add `globalStatusAutoClose` prop to DataContext.Provider

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1777373660252399

Add a new 'globalStatusAutoClose' prop (default: true) to DataContext.Provider and Form.Handler. When set to `false`, the GlobalStatus error summary remains visible after the user starts editing fields, instead of disappearing immediately. This allows consumers to keep the GlobalStatus banner persistent until all validation errors are resolved, matching the behavior of the classic Input component approach. 

Changes: 
- DataContext.Provider: add globalStatusAutoClose prop, guard the showAllErrorsRef reset in handlePathChange 
- Form.Handler: forward globalStatusAutoClose to Provider 
- FormStatus: also check for content when deciding to update GlobalStatus, ensuring items are removed when errors resolve

### DIFF
--- a/packages/dnb-eufemia/src/components/form-status/FormStatus.tsx
+++ b/packages/dnb-eufemia/src/components/form-status/FormStatus.tsx
@@ -433,7 +433,7 @@ function FormStatusComponent(
       fillCache()
 
       if (state === 'error') {
-        if (show) {
+        if (show && getContent(restOwnProps)) {
           globalStatusRef.current?.update(
             statusId,
             {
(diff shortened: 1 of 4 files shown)


tujoworker:
So maybe the question is; should we change the default behavior?

langz:
> So maybe the question is; should we change the default behavior?

Yes, that's a valid question as well.
I'm not sure if the code today is made with a purpose in mind or not. I would hope it is 😓

langz:
Fixed in https://github.com/dnbexperience/eufemia/pull/7877

PR #7833, opened 2026-04-28
fix(ScrollView): guard ResizeObserver.observe against null ref

tujoworker:


langz:
Thanks so much ❤️ 
I do get a bit eager from time to time, and don't want to spam PR reviews for "simple refactoring" where tests go green, but here I made a mistake. Here is the history:
- https://github.com/dnbexperience/eufemia/pull/7831
- https://github.com/dnbexperience/eufemia/pull/7834

Closing this for now, sorry for the inconvenience and thanks for the help ❤️

PR #7839, opened 2026-04-28
fix: prevent focus trap infinite loop with dynamically added elements

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1777384168492689

Not sure about this fix, seems complicated...

[StackBlitz of the issue](https://stackblitz.com/edit/u6dlg7ok-11jwk6xi?file=package.json,src%2FApp.tsx)
[StackBlitz with the fix](https://stackblitz.com/edit/u6dlg7ok-rsgtofvq?file=src%2FApp.tsx) (Not working)

When a Dialog is open, elements added to the DOM dynamically (e.g. toast close buttons rendered via React portals) could remain focusable since InteractionInvalidation only processed elements present at activation time. This caused an infinite focus-fighting loop between the Dialog's focus trap and the portal, freezing the browser tab.

Add a MutationObserver to InteractionInvalidation that automatically invalidates any new elements added to the DOM while the invalidation is active, respecting existing bypass selectors and bypass elements.

### DIFF
--- a/packages/dnb-eufemia/src/shared/helpers/InteractionInvalidation.ts
+++ b/packages/dnb-eufemia/src/shared/helpers/InteractionInvalidation.ts
@@ -21,11 +21,13 @@ export class InteractionInvalidation {
   bypassElements: Array<TargetElement>
   bypassSelectors: Array<TargetSelector>
   _nodesToInvalidate: Array<HTMLElementNode>
+  _observer: MutationObserver | null
   options: InteractionInvalidationOptions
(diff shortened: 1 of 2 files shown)


tujoworker:
Yeah, that was a lot of code. 
I also look into the issue now.

langz:
> Yeah, that was a lot of code. 
> 
> I also look into the issue now.

Also it did not work 🤣

PR #7851, opened 2026-04-29
refactor: add border radius tokens to components

joakbjerk:
Removing redundant theme files and verifying that components are using the correct tokens will be done in seperate pull requests after this one is merged 🙏


langz:
Wow, great work 🚀 

Which commit decorator should we use for this?
I'm not against chore, but I also do like that this is sort of visible in our release logs, like we've done with tokens, by using refactor(X), so it shows like so https://github.com/dnbexperience/eufemia/releases/tag/v11.0.3

joakbjerk:
> Wow, great work 🚀
> 
> Which commit decorator should we use for this? I'm not against chore, but I also do like that this is sort of visible in our release logs, like we've done with tokens, by using refactor(X), so it shows like so https://github.com/dnbexperience/eufemia/releases/tag/v11.0.3

I guess we could do refactor? 🤔

langz (approved):
Looks great to me, I guess there's probably other places we could consider using these tokens as well:
- code-mixins.scss?
- possibly other places.

But I guess it's quite difficult to find all occurrences, and if they should be changed to be a token or not. How did you attack this challenge? Did you check Figma?

joakbjerk:
> Looks great to me, I guess there's probably other places we could consider using these tokens as well:
> 
>     * code-mixins.scss?
> 
>     * possibly other places.
> 
> 
> But I guess it's quite difficult to find all occurrences, and if they should be changed to be a token or not. How did you attack this challenge? Did you check Figma?

Used AI to replace the hard coded border radius values with their token equivalent sizes, and gone through figma quickly to double check if they are correct. Looks like 90% ish of them are correct, but some components will need to use a different token to achieve the correct border radius. I'll create separate pull request for the components in question, just to prevent a bunch of smaller pull requests that adds border radius per component 😅

tujoworker (approved):

PR #7742, opened 2026-04-24
refactor(Portal): use design tokens and implement new Eufemia logo

tujoworker:
Preview: https://chore-portal-tokens.eufemia-e25.pages.dev/

PR #7854, opened 2026-04-29
chore(Portal): prevent search autocomplete animation glitch on first interaction

langz:
Motivation: https://github.com/dnbexperience/eufemia/pull/7742#issuecomment-4315919628
Deploy preview: https://fix-portal-search-animation.eufemia-e25.pages.dev/

- Add noAnimation prop to the search Autocomplete to prevent the
  drawer-list slide animation from firing on an empty container when the
  input is clicked before any search results have loaded.
- Use showIndicatorItem instead of showIndicator in the onType handler so
  the dropdown displays a loading indicator item while waiting for search
  results, rather than opening empty.
- Close the dropdown on search errors via setHidden to avoid leaving the
  indicator item visible.

### DIFF
--- a/packages/dnb-design-system-portal/src/shared/menu/SearchBar.tsx
+++ b/packages/dnb-design-system-portal/src/shared/menu/SearchBar.tsx
@@ -92,6 +92,7 @@ export const SearchBarInput = () => {
       mode="async"
       showClearButton
       noScrollAnimation
+      noAnimation
       preventSelection
       disableFilter
       fixedPosition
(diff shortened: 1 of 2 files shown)


tujoworker (changes requested):
The real issue is this visible all the time. I look into this.

Image

langz:
> The real issue is this visible all the time. I look into this.
> 
> Image

Good good, yeah, I've tried a few different approaches now, but never solved it 😢

PR #7862, opened 2026-04-29
fix(Dialog, Drawer, ScrollView): handle content movement when scrollbar appears and add `scrollbarGutter` property

tujoworker:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1746779578744739

Deploy preview Dialog: https://feat-modal-scroll-view.eufemia-e25.pages.dev/uilib/components/dialog/demos/
Deploy preview Drawer: https://feat-modal-scroll-view.eufemia-e25.pages.dev/uilib/components/drawer/demos/

OK support for this UX enhancement feature: https://caniuse.com/mdn-css_properties_scrollbar-gutter

The reason why we can't enable it by default are such cases:
Screenshot 2026-04-29 at 13 07 01

But for the Dialog and Drawer I think its save to enable it by default, as long as spacing is not disabled, because then it means, there is something special/custom content.

### DIFF
--- a/packages/dnb-eufemia/src/components/dialog/DialogDocs.ts
+++ b/packages/dnb-eufemia/src/components/dialog/DialogDocs.ts
@@ -106,6 +106,11 @@ export const DialogProperties: PropertiesTableProps = {
     type: 'React.RefObject',
     status: 'optional',
   },
+  scrollbarGutter: {
+    doc: 'Reserves space for the scrollbar gutter, preventing layout shifts when content overflows. When set to `auto`, it enables `stable` for the `information` variant with spacing. Defaults to `auto`.',
+    type: ['"auto"', '"stable"'],
+    status: 'optional',
+  },
   contentRef: {
     doc: 'To get the inner content Element, pass in your own React ref.',
     type: 'React.RefObject',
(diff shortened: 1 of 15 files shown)


packages/dnb-eufemia/src/components/dialog/DialogDocs.ts
```diff
@@ -106,6 +106,11 @@ export const DialogProperties: PropertiesTableProps = {
     type: 'React.RefObject',
     status: 'optional',
   },
+  scrollbarGutter: {
+    doc: 'Reserves space for the scrollbar gutter, preventing layout shifts when content overflows. When set to `auto`, it enables `stable` for the `information` variant with spacing. Defaults to `auto`.',
+    type: ['"auto"', '"stable"'],
+    status: 'optional',
+  },
```
langz, line 113: Good good, what auto does makes sense to me. It could also potentially collide with some devs thinking it means `scrollbar-gutter: auto` https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/scrollbar-gutter#syntax

Which it does not mean, but I'm fine with this.

Instead of having the value auto, it could just be the default behavior when not providing any value (undefined). But I'm fine with the existing code as well.

Or value "smart"? 😂

langz (approved):
Great great great, I can test once on my iphone.

EDIT: I did not find any issues or interesting stuff.

PR #7870, opened 2026-04-30
feat(Provider): add `translationsLoader` for async dynamic translations

tujoworker:
Motivation: https://dnb-it.slack.com/archives/C05SQ17NEJH/p1774274656919459

Add `translationsLoader` prop to Provider that accepts an async function (locale) => Promise<Translations>. Translations are loaded on mount and locale change, merged with static translations, with cancellation guard and error handling.

In a following up PR I will use this in the Portal docs.

### DIFF
--- a/packages/dnb-eufemia/src/shared/Context.tsx
+++ b/packages/dnb-eufemia/src/shared/Context.tsx
@@ -179,6 +179,13 @@ export type ContextProps = ContextComponents & {
    */
   translations?: Translations | TranslationCustomLocales
 
+  /**
+   * Async function to load translations for a given locale.
+   * Called on mount and whenever the locale changes.
+   * The returned translations are merged with any existing translations.
+   */
+  translationsLoader?: TranslationsLoader
+
   // -- For internal use --
   __context__?: Record<string, unknown>
   updateTranslation?: (
@@ -202,6 +209,7 @@ export type InternalLocale =
 export type Translations =
   | Partial<Record<InternalLocale, Translation | TranslationFlat>>
   | Partial<Record<InternalLocale, FormsTranslation>>
+export type TranslationsLoader = (locale: string) => Promise<Translations>
 export type TranslationDefaultLocales = typeof defaultLocales
 export type TranslationLocale = keyof TranslationDefaultLocales
 export type TranslationValues =
--- a/packages/dnb-eufemia/src/shared/Context.tsx
+++ b/packages/dnb-eufemia/src/shared/Context.tsx
@@ -209,7 +209,9 @@ export type InternalLocale =
 export type Translations =
   | Partial<Record<InternalLocale, Translation | TranslationFlat>>
   | Partial<Record<InternalLocale, FormsTranslation>>
-export type TranslationsLoader = (locale: string) => Promise<Translations>
+export type TranslationsLoader = (
+  locale: InternalLocale
+) => Promise<Translations>
 export type TranslationDefaultLocales = typeof defaultLocales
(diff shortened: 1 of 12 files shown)


packages/dnb-eufemia/src/shared/Provider.tsx
```diff
@@ -50,6 +57,48 @@ export default function Provider<Props>(
...
+    if (!translationsLoader) {
+      return undefined // stop here
+    }
+
+    let cancelled = false
+
+    translationsLoader(effectiveLocale)
+      .then((loaded) => {
+        if (!cancelled && loaded) {
+          const base = propTranslations || {}
```
langz: Should we add `propTranslations` to the useEffect's dependency array?

langz:
Should the consumer be able to know if translations are still loading, or if the loading of translations failed?
We could potentially expose isLoading and/or error callbacks or such?

What about a retry mechanism?

Should we display a skeleton?

packages/dnb-eufemia/src/shared/Context.tsx
```diff
@@ -202,6 +209,7 @@ export type InternalLocale =
 export type Translations =
   | Partial<Record<InternalLocale, Translation | TranslationFlat>>
   | Partial<Record<InternalLocale, FormsTranslation>>
+export type TranslationsLoader = (locale: string) => Promise<Translations>
```
langz:
```suggestion
export type TranslationsLoader = (locale: InternalLocale) => Promise<Translations>
```

langz:
One agent flags this as a potential concern:

Unstable function reference causes re-fetching on every render translationsLoader is in the useEffect dependency array. If a consumer writes:  `<Provider translationsLoader={async (locale) => fetch(...)} />`


Every parent render creates a new function reference → the effect re-runs → the loader is called again. This causes infinite re-fetching in common usage patterns. 

Solutions: 
- Use a useRef internally to store the loader, and remove it from the dep array (call via loaderRef.current). 
- Or document prominently that the function must be referentially stable (wrapped in useCallback or defined outside the component). The current docs show module-level functions but don't warn about this pitfall. T

packages/dnb-eufemia/src/shared/Provider.tsx
```diff
@@ -50,6 +57,48 @@ export default function Provider<Props>(
     [update]
   )
 
+  const { translationsLoader, translations: propTranslations } = localProps
+
+  const effectiveLocale =
+    localContext?.__context__?.locale ||
+    localProps.locale ||
+    nestedContext.locale
```
langz: Should we fallback to `LOCALE` if no other .locale is actually set. Or perhaps that's just a strange edge-case

langz:
How should it work if one does nest Providers? How would it merge translations? Will the child inherit the translations from the parent?
```
<Provider translations={{ 'nb-NO': { Modal: { closeTitle: 'Parent' } } }}>
  <Provider translationsLoader={loader}>
```

langz (approved):
Will be interesting to see this in play in the Portal, if there will be any visible flash when loading async translations, etc 🚀

langz:
Naming discussion:

`<Provider translations={...} translationsLoader={...} />` vs.  `<Provider translations={...} loadTranslations={...} />`.

langz:
Because of async nature, and re-renders, and prop changes, I guess there must be some kind of scary race-conditions, etc, that can happen. Will try to think of some 💭 
To further complicate if one adds nesting of Providers as well 😨

langz (commented):
Did we add a property in some docs for this new feature?

PR #7876, opened 2026-04-30
fix(Field.Selection, Field.ArraySelection): pass `width` to FieldBlock to prevent label truncation

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1777373660252399

When width='stretch' was set on Field.Selection or Field.ArraySelection, only contentWidth was forwarded to FieldBlock. This meant the CSS class 'dnb-forms-field-block--width-stretch' was never applied, and the label kept its max-width constraint, causing long labels to be cut off. 

Both components now pass width='stretch' to FieldBlock, matching the pattern used by Field.String, Field.SelectCountry, and Field.SelectCurrency. This applies to all Selection variants (dropdown, autocomplete, radio, button) and ArraySelection (checkbox, button, checkbox-button).


tujoworker (approved):
🙏

PR #7881, opened 2026-04-30
chore(Button): add border radius token

joakbjerk:


packages/dnb-eufemia/src/components/button/style/dnb-button.scss
```diff
@@ -41,8 +41,9 @@
   --button-border-width: var(--button-border-width--default);
   --button-border-width--default: 0.0625rem;
   --button-border-width--hover: 0.125rem;
-  --button-border-width--focus: var(--focus-ring-width);
-  --button-border-radius: calc(var(--button-height) / 2);
+  --button-border-width--focus: 0.1875rem;
```
langz: Can this be a token as well?

langz (approved):

packages/dnb-eufemia/src/components/button/style/dnb-button.scss
joakbjerk: I think that should maybe be a token set somewhere else, so this can be removed 😄

joakbjerk: it should be `--focus-ring-width`, no idea why this was changed 😅 it is set to be `0.25rem` which is the focus border width defined in Figma, and is not tokenized yet 🫡

tujoworker (approved):

PR #7893, opened 2026-05-03
chore: migrate test runner from Jest to Vitest

tujoworker:
For native ESM support. First up are just the @dnb/eufemia tests. Next is the portal done in a follow up PR.

PR #7894, opened 2026-05-03
feat: add opt-in ICU MessageFormat support for translations

tujoworker:
Preview: https://feat-translation-icu.eufemia-e25.pages.dev/uilib/usage/customisation/localization/#icu-message-format

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/getting-started.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/getting-started.mdx
@@ -1112,6 +1112,43 @@ Here is a [demo](/uilib/extensions/forms/Form/Handler/demos/#locale-and-translat
 
 When creating [your own field](/uilib/extensions/forms/create-component/#localization-and-translations), you can use the `Form.useTranslation` hook to localize your field.
 
+### ICU Message Format
+
+Forms translations support [ICU MessageFormat](/uilib/usage/customisation/localization/#icu-message-format) for pluralization, gender selection, and other locale-aware formatting. Enable it by passing the `icu` handler to `Form.Handler`:
+
+```tsx
+import { icu } from '@dnb/eufemia/shared'
+import { Form } from '@dnb/eufemia/extensions/forms'
+
+const translations = {
+  'en-GB': {
+    MyForm: {
+      summary:
+        'You have {count, plural, =0 {no items} one {# item} other {# items}} in your cart.',
+    },
+  },
+  'nb-NO': {
+    MyForm: {
+      summary:
+        'Du har {count, plural, =0 {ingen varer} one {# vare} other {# varer}} i handlekurven.',
+    },
+  },
+}
+
+function MyForm() {
+  const { formatMessage } = Form.useTranslation()
+  return <>{formatMessage('MyForm.summary', { count: 3 })}</>
+}
+
+render(
+  <Form.Handler icu={icu} translations={translations} locale="en-GB">
+    <MyForm />
+  </Form.Handler>
+)
+```
+
+See the [ICU Message Format docs](/uilib/usage/customisation/localization/#icu-message-format) for the full list of supported formatters and syntax.
+
 ## Layout
 
 When building your application forms, preferably use the following layout components. They seamlessly places all the fields and components of Eufemia Forms correctly into place.
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/getting-started.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/getting-started.mdx
@@ -1114,7 +1114,7 @@ When creating [your own field](/uilib/extensions/forms/create-component/#localiz
 
 ### ICU Message Format
 
-Forms translations support [ICU MessageFormat](/uilib/usage/customisation/localization/#icu-message-format) for pluralization, gender selection, and other locale-aware formatting. Enable it by passing the `icu` handler to `Form.Handler`:
+Forms translations support [ICU MessageFormat](/uilib/usage/customisation/localization/#icu-message-format) for pluralization, gender selection, and other locale-aware formatting. Enable it by passing the `icu` message formatter to `Form.Handler`:
 
 ```tsx
 import { icu } from '@dnb/eufemia/shared'
@@ -1141,7 +1141,11 @@ function MyForm() {
 }
 
 render(
-  <Form.Handler icu={icu} translations={translations} locale="en-GB">
+  <Form.Handler
+    messageFormatter={icu}
+    translations={translations}
+    locale="en-GB"
+  >
     <MyForm />
   </Form.Handler>
 )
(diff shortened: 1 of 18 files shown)


packages/dnb-eufemia/package.json
```diff
@@ -113,6 +113,7 @@
     "clsx": "2.1.1",
     "core-js-pure": "3.47.0",
     "date-fns": "4.1.0",
+    "intl-messageformat": "^10",
```
tujoworker: With this code/[PR](https://github.com/dnbexperience/eufemia/pull/7893) in place, we can upgrade to v11 as it is ESM only.

langz: Will this package and its transitive packages also end up in all consumers node_modules, even if they don't use this ICU feature? Or perhaps all users has to use it?

packages/dnb-eufemia/src/shared/icuFormatMessage.ts
```diff
@@ -0,0 +1,71 @@
+import IntlMessageFormat from 'intl-messageformat'
+
+const ICU_PATTERN =
+  /\{[^}]+,\s*(?:plural|select|selectordinal|number|date|time)\b/
+
+const cache = new Map<string, IntlMessageFormat>()
```
langz, line 7: Does a cache need some kind of caching strategy?
Could this potentially grow without bound in long-running SPAs or SSR/SSG environments?

packages/dnb-eufemia/src/shared/useTranslation.tsx
```diff
@@ -330,16 +349,67 @@ export function formatMessage(
       }
     }
     if (!found && typeof id === 'string') {
+      // Only warn when the id looks like a message path (e.g. "MyForm.title").
+      // Raw strings passed directly (e.g. already-resolved translations) are not missing keys.
+      if (id.includes('.')) {
+        warn(`formatMessage: Missing translation for "${id}".`)
```
langz: Can't a id look like` MyForm.title`? 🤔 
If so, then should the message be something like "wrongly formatted mesasge", etc?

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/getting-started.mdx
```diff
@@ -1112,6 +1112,43 @@ Here is a [demo](/uilib/extensions/forms/Form/Handler/demos/#locale-and-translat
...
+  },
+}
+
+function MyForm() {
+  const { formatMessage } = Form.useTranslation()
+  return <>{formatMessage('MyForm.summary', { count: 3 })}</>
+}
+
+render(
+  <Form.Handler icu={icu} translations={translations} locale="en-GB">
```
langz: Personally, I did not easily understand `icu={icu}`, but programatically it might make sense.
Could it be more descriptive of what it does?

Like `<Form.Handler messageFormatter={icu}>`, or perhaps `icu={icu}` is good enough 👍

langz:
Not sure what we do in our codebase today, but for TS naming, I'm not sure if we should do `ICUFormatMessage` or `IcuFormatMessage`, [this](https://google.github.io/styleguide/tsguide.html#camel-case) is Google's guide, but not sure what we do in our codebase. It's not too important.

packages/dnb-eufemia/src/shared/icuFormatMessage.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export type ICUFormatMessage = {
+  isICU: (message: string) => boolean
```
langz, line 10:
```suggestion
  isIcuMessage: (message: string) => boolean
```

Just a suggestion, I'm fine with the existing one as well.

packages/dnb-eufemia/src/shared/useTranslation.tsx
```diff
@@ -366,3 +436,46 @@ export function renderMessage(
 
   return text
 }
+
+function renderTags(
+  str: string,
+  args: Record<string, unknown>
+): React.ReactNode {
+  const TAG_RE = /<(\w+)>([\s\S]*?)<\/\1>/
```
langz, line 449: Does this also support nested tags?:
-  `<bold>Hello <em>world</em></bold>`
-  `<bold>Hello <bold>world</bold></bold>`

packages/dnb-eufemia/package.json
tujoworker: no, it should not end up when not using it. That's why we introduced the icu prop in the provider and form handler.

packages/dnb-eufemia/src/shared/useTranslation.tsx
```diff
@@ -252,13 +254,22 @@ export type CombineWithExternalTranslationsReturn = Translation &
   TranslationCustomLocales &
   AdditionalReturnUtils
 
-export function useAdditionalUtils() {
+export function useAdditionalUtils(icu?: ICUFormatMessage) {
   const translationsRef =
     useRef<CombineWithExternalTranslationsReturn>(undefined)
+  const localeRef = useRef<string>('nb-NO')
```
langz:
```suggestion
  const localeRef = useRef<string>(LOCALE)
```

Could potentially use LOCALE from packages/dnb-eufemia/src/shared/defaults.ts

packages/dnb-eufemia/src/shared/useTranslation.tsx
```diff
@@ -330,16 +349,67 @@ export function formatMessage(
...
       str = id
     }
   } else if (typeof id === 'function') {
     str = id(messages)
   }
 
   if (typeof str === 'string') {
+    if (args && icu && icu.isICU(str)) {
+      try {
+        const result = icu.format(str, args, locale || 'nb-NO')
```
langz:
```suggestion
        const result = icu.format(str, args, locale || LOCALE)
```

Could potentially use LOCALE from packages/dnb-eufemia/src/shared/defaults.ts

packages/dnb-eufemia/package.json
langz: Not sure if we should consider pinning to a specific version or not.
I see that quite a lot of our other deps are pinned to a specific version.

langz (approved):

packages/dnb-eufemia/src/shared/useTranslation.tsx
tujoworker: Do we want it to support? I think we may add that later if needed.

packages/dnb-eufemia/package.json
tujoworker: yes, now with v11 we do that 🙏

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/getting-started.mdx
tujoworker: Good idea!

packages/dnb-eufemia/src/shared/icuFormatMessage.ts
tujoworker: Now we have a max size of 1000 entries in the cache as once.

langz: Thanks :)

PR #7901, opened 2026-05-04
chore(DateFormat): improve flaky tooltip test race condition

tujoworker:
Move tooltip element lookup and assertion inside waitFor to ensure the tooltip DOM node exists before querying it.


langz (approved):
Image

PR #7914, opened 2026-05-04
chore: prevent formatMessage from treating error messages as translation keys

langz:
Motivation: https://github.com/dnbexperience/eufemia/actions/runs/25319551028/job/74224924734#step:9:278

Messages containing spaces (e.g. 'Dette feltet må fylles ut.') are natural language strings, not translation keys. Only attempt formatMessage lookup when the string contains a dot but no spaces.


tujoworker (approved):
I think we need a test for that case, even it is meant to fix tests.

langz:
> I think we need a test for that case, even it is meant to fix tests.

Yes yes, I'm also not sure if this fix is correct. Guessing this was introduced in the icu feature pr or async translations, as the v11 release test run went fine without warnings.

PR #7923, opened 2026-05-05
chore: fix command `yarn test:update X`

langz:
vitest's --update flag accepts an optional [type] parameter. When running 'yarn test:update FormStatus', the CLI parser (cac) consumed 'FormStatus' as the value for --update instead of treating it as a test filter, causing all tests to run.

Using --update=true explicitly sets the flag value, so any additional arguments are correctly treated as test file name filters.


langz:
I've tested it, and it works on "my machine" :) 

image

joakbjerk (approved):

PR #7925, opened 2026-05-05
fix(Heading, Ul, Ol, Dl, FormLabel, Hr): ensure surface="dark" support

tujoworker:
Screenshot 2026-05-05 at 11 50 28

vs 

Screenshot 2026-05-05 at 11 50 43

### DIFF
--- a/packages/dnb-eufemia/src/style/__tests__/__snapshots__/Elements.test.ts.snap
+++ b/packages/dnb-eufemia/src/style/__tests__/__snapshots__/Elements.test.ts.snap
@@ -219,6 +219,11 @@ del .dnb-code {
  */
 .dnb-hr {
   --hr-color: var(--token-color-stroke-neutral-subtle);
+}
+.dnb-hr--surface-dark {
+  --hr-color: var(--token-color-stroke-neutral-ondark);
+}
+.dnb-hr {
   position: relative;
   padding: 0;
   width: 100%;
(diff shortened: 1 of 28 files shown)


langz:
How do we test or find out issues like this?
Any way to do it in the portal?

Should we add darksurface example for all of our components?

langz (approved):

tujoworker:
I don't think so. Dark surfaces are uses with typography for promotion. Even label is not really within the scope. I just think I add it as labels could be used for something different than inputs too.

packages/dnb-eufemia/src/style/__tests__/__snapshots__/Elements.test.ts.snap
```diff
@@ -468,6 +479,9 @@ del .dnb-code {
 .dnb-ol {
   color: var(--token-color-text-neutral, currentColor);
 }
+.dnb-ol--surface-dark {
+  color: var(--token-color-text-neutral-ondark);
+}
```
langz, line 484: Is it a problem that it's listed twice in this snapshot?

tujoworker: No, it's not possible in a different way when compiled. CSS does support nested selectors I think, but for modern browsers only.

langz: image

tujoworker: Ahh, it look ok. One for OL and one for UL.

langz: yeah, screenshot was wrong, see here:

Image

But I guess it's okay, it's just duplication.

langz: Image

PR #7933, opened 2026-05-05
chore(CodeBlock): fix act import in test file

tujoworker:


packages/dnb-design-system-portal/src/shared/tags/__tests__/CodeBlock.test.tsx
```diff
@@ -7,8 +7,8 @@ import {
   beforeEach,
   afterAll,
 } from 'vitest'
-import React, { act } from 'react'
-import { render } from '@testing-library/react'
+import React from 'react'
+import { act, render } from '@testing-library/react'
```
langz, line 11: If this is important, that we use act from testing-library instead of from react, should we enforce this in some kind of linting rule? Or perhaps this is scenario specific.

langz (approved):

packages/dnb-design-system-portal/src/shared/tags/__tests__/CodeBlock.test.tsx
tujoworker: I don't think we should use it from react. Its a testing artifact.

PR #7937, opened 2026-05-05
chore: support filter env var in Vite dev server

langz:
Motivation: `filter=button yarn start`

Could be that this is not needed? As it's not much of a build time difference doing filter than running the whole app 🤔

### DIFF
--- a/packages/dnb-design-system-portal/vite/README.md
+++ b/packages/dnb-design-system-portal/vite/README.md
@@ -13,6 +13,17 @@ yarn preview            # Preview the production build locally
 yarn test:e2e:portal    # Run e2e tests against a Vite preview server
 ```
 
+### Filtering pages
+
+To speed up the dev server by only loading specific pages, use the `filter` env var:
+
+```bash
+filter=button yarn start          # Only pages with "button" in the path
+filter=button,accordion yarn start # Multiple filters (comma or space separated)
+```
+
+This is useful when working on a single component — it reduces the number of routes and speeds up startup.
+
 ## Directory Structure
 
 ```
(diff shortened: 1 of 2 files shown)


tujoworker:
I kept this out in purpose. Because does it make it faster? How many seconds?

langz:
I've not measured it. It was just that my workflow, where I usually use filter, was not working anymore. But I'm happy if it's not worth having. 

I can test on my older machine this evening.

tujoworker:
Yes, would be good to know. It really should not be an issue anymore.

langz:
## Analysis: `filter` env var for Vite dev server

### What the PR does

Adds a `filter` env var to the `portal-pages.ts` plugin so that `filter=button yarn start` only registers routes matching the filter term. This reduces the route count from **~1235 pages to ~60** (for `filter=button`).

### Benchmark results

Machine: MacBook (macOS), Vite v8.0.10

| Scenario | Cold start | Warm start |
|---|---|---|
| **No filter** (all routes) | 13,662 ms | 1,710 ms |
| **`filter=button`** (~60 routes) | 2,551 ms | 2,015 ms |

**Cold-start savings: ~11 seconds** (13.7s → 2.6s).

Warm-start difference is negligible (~0.3s, within noise).

### Verdict

The filter **saves ~11 seconds on cold starts** (first start or after clearing Vite cache). This comes from Vite having to pre-bundle fewer dependencies and process fewer route definitions on first boot.

On warm starts (typical dev workflow after the first launch), the difference is negligible since Vite caches aggressively. The benefit is most felt when:
- Starting the dev server for the first time
- After `node_modules` changes / cache invalidation
- On slower machines where the cold start gap could be even larger

tujoworker:
yeah right. The thing is, that you will probably never go and clean the Vite cache manually like we did with Gatsby. So it will almost always be warm.

langz:
> yeah right. The thing is, that you will probably never go and clean the Vite cache manually like we did with Gatsby. So it will almost always be warn.

Great, speedy speedy ⚡

PR #7941, opened 2026-05-05
chore(Portal): improve portal tools and theme switching

tujoworker:
When switching from DNB to Sbanken, it closes and re-opens. Which is not nice UX.


langz:
Based on my experience, this changes from:

https://github.com/user-attachments/assets/c61c1db5-84c3-4b15-a4d6-1153bcbc854b

to:

https://github.com/user-attachments/assets/f9789ad2-c1df-4053-b686-d530de0d22be


I guess there’s a subtle difference, but I’m not 100% sure what it is. Hehe

This compares main vs. this deploy preview. Could be that I should have tested release v11.1.1 deploy preview vs this pr deploy preview.

langz (approved):

tujoworker:
It only happens once and on desktop.

PR #7950, opened 2026-05-06
chore(Portal): prefer local Eufemia (pre) build when it exists

tujoworker:
We did only support it partially.


packages/dnb-design-system-portal/vite.config.ts
```diff
@@ -134,7 +147,6 @@ export default defineConfig({
             }
             if (fs.existsSync(candidate)) {
               examplesPath = candidate
-              break
```
langz, line 137: 🤔

langz (approved):
Some duplicated logic between .ts and .cjs — `EUFEMIA_PACKAGE_NAME`, `EXPLICIT_PREBUILD_TARGETS`, `rewriteToPrebuild`, and `hasPrebuild`, etc, in eufemia-prebuild.ts and eufemia-prebuild-paths.cjs.

But I guess that's fine ☘️

PR #7954, opened 2026-05-06
chore: remove stale lmdb and msgpackr resolutions

langz:
These were Gatsby-era workarounds. The portal has since migrated to Vite and neither package is depended on by anything in the workspace (yarn why returns empty for both).

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,8 +41,6 @@
     "prettier": "3.8.1"
   },
   "resolutions": {
-    "lmdb": "3.5.1",
-    "msgpackr": "1.11.8",
     "sharp": "0.33.5",
     "@types/react": "19.2.14",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -41,8 +41,6 @@
     "prettier": "3.8.1"
   },
   "resolutions": {
-    "lmdb": "3.5.1",
-    "msgpackr": "1.11.8",
     "sharp": "0.33.5",
```
tujoworker, line 44: sharp is also a thing we can remove

tujoworker (approved):

package.json
langz: Great, thanks :) 

https://github.com/dnbexperience/eufemia/pull/7961

PR #7964, opened 2026-05-06
refactor(Tooltip): add correct border radius token

joakbjerk:
Using `--token-radius-full` as defined in the sketches, required a small rewrite of the `borderRadius` value used in `PopoverContainer`, since `--token-radius-full` resolves to `9999px` which threw off the position of the `Tooltip`

### DIFF
--- a/packages/dnb-eufemia/src/components/popover/PopoverContainer.tsx
+++ b/packages/dnb-eufemia/src/components/popover/PopoverContainer.tsx
@@ -682,9 +682,17 @@ function PopoverContainer(props: PopoverContainerProps) {
     const marginRight = computedElementStyle
       ? parseFloat(computedElementStyle.marginRight) || 0
       : 0
-    const borderRadius = computedElementStyle
+    const rawBorderRadius = computedElementStyle
       ? parseFloat(computedElementStyle.borderRadius) || 0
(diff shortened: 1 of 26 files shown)


packages/dnb-eufemia/src/components/popover/PopoverContainer.tsx
```diff
@@ -682,9 +677,17 @@ function PopoverContainer(props: PopoverContainerProps) {
...
       ? parseFloat(computedElementStyle.borderRadius) || 0
       : 0
+    // Cap border-radius to actual visual maximum, since the browser
+    // clamps it to half the element size but getComputedStyle returns
+    // the raw value (e.g. 9999px from --token-radius-full).
+    const borderRadius = Math.min(
+      rawBorderRadius,
+      elementWidth / 2,
+      elementHeight / 2
+    )
```
langz, line 695: Looks smart.
Should we add any unit tests for this new clamping logic.
Like that:
- Arrow stays within bounds with large border-radius
- Arrow position matches clamped behavior

Not sure if it's complicated to test or not, just a suggestion.

langz (approved):

tujoworker (approved):

PR #7971, opened 2026-05-06
feat: add ESLint and Stylelint Plugin to detecting deprecated color variable usage

tujoworker:
Preview: https://feat-eufemia-lint-plugin-col.eufemia-e25.pages.dev/uilib/usage/first-steps/tools/#lint-plugins

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/usage/first-steps/tools.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/usage/first-steps/tools.mdx
@@ -100,9 +100,6 @@ Use the recommended flat config preset:
 import eufemiaEslint from '@dnb/eufemia/plugins/eslint.js'
 
 export default [eufemiaEslint.recommended]
-
-// or for legacy config (ESLint v7 and earlier):
-export default [eufemiaEslint.legacyRecommended]
 ```
 
 If you need full control, register the plugin and configure the rules yourself:
@@ -116,7 +113,11 @@ export default [
       eufemia: eufemiaEslint,
     },
     rules: {
-      'eufemia/no-deprecated-color-variables': 'warn',
+      // All rules
+      ...eufemiaEslint.recommended.rules,
+
+      // Or specific rules
+      'eufemia/no-deprecated-color-variables': 'error',
     },
   },
 ]
@@ -147,4 +148,4 @@ export default {
 
 For SCSS files, configure Stylelint with `postcss-scss` as the custom syntax.
 
-The first shipped ESLint and Stylelint rule reports deprecated `--color-*` CSS variables. Use design tokens instead.
+Both plugins ship with one rule: `no-deprecated-color-variables`. It reports deprecated `--color-*` CSS variables and guides towards design tokens instead.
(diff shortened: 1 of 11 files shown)


langz:
`The first shipped ESLint and Stylelint rule reports deprecated --color-* CSS variables. Use design tokens instead.` -> `Both plugins ship with one rule: no-deprecated-color-variables. It reports deprecated --color-* CSS variables and guides towards design tokens instead.`

langz:
`For SCSS files, configure Stylelint with `postcss-scss` as the custom syntax.`

We could potentially also mention that users will have to install postcss-scss, like `yarn add --dev postcss-scss`

langz:
`const COLOR_VARIABLE_REGEX = /--color-[a-z0-9-]+/gi`, I guess the `i` makes it case-insensitive.
But I guess all our tokens are case-sensitive. So we could propably remove the `i` flag, or else we could potentially end up flagging other variables I guess? false-positives

langz:
`legacyRecommendedConfig` and `flatRecommendedConfig` is looking identical, I guess they would look different.
So I assume the legacy preset will not work correctly if used in older versions of eslint? as of right now.

langz (approved):
I could not add comments in line or in files, because of issues with GitHub, so added them as regular comments.

https://www.githubstatus.com/

langz:
Are we sure that the plugins will be available from `@dnb/eufemia/plugins/eslint.js`?
How can we make sure that will work 🤔

langz:
I see the POSTcss plugins says: `"Deprecated CSS color variable \"${variable}\" detected. Use a design token instead."`

But eslint says: `"Deprecated CSS color variable \"{{ variable }}\" detected. Use a design token instead; the correct replacement depends on context."`

Should they be more similar? Should PostCSS also mention context?

tujoworker:
> Are we sure that the plugins will be available from `@dnb/eufemia/plugins/eslint.js`?
> 
> How can we make sure that will work 🤔 

Yes. Everything from /src is part of the bundle.

I testet it in the repo and in the starter. But directly from source. So no 100% confirmation. But 99% 😅

packages/dnb-design-system-portal/src/docs/uilib/usage/first-steps/tools.mdx
```diff
@@ -86,6 +86,66 @@ Install the [VSCode Extension](https://marketplace.visualstudio.com/items?itemNa
...
+
+export default {
+  plugins: [eufemiaStylelint],
+  rules: {
+    'eufemia/no-deprecated-color-variables': true,
+  },
+}
+```
+
+For SCSS files, configure Stylelint with `postcss-scss` as the custom syntax.
```
langz:
```suggestion
For SCSS files, configure Stylelint with [postcss-scss](https://www.npmjs.com/package/postcss-scss) as the custom syntax.
```

langz: Perhaps not needed, but would be helpful for users reading this who don't necessarily use or know what postcss-scss is, to have a link to read more about it, or just to understand that it's a package.

langz:
Could potentially add a `meta.docs.url` to the ESLint rule, with a link to the design tokens guide so that IDEs display a clickable "learn more" link.

https://eslint.org/docs/latest/extend/custom-rules

PR #7972, opened 2026-05-06
feat(StyleIsolation): warn when deprecated color variables are used via PostCSS plugin (#7972)

tujoworker:
Based on #7971

Screenshot 2026-05-06 at 16 58 15

### DIFF
--- a/packages/dnb-eufemia/src/plugins/postcss-isolated-style-scope/__tests__/isolated-style-scope-plugin.test.ts
+++ b/packages/dnb-eufemia/src/plugins/postcss-isolated-style-scope/__tests__/isolated-style-scope-plugin.test.ts
@@ -1467,6 +1467,9 @@ describe('isolated-style-scope-plugin', () => {
 
       expect(warnings).toHaveLength(1)
       expect(warnings[0].text).toContain('--color-sea-green')
+      expect(warnings[0].text).toContain(
+        'https://eufemia.dnb.no/uilib/usage/customisation/theming/design-tokens/guide/'
+      )
     })
 
     it('should warn on deprecated color variable in property', async () => {
--- a/packages/dnb-eufemia/src/plugins/postcss-isolated-style-scope/__tests__/isolated-style-scope-plugin.test.ts
+++ b/packages/dnb-eufemia/src/plugins/postcss-isolated-style-scope/__tests__/isolated-style-scope-plugin.test.ts
@@ -1429,4 +1429,81 @@ describe('isolated-style-scope-plugin', () => {
       })
     })
   })
+
+  describe('warnOnDeprecatedColorVariables', () => {
+    async function runWithWarnings(
+      input: string,
+      options: Options & { warnOnDeprecatedColorVariables?: boolean } = {}
+    ) {
+      options.postcssOptions = options.postcssOptions ?? {
+        from: '/file.css',
+      }
+
+      const result = await postcss([plugin(options)]).process(
+        input,
+        options.postcssOptions
+      )
+
+      return result.warnings()
+    }
+
+    it('should not warn when opted out', async () => {
+      const warnings = await runWithWarnings(
+        '.foo { color: var(--color-sea-green); }',
+        {
+          scopeHash: 'test-scope',
+          warnOnDeprecatedColorVariables: false,
+        }
+      )
+
+      expect(warnings).toHaveLength(0)
+    })
+
+    it('should warn on deprecated color variable in value', async () => {
+      const warnings = await runWithWarnings(
+        '.foo { color: var(--color-sea-green); }',
+        { scopeHash: 'test-scope', warnOnDeprecatedColorVariables: true }
+      )
+
+      expect(warnings).toHaveLength(1)
+      expect(warnings[0].text).toContain('--color-sea-green')
+    })
+
+    it('should warn on deprecated color variable in property', async () => {
+      const warnings = await runWithWarnings(
+        '.foo { --color-mint: #00f; }',
+        { scopeHash: 'test-scope', warnOnDeprecatedColorVariables: true }
+      )
+
+      expect(warnings).toHaveLength(1)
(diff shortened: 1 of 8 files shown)


boyum:
Can we add a "see link/to/docs for more information"?

packages/dnb-eufemia/src/plugins/postcss-isolated-style-scope/__tests__/isolated-style-scope-plugin.test.ts
```diff
@@ -1429,4 +1429,81 @@ describe('isolated-style-scope-plugin', () => {
...
+        { scopeHash: 'test-scope', warnOnDeprecatedColorVariables: true }
+      )
+
+      expect(warnings).toHaveLength(2)
+    })
+
+    it('should not warn on non-deprecated variables', async () => {
+      const warnings = await runWithWarnings(
+        '.foo { color: var(--token-color-text-neutral); }',
+        { scopeHash: 'test-scope' }
```
langz, line 1516:
```suggestion
        { scopeHash: 'test-scope', warnOnDeprecatedColorVariables: true }
```

langz:
> Can we add a "see link/to/docs for more information"? 

Sounds like a great idea 💡

packages/dnb-eufemia/src/plugins/postcss-isolated-style-scope/__tests__/isolated-style-scope-plugin.test.ts
tujoworker: "should not warn ..." by default.

langz: Aha, based on the it statement when I added the comment (should not warn on non-deprecated variables), I understood it differently. Hehe.

boyum:
What's the reasoning for not making it run by default?

tujoworker:
We can do that. I just thought many warnings could become overwhelming, especially since there will be apps that do not want or need to use tokens right now.

Even if we mark the old colors as deprecated, we will still need to offer a subset of brand colors in the future. They may just be prefixed and named differently, for example:

--dnb-color-primary-1

So maybe it is a good idea to enable this by default after all.

boyum:
>I just thought many warnings could become overwhelming

It's definitely a good point! I'm just worried people will miss the announcement when they are removed, and since using non-existing CSS vars will never fail in build or even runtime, the issue might be caught very late.

Do we know which vars will be replaced by tokens and which will be renamed? If we knew, we could even create a migration tool for this, and the error could simply say "run npx @dnb/eufemia/migrations/replace-vars-v1".

tujoworker:
> > I just thought many warnings could become overwhelming
> 
> It's definitely a good point! I'm just worried people will miss the announcement when they are removed, and since using non-existing CSS vars will never fail in build or even runtime, the issue might be caught very late.
> 
> Do we know which vars will be replaced by tokens and which will be renamed? If we knew, we could even create a migration tool for this, and the error could simply say "run npx @dnb/eufemia/migrations/replace-vars-v1".

No, we don't know at the moment.

boyum:
> > > I just thought many warnings could become overwhelming
> > 
> > 
> > It's definitely a good point! I'm just worried people will miss the announcement when they are removed, and since using non-existing CSS vars will never fail in build or even runtime, the issue might be caught very late.
> > Do we know which vars will be replaced by tokens and which will be renamed? If we knew, we could even create a migration tool for this, and the error could simply say "run npx @dnb/eufemia/migrations/replace-vars-v1".
> 
> No, we don't know at the moment.

Then maybe we shouldn't announce that consumers should make the shift at all? Sorry for sticking my nose into this, I'm sure you have control :')

tujoworker:
Such brand colors should then only be used in special cases. We would still 100% recommend using color tokens instead of raw colors — especially within microfrontends, which this PR is about.

langz:
Should we also add guide link like https://github.com/dnbexperience/eufemia/pull/7972/changes/7f195af9bd56b69a0a766bf27ba5338a277e6128 in to the PostCSS rule?

tujoworker:
> Should we also add guide link like [7f195af](https://github.com/dnbexperience/eufemia/commit/7f195af9bd56b69a0a766bf27ba5338a277e6128) in to the PostCSS rule?

We have added the link, yes. Something different you think about?

langz:
> > Should we also add guide link like [7f195af](https://github.com/dnbexperience/eufemia/commit/7f195af9bd56b69a0a766bf27ba5338a277e6128) in to the PostCSS rule?
> 
> We have added the link, yes. Something different you think about?

The PostCSS plugin already includes the guide link. However, the ESLint and Stylelint rules (no-deprecated-color-variables) are missing it — their messages end with "the correct replacement depends on context" without linking to the guide. If this is intended, I'm fine with it :)

langz (approved):
I guess we should first merge https://github.com/dnbexperience/eufemia/pull/7971 and then rebase this current PR 🥇

packages/dnb-eufemia/src/plugins/postcss-isolated-style-scope/isolated-style-scope-plugin.js
```diff
@@ -38,10 +42,29 @@ const postcssIsolateStyle = (opts = {}) => {
   return {
     postcssPlugin: 'isolated-style-scope-plugin',
 
-    Once(root) {
+    Once(root, { result }) {
       const file = root.source?.input?.file ?? ''
       const isCssModule = runAsCssModule || file.includes('.module.')
 
+      if (warnOnDeprecatedColorVariables) {
+        const colorVariableRegex = /--color-[a-z0-9-]+/gi
```
langz:
```suggestion
        const colorVariableRegex = /--color-[a-z0-9-]+/g
```

langz (approved):

PR #7974, opened 2026-05-06
style(ToggleButton, Radio): ToggleButton redesign

snorrekim:
- redesign `ToggleButton`
- `Radio` design fixes


packages/dnb-eufemia/src/components/radio/style/dnb-radio.scss
```diff
@@ -46,30 +45,23 @@
   --radio-color-border--active: var(--token-color-stroke-action-pressed);
 
   // Hover state
+  --radio-border-inset--hover: ;
```
langz, line 48: How does this work?
Empty value initially, then set at a later stage? 🤔

packages/dnb-eufemia/src/components/toggle-button/style/dnb-toggle-button.scss
```diff
@@ -95,10 +116,179 @@
...
+        .dnb-checkbox__gfx {
+          --checkbox-color-gfx-on: var(--checkbox-color-gfx-on--active);
+        }
+      }
+    }
+
+    @include utilities.focusVisible() {
+      .dnb-radio {
+        .dnb-radio__button {
+          // todo: missing default in radio
```
langz: Image

langz:
`Radio.tsx` looks to still render `<span className="dnb-radio__focus" aria-hidden /> but all .`dnb-radio__focus` CSS has been removed in this PR. I guess we can remove the span then, as I think it does not serve any purpose anymore.

langz (approved):
I see it's still marked as under draft, but I think it looks very very good :)

packages/dnb-eufemia/src/components/radio/style/dnb-radio.scss
snorrekim: it's how we can swap between `box-shadow: inset 0 0 0 1px red` and `box-shadow: 0 0 0 1px red` using variables. We can't make the variable invalid, because then the whole style fails. So it need a value even when "empty"

snorrekim:
the screenshots for radio group aren't working, so I'll delay this PR until that's fixed

snorrekim:
I added an option to simulate a "realhover" for screenshots as that is needed when checking input hover effects when its label in hovered.

tujoworker (approved):

PR #7984, opened 2026-05-07
chore(Portal): recover from stale chunk URLs after deploy

tujoworker:
Returning users may keep the previous app shell in memory or restore it from bfcache on a soft reload, then try to dynamically import chunk URLs that no longer exist on the new build (this surfaced when the portal moved from Gatsby to Vite, but applies to any future hash rotation).

- Listen for ChunkLoadError / 'Failed to fetch dynamically imported module' / 'Importing a module script failed' / 'Unable to preload CSS' and force a single hard reload, guarded by a sessionStorage flag to avoid loops.
- Unregister any service workers from earlier portal versions (gatsby-plugin-offline) so they stop intercepting fetches with stale cached responses.
- Ship a kill-switch /sw.js that replaces the old worker, drops all caches and self-unregisters.

### DIFF
--- a/packages/dnb-design-system-portal/static/sw.js
+++ b/packages/dnb-design-system-portal/static/sw.js
@@ -1,0 +1,52 @@
+/**
+ * Kill-switch service worker.
+ *
+ * The Eufemia portal does not use a service worker, but earlier
+ * versions (running on Gatsby) may have shipped one to production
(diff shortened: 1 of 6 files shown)


langz (approved):
Great, thanks, I've encountered some issues/errors related to this lately 💯

PR #7986, opened 2026-05-07
chore: move misplaced dependencies to devDependencies

langz:
- Move postcss-selector-parser from dependencies to devDependencies in @dnb/eufemia (only used in a build-time PostCSS plugin)
- Move algoliasearch from dependencies to devDependencies in dnb-design-system-portal (only used in build script push-algolia.mjs)

### DIFF
--- a/packages/dnb-eufemia/package.json
+++ b/packages/dnb-eufemia/package.json
@@ -114,6 +114,7 @@
     "core-js-pure": "3.47.0",
     "date-fns": "4.1.0",
     "intl-messageformat": "11.2.3",
+    "postcss-selector-parser": "7.1.0",
     "zod": "4.1.8"
   },
   "peerDependencies": {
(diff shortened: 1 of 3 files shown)


packages/dnb-eufemia/package.json
```diff
@@ -216,6 +215,7 @@
     "packpath": "0.1.0",
     "postcss": "8.4.32",
     "postcss-preset-env": "9.6.0",
+    "postcss-selector-parser": "7.1.0",
```
tujoworker: That change would break apps using style isolation. They would need to install it in addition.

langz: I added a comment here, so that I will not make the same mistake in the future.

https://github.com/dnbexperience/eufemia/pull/7986/changes#diff-564a81686969ead1adcc2d14fbd2ceabdf38231ee5fa38382332258825cffcabR4-R6

PR #7996, opened 2026-05-07
feat(Table): add `selected` state to Table `Td`

tujoworker:
Support both controlled (selected prop) and uncontrolled (`setSelected` callback) selection. The `onClick` callback now provides `isSelected` and `setSelected` for toggling. Uses aria-pressed for accessibility and selected color tokens for styling.

Preview: https://feat-table-seleted.eufemia-e25.pages.dev/uilib/components/table/#table-with-clickable-cells
Motivation: https://dnb-it.slack.com/archives/C081G91GCT0/p1778145620132619?thread_ts=1778072480.217599&cid=C081G91GCT0


packages/dnb-eufemia/src/components/table/TableTd.tsx
```diff
@@ -29,10 +38,18 @@ export type TableTdProps = {
    */
   verticalAlign?: 'top' | 'middle' | 'bottom'
 
+  /**
+   * If set to `true`, the cell will be styled as selected.
+   * When provided (either `true` or `false`), the cell button will be
+   * announced as a toggle button by screen readers.
+   * Default: `undefined`
+   */
+  selected?: boolean
```
langz, line 47: Should `selected` work without `onClick`?
When `selected={true}` and `onClick`:

Image

When `selected={true}` and no `onClick`:

Image

langz (approved):
Wow, fancy tables now 🥇 🥈 🥉

packages/dnb-eufemia/src/components/table/TableTd.tsx
tujoworker: At the moment we relay on both. I think it makes sense. Its documented and tested now.

joakbjerk (approved):

PR #8001, opened 2026-05-07
chore(Portal): use viewTransition on RouterLink for page transitions

tujoworker:
Use React Router's built-in viewTransition prop on RouterLink instead of our custom startPageTransition wrapper. Remove Transition.ts — it is no longer needed since React Router handles startViewTransition() internally with correct timing.


packages/dnb-design-system-portal/src/shared/tags/Intro.tsx
```diff
@@ -23,7 +22,6 @@ const Intro = ({ children }: IntroProps) => {
       if (/textarea|input/i.test(document.activeElement.tagName)) {
         return
       }
-      startPageTransition()
```
langz, line 26: I guess we lose some functionality by removing this?
But it's most likely not very important, I'm fine with it.

langz (approved):

packages/dnb-design-system-portal/src/shared/tags/Intro.tsx
tujoworker: Its just the opacity page fade that we have else on navigation. But I think we don't keep this extra code for that place.

PR #8002, opened 2026-05-07
fix(HeightAnimation): fix hydration mismatches for SSR compatibility

tujoworker:
Fix hydration mismatches in HeightAnimation when used in SSR/SSG applications.

## Changes

- **Move CSS custom properties to useEffect**: React's SSR serializes CSS custom properties without spaces (e.g. `--duration:600ms`) while the browser normalizes them with spaces and a trailing semicolon (`--duration: 600ms;`), causing a hydration mismatch. Setting them via `el.style.setProperty()` in a useEffect avoids the discrepancy.

- **Initialize `isVisible` from `open` prop**: `useHeightAnimation` initialized `isVisible` as `false` regardless of the `open` prop, causing a flash where the component renders without `dnb-height-animation--is-visible` before the effect kicks in.

## Tests

- SSR tests verifying custom properties are not in server-rendered HTML
- SSR test for `is-visible` class when `open` is true
- SSR test for element omission when `open` is false

## Preview

https://fix-height-animation-ssr.eufemia-e25.pages.dev/uilib/components/height-animation/


packages/dnb-eufemia/src/components/height-animation/HeightAnimation.tsx
```diff
@@ -91,17 +91,28 @@ function HeightAnimation({
...
+  // (e.g. "--duration:600ms") while the client serializes them with
+  // spaces and a trailing semicolon ("--duration: 600ms;"), causing
+  // a hydration mismatch.
+  useEffect(() => {
+    const el = targetRef.current
+    if (!el) {
+      return
+    }
+    if (duration) {
+      el.style.setProperty('--duration', `${duration}ms`)
```
langz, line 105: Could it be that we should always remove   `el.style.removeProperty('--duration')` when duration is not set?

Let's say if duration was 600 and the parent re-renders without passing duration, could the old --duration: 600ms stay on the element's style attribute for ever?

Same potential thought applies to delay as well.

langz (approved):

Issue #4558, opened 2025-02-11
Iterate.Array examples produce 70+ warnings in the console


langz:
## 🐛 Bug Report

Iterate.Array examples produce 70+ warnings in the console, like:
- https://react.dev/errors/418?invariant=418
- https://react.dev/errors/418?invariant=423

## To Reproduce

Visit https://eufemia.dnb.no/uilib/extensions/forms/Iterate/Array/demos/

Do a hard refresh, command +shift + r, and then open the console.

## Expected behavior

No errors

tujoworker:
Now that we use Vite with our own hydration logic, its not present anymore 😆 That means not it can be an issue. Will probably dig deeper into that and try to create a replication.

tujoworker:
Looks good now.

PR #8005, opened 2026-05-07
chore(Portal): enable React hydration and fix SSR/SSG mismatches

tujoworker:
Enable React hydration for pre-rendered portal pages and fix all SSR/SSG hydration mismatches so the client can reuse server-rendered HTML without re-creating the DOM.

Preview: https://chore-portal-hydration.eufemia-e25.pages.dev/

This change means that we now get a dev-console error when a component interferes with hydration in production builds.

The app may still appear to “work”, but React can no longer safely hydrate the existing server-rendered markup. Instead, it falls back to client rendering and injects a re-rendered app into the DOM.

That makes the issue visible, but it also means the component should be fixed so the initial server render and the first client render produce the same markup.

### DIFF
--- a/packages/dnb-design-system-portal/vite/client/render-portal-app.tsx
+++ b/packages/dnb-design-system-portal/vite/client/render-portal-app.tsx
@@ -1,4 +1,4 @@
-import { createRoot } from 'react-dom/client'
+import { createRoot, hydrateRoot } from 'react-dom/client'
 import type { Root } from 'react-dom/client'
 
 type RootStore = {
@@ -6,29 +6,13 @@ type RootStore = {
 }
 
 type CreateRootFn = (container: Element) => Root
+type HydrateRootFn = typeof hydrateRoot
 
 type RenderPortalAppOptions = {
   container?: Element | null
   rootStore?: RootStore
   createRootFn?: CreateRootFn
-}
-
-export function getOrCreatePortalRoot(
-  container: Element,
-  {
-    rootStore = globalThis as RootStore,
-    createRootFn = createRoot,
-  }: Omit<RenderPortalAppOptions, 'container'> = {}
-) {
-  const existingRoot = rootStore.__portalRoot
-
-  if (existingRoot) {
-    return existingRoot
-  }
-
-  const nextRoot = createRootFn(container)
-  rootStore.__portalRoot = nextRoot
-  return nextRoot
+  hydrateRootFn?: HydrateRootFn
 }
 
 export function renderPortalApp<Props extends object>(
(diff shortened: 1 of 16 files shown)


tujoworker:
I have not seen any hydration issues locally in regards to #4558

packages/dnb-design-system-portal/vite/prod/prerender-utils.ts
```diff
@@ -194,13 +213,59 @@ export function injectHtml(
...
+        .filter(Boolean)
+        .map((decl: string) => {
+          const colonIdx = decl.indexOf(':')
+          if (colonIdx === -1) return decl
+          const prop = decl.slice(0, colonIdx).trim()
+          const value = decl.slice(colonIdx + 1).trim()
+          return `${prop}: ${value}`
+        })
+        .join('; ')
+      return `style="${normalized};"`
```
langz: Could this every be empty and result in `style=";"`? Or perhaps it's only theoretical.

packages/dnb-design-system-portal/vite/prod/prerender.mjs
```diff
@@ -360,9 +378,50 @@ function injectHtml(
...
+      .filter(Boolean)
+      .map((decl) => {
+        const colonIdx = decl.indexOf(':')
+        if (colonIdx === -1) return decl
+        const prop = decl.slice(0, colonIdx).trim()
+        const value = decl.slice(colonIdx + 1).trim()
+        return `${prop}: ${value}`
+      })
+      .join('; ')
+    return `style="${normalized};"`
```
langz: Could this every be empty and result in `style=";"`? Or perhaps it's only theoretical.

packages/dnb-design-system-portal/vite/client/render-portal-app.tsx
```diff
@@ -37,19 +22,60 @@ export function renderPortalApp<Props extends object>(
...
+    // render a HydrateFallback.
+    //
+    // Minor hydration mismatches (e.g. AriaLive spans from Tooltips
+    // that only render client-side) are expected and handled
+    // gracefully by React's recovery mechanism.
+    root = hydrateRootFn(container, element, {
+      onRecoverableError(error, errorInfo) {
+        // Only log hydration mismatches on local and preview builds
+        // so they surface during development without cluttering
+        // production error reporting.
```
langz, line 60: I'll do some testing in a preview build to look for errors then 🦯

tujoworker: You don't need to test too much in this PR. We can fix component issues later as well. But I don't think there are any actually.

langz: Seems too good to be true 😂

langz (approved):

PR #8007, opened 2026-05-07
chore(Portal): remove viewTransition from RouterLink in Anchor

langz:
Motivation: As of now when testing and navigating between pages in https://main.eufemia-e25.pages.dev the app/portal feels sluggish on my computer. Slow animations, feels unresponsive, etc.
But on my iPhone it behaves as expected. Perhaps we should test this in a browser with reduced hardware simulation, or perhaps if someone has access to old machines.

Deploy preview: https://revert-remove-view-transitio-pj9w.eufemia-e25.pages.dev/

Without `viewTransition`, it feels responsive, and not lagging, etc.


tujoworker (approved):
I'm fine with removing it. It gives us not a real value or better UX at the end.

PR #8009, opened 2026-05-08
fix(Forms): ensure Form.useData is in sync when data prop changes (using id)

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1778156025744859
Issue reprod: https://stackblitz.com/edit/u6dlg7ok-wdy82dhk?file=src%2Fmain.tsx
Fix reprod: https://stackblitz.com/edit/u6dlg7ok-taninbxc?file=package.json

When a Provider has both an `id` and a `data` prop, updating the `data` prop did not propagate to the internal shared state. This caused fields to show stale values after unmount/remount cycles, and Form.useData to return outdated data.


tujoworker (approved):

langz:
Thanks for the assistance @tujoworker, looks good to me now :) 
Merging 🚀

PR #8015, opened 2026-05-08
chore(Portal): persist and restore window scroll position on refresh

tujoworker:
Save window.scrollY to sessionStorage alongside the sidebar scroll position. On refresh the saved position is restored; on SPA navigation the window scrolls to top while only the sidebar position is restored.

Switch all scroll storage from localStorage to sessionStorage so positions are tab-scoped and cleared when the tab closes. Update the inline prerender scripts to match.

This helps keeping the scroll position during development.

### DIFF
--- a/packages/dnb-design-system-portal/vite/client/plugins/scroll-position.ts
+++ b/packages/dnb-design-system-portal/vite/client/plugins/scroll-position.ts
@@ -48,12 +48,16 @@ const SCROLL_ELEMENTS = [
   },
 ]
 
+const WINDOW_SCROLL_KEY = 'scroll-window'
+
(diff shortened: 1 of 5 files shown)


langz (approved):

packages/dnb-design-system-portal/vite/client/plugins/scroll-position.ts
langz:
```suggestion
 * sessionStorage. Handles iOS Safari's pagehide event for back-forward
```

langz: Or perhaps remove the comment all together? 🤔

PR #8022, opened 2026-05-08
feat(Icon): add filled icon support to Icon component (`fill` prop)

tujoworker:
Adds a `filled` prop and `filled` wrapper that fills SVG icon paths with `currentColor` via CSS.

Preview: https://feat-icon-filled.eufemia-e25.pages.dev/icons/


packages/dnb-eufemia/src/components/icon/__tests__/Icon.test.tsx
```diff
@@ -278,6 +278,76 @@ describe('Icon component', () => {
...
   })
+
+  it('should have filled class when filled prop is true', () => {
+    render(<Icon icon={question} filled />)
+    expect(document.querySelector('span.dnb-icon').classList).toContain(
+      'dnb-icon--filled'
+    )
+  })
+
+  it('should apply filled prop even for icons not in the allowlist', () => {
```
langz: This test looks to be identical as the one above, `should have filled class when filled prop is true`

langz:
How does this work together with IconPrimary?
Can I use `filled` property and `filled` function with those?
Or perhaps these functionalities is only for `Icon`?

langz:
Not sure if it's worth adding a11y tests for filled prop and filled function, but could be worth considering 🙏

langz (approved):
Cool feature 🚀 

These donuts are filled 🍰 

Image

langz:
The filled function is quite cool and nice, so that others can use it where we use icon property, like in Button, etc.
But the API feels kind of different than how we usually do it. Usually we would do a `iconFilled` property or such.

tujoworker:
> How does this work together with IconPrimary?
> 
> Can I use `filled` property and `filled` function with those?
> 
> Or perhaps these functionalities is only for `Icon`?

You can always import them and do the same.

tujoworker:
> The filled function is quite cool and nice, so that others can use it where we use icon property, like in Button, etc.
> 
> But the API feels kind of different than how we usually do it. Usually we would do a `iconFilled` property or such.

True. But to save us some time and extra boilerplate code, I think this is a better option for this feature. Also, such icons will be used very rarely. And the support will always be low and for a limited selection of icons.

packages/dnb-eufemia/src/components/icon/__tests__/Icon.test.tsx
```diff
@@ -278,6 +278,55 @@ describe('Icon component', () => {
...
+      'dnb-icon--filled'
+    )
+  })
+
+  it('should not have filled class for unwrapped icons', () => {
+    render(<Icon icon={question} />)
+    expect(
+      document.querySelector('span.dnb-icon').classList
+    ).not.toContain('dnb-icon--filled')
+  })
```
langz: This test looks to do the exact same as `should not have filled class by default`

packages/dnb-eufemia/src/components/icon/style/dnb-icon.scss
```diff
@@ -39,6 +39,16 @@
     height: inherit;
   }
 
+  // filled
+  &--filled svg {
+    path,
+    circle,
+    rect,
+    polygon {
```
langz: Not sure if it exists more shapes in our icons, ellipse? 🤔

tujoworker: Nope, it does not.

langz: Thanks for checking 🙌

PR #8025, opened 2026-05-08
feat(Table): add `highlight` prop to `Th` and `Td` for highlighting

tujoworker:
Preview: https://feat-table-th-highlight.eufemia-e25.pages.dev/uilib/components/table#column-highlight

Based on:
Screenshot 2026-05-08 at 20 46 04


packages/dnb-eufemia/src/components/table/__tests__/__snapshots__/Table.test.tsx.snap
```diff
@@ -581,6 +605,20 @@ html:not([data-whatintent=touch]) .dnb-table__td__button:hover:not([disabled]) .
   min-width: 0;
 }
 
+.dnb-table__td--highlight {
+  --background-color: var(--table-th-highlight-background);
+}
+
+.dnb-table__td--highlight-border {
+  --table-border-top: var(--table-border-width) solid
+    var(--table-th-highlight-border);
```
langz, line 620: Could we just call it `var(--table-highlight-border);`?
To me, it seems a bit strange that we use -th variables in a `td`

packages/dnb-eufemia/src/components/table/__tests__/__snapshots__/Table.test.tsx.snap
```diff
@@ -581,6 +605,20 @@ html:not([data-whatintent=touch]) .dnb-table__td__button:hover:not([disabled]) .
   min-width: 0;
 }
 
+.dnb-table__td--highlight {
+  --background-color: var(--table-th-highlight-background);
```
langz: Could we just call it `var(--table-highlight-background);`?
To me, it seems a bit strange that we use -th variables in a `td`

packages/dnb-eufemia/src/components/table/TableTd.tsx
```diff
@@ -110,6 +126,7 @@ export default function Td(
 
   return (
     <td
+      ref={tdRef}
```
langz: I'm not 100% sure what we are using this useRef for, what purpose does it serve?

packages/dnb-eufemia/src/components/table/plugins/TableHighlightPlugin.ts
```diff
@@ -0,0 +1,71 @@
...
+ * highlighted cells.
+ *
+ * Usage:
+ * ```tsx
+ * import { highlightPlugin } from '@dnb/eufemia/components/Table'
+ *
+ * <Table plugins={[highlightPlugin]}>…</Table>
+ * ```
+ */
+export const highlightPlugin: TablePlugin = (tableElement) => {
```
langz: Just curious but why does one have to import this? Why isn't this just a default functionality?

packages/dnb-eufemia/src/components/table/TableTh.tsx
```diff
@@ -33,6 +35,12 @@ export type TableThProps = {
    */
   noWrap?: boolean
 
+  /**
+   * Highlights the entire column with a subtle background overlay.
```
langz: I guess this is only true if one would pass the plugin as well?
I tried without the plugin, and then it does not `Highlights the entire column with a subtle background overlay.`

langz (approved):

packages/dnb-eufemia/src/components/table/plugins/TableHighlightPlugin.ts
tujoworker: The hook exists because column highlighting requires DOM traversal after render — React has no built-in way for a Th in column 2 to tell all Td cells in that column to also get highlighted. Each cell renders independently.

packages/dnb-eufemia/src/components/table/TableTd.tsx
tujoworker: Its not used 🙏

tujoworker:
Here is the border situation 😀 It's correct. But looks weird with the colors.

Screenshot 2026-05-12 at 20 03 10

PR #8037, opened 2026-05-10
fix(Forms): ensure Form.useData is in sync when `data` prop changes

tujoworker:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1778156025744859
Reprod with fix: https://stackblitz.com/edit/u6dlg7ok-cvujuth2?file=src%2FApp.tsx

Target PR #8009 I think we should avoid using deepEqual here. It is expensive from a performance perspective.

Fixes this issue as well #5939
Reprod with fix: https://codesandbox.io/p/devbox/mutable-dream-8f2wtk?file=/src/App.tsx&workspaceId=ws_LdVovV1oCo5S5p2KEnsf6m


langz (approved):

tujoworker:
Added a new commit with a better approach.

I rewrote the shared state hook to use [useSyncExternalStore](https://react.dev/reference/react/useSyncExternalStore) – something I had tried before to fix the issue reported in PR #5939. This time, I think we have a solution that also covers that issue.

packages/dnb-eufemia/src/extensions/forms/DataContext/Provider/__tests__/Provider.test.tsx
```diff
@@ -163,354 +163,6 @@ describe('DataContext.Provider', () => {
...
-          <Field.String path="/foo" />
-          <MockConsumer />
-        </DataContext.Provider>
-      )
-
-      expect(input).toHaveValue('changed')
-      expect(span).toHaveTextContent('{"foo":"changed"}')
-    })
-
-    it('should not carry over removed keys when data prop loses a key with id', () => {
```
langz, line 250: I guess this is fine to remove, and we do not need this functionality?

What do we expect to happen when data prop changes from `{ foo: 'a', bar: 'b' }` to `{ foo: 'a' }`?
In this PR, the `internalDataRef.current` correctly becomes `{ foo: 'a' }`, but extendSharedData produces `{ foo: 'a', bar: 'b' } `— bar is never removed from the shared state. 

So Fields inside the Provider render correctly (they use internalDataRef), but any `Form.useData(id)` consumer outside the Provider would see the stale key.

tujoworker: 100% – reverted the tests and aligned the implementation.

PR #8053, opened 2026-05-12
feat(Autocomplete, MultiSelection): add unified search config prop

langz:
Extract shared search utilities (prepareSearchWords, findMatchingWords,
calculateTotalScore, checkMultipleNumericTerms) into shared/search/ and
introduce a SearchOptions type used by both Autocomplete and MultiSelection.

Autocomplete:
- Add search prop (SearchOptions) replacing deprecated individual props
  (disableFilter, disableReorder, disableHighlighting, searchNumbers,
  searchInWordIndex, searchMatch)
- Deprecated props still work with console warnings

MultiSelection:
- Add search prop to configure filtering when showSearchField is enabled
- Replace simple includes()-based filtering with shared search utilities
- Support numbers, matchInsideWordsFrom, match, filter, and reorder options
- Results are scored and reordered by relevance by default

Deploy preview [Autocomplete](https://feat-autocomplete-search-obj.eufemia-e25.pages.dev/uilib/components/autocomplete/demos)
Deploy preview [MultiSelection](https://feat-autocomplete-search-obj.eufemia-e25.pages.dev/uilib/extensions/forms/base-fields/MultiSelection/demos)

### DIFF
--- a/packages/dnb-eufemia/src/components/autocomplete/Autocomplete.tsx
+++ b/packages/dnb-eufemia/src/components/autocomplete/Autocomplete.tsx
@@ -41,6 +41,13 @@ import type { FormStatusBaseProps } from '../FormStatus'
 import type { IconIcon, IconSize } from '../Icon'
 import type { SkeletonShow } from '../Skeleton'
 import type { SpacingProps } from '../../shared/types'
+import type { SearchOptions } from '../../shared/search'
+import {
+  prepareSearchWords,
+  findMatchingWords,
+  calculateTotalScore,
+  passesNumericTermsCheck,
+} from '../../shared/search'
 import {
   warn,
   extendPropsWithContext,
@@ -48,7 +55,6 @@ import {
   dispatchCustomElementEvent,
   getStatusState,
   combineDescribedBy,
-  escapeRegexChars,
   getClosestParent,
 } from '../../shared/component-helper'
 import { IS_MAC, debounce, hasSelectedText } from '../../shared/helpers'
@@ -267,15 +273,22 @@ export type AutocompleteProps = {
    */
   showClearButton?: boolean
   /**
-   * If set to `true`, word highlighting will still be active, but no options will be filtered out. Defaults to `false`.
+   * Configure search behavior with a single config object. An object with optional keys: `filter` (enable result filtering, default `true`), `reorder` (enable relevance reordering, default `true`), `highlight` (enable text highlighting, default `true`), `numbers` (enable number search, default `false`), `matchInsideWordsFrom` (threshold for in-word search, default `3`), and `match` (matching mode `"word"` or `"starts-with"`, default `"word"`). Example: `search={{ filter: false }}` to disable filtering while keeping highlighting.
+   */
+  search?: SearchOptions
+  /**
+   * If set to `true`, word highlighting will still be active, but no options will be filtered out. Defaults to `false`. **Deprecated**: Use `search={{ filter: false }}` instead.
+   * @deprecated Use `search={{ filter: false }}` instead.
    */
   disableFilter?: boolean
   /**
-   * If set to `true`, reordering of search results will be disabled. Defaults to `false`.
+   * If set to `true`, reordering of search results will be disabled. Defaults to `false`. **Deprecated**: Use `search={{ reorder: false }}` instead.
+   * @deprecated Use `search={{ reorder: false }}` instead.
    */
   disableReorder?: boolean
   /**
-   * If set to `true`, word highlighting will be disabled, but the options will still get filtered. Defaults to `false`.
+   * If set to `true`, word highlighting will be disabled, but the options will still get filtered. Defaults to `false`. **Deprecated**: Use `search={{ highlight: false }}` instead.
+   * @deprecated Use `search={{ highlight: false }}` instead.
    */
   disableHighlighting?: boolean
   /**
@@ -299,15 +312,18 @@ export type AutocompleteProps = {
    */
   inputElement?: AutocompleteInputElement
   /**
-   * This gives you the possibility to change the threshold number, which defines from what word on we search "inside words". Defaults to `3`.
+   * This gives you the possibility to change the threshold number, which defines from what word on we search "inside words". Defaults to `3`. **Deprecated**: Use `search={{ matchInsideWordsFrom: number }}` instead.
+   * @deprecated Use `search={{ matchInsideWordsFrom: number }}` instead.
    */
   searchInWordIndex?: AutocompleteSearchInWordIndex
   /**
-   * Defines how search matching is performed. Use `starts-with` to only match items that begin with the first typed word. Defaults to `word`.
+   * Defines how search matching is performed. Use `starts-with` to only match items that begin with the first typed word. Defaults to `word`. **Deprecated**: Use `search={{ match: "word" | "starts-with" }}` instead.
+   * @deprecated Use `search={{ match: "word" | "starts-with" }}` instead.
    */
   searchMatch?: AutocompleteSearchMatch
   /**
-   * If set to `true` and `searchInWordIndex` is not set, the user will be able to more easily search and filter e.g. bank account numbers. Defaults to `false`.
+   * If set to `true` and `searchInWordIndex` is not set, the user will be able to more easily search and filter e.g. bank account numbers. Defaults to `false`. **Deprecated**: Use `search={{ numbers: true }}` instead.
+   * @deprecated Use `search={{ numbers: true }}` instead.
    */
   searchNumbers?: boolean
   /**
@@ -611,6 +627,7 @@ function AutocompleteComponent(ownProps: AutocompleteAllProps) {
     searchNumbers,
     searchInWordIndex,
     searchMatch,
+    search,
     showOptionsSr,
     selectedSr,
     submitButtonTitle,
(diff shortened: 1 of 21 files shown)


tujoworker:
I like the grouped `search` prop, but I’m not sure about the name `SearchConfig`.

To me, “config” sounds a bit static/global, while this is really per-instance search behavior that can change with props. I think `SearchOptions` would read more naturally:

> search?: SearchOptions

Especially since the object contains optional behavior toggles like `filter`, `reorder`, `highlight`, etc.

tujoworker:
I like the direction of consolidating the search-related props into `search={{ ... }}`. It makes the API much easier to extend and avoids more top-level props.

One thing I would reconsider is some of the option names:

- `filter`, `reorder` and `highlight` read well as positive booleans.
- `matchNumbers` feels a bit ambiguous. It sounds like “only match numbers”, while the behavior is more number-optimized/normalized search. Maybe `numbers: true`, `numeric: true` or `normalizeNumbers: true`?
- `inWordIndex` feels quite implementation-oriented. Something like `matchInsideWordsFrom` may be easier to understand from a consumer perspective.
- `match: 'starts-with'` is okay, though `strategy`/`mode` could also work if we expect more matching strategies later.

So maybe:

```
search={{
  filter: false,
  highlight: false,
  numbers: true,
  match: 'starts-with',
  matchInsideWordsFrom: 3,
}}
```

But overall, `search` as the new grouped prop feels like the right direction.

langz:
> I like the direction of consolidating the search-related props into `search={{ ... }}`. It makes the API much easier to extend and avoids more top-level props.
> 
> One thing I would reconsider is some of the option names:
> 
> * `filter`, `reorder` and `highlight` read well as positive booleans.
> * `matchNumbers` feels a bit ambiguous. It sounds like “only match numbers”, while the behavior is more number-optimized/normalized search. Maybe `numbers: true`, `numeric: true` or `normalizeNumbers: true`?
> * `inWordIndex` feels quite implementation-oriented. Something like `matchInsideWordsFrom` may be easier to understand from a consumer perspective.
> * `match: 'starts-with'` is okay, though `strategy`/`mode` could also work if we expect more matching strategies later.
> 
> So maybe:
> 
> ```
> search={{
>   filter: false,
>   highlight: false,
>   numbers: true,
>   match: 'starts-with',
>   matchInsideWordsFrom: 3,
> }}
> ```
> 
> But overall, `search` as the new grouped prop feels like the right direction.

Thanks :) 
I'll start looking at this again now 🤞

langz:
> > I like the direction of consolidating the search-related props into `search={{ ... }}`. It makes the API much easier to extend and avoids more top-level props.
> > One thing I would reconsider is some of the option names:
> > 
> > * `filter`, `reorder` and `highlight` read well as positive booleans.
> > * `matchNumbers` feels a bit ambiguous. It sounds like “only match numbers”, while the behavior is more number-optimized/normalized search. Maybe `numbers: true`, `numeric: true` or `normalizeNumbers: true`?
> > * `inWordIndex` feels quite implementation-oriented. Something like `matchInsideWordsFrom` may be easier to understand from a consumer perspective.
> > * `match: 'starts-with'` is okay, though `strategy`/`mode` could also work if we expect more matching strategies later.
> > 
> > So maybe:
> > ```
> > search={{
> >   filter: false,
> >   highlight: false,
> >   numbers: true,
> >   match: 'starts-with',
> >   matchInsideWordsFrom: 3,
> > }}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > But overall, `search` as the new grouped prop feels like the right direction.
> 
> Thanks :) I'll start looking at this again now 🤞

I've tried my best to address this feedback in the latest code now 🤞

tujoworker (changes requested):
The extracted utility changes the existing starts-with behavior. Previously only the first search term was anchored to the start of the item; later terms still followed matchInsideWordsFrom and could match inside words. The new !startsWithMatch condition disables inside-word matching for every term, and the new test explicitly locks in that incompatible behavior. Please preserve the previous semantics and add a regression test covering a later term after the threshold.

langz:
Thanks @tujoworker, good catch! Restored the previous starts-with semantics: only the first term is anchored to the start of the item, while later terms past `matchInsideWordsFrom` can again match inside words.

- Removed the `!startsWithMatch` condition in `prepareSearchWords` so the threshold behaves as before.
- Replaced the test that locked in the old behavior with a regression test covering a later term after the threshold (plus a term before it still requiring a word boundary).

Also rebased onto the latest `main`.

langz:
Small follow-up cleanup: removed the unused `getWordBoundary` from the `PreparedSearch` return (it was never consumed — callers that need boundaries compute their own). No behavior change; pushed as a separate commit.

tujoworker:
This also needs an Autocomplete migration entry in `v13-info.mdx`. The PR deprecates six public props that are expected to be removed in the next major, so consumers need the replacement mapping:

- `disableFilter` → `search.filter`
- `disableReorder` → `search.reorder`
- `disableHighlighting` → `search.highlight`
- `searchNumbers` → `search.numbers`
- `searchInWordIndex` → `search.matchInsideWordsFrom`
- `searchMatch` → `search.match`

Could you add a short before/after example under Component changes, consistent with the other v13 migrations? MultiSelection only gains the new API, so it does not need migration guidance.

langz:
Added an Autocomplete entry under **Component changes** in `v13-info.mdx` (plus a TOC link) with the six prop → `search.*` mappings and a before/after example. I also called out that the `disable*` booleans map to their positive counterpart (so a disabled feature becomes `false`), to avoid a subtle inversion mistake during migration.

Left MultiSelection out since it only gains the new API. Pushed as `59a4177`.

packages/dnb-eufemia/src/components/autocomplete/Autocomplete.tsx
```diff
@@ -664,6 +681,71 @@ function AutocompleteComponent(ownProps: AutocompleteAllProps) {
...
   } = props
 
+  // Resolve search config: `search` prop takes precedence over deprecated props.
+  // @deprecated - The fallbacks to deprecated props below ensure backward compatibility.
+  // Remove these fallbacks when the deprecated props are removed.
+  const disableFilter =
+    search?.filter !== undefined
+      ? !search.filter
+      : _deprecatedDisableFilter // @deprecated fallback
+  const disableReorder =
```
tujoworker, line 691: `disableFilter` and `disableReorder` are recomputed here, but `skipFilterRef` and `skipReorderRef` below capture them only on mount. Changing `search.filter` or `search.reorder` after mount therefore leaves filtering/reordering on the previous values, unlike the other search options. Could we keep those refs synchronized (or read the current resolved values directly) and add a rerender regression test that toggles both options?

tujoworker:
I pushed a small fix that keeps `search.filter` and `search.reorder` in sync when the prop changes, with a regression test for both.

tujoworker (approved):
Looks good now. The search options stay in sync when changed, and the regression test covers the behavior.

langz:
Thanks for your improvements @tujoworker 🚀 
I took a look through this, and I think it looks great 😄 ✅ 

An agent added two comments, that we potentially should look into. I guess we'll evaluate it.
See comment below:
https://github.com/dnbexperience/eufemia/pull/8053#pullrequestreview-5029922884

packages/dnb-eufemia/src/extensions/forms/Field/MultiSelection/MultiSelection.tsx
```diff
@@ -280,37 +316,100 @@ function MultiSelection(props: FieldMultiSelectionProps) {
...
+            }
+
+            return null
+          }
+
           const matches =
-            title.includes(searchLower) ||
-            text.includes(searchLower) ||
-            description.includes(searchLower)
+            matchedWords.length === prepared.searchWordsData.length
```
langz: This requires *all* search words to match (AND), but Autocomplete keeps items that match *any* word (OR) and just reorders. So `banana cake` filters differently in the two components even though they now share the same `search` prop. Should we align them, or at least document the difference? 🤔

**Update:** attempted a fix in #9132 — aligned MultiSelection to OR + relevance reorder so it matches Autocomplete. Multiple numeric terms still require all to match.

packages/dnb-eufemia/src/extensions/forms/Field/MultiSelection/MultiSelection.tsx
```diff
@@ -237,9 +252,27 @@ function MultiSelection(props: FieldMultiSelectionProps) {
...
+        'ArrowDown',
+        'PageUp',
+        'PageDown',
+        'End',
+        'Home',
+      ])
+    }
+
+    return () => {
+      whatInput.specificKeys(['Tab'])
```
langz: Two issues here: this is unrelated to the search config, and the cleanup resets the global to `['Tab']` while the app default (set once in `component-helper`) is `[]`. Nothing else touches `specificKeys` at runtime, so after the first open/close every non-Tab key stops registering as keyboard input app-wide (breaks focus-visible everywhere). Verified: before, `a` → `data-whatinput="keyboard"`; after open+unmount, `a` → `"mouse"` and only `Tab` still counts. We should capture + restore the previous value, or drop this from the PR. 😬

**Update:** attempted a fix in #9132 — removed the effect entirely (your call in the review, thanks). The pre-existing MultiSelection keyboard tests still pass without it.

langz (commented):
A couple of things I noticed while going back over this 👇

tujoworker (changes requested):
You are right. The `what-input` effect is unrelated to the search change and restores the wrong global state. This should be removed from this PR.

langz:
> You are right. The `what-input` effect is unrelated to the search change and restores the wrong global state. This should be removed from this PR.

Great, I'll do so

tujoworker (approved):

PR #8055, opened 2026-05-12
chore: align TSDoc comments with Docs.ts property documentation 

langz:


packages/dnb-eufemia/src/components/radio/Radio.tsx
```diff
@@ -88,7 +88,7 @@ export type RadioProps = {
    */
   group?: string
   /**
-   * The size of the Radio button. For now there is **medium** (default) and **large**.
+   * The size of the Radio button. For now there is `default`, `medium` (default) and `large`.
```
joakbjerk: Do we need a value that is default, if medium is set as the default value?

packages/dnb-eufemia/src/components/height-animation/HeightAnimation.tsx
```diff
@@ -22,14 +22,14 @@ export type HeightAnimationProps = {
   showOverflow?: boolean
 
   /**
-   * Defines the duration of the animation in milliseconds.
-   * Default: `400`
+   * Custom duration of the animation in milliseconds.
+   * Default: `400ms`
```
joakbjerk: Should this stay as `400` instead of `400ms`, as the value the user sets is a number and not number + ms?

packages/dnb-eufemia/src/components/height-animation/HeightAnimation.tsx
```diff
@@ -22,14 +22,14 @@ export type HeightAnimationProps = {
...
    */
   duration?: number
 
   /**
-   * Defines the delay of the animation in milliseconds.
-   * Default: `0`
+   * Custom delay of the animation in milliseconds.
+   * Default: `0ms`
```
joakbjerk: Should this stay as `+` instead of `0`, as the value the user sets is a number and not number + ms?

joakbjerk (approved):

PR #8058, opened 2026-05-12
chore: add sync-docs-jsdoc ESLint rule

tujoworker:
Ensures JSDoc comments on type/interface properties match the doc strings defined in sibling *Docs files. Mismatches are auto-fixable.

Related to #8057

In the editor

https://github.com/user-attachments/assets/6050cd96-9b8e-46db-a2a8-6ecf8765c5e3

Or we could run it in the CLI as well in a bulk run.


langz (approved):
Thanks, I like it :)

PR #8059, opened 2026-05-12
chore(Radio): fix cloneNode losing radio checked state in screenshot tests

tujoworker:
measureWrapperWidth() cloned the element to pre-measure inline-block width. The clone duplicated radio inputs with the same name attribute, causing the browser to uncheck the originals. Replaced with direct measurement on the real wrapper after reparenting, eliminating the clone entirely. Update affected snapshots.


langz (approved):

snorrekim:
Hm, looks like every test we set `pageViewport.width` had screenshots that were 33px smaller, missing the left and right padding. And now they are are 65px smaller. So some kind of 4rem+1px error.

We used to take the width of the smallest of the parent element, or the element in a floating infinite space box.

Now we take the smallest of the parent element or the element "when it is inside the wrapper". 

Interestingly, we get a more correct screenshot on small viewports now, we used to basically stretch the component by 2rem (the padding on the parent "example box" element).

But since we are already rendering a slightly different page for screenshot tests, perhaps we should also remove the left and right padding from the example box?

This is the element that is screenshotted (the highlighted copy above the original): 
image

But the example box adds 64px of padding that isn't expected in a usual situation. If we take a screenshot with viewport 400, I should expect that the element has "some" padding, but not 96px.

More correct without example box padding: 
image


Also, as a side note, we are adding a 1px border on the `.dnb-app-content` div, we should change it to something that doesn't change the content area size.

tujoworker:
> But since we are already rendering a slightly different page for screenshot tests, perhaps we should also remove the left and right padding from the example box?

Removing it results in some cases that the element is not "centered", like when taking screenshots of Tooltip (Button).
But, I we are able to improve on it. Will add another commit.

PR #8063, opened 2026-05-12
chore(Portal): migrate from transformWithEsbuild to transformWithOxc

tujoworker:


langz (approved):
Faster, faster faster ⚡

PR #8079, opened 2026-05-13
feat(Field.Time): add `Time` field

joakbjerk:
[Motivation](https://dnb-it.slack.com/archives/CMXABCHEY/p1769691362432079)

Deploy preview: https://feat-field-time.eufemia-e25.pages.dev/uilib/extensions/forms/feature-fields/Time


langz:
Initial thoughts:
- is 24:00 a valid time? As well as 00:00?
- I was able to experience the same issue as solved [here ](https://github.com/dnbexperience/eufemia/pull/7958)
- onblur validation when entering  12:1 so the input field looks like 12:1m

joakbjerk:
> Initial thoughts:
> 
>     * is 24:00 a valid time? As well as 00:00?
> 
>     * I was able to experience the same issue as solved [here ](https://github.com/dnbexperience/eufemia/pull/7958)
> 
>     * onblur validation when entering  12:1 so the input field looks like 12:1m

Hmm, I feel like it would add more complexity to the time validation if we accept 24:00 and 00:00. I don't think many people write 24:30 instead of 00:30, or maybe in some cultures? 🤔 So then we would have to do a check on the minute value to make sure its not higher than 00 when hour is 24?

The issue you experienced in `Expiry` should be fixed for `Time` as well now 🫡

I have found a fix for point number 3, but what should the expected behavior be here? Should the user type the leading 0 themselves, or should add the 0 ourselves if the user only gives us a single digit? The last approach seems to go against what we have done earlier with not controlling the users input 🤷‍♂️

Edit: Went for the solution where users have to input the leading zero themselves, to reduce code complexity and UX-confusion, but that might not be the right approach?

langz:
Do we need a corresponding Value.Time component? Or can it reuse some other Value.X component? If it can reuse, then we can just link to it from the Field.Time page, like we do in other Field.X components.

I think this could be worth looking into before releasing this, as it could be that one would need or expect both to be available together.

packages/dnb-eufemia/src/extensions/forms/Field/Time/Time.tsx
```diff
@@ -0,0 +1,380 @@
+import { useCallback, useMemo } from 'react'
+import type {
+  FieldProps,
+  Validator,
+  ValidatorWithCustomValidators,
+} from '../../types'
+import { pickSpacingProps } from '../../../../components/flex/utils'
+import { useFieldProps } from '../../hooks'
+import clsx from 'clsx'
```
langz:
```suggestion
import { clsx } from 'clsx'
```

Nitpick https://github.com/dnbexperience/eufemia/pull/8169

langz:
Potential future feature request:
- `minTime`
- `maxTime` 

Not for now I guess.

langz:
> Initial thoughts:
> 
> * is 24:00 a valid time? As well as 00:00?
> * I was able to experience the same issue as solved [here ](https://github.com/dnbexperience/eufemia/pull/7958)
> * onblur validation when entering  12:1 so the input field looks like 12:1m

I think 24:00  is technically a valid ISO 8601 time representing midnight at the end of a day.

joakbjerk:
> Do we need a corresponding Value.Time component? Or can it reuse some other Value.X component? If it can reuse, then we can just link to it from the Field.Time page, like we do in other Field.X components.
> 
> I think this could be worth looking into before releasing this, as it could be that one would need or expect both to be available together.

Great point, I'll take a look 🫡

packages/dnb-eufemia/src/extensions/forms/Field/Time/TimeDocs.ts
```diff
@@ -0,0 +1,19 @@
+import type { PropertiesTableProps } from '../../../../shared/types'
+import { InputProperties } from '../../../../components/input/InputDocs'
+
+export const TimeProperties: PropertiesTableProps = {
```
langz, line 5: I think the `onChange` [docs](https://feat-field-time.eufemia-e25.pages.dev/uilib/extensions/forms/feature-fields/Time/events#events) could be improved, like we have in https://feat-field-time.eufemia-e25.pages.dev/uilib/extensions/forms/feature-fields/DateOfBirth/events

To better describe the output.

joakbjerk:
> > Initial thoughts:
> > 
> > * is 24:00 a valid time? As well as 00:00?
> > * I was able to experience the same issue as solved [here ](https://github.com/dnbexperience/eufemia/pull/7958)
> > * onblur validation when entering  12:1 so the input field looks like 12:1m
> 
> I think 24:00 is technically a valid ISO 8601 time representing midnight at the end of a day.

24:00 is valid ISO, so I could do a quick iteration of supporting that, just to see if the complicates the time validation more or not, because then we would need to warn the user if they input "24:05" for example, as that is not valid 🤔

langz:
> > > Initial thoughts:
> > > 
> > > * is 24:00 a valid time? As well as 00:00?
> > > * I was able to experience the same issue as solved [here ](https://github.com/dnbexperience/eufemia/pull/7958)
> > > * onblur validation when entering  12:1 so the input field looks like 12:1m
> > 
> > 
> > I think 24:00 is technically a valid ISO 8601 time representing midnight at the end of a day.
> 
> 24:00 is valid ISO, so I could do a quick iteration of supporting that, just to see if the complicates the time validation more or not, because then we would need to warn the user if they input "24:05" for example, as that is not valid 🤔

Yeah, I get it, it complicates a bit more, hehe 😂

tujoworker:
> > > Initial thoughts:
> > > 
> > > * is 24:00 a valid time? As well as 00:00?
> > > * I was able to experience the same issue as solved [here ](https://github.com/dnbexperience/eufemia/pull/7958)
> > > * onblur validation when entering  12:1 so the input field looks like 12:1m
> > 
> > 
> > I think 24:00 is technically a valid ISO 8601 time representing midnight at the end of a day.
> 
> 24:00 is valid ISO, so I could do a quick iteration of supporting that, just to see if the complicates the time validation more or not, because then we would need to warn the user if they input "24:05" for example, as that is not valid 🤔

24:05 is not valid ISO 8601 time.

24:01 ❌ invalid
00:01 ✅ valid

https://www.iso.org/iso-8601-date-and-time-format.html

langz:
`<input type="time" />` does not support 24:00.
AI says `Intl.DateTimeFormat doesn't handle it — new Date().setHours(24, 0, 0) rolls over to the next day. The Value.Time component would display it incorrectly.`

Let's discuss it together 🤔 
Perhaps we should involve some actual back-end devs who have dealt with time, and those expectations.

langz:
> Potential future feature request:
> 
> * `minTime`
> * `maxTime`
> 
> Not for now I guess.

But this could also be a separate component, `Field.TimeRange` 😂

langz (commented):
Could we remove the `// @ts-expect-error - strictFunctionTypes`, not sure if we usually add these? If not possible, please ignore this comment :)

langz (approved):
To me it looks great ⏲️ 

Not entirely sure myself if we should support 24:00 or not.
Either way, I think we should document the current decision in our docs, if we support ISO X or if we do not support 24:00. I do however think it's quite normal to write 24:00 🤔

tujoworker (approved):

joakbjerk:
> To me it looks great ⏲️
> 
> Not entirely sure myself if we should support 24:00 or not. Either way, I think we should document the current decision in our docs, if we support ISO X or if we do not support 24:00. I do however think it's quite normal to write 24:00 🤔

Added support for it now, as it could be a valid way of the user sayings its at the end of the day, instead of the "start" 08:00-24:00 makes more sense than 08:00-00:00, maybe? I don't know

tujoworker:
I think we should add the "new" badge to that component.

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/feature-fields/Time.mdx
```diff
@@ -0,0 +1,26 @@
+---
+title: 'Field.Time'
+description: '`Field.Time` is a wrapper component for the input of strings, with user experience tailored for time values (hours and minutes, optionally seconds).'
+showTabs: true
```
langz, line 4:
```suggestion
showTabs: true
status: 'new'
```

langz:
> I think we should add the "new" badge to that component.

https://github.com/dnbexperience/eufemia/pull/8079/changes/641bc783decfbd4c1e88db43438526705baad1ca

PR #8089, opened 2026-05-13
feat(Space): add useSpacing hook and apply to all components and deprecate applySpacing

tujoworker:
So we can use dynamic spacing with React Context instead of div wrapper.

The hook is as simple as this for now:
```tsx
export const useSpacing = <T extends ApplySpacingTarget>(
  props: SpacingProps | SpacingUnknownProps,
  target: T,
  elementName: string | null = null
): T => {
  return applySpacing(props, target, elementName)
}
```

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/layout/space/info.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/layout/space/info.mdx
@@ -13,7 +13,7 @@ import { Space } from '@dnb/eufemia'
 
 ## Description
 
-The Space component provides `margins` within the [provided spacing patterns](/uilib/usage/layout/spacing#spacing-helpers).
+The Space component provides `margins` and inner `padding` within the [provided spacing patterns](/uilib/usage/layout/spacing#spacing-helpers).
 
 ## Relevant links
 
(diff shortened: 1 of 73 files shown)


packages/dnb-design-system-portal/src/docs/uilib/layout/space/info.mdx
```diff
@@ -13,7 +13,7 @@ import { Space } from '@dnb/eufemia'
 
 ## Description
 
-The Space component provides `margins` within the [provided spacing patterns](/uilib/usage/layout/spacing#spacing-helpers).
+The Space component provides `margins` and inner `padding` within the [provided spacing patterns](/uilib/usage/layout/spacing#spacing-helpers).
```
langz, line 16: 👍

langz (approved):
Looks great, I have nothing to complain about 🥇 

Should we use `useSpacing` in:
- AccordionHeader.tsx
- Badge.tsx 
- DateFormat.tsx 
- Icon.tsx 
- Modal.tsx 
- Pagination.tsx 
- Tabs.tsx 
- TabsContentWrapper.tsx 
- TabsCustomContent.tsx 
- Tooltip.tsx

tujoworker:
> Should we use `useSpacing` in:

I’m pretty sure I searched for more places without finding any, but I can see now that there are still some leftovers. Nice finding 💛

PR #8107, opened 2026-05-15
fix(Provider): use createSharedState to prevent infinite render loop

langz:
Motivation: https://main.eufemia-e25.pages.dev/uilib/extensions/forms/Form/Handler/
Deploy preview: https://fix-provider-shared-state-lo.eufemia-e25.pages.dev/uilib/extensions/forms/Form/Handler/

This PR fixes `Error: Minified React error #185` in https://main.eufemia-e25.pages.dev/uilib/extensions/forms/Form/Handler/

The Provider used useSharedState for sharedDataContext, which subscribes via useSyncExternalStore. Since the Provider only writes to this store (never reads), the reactive subscription was unnecessary. Because the Provider updates the store during render with a new contextValue object each time, useSyncExternalStore detects the snapshot change during commit and forces a re-render, causing 'Maximum update depth exceeded'.

Switch to createSharedState which creates/returns the store instance without subscribing to it.


tujoworker (approved):
Very good you found this! 💯

PR #8109, opened 2026-05-15
chore(Portal): stabilize focusmode ID across HMR remounts

tujoworker:
Adds a babel plugin (`injectStableName`) that injects a `stableName` prop on each `<ComponentBox>`, derived from the enclosing export function name. This gives each code example a stable identity that survives hot module replacement, fixing focus mode breaking during HMR.


langz (approved):
Fancy fancy

PR #8115, opened 2026-05-16
chore: add test:screenshots:renew script for targeted snapshot regeneration

tujoworker:
Adds a `test:screenshots:renew` script that deletes existing screenshot snapshots for specific components and regenerates them.

### Usage

```sh
yarn test:screenshots:renew phone
yarn test:screenshots:renew phone radio
```

### How it works

1. Finds `__image_snapshots__` directories matching the given filters and deletes the PNGs
2. Finds matching `*.screenshot.test.{ts,tsx}` files
3. Runs vitest with a `SCREENSHOT_INCLUDE` env var to scope the run to only the matching test files (vitest browser mode ignores CLI filter args)

### Changes

- **`scripts/tools/renewScreenshots.ts`** — new script handling deletion + scoped vitest invocation
- **`vitest.config.screenshots.ts`** — reads `SCREENSHOT_INCLUDE` env var to override the `include` pattern
- **`package.json`** — adds the `test:screenshots:renew` script entry
- **`make-and-run-tests.mdx`** — documents the new command


langz (approved):
Good good :)

I'm assuming we will not run this script very often. 
Or is it likely that we'll run this now instead of :update ?

tujoworker:
No, its not something we use often. I just have seen the need over the years. Its when we rename or remove a test. Instead of finding the pngs manually, you can just run this.

PR #8118, opened 2026-05-18
chore(Portal): use automatic JSX runtime in react-live-ssr

langz:
Motivation: console warning: `Your app (or one of its dependencies) is using an outdated JSX transform. Update to the modern JSX transform for faster performance: https://react.dev/link/new-jsx-transform`

Screenshot 2026-05-18 at 10 02 16

Deploy preview: https://fix-jsx-transform-warning.eufemia-e25.pages.dev/uilib/extensions/forms/base-fields/String/demos


tujoworker (approved):
We drift more and more away from the original. I think we need to make an attempt to fix it.

langz:
> We drift more and more away from the original. I think we need to make an attempt to fix it.

Yeah, agree 🤝

PR #8127, opened 2026-05-18
fix(Tabs): prevent setState during render in shared state initialization

langz:
Use silent: true when setting initial shared state data during render to avoid triggering subscribers. This fixes the React warning 'Cannot update a component while rendering a different component' that occurred when navigating between tab pages.

Can be experienced when navigation from http://localhost:8000/uilib/components/fragments to http://localhost:8000/uilib/components/fragments/drawer-list by clicking the DrawerList link/menu item in the menu:

Error message: `Cannot update a component (`ContentWrapper`) while rendering a different component (`Tabs`). To locate the bad setState() call inside `Tabs`, follow the stack trace as described in https://react.dev/link/setstate-in-render`

Image:
image

### DIFF
--- a/packages/dnb-eufemia/src/components/tabs/__tests__/Tabs.test.tsx
+++ b/packages/dnb-eufemia/src/components/tabs/__tests__/Tabs.test.tsx
@@ -33,6 +33,44 @@ const contentWrapperData = {
 }
 
 describe('Tabs component', () => {
+  it('should not trigger setState warnings when using shared state with ContentWrapper', () => {
+    const consoleError = jest.spyOn(console, 'error').mockImplementation()
+
+    const sharedId = 'shared-tabs-id'
+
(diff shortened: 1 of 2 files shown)


tujoworker (approved):

packages/dnb-eufemia/src/components/tabs/__tests__/Tabs.test.tsx
```diff
@@ -33,6 +33,35 @@ const contentWrapperData = {
 }
 
 describe('Tabs component', () => {
+  it('should not trigger setState warnings when using shared state with ContentWrapper', () => {
```
langz, line 36: I'll do some modifications to the test, to make sure that it will fail when the actual fix is not provided.

langz: Done so now

PR #8129, opened 2026-05-18
chore(Skeleton): remove useless concat() call in figure components

langz:
The concat() return value was being discarded. Since reverse() already mutates the array in place, the concat() call served no purpose.

Also modernized new Array(n).fill().map() to Array.from() for clarity.


tujoworker (approved):
😆

PR #8140, opened 2026-05-19
chore: migrate Jest APIs to Vitest

tujoworker:


packages/dnb-eufemia/src/shared/useMedia.ts
```diff
@@ -200,7 +203,10 @@ export default function useMedia(
 
         return acc
       },
-      { isSSR: !isMatchMediaSupported(), key: null } as UseMediaResult
+      {
+        isSSR: isSSRTest() || !isMatchMediaSupported(),
```
langz: Feels a bit iffy to add checks for test to production code 🤔 
Could this instead be solved by mocking `isMatchMediaSupported() `in the relevant tests?
Or is the overhead of doing that too much?

packages/dnb-eufemia/eslint.config.mjs
```diff
@@ -366,27 +374,14 @@ export default [
...
-      jest: jestPlugin,
-    },
-    settings: {
-      // See comment above — jest package is removed; version must be explicit.
-      jest: { version: 30 },
+      vitest: vitestPlugin,
     },
     rules: {
+      ...vitestPlugin.configs.recommended.rules,
+      'vitest/expect-expect': 'off',
```
langz: Do we need this?
I generally like that tests must have expect assertions

langz (approved):

langz:
Super nice work, I love reducing the number of packages we use 🥇 🥈 🥉

langz:
Not important, but do we know if tests will run faster or slower with these changes?
I guess it does not matter too much, if it's not a significant difference, as I think there's such a great upside to only using one package for testing instead of multiple

packages/dnb-eufemia/src/core/vitest/setupVitest.ts
```diff
@@ -66,6 +52,16 @@ expect.extend({
   },
 })
 
+expect.addEqualityTesters([
```
langz: Is it safe to add this?
If I understand it correctly, it will make all toEqual matchings with an array of strings to be order-insensitive.

If this is how it previously worked with jest, then of course it's fine.

Potentially we could make it opt-in via an explicit custom matcher like toEqualClassNames()?

packages/dnb-eufemia/eslint.config.mjs
tujoworker: Did remove it.

packages/dnb-eufemia/src/shared/useMedia.ts
tujoworker: Agree 💯 Did remove it 👍

packages/dnb-eufemia/src/core/vitest/setupVitest.ts
tujoworker: We, now we use `toHaveClass` and some places with "exact": `expect(element).toHaveClass('foo bar', { exact: true })`

tujoworker: `{ exact: true }` is from `@testing-library/jest-dom`’s `toHaveClass` matcher.

tujoworker: We use now toHaveClass directly.

packages/dnb-eufemia/eslint.config.mjs
tujoworker: Did remove it and align everything.

packages/dnb-eufemia/src/shared/useMedia.ts
tujoworker: Nice you cached this. I agree 💯. Its aligned now.

PR #8149, opened 2026-05-19
fix(ToggleButton): fix checked border width and make toggle state visible when focused

snorrekim:
- increase border width when checked to `2px`
- no longer changes background and text color when focused, so the user can still see if the toggle is checked or not


snorrekim:
After a quick chat with UX we agreed that this is an acceptable quick fix for now.

tujoworker (approved):

langz (approved):

PR #8159, opened 2026-05-20
chore(Portal): add copy code button to live toolbar

langz:
Deploy preview: https://feat-copy-code-button.eufemia-e25.pages.dev/uilib/components/button/demos
https://github.com/user-attachments/assets/2f64346a-fae6-4f8f-84fb-7e29249d65fb


tujoworker:
Nice – very similar to #8164 but not overlapping. But maybe we should align it somehow? or do they work just fine in parallel?

langz:
> Nice – very similar to #8164 but not overlapping. But maybe we should align it somehow? or do they work just fine in parallel?

Yeah, I've adjusted now, so that both use a shared component, making sure the experience is quite the same.

tujoworker (approved):

PR #8165, opened 2026-05-20
fix(Tooltip): reposition when children content changes

tujoworker:
Recalculates Popover/Tooltip position when `children` content changes while the popover is active. This ensures the tooltip stays correctly positioned when the content it's anchored to changes size.

### DIFF
--- a/packages/dnb-eufemia/src/components/popover/PopoverContainer.tsx
+++ b/packages/dnb-eufemia/src/components/popover/PopoverContainer.tsx
@@ -250,6 +250,12 @@ function PopoverContainer(props: PopoverContainerProps) {
     setResolvedPlacement(placement)
   }, [placement])
 
+  useEffect(() => {
+    if (isActive) {
+      requestRecalculation()
+    }
+  }, [children, isActive, requestRecalculation])
+
   useLayoutEffect(() => {
     if (typeof window === 'undefined') {
       return undefined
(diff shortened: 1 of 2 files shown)


packages/dnb-eufemia/src/components/popover/PopoverContainer.tsx
```diff
@@ -250,6 +250,12 @@ function PopoverContainer(props: PopoverContainerProps) {
     setResolvedPlacement(placement)
   }, [placement])
 
+  useEffect(() => {
+    if (isActive) {
+      requestRecalculation()
+    }
+  }, [children, isActive, requestRecalculation])
```
langz, line 257: Could it be problematic using children as a dep in useEffect?
Leading to infinite loops or unnecessary re-renders?

Not sure if it's the case gere.

langz (approved):

packages/dnb-eufemia/src/components/popover/PopoverContainer.tsx
tujoworker: Yeah, definitively unnecessary re-renders, but then it happens only when "active".

tujoworker: We don't have another option. We could potentially add a prop to enable this. Would you prefer that?

langz: resizeobserver? Not sure.
Probably not too important either.

PR #8167, opened 2026-05-20
chore(Portal): add compact locale switcher to live code toolbar

langz:


tujoworker:
I have thought about this before, but concluded that we not often switch locale during development. But maybe its just me. I just don't want too much stuff in the toolbar at the end.

PR #8179, opened 2026-05-21
feat: add `triggerProps` support to DatePicker and Field.Date

tujoworker:


langz (approved):

langz (approved):
Could it be worth standardizing the type for triggerProps? Making it a shared type if possible?

tujoworker:
Yeah, On the Dialog component we call it `triggerAttributes` – but props is more accurate, is it not? So maybe we should rename it there? happy to get inn a suggestion/PR on that.

langz:
> Yeah, On the Dialog component we call it `triggerAttributes` – but props is more accurate, is it not? So maybe we should rename it there? happy to get inn a suggestion/PR on that. 

Not sure what makes most sense. 
I would just continue down the existing path that we have, and bulk change it later in v12 for all props, instead of having multiple props that do the same with different naming in the codebase at once.

PR #8185, opened 2026-05-21
chore(CodeBlock): remove duplicate fullscreen icon import

langz:
The fullscreen icon was imported twice under different aliases (focusModeIcon and focusModePaddingIcon). Both resolved to the same icon. Reuse focusModeIcon for both purposes.


tujoworker (approved):
I know about that. But at the end, we do not end up having two svgs, but only one anyway. But I'm fine with the change too.

PR #8192, opened 2026-05-21
feat(DatePicker): show only one calendar in `range` and add `sideBySide` option for two calendars

tujoworker:
Motivation: Enhance UX
Preview: https://feat-date-picker-single-cale-rt8o.eufemia-e25.pages.dev/uilib/components/date-picker

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/components/date-picker/Examples.tsx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/components/date-picker/Examples.tsx
@@ -221,7 +221,13 @@ export const DatePickerErrorStatus = () => (
 
 export const DatePickerCalendarInline = () => (
   <ComponentBox background="plain">
-    <DatePicker inline range startDate="2019-05-05" endDate="2019-06-05" />
+    <DatePicker
+      inline
+      range
+      sideBySide
+      startDate="2019-05-05"
+      endDate="2019-06-05"
+    />
   </ComponentBox>
 )
 
--- a/packages/dnb-design-system-portal/src/docs/uilib/components/date-picker/visual-tests.tsx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/components/date-picker/visual-tests.tsx
@@ -164,6 +164,7 @@ const DatePickerRangeMode = () => (
     <DatePicker
       inline
       range
+      sideBySide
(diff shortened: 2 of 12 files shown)


packages/dnb-design-system-portal/src/docs/uilib/components/date-picker/demos.mdx
```diff
@@ -16,6 +16,7 @@ import {
   DatePickerErrorStatus,
   DatePickerCalendarInline,
   DatePickerYearNavigation,
+  DatePickerSingleCalendar,
```
snorrekim: This is never used, and doesn't exist in "Examples"?

snorrekim: How did it pass the checks?

tujoworker: We can't run eslint on mdx files.

snorrekim:
There was a bug with the range keyboard navigation, and it's worse with a single pane.

**Existing bug:**
If you use keyboard (tab , arrow keys, and enter) you can change the date to something that is reversed, just move one of the dates to the other side:
image

**New bug:**
When there is only one pane, you can only keyboard move the "first" data. Before, you moved the "first" date by tabbing to the first pane, and the "last" date by tabbing to the second pane.

**Suggestions:**
The currently focused day-marker should probably get the focus effect. 

But I am unsure what is best accessibility for when the user reverses the order. Should we swap the selectors and change focus? Or would that be confusing because now the next element isn't the one you expected?

tujoworker:
Thanks for the findings. This is fixed now.

snorrekim:
It's still buggy. As far as I can tell you can now use "enter" to change the date being manipulated. But if you move the date across, the focus swaps to the wrong element. And you ned to press enter twice to actually update and get the correct focus+active date. If you just use keyboard it's still possible to select an invalid reversed range.

Also, it's not very clear that enter changes the focused date.


https://github.com/user-attachments/assets/525f61c6-ccdd-4836-9f7d-4d24d9e81ad8

tujoworker:
Closing this in to split up feature and bug.

tujoworker:
Here is a following up PR #8232

PR #8195, opened 2026-05-21
chore(Field.Date): derive props from single source of truth

tujoworker:
Replace the duplicated `Pick` type and runtime `datePickerPropKeys` array in `Field.Date` with a single `as const satisfies` array. The `PickedDatePickerProps` type is now derived from the array, so adding a new `DatePicker` prop to forward only requires one change.

Also removes stale duplicate `onDaysRender` and `showInput` entries from the runtime array.

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/Date/Date.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/Date/Date.tsx
@@ -72,47 +72,7 @@ export type DateProps = Omit<
    */
   showResetButton?: DatePickerProps['showResetButton']
   onBlurValidator?: DateValidator | false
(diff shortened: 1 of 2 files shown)


langz (approved):
Looks good, just need to run prettier 🖌️

langz:
But now since it's "automatic" there's potentially new props that will be supported which was not supported before, and perhaps is not documented for the Field.Date docs/properties as well or? 🤔 
Will we need to make changes to packages/dnb-eufemia/src/extensions/forms/Field/Date/DateDocs.ts based on the changes in this PR?

tujoworker:
The change is that we derive it form the array with pros. So its not really automatic. We still define it there as well.

PR #8209, opened 2026-05-22
chore(Starter): remove `transformWithOxc` from vite config

tujoworker:
With Vite v8 we don't need `transformWithOxc` anymore it seems. Works well at least when testing.

Also, replace `path.resolve(__dirname, ...)` with `new URL(..., import.meta.url).pathname` in the eufemia-starter vite config, removing the `path` import.


langz (approved):
🚀

langz:
I tried the following stackblitz link from this example:
https://chore-vite-config-import-met.eufemia-e25.pages.dev/uilib/extensions/forms/blocks/ChildrenWithAge/demos#in-wizard

It does not work as it no longer imports `<Blocks.ChildrenWithAge />` from `import { ChildrenWithAge } from '@dnb/eufemia/extensions/forms/blocks'`

Not exactly sure which of the latest PRs that removed this functionality

PR #8211, opened 2026-05-22
feat(Forms): enable `allowOverflow` for fields where you can type more than the limit

joakbjerk:


tujoworker (commented):
Would be good to split this up in individual PR with one test at least for every case.

joakbjerk:
> Would be good to split this up in individual PR with one test at least for every case.

Okay, can do that 😄

PR #8220, opened 2026-05-22
feat(Table): add support for `border="horizontal"` for horizontal borders only

tujoworker:
Motivation: https://dnb-it.slack.com/archives/C081G91GCT0/p1779365998336979
Preview: https://feat-table-border-horizontal.eufemia-e25.pages.dev/uilib/components/table#multiple-table-body-sections


packages/dnb-eufemia/src/components/table/TableDocs.ts
```diff
@@ -14,8 +14,8 @@ export const TableProperties: PropertiesTableProps = {
     status: 'optional',
   },
   border: {
-    doc: 'Use `true` to show borders between table data cells.',
-    type: 'boolean',
+    doc: 'Use `true` to show borders between table data cells. Use `"horizontal"` to show only horizontal borders between rows.',
```
langz, line 17: When should I use `"inline"`, it's not clear to me when reading this doc.

tujoworker: Yeah, maybe we should call it horizontal.

langz:
I don't fully understand why one would use border="inline" for horizontal borders only, as the pr title says.
If I would want horizontal borders, wouldn't it make sense to use `border="horizontal"`?

langz:
Also border="inline" and border="horizontal" as of now produce the same CSS if I'm not mistaken? 🤔

langz (approved):
Good good 🚀 
I think we should just pick one horizontal or inline.
I think horizontal makes most sense for me.
Then updated docs, jsdocs, etc.

PR #8224, opened 2026-05-22
feat(Field.NationalIdentityNumber): allow typing beyond the 11-digit mask length

joakbjerk:

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/NationalIdentityNumber/NationalIdentityNumber.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/NationalIdentityNumber/NationalIdentityNumber.tsx
@@ -146,6 +146,7 @@ function NationalIdentityNumber(props: FieldNationalIdentityNumberProps) {
     label: labelProp ?? label,
     errorMessages,
     mask,
+    allowOverflow: true,
     width: width ?? 'medium',
     inputMode: 'numeric',
     onChangeValidator: validate ? onChangeValidator : undefined,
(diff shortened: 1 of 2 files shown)


tujoworker (approved):

langz:
Do we need to add any docs for this?
Or perhaps that's in a seperate PR, adding it for all certain Field.X's?

EDIT: Actually this is not needed I would guess, as it's just the default behavior now. Somehow I thought by reading the pr description, that this PR adds a new property allowOverflow, but it's actually using allowOverflow under the hood.

langz:
I also guess we should do a small check if the validators used in all these components is caring for the maximum/minimum length?
I'm not sure, but potentially it could exist new values now when used with `allowOverflow` that the validators never has been designed to test as it's always had a "max length" restriction in the field.
Worth a consideration, just to look at the validators for all these fields that we now open `allowOverflow` to.

langz:
Good good 💯 

So with `allowOverflow` I'm able to enter `01017501234se` which I don't think we should allow? Or if then we must reconsider if validators work as expected, etc. Further if entering `01017501234se`, then blurring the letters are removed. Which I find is a bit strange user experience.

joakbjerk:
> I also guess we should do a small check if the validators used in all these components is caring for the maximum/minimum length? I'm not sure, but potentially it could exist new values now when used with `allowOverflow` that the validators never has been designed to test as it's always had a "max length" restriction in the field. Worth a consideration, just to look at the validators for all these fields that we now open `allowOverflow` to.

The validator for this component warns the user if the value is to short or too long with the "It has to be 11 numbers dummy!" message.

langz (approved):
Image

PR #8225, opened 2026-05-22
feat(Field.BankAccountNumber): allow typing beyond the 11-digit mask length

joakbjerk:


tujoworker:
Here is the PR preview: https://chore-national-identity-numb.eufemia-e25.pages.dev/uilib/extensions/forms/feature-fields/NationalIdentityNumber

langz:
Good good 💯 

So with `allowOverflow` I'm able to enter `20001234567se` which I don't think we should allow? Or if then we must reconsider if validators work as expected, etc. Further if entering `20001234567se`, then blurring the letters are removed. Which I find is a bit strange user experience.

langz (approved):

joakbjerk:
This field only validates the length of the `norwegianBban`, but not the other types, so that should be fixed before merge, or maybe not? We have a test that says "should not validate non-norwegianBban types"

langz (approved):
Image

PR #8226, opened 2026-05-22
feat(Field.OrganizationNumber): allow typing beyond the 9-digit mask length

joakbjerk:


langz (commented):
Good good 💯 

So with `allowOverflow` I'm able to enter `987654321se` which I don't think we should allow? Or if then we must reconsider if validators work as expected, etc. Further if entering `987654321se`, then blurring the letters are removed. Which I find is a bit strange user experience.

langz (approved):

langz (approved):
Image

PR #8227, opened 2026-05-22
feat(Field.PhoneNumber): allow typing beyond the 8-digit mask length

joakbjerk:

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber/PhoneNumber.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber/PhoneNumber.tsx
@@ -407,33 +407,10 @@ function PhoneNumber(props: FieldPhoneNumberProps = {}) {
         dataObj?.selectedKey?.trim() || emptyValue)
       currentCountryRef.current = dataObj?.country
 
-      // If the phone number is more than 8 digits, and the country code is the default one (+47),
-      // we truncate the phone number to 8 digits.
-      // This is to ensure that the phone number does not exceed the maximum length set by the mask.
-      // TODO: This is a temporary solution, and should be removed once the mask is updated to handle this case.
-      if (
-        !numberMask &&
-        countryCodeRef.current?.includes(defaultCountryCode) &&
-        numberRef.current?.length > 8
-      ) {
-        const truncatedNumber = numberRef.current.substring(0, 8)
-        callOnChange({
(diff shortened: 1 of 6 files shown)


langz:
Good good 💯 

So with `allowOverflow` I'm able to enter `98765432se` which I don't think we should allow? Or if then we must reconsider if validators work as expected, etc. Further if entering `98765432se`, then blurring the letters are removed. Which I find is a bit strange user experience.



https://github.com/user-attachments/assets/0844529a-e18a-422a-8ecd-765244b05180

joakbjerk:
Should be fixed now @langz

langz:
In the other Field.X where we've allowed typing beyond their mask length we've had validators which will prevent submitting "non valid" values, which sort of make sure that we remain the existing functionality/validation that we have as of today.

_Norwegian phone numbers longer than 8 digits were previously impossible to enter and submit. After this PR, they can be entered and submitted without a validation error, since Field.PhoneNumber has no built-in max-length validator for Norwegian numbers._

Not exactly sure what we want to do, implement a validator?

joakbjerk:
> In the other Field.X where we've allowed typing beyond their mask length we've had validators which will prevent submitting "non valid" values, which sort of make sure that we remain the existing functionality/validation that we have as of today.
> 
> _Norwegian phone numbers longer than 8 digits were previously impossible to enter and submit. After this PR, they can be entered and submitted without a validation error, since Field.PhoneNumber has no built-in max-length validator for Norwegian numbers._
> 
> Not exactly sure what we want to do, implement a validator?

Looks like that might be our best bet?

langz:
> > In the other Field.X where we've allowed typing beyond their mask length we've had validators which will prevent submitting "non valid" values, which sort of make sure that we remain the existing functionality/validation that we have as of today.
> > _Norwegian phone numbers longer than 8 digits were previously impossible to enter and submit. After this PR, they can be entered and submitted without a validation error, since Field.PhoneNumber has no built-in max-length validator for Norwegian numbers._
> > Not exactly sure what we want to do, implement a validator?
> 
> Looks like that might be our best bet?

I've added a few commits now, where I've attempted adding an "internal validator" that can be used to maintain the same functionality as we had "before". Let's look at it together, and see if it handles all scenarios.

langz (approved):

langz (approved):
Image

tujoworker (approved):

PR #8229, opened 2026-05-22
feat(Field.PostalCodeAndCIty): allow typing beyond the 4-digit postal code mask length

joakbjerk:


langz:
Good good 💯 

Now when using `allowOverflow`, for the postalcode field after filling out the "mask" I'm suddenly allowed to enter regular letters, etc. 
So with `allowOverflow` I'm able to enter `1234se` which I don't think we should allow? Or if then we must reconsider if validators work as expected, etc. Further if entering `1234se`, then blurring the letters are removed. Which I find is a bit strange user experience.

langz (approved):

langz (approved):
Image

PR #8232, opened 2026-05-22
fix(DatePicker): improve keyboard navigation in calendar

tujoworker:
Preview: https://fix-date-picker-keyboard-nav.eufemia-e25.pages.dev/uilib/components/date-picker/demos

## Summary

Improves keyboard navigation in the DatePicker calendar by separating visual focus movement from date selection.

### DatePicker
- Focus the currently selected date button when opening the calendar (instead of the table element)

### DatePickerCalendar
- Arrow keys now move visual focus between day buttons without changing the selected date
- Enter/Space confirms the focused date as the selection
- In range mode, Enter/Space sets start or end date based on selection state (not calendar position), and keeps focus on the selected date
- Selecting in either calendar as the first pick sets the start date; the second pick completes the range
- Month navigation via arrow keys triggers a re-render and restores focus via `useLayoutEffect`
- Added `data-date` attributes to day cells for focus targeting
- Skip hover effect on already-selected dates

### Popover
- Prevent `setTimeout` re-focus from stealing focus when it has already moved to another element inside the popover
- Ignore keyboard events targeting `document.body` during React re-renders to avoid false popover close

### Styles
- Added `focus-visible` styles for keyboard-navigated day buttons with hover visual + focus outline

### DIFF
--- a/packages/dnb-eufemia/src/components/date-picker/DatePickerCalendar.tsx
+++ b/packages/dnb-eufemia/src/components/date-picker/DatePickerCalendar.tsx
@@ -25,7 +25,6 @@ import {
   startOfDay,
   differenceInCalendarDays,
   differenceInMonths,
-  lastDayOfMonth,
   setDate,
 } from 'date-fns'
 
@@ -42,6 +41,7 @@ import Button from '../button/Button'
 import type { DatePickerContextValue } from './DatePickerContext'
 import DatePickerContext from './DatePickerContext'
 import type { InternalLocale } from '../../shared/Context'
+import useIsomorphicLayoutEffect from '../../shared/helpers/useIsomorphicLayoutEffect'
 import type { DatePickerChangeEvent } from './DatePickerProvider'
 import type { DatePickerDates } from './hooks/useDates'
 import type { CalendarNavButtonProps } from './DatePickerCalendarNavigator'
@@ -146,6 +146,8 @@ function DatePickerCalendar(restOfProps: DatePickerCalendarProps) {
     hoverDate,
     setHoverDate,
     setSubmittedDates,
+    views,
+    setViews,
     props: { onDaysRender, yearNavigation },
     translation: {
       DatePicker: { firstDay: defaultFirstDayOfWeek, selectedMonth },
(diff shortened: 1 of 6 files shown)


langz (commented):
Great great, I think it works great :)
I think it sort of solves the problem we discussed at work before the weekend 🚀

langz:
I have some thoughts:
Why does the next/right calendar get focused after I select a date in the first/left calendar?
How do we know that the user will want to make a selection there? What if I would want to select an end date in the first calendar?
Isn't it safest to just keep the focus on the same, now selected, date? As we don't know which date the user might select next.


https://github.com/user-attachments/assets/ab779297-d50a-4481-8812-265f71926f75

langz:
If I select my startdate (first date I select) in the second/right calendar, then it closes immediately, and the start day which I selected is actually marked as my end date 🤔 


https://github.com/user-attachments/assets/4de94188-22d0-49e4-abd1-b7683ca51d5c




https://fix-date-picker-keyboard-nav.eufemia-e25.pages.dev/uilib/components/date-picker/demos/#linked-datepickers

tujoworker:
Now it works much more like if using mouse.

langz:
https://fix-date-picker-keyboard-nav.eufemia-e25.pages.dev/uilib/components/date-picker/demos/#linked-datepickers

If I select a start date (my first selected date) in the "second/last calendar" by using space, then press right, I would expect the next date selected to be the date after the first selected.
But in this [example](https://fix-date-picker-keyboard-nav.eufemia-e25.pages.dev/uilib/components/date-picker/demos/#linked-datepickers) when I select 16.06, then press right, then 30.06 is selected:


https://github.com/user-attachments/assets/13b63db2-edbd-4f9e-8fed-7b29a89999e9



Also, if I initially press the down button in both calendars, I get a different behavior from both.
The first calendar selects the first value, as I would expect, it selects 1.
But the second calendar selects the last value, 30.



https://github.com/user-attachments/assets/c163d387-a721-4864-a0f0-1ec4e9d615d0

packages/dnb-eufemia/src/components/date-picker/DatePickerCalendar.tsx
```diff
@@ -197,6 +208,44 @@ function DatePickerCalendar(restOfProps: DatePickerCalendarProps) {
...
+    if (endButton) {
+      endButton.focus({ preventScroll: true })
+    } else {
+      ;(endTable as HTMLElement | undefined)?.focus({
+        preventScroll: true,
+      })
+    }
+  }, [])
+
+  useLayoutEffect(() => {
```
langz: `useLayoutEffect` or `useIsomorphicLayoutEffect` 🤔

langz (approved):

packages/dnb-eufemia/src/components/date-picker/DatePickerCalendar.tsx
tujoworker: Fixed!

tujoworker:
The wired second calendar behavior is now solved.

PR #8241, opened 2026-05-23
style(Button): replace `tertiary` variant pseudo underline with text-decoration

tujoworker:
Its pure legacy that we still used pseudo elements.

Screenshot 2026-05-23 at 14 21 04
Screenshot 2026-05-23 at 14 24 02


langz (approved):
Visually, it looks much better :)

PR #8250, opened 2026-05-24
fix(Accordion): only focus content when opening tertiary variant

tujoworker:
The tertiary accordion's `shouldFocusContent` was being set on both open and close. This meant closing via keyboard would incorrectly attempt to focus the content area.

**Fix:** Gate `shouldFocusContent` on `next` being `true`, so focus is only moved when opening.

**Test:** Added a test that verifies closing via keyboard does not focus the content.


langz (approved):
🍗

PR #8260, opened 2026-05-26
fix(InputMasked, Field.Number): prevent input beyond MAX_SAFE_INTEGER

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1779453021233979?thread_ts=1779175833.920299&cid=CMXABCHEY

Personally, I like this fix, as we previously clamped to `MAX_SAFE_INTEGER`, which is strictly worse UX than rejecting a new digit.

Deploy preview: https://fix-prevent-input-beyond-max.eufemia-e25.pages.dev/uilib/extensions/forms/feature-fields/Currency/demos#empty

Deploy preview from main for existing functionality:
https://main.eufemia-e25.pages.dev/uilib/extensions/forms/feature-fields/Currency/demos#empty


joakbjerk (approved):
🔥

packages/dnb-eufemia/src/extensions/forms/Field/Number/Number.tsx
```diff
@@ -414,20 +428,38 @@ function NumberComponent(props: FieldNumberProps) {
...
+      setPasteWarning(undefined)
+    }
+
+    handleBlur()
+  }, [handleBlur, pasteWarning])
+
+  const onPasteHandler = useCallback(
+    (e: ClipboardEvent<HTMLInputElement>) => {
+      const pasted = e.clipboardData?.getData('text/plain') ?? ''
+      const raw = pasted.replace(/[\s\u00A0,]/g, '')
```
tujoworker: A comment why we do that would be nice.

tujoworker (changes requested):
That was a lot of code added just for this edge case. This code runs on every re-render and keystroke. I think we should rethink it.

langz:
> That was a lot of code added just for this edge case. This code runs on every re-render and keystroke. I think we should rethink it.

Sure, that's fine :)
Let's discuss it once more on Thursday or so 🙏

tujoworker:
Is it on purpose that the user can't paste anything, when the number is too long?

langz:
> Is it on purpose that the user can't paste anything, when the number is too long?

Yes, the intention was to not display any value pasted in as it's bigger than what we support, and display a warning message.

I see the code and deploy preview is changed now.

tujoworker:
Yes, added some commits to get there. The overall goal is to have a performant and as little extra code as possible. Works as expected for me now.

tujoworker (approved):

langz:
> Yes, added some commits to get there. The overall goal is to have a performant and as little extra code as possible. Works as expected for me now.

Great, thanks :) 
Looks and works great now 🚀

langz (commented):
✅

PR #8262, opened 2026-05-26
chore(Icon): remove file extension icons

joakbjerk:
Motivation from task list:
Remove file extensions icons from icons library 

Unsure about what decorator to use here 😄


tujoworker:
That is technically a breaking change, so I think we need to close this PR and rather move the task into v12.

joakbjerk:
> That is technically a breaking change, so I think we need to close this PR and rather move the task into v12.

I'll move the task from Gjøremål to v12 then 🫡

PR #8265, opened 2026-05-26
chore(Portal): keep default theme CSS active until non-default theme loads

langz:
The prerender early-inline script previously disabled the default theme's CSS link at the same time as enabling the non-default theme's link. Since Chrome doesn't pre-fetch disabled links, this caused a window where no component CSS was active, producing completely unstyled cards on the homepage for non-default themes (sbanken, eiendom, carnegie).

The script now keeps the default theme CSS enabled as a fallback and only disables it after the new theme's onload event fires. This works because the PostCSS scoping gives non-default themes higher CSS specificity, so the correct styles automatically win once loaded.

### DIFF
--- a/packages/dnb-design-system-portal/vite/__tests__/prod/prerender-utils.test.ts
+++ b/packages/dnb-design-system-portal/vite/__tests__/prod/prerender-utils.test.ts
@@ -514,6 +514,68 @@ describe('prerender-utils', () => {
       expect(cssPos).toBeLessThan(headClosePos)
     })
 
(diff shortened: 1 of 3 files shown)


tujoworker:
Since #8267 is a superset (same prod fix + dev mode fix + cleaner script placement), merging only #8267 and closing #8265 would be the clean path. #8265 becomes redundant.

langz:
> Since #8267 is a superset (same prod fix + dev mode fix + cleaner script placement), merging only #8267 and closing #8265 would be the clean path. #8265 becomes redundant.

🦀

PR #8266, opened 2026-05-26
chore(Portal): use :where() wrapping in PostCSS theme scope plugin

tujoworker:
Wrap the brand class in `:where()` for the catch-all pattern in the PostCSS theme scope plugin, so non-default theme component styles don't increase specificity and override portal-specific styles (e.g. sidebar menu anchor colors, icon sizes).

**Changes:**
- Use `:where(.eufemia-theme__<brand>)` instead of `.eufemia-theme__<brand>` in the catch-all selector rewrite — this contributes zero specificity
- Skip the generic `.eufemia-theme` class (without `__` suffix) that lives on the same DOM element as the brand class
- Handle `:where()`-wrapped component selectors from anchor mixins (e.g. `:where(:not(.dnb-anchor--no-style)).dnb-anchor`)
- Updated and added tests for all new patterns

Caused by PR #8251

image (13)

### DIFF
--- a/packages/dnb-design-system-portal/postcss-eufemia-theme-scope.cjs
+++ b/packages/dnb-design-system-portal/postcss-eufemia-theme-scope.cjs
@@ -16,7 +16,10 @@
  * And component selectors like:
  *   .eufemia-scope--portal .dnb-accordion
  * to:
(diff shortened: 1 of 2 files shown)


langz (approved):
Weee 🚀 
Why must it work different in dev vs. in portal(prod)?
Could it work in the same way?

PR #8267, opened 2026-05-26
chore(Portal): add brand class to body in dev mode theme switching

tujoworker:
In dev mode, `applyThemeStyles` disabled non-active theme `<style>` elements but did not add the brand class (e.g. `eufemia-theme__sbanken`) to `<body>`. Since theme tokens are scoped under the brand class by the PostCSS plugin, CSS custom properties like `--token-color-background-neutral` did not resolve at the body level — causing a white background in dark mode with non-default themes.

Build mode already handled this via `activateTheme()`. This adds the same body class management to the dev mode `applyThemeStyles` function.

**Changes:**
- Add brand class to `<body>` in dev mode's `applyThemeStyles`, replacing any previous brand class while preserving color-scheme classes
- 3 new tests: adding brand class, replacing when switching, preserving color-scheme

### DIFF
--- a/packages/dnb-design-system-portal/vite/__tests__/plugins/eufemia-theme.test.ts
+++ b/packages/dnb-design-system-portal/vite/__tests__/plugins/eufemia-theme.test.ts
@@ -235,6 +235,100 @@ describe('eufemia-theme plugin', () => {
     })
   })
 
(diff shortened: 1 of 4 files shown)


langz (approved):
Weee 🚀 
Why must it work different in dev vs. in portal(prod)?
Could it work in the same way?

PR #8268, opened 2026-05-26
fix: prevent xmlns specificity boost from overriding text-align modifiers

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1779786660617699

The resetLegacyStyles mixin wraps .dnb-p rules under html[xmlns='...'], giving them specificity (0,2,1). This overrides the alignment modifier classes like .dnb-t__align--center (0,1,0) on e-Platform where <html> has the xmlns attribute.

Split the text-align reset into a separate rule with :not() exclusions so it yields when alignment modifiers are present.


joakbjerk (approved):

tujoworker (changes requested):
I don’t think these tests give us any additional help or safety.

They mostly extract everything and check whether a certain string exists. To me, that feels more like testing implementation details than actual behavior.

A comment above the code explaining the “why” would probably give more value than these tests.

langz:
> I don’t think these tests give us any additional help or safety.
> 
> They mostly extract everything and check whether a certain string exists. To me, that feels more like testing implementation details than actual behavior.
> 
> A comment above the code explaining the “why” would probably give more value than these tests.

It's removed now 🤞

tujoworker (approved):

packages/dnb-eufemia/src/style/core/reset.scss
```diff
@@ -394,9 +394,14 @@
     // reset paragraphs
     .dnb-p {
       overflow: initial;
-      text-align: inherit;
       background-color: initial;
     }
+    // Separate text-align reset so alignment modifiers (.dnb-t__align--center
+    // etc.) are not overridden by the xmlns specificity boost (0,2,1) that
+    // the resetLegacyStyles mixin adds via html[xmlns='...'].
```
tujoworker, line 401: 🙏

PR #8269, opened 2026-05-26
fix: use CSS variable chain for primary button color in xmlns reset

langz:
Motivation: Found in the light of https://github.com/dnbexperience/eufemia/pull/8268

The legacy reset hardcoded color: var(--color-white) on a.dnb-button--primary under html[xmlns], bypassing the component's --button-color-text variable chain. This caused disabled, focus, and themed states to be ignored for <a> primary buttons on e-Platform.

Use the same var(--button-color-text) reference the component uses, so state overrides (hover, focus, disabled) and theme customizations are respected while still winning the specificity battle against legacy anchor styles.


joakbjerk (approved):

tujoworker (approved):
🌚

PR #8272, opened 2026-05-26
chore: replace dead ::after selectors with text-decoration equivalents

langz:
The tertiary button underline moved from ::after pseudo-element to text-decoration, but Table sort buttons, Breadcrumb, and Sbanken theme still had ::after selectors targeting the now-removed pseudo-element.

Replace with text-decoration-color / text-decoration equivalents to match the current tertiary button implementation.


tujoworker (approved):
Nice findings 😍

PR #8273, opened 2026-05-26
chore(Card.Action): warn about nested interactive elements

langz:


packages/dnb-eufemia/src/components/card/CardAction.tsx
```diff
@@ -60,6 +62,25 @@ function CardAction(props: CardActionProps) {
...
+  const internalRef = useRef<HTMLElement>(null)
+  const combinedRef = useCombinedRef(ref, internalRef)
+
+  useEffect(() => {
+    const el = internalRef.current
+    if (!el) {
+      return // stop here
+    }
+
+    const nested = el.querySelector(
```
tujoworker, line 78: I think we should check if it runs in dev/test environment before we run the query.

langz: Smart, I've added this now :)

https://github.com/dnbexperience/eufemia/pull/8273/changes/890e6e490182efe61fd9b81c2616fdf8bb3eda69

tujoworker (approved):

PR #8279, opened 2026-05-27
feat(DatePicker): add `rangeSingleCalendar` prop

tujoworker:
Preview: https://feat-date-picker-range-dual.eufemia-e25.pages.dev/uilib/components/date-picker/demos#range-datepicker

### DIFF
--- a/packages/dnb-eufemia/src/components/date-picker/DatePicker.tsx
+++ b/packages/dnb-eufemia/src/components/date-picker/DatePicker.tsx
@@ -204,6 +204,10 @@ export type DatePickerProps = {
    */
   range?: boolean
   /**
+   * If set to `false`, only one calendar is shown in range mode instead of two side-by-side calendars. Defaults to `true`.
+   */
+  rangeDualCalendar?: boolean
+  /**
    * Link both calendars, once to the user is navigating between months. Only meant to use if the range is set to `true`. Defaults to `false`.
    */
   link?: boolean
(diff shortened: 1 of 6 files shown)


packages/dnb-eufemia/src/components/date-picker/DatePicker.tsx
```diff
@@ -203,6 +203,10 @@ export type DatePickerProps = {
    * If the date picker should support a range of two dates (starting and ending date). Defaults to `false`.
    */
   range?: boolean
+  /**
+   * If set to `false`, only one calendar is shown in range mode instead of two side-by-side calendars. Defaults to `true`.
```
langz: I feel like it would be more common to describe what happens if one set the prop to be true. But I see why it's written like it is. Nothing to improve I think.

packages/dnb-eufemia/src/components/date-picker/DatePicker.tsx
```diff
@@ -203,6 +203,10 @@ export type DatePickerProps = {
...
+   */
+  rangeDualCalendar?: boolean
   /**
    * Link both calendars, once to the user is navigating between months. Only meant to use if the range is set to `true`. Defaults to `false`.
```
langz, line 211: Could perhaps mention in the docs that this only makes sense when `rangeDualCalendar` is `true`.

langz (approved):
I find this hard to name, and am not able to come up with much better alternatives, so please go ahead with what you suggest or @snorrekim suggestion previously (rangeSideBySide or such).


I personally, think `rangeSingleCalendar` reads a bit better:
`<DatePicker range rangeSingleCalendar startDate="2024-10-01" endDate="2024-10-15" />` 

compared to:

`<DatePicker range rangeDualCalendar={false} startDate="2024-10-01" endDate="2024-10-15" />`

PR #8280, opened 2026-05-27
fix(DateFormat): preserve leading zeros when `hideYear`/`hideCurrentYear` used with `dateStyle="short"`

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1779859766335859

Replace manual Intl.DateTimeFormatOptions mapping with formatToParts-based approach that strips the year from the original dateStyle output. This preserves locale-specific formatting (e.g. leading zeros for short style) instead of losing them by using { month: 'numeric', day: 'numeric' }.

Before: <DateFormat dateStyle="short" hideYear>2026-08-01</DateFormat> → 1.8
After:  <DateFormat dateStyle="short" hideYear>2026-08-01</DateFormat> → 01.08


tujoworker (approved):
Nice 🫨

PR #8303, opened 2026-05-27
chore(CI): prevent Playwright install hanging

langz:
Motivation: https://github.com/dnbexperience/eufemia/actions/runs/26507991157/job/78065316551

Split 'playwright install --with-deps' into separate 'playwright install'
and 'playwright install-deps' steps so the browser download no longer
blocks on system dependency installation.

- Remove --with-deps from the browser download step
- Run install-deps unconditionally (system deps are not cached)
- Keep HOMEBREW_NO_AUTO_UPDATE=1 on install-deps for macOS runners

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -76,11 +76,10 @@ jobs:
             %USERPROFILE%\AppData\Local\ms-playwright
           key: ${{ secrets.CACHE_VERSION }}-${{ runner.os }}-playwright-${{ hashFiles('**/yarn.lock') }}
       - name: Install Playwright browsers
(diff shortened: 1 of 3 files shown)


tujoworker:
They still hang in there. This PR #8304 seems to solve the problem. Not sure why.

langz:
> They still hang in there. This PR #8304 seems to solve the problem. Not sure why.

Yeah, I saw that, trying a new approach now :)
But if that fixes it, then it's all good 📽️

tujoworker:
E2E did run green. Visual snapshots will take some more time. The card component gets a wired result. On it.

PR #8308, opened 2026-05-28
fix(Dialog): use role="dialog" on all platforms to prevent VoiceOver double titles

### LINKED ISSUE #5919: Dialog's title is presented twice to user when using VoiceOver
https://eufemia.dnb.no/uilib/components/dialog/demos/

Image

langz:
On macOS/Safari/iOS, ModalContent downgraded role="dialog" to role="region" and omitted aria-modal as a workaround for older VoiceOver bugs. This caused VoiceOver to announce the title twice: once as the region's accessible name (via aria-labelledby) when entering the landmark, and again when focus moved to the h1 heading.

Modern VoiceOver handles role="dialog" with aria-modal correctly. Remove the platform-specific workaround so all platforms use the standard dialog role.

Deploy preview: https://fix-dialog-voiceover-double.eufemia-e25.pages.dev/uilib/components/dialog/demos

Closes #5919

From:
Screenshot 2026-05-29 at 12 52 24

To:

Screenshot 2026-05-29 at 12 53 24

### DIFF
--- a/packages/dnb-eufemia/src/components/dialog/__tests__/Dialog.test.tsx
+++ b/packages/dnb-eufemia/src/components/dialog/__tests__/Dialog.test.tsx
@@ -296,7 +296,7 @@ describe('Dialog', () => {
     })
   })
 
-  it('respects focusSelector over close button', async () => {
+  it('respects focusSelector over dialog container and close button', async () => {
     render(
       <Dialog noAnimation open title="Title" focusSelector="#focus-me">
         <Dialog.Body>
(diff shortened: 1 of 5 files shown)


joakbjerk (approved):
oꡮ

tujoworker:
I tested it with VoiceOver – especially the nested dialog situation, because the reason we did not set role="dialog" on Safari, was that Safari did not support nested role="dialog" – but now it seems it works as expected. 🙏

tujoworker (approved):

tujoworker:
You may have a look at my extra commits.

langz:
> I tested it with VoiceOver – especially the nested dialog situation, because the reason we did not set role="dialog" on Safari, was that Safari did not support nested role="dialog" – but now it seems it works as expected. 🙏

Great, thanks for taking the time to test this ❤️

langz:
> You may have a look at my extra commits.

Took a look through it, and it seems great to me 🥇 
Thanks 🦅

PR #8314, opened 2026-05-28
chore(TermDefinition): remove 'new' status

langz:
TermDefinition was added in November 2025 and could be considered to not be new anymore.


tujoworker:
Lets keep it for a year. Its an interesting component many don't know about could exist or know the term if they would like to use something like that.

PR #8341, opened 2026-05-29
fix: use plain objects instead of module namespace for component sub-exports

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1779450871937159

I guess this will affect tree-shaking, so might not be that this is a desired approach 🤔 

_Replace `import * as _default` (Module Namespace exotic object) with
explicit imports and plain object exports in all namespace-style
components (NumberFormat, Flex, Grid, Menu, List, Stat)._

_The namespace form can break when re-exported through barrel files and
consumed via named imports (`import { NumberFormat } from '@dnb/eufemia'`)
in bundlers that perform CJS/ESM interop or dependency pre-bundling,
causing sub-components like NumberFormat.BankAccountNumber to be undefined
at runtime._

Perhaps this could also be considered to be fixed in the app's bundler. Not exactly sure how.


tujoworker:
I don’t think this approach is the right way to go. It would make all of the components non-tree-shakeable.

The whole Eufemia Forms package is built around that pattern, so if this was the actual issue, I would expect it to have been reported many times before.

I still think there may be a different issue in this case, and that we should look closer at what is actually being imported, bundled, or executed before changing the export structure.

PR #8354, opened 2026-06-01
chore(NumberFormat.BankAccountNumber): remove unsupported types `decimals`, `rounding` & `signDisplay`

langz:
These props have no effect on BankAccountNumber since the formatter uses pure string manipulation, not Intl.NumberFormat. Omitting them from the type prevents confusion.


tujoworker (approved):

tujoworker:
I don’t think any developers have used it, since it doesn’t do anything. So I don’t think we need to list it in the release log.

PR #8355, opened 2026-06-01
fix(Textarea): use consistent disabled background token with input 

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1780305095629889

From:
Screenshot 2026-06-01 at 12 23 47

To:
Screenshot 2026-06-01 at 12 24 13


langz:
I'm a bit unsure about this, or which approach that's correct.
Think I'll have to check with UX, as the Figma designs show the darker one:

image

Based on the Figma designs, I guess I should do the opposite, to change the input's disabled token?

langz:
> I'm a bit unsure about this, or which approach that's correct. Think I'll have to check with UX, as the Figma designs show the darker one:
> 
> image
> Based on the Figma designs, I guess I should do the opposite, to change the input's disabled token?

On a second thought, I do think the change here in this current PR is correct, and that we should rather update the designs in Figma. Reason for me thinking this is correct, is because textarea's disabled styling in Figma is different than compared to others(it's darker). Here's images from Figma where TextArea is the outlier:


Screenshot 2026-06-01 at 14 40 12
Screenshot 2026-06-01 at 14 39 32
Screenshot 2026-06-01 at 14 37 01
Screenshot 2026-06-01 at 14 36 47

tujoworker (approved):

PR #8357, opened 2026-06-01
fix(Breadcrumb): set background color for non collapsable breadcrumbs to transparent and correct paddings and indents

joakbjerk:
## TODO BEFORE MERGE!

- [x] Fix padding for collapsed header
- [x] Fix background color for collapsed body
- [x] Fix collapse body content indent
- [x] Double check for other design discrepancies


langz:
Do we need to set background as transparent by using both "props"(variant="transparent") and in SCSS? Or would one strategy suffice?

tujoworker (approved):

packages/dnb-eufemia/src/components/breadcrumb/BreadcrumbMultiple.tsx
```diff
@@ -27,7 +27,11 @@ export const BreadcrumbMultiple = ({
       animate={!noAnimation}
       className="dnb-breadcrumb__multiple"
     >
-      <Section className="dnb-breadcrumb__list" element="ol">
+      <Section
+        className="dnb-breadcrumb__list"
+        variant="transparent"
+        element="ol"
+      >
```
tujoworker: Do we use the section actually? can it be a div?

joakbjerk: I'll double check 🫡

joakbjerk:
> Do we need to set background as transparent by using both "props"(variant="transparent") and in SCSS? Or would one strategy suffice?

If we remove the section as @tujoworker suggests, then we would only need to do it in scss

PR #8358, opened 2026-06-01
fix(DatePicker): remove calendar padding in inline mode to avoid unnecessary whitespace

tujoworker:
The inline DatePicker calendar inherits the same 1rem padding as the popover variant, which adds unwanted whitespace when the calendar is rendered directly in the page layout. This removes that padding so inline calendars sit flush with their surroundings.


langz (approved):
Should the gap be a spacing token?

tujoworker:
No, we should avoid them inside components

PR #8360, opened 2026-06-01
design: update color design tokens

snorrekim:
- focus effect background is now lighter for correct color contrast
- remove `--token-color-text-destructive` (use `--token-color-text-error` instead)
- rename `--token-color-text-destructive-inverse`  to `--token-color-text-error-inverse`
- remove `--token-color-icon-destructive` (use `--token-color-icon-error` instead)
- remove `--token-color-component-table-background-neutral-alternative` (unused)


snorrekim:
This is just an update of the existing tokens. New spacing tokens will be in a separate PR.

snorrekim:
I don't think we can remove tokens like `--color-icon-error-subtle` unless we remove them from the source .json file.

snorrekim:
Looking at the other tokens, `--color-icon-error-subtle` should actually be renamed `--color-icon-destructive-subtle`. I'll talk with UX.

langz (approved):
Looks good to me :)
I've quickly looked through the updated screenshot tests, and I think it looks good.

There's of course a few "breaking changes" in removal of tokens, but since that's marked as beta, we are all good 🤞

tujoworker (approved):

PR #8365, opened 2026-06-02
refactor: rename `status` to `statusMessage` and repurpose `status` to replace `statusState`

tujoworker:
## Motivation

https://dnb-it.slack.com/archives/CMXABCHEY/p1779868765841559?thread_ts=1779784812.244889&cid=CMXABCHEY

## Why

The existing `status` / `statusState` API is confusing: `status` holds the *message text*, while `statusState` holds the *actual status*. This is counterintuitive — most developers expect `status` to describe the state, not the message.

This PR fixes the naming so each prop means what its name says:

| Before | After | Purpose |
|---|---|---|
| `status` | `statusMessage` | The status message text |
| `statusState` | `status` | The visual status state (`error`, `warning`, etc.) |

The rename also enables setting a visual status (e.g. error border) *without* requiring a message — something the old API couldn't express cleanly.

## Changes

- **`statusMessage` prop** – replaces the old `status` prop for providing status message text.
- **`status` prop repurposed** – now accepts a status state value (`"error"`, `"warning"`, `"information"`, `"success"`, `"marketing"`).
- **`statusState` deprecated** – falls back to `statusState` when the new `status` is not set. Runtime deprecation warning in dev mode.
- **`aria-invalid`** – automatically set when `status="error"`.
- **v12 migration entry** – documents the rename in v12-info.mdx.

## Affected components

Autocomplete, Button, Checkbox, DatePicker, Dropdown, Input, Radio, RadioGroup, SegmentedField, Slider, StepIndicator, Switch, Textarea, ToggleButton, ToggleButtonGroup.

## Migration

```diff
-<Input status="Error message" statusState="error" />
+<Input statusMessage="Error message" status="error" />
```

## Preview:

Button: https://feat-intent-prop.eufemia-e25.pages.dev/uilib/components/button/demos
Input: https://feat-intent-prop.eufemia-e25.pages.dev/uilib/components/input/demos

```tsx
<Input label="Label" status="warning" statusMessage="Message" />
```

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/about-the-lib/releases/eufemia/v12-info.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/about-the-lib/releases/eufemia/v12-info.mdx
@@ -13,6 +13,7 @@ draft: true
   - [CSS custom properties](#css-custom-properties)
   - [Component changes](#component-changes)
     - [Accordion](#accordion)
+    - [statusState → intent](#statusstate--intent)
     - [Modal, Dialog and Drawer](#modal-dialog-and-drawer)
     - [InfoCard](#infocard)
     - [Popover](#popover)
@@ -66,6 +67,23 @@ When removing these properties, also update the documentation in `classes.mdx` (
 ### [Accordion](/uilib/components/accordion/)
 
 - Replace `id` with `connectedTo` on `Accordion.Content` when connecting to a standalone tertiary button.
+
+### statusState → intent
+
+The `statusState` prop has been deprecated in favor of `intent`. Replace `statusState` with `intent` on all form components that support it.
+
+```diff
+-<Input status="Error message" statusState="error" />
++<Input status="Error message" intent="error" />
+```
+
+The `intent` prop also supports setting visual intent without a status message. When used this way, consider adding `aria-label` or `aria-labelledby` to convey the intent to assistive technologies:
+
+```jsx
+<Input intent="error" aria-label="Error in field" />
+```
+
+Affected components: Autocomplete, Button, Checkbox, DatePicker, Dropdown, Input, Radio, RadioGroup, SegmentedField, Slider, StepIndicator, Switch, Textarea, ToggleButton, ToggleButtonGroup.
 
 ### [Modal](/uilib/components/modal/), [Dialog](/uilib/components/dialog/) and [Drawer](/uilib/components/drawer/)
 
--- a/packages/dnb-design-system-portal/src/docs/uilib/about-the-lib/releases/eufemia/v12-info.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/about-the-lib/releases/eufemia/v12-info.mdx
@@ -13,7 +13,7 @@ draft: true
   - [CSS custom properties](#css-custom-properties)
   - [Component changes](#component-changes)
     - [Accordion](#accordion)
-    - [statusState → intent](#statusstate--intent)
+    - [statusState → status, status → statusMessage](#statusstate--status-status--statusmessage)
     - [Modal, Dialog and Drawer](#modal-dialog-and-drawer)
     - [InfoCard](#infocard)
     - [Popover](#popover)
@@ -68,19 +68,19 @@ When removing these properties, also update the documentation in `classes.mdx` (
 
 - Replace `id` with `connectedTo` on `Accordion.Content` when connecting to a standalone tertiary button.
 
-### statusState → intent
+### statusState → status, status → statusMessage
 
-The `statusState` prop has been deprecated in favor of `intent`. Replace `statusState` with `intent` on all form components that support it.
+The `statusState` prop has been deprecated in favor of `status`. The old `status` prop (which held the message text) has been renamed to `statusMessage`. Replace `statusState` with `status` and `status` with `statusMessage` on all form components that support it.
 
 ```diff
 -<Input status="Error message" statusState="error" />
-+<Input status="Error message" intent="error" />
++<Input statusMessage="Error message" status="error" />
 ```
 
-The `intent` prop also supports setting visual intent without a status message. When used this way, consider adding `aria-label` or `aria-labelledby` to convey the intent to assistive technologies:
+The `status` prop also supports setting visual status without a status message. When used this way, consider adding `aria-label` or `aria-labelledby` to convey the status to assistive technologies:
 
 ```jsx
-<Input intent="error" aria-label="Error in field" />
+<Input status="error" aria-label="Error in field" />
 ```
 
 Affected components: Autocomplete, Button, Checkbox, DatePicker, Dropdown, Input, Radio, RadioGroup, SegmentedField, Slider, StepIndicator, Switch, Textarea, ToggleButton, ToggleButtonGroup.
--- a/packages/dnb-design-system-portal/src/docs/uilib/about-the-lib/releases/eufemia/v12-info.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/about-the-lib/releases/eufemia/v12-info.mdx
@@ -77,6 +77,13 @@ The `statusState` prop has been deprecated in favor of `status`. The old `status
 +<Input statusMessage="Error message" status="error" />
 ```
 
+If you used `status={true}` to only get the status color without a message, replace it with the status state you want to show. Do not move `true` to `statusMessage`.
+
+```diff
+-<Input status={true} />
++<Input status="error" />
+```
+
 The `status` prop also supports setting visual status without a status message. When used this way, consider adding `aria-label` or `aria-labelledby` to convey the status to assistive technologies:
 
 ```jsx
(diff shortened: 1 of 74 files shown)


packages/dnb-eufemia/src/components/toggle-button/ToggleButtonGroup.tsx
```diff
@@ -191,12 +195,23 @@ function ToggleButtonGroup(ownProps: ToggleButtonGroupProps) {
...
+    statusMessage: statusMessageProp,
+  })
+
+  const showStatus = getStatusState(statusMessage)
 
   const rootProps = useSpacing(props, {
     className: clsx(
       'dnb-toggle-button-group',
-      status && `dnb-toggle-button-group__status--${statusState}`,
+      statusMessage &&
```
langz, line 213:
```suggestion
      effectiveStatus &&
```

packages/dnb-design-system-portal/src/docs/uilib/about-the-lib/releases/eufemia/v12-info.mdx
```diff
@@ -12,6 +12,7 @@ draft: true
   - [Install](#install)
   - [Component changes](#component-changes)
     - [Accordion](#accordion)
+    - [statusState → intent](#statusstate--intent)
```
langz:
```suggestion
    - [statusState → status](#statusstate--intent)
```

I don't believe this heading and the link is correct.

langz:
How should we handle TypeScript types for this change?
Won't it be kind of breaking when changing `status` from `string | boolean | ReactNode` as of today, and to '`error' | 'warning' | 'information' | 'success' | 'marketing'`?
App's build will most likely break, and then they'll have to do the "breaking change migration" for it to work as expected. If that will be the case, then it's not really worth it to deprecate 😂

packages/dnb-eufemia/src/components/toggle-button/ToggleButtonGroup.tsx
tujoworker: It works, but `showStatus` would be more consistent — it's already defined a few lines below and used for the `aria-describedby` and `FormStatus` visibility in the same component:

```tsx
const showStatus = getStatusState(statusMessage)
```

Using `showStatus` here aligns with the rest of the component:

```tsx
showStatus &&
  `dnb-toggle-button-group__status--${effectiveStatus}`,
```

In practice they behave the same (since `statusMessage` is always a message string after the refactoring, not a state string), but `showStatus` is the semantic "should we display status UI" flag.

packages/dnb-design-system-portal/src/docs/uilib/about-the-lib/releases/eufemia/v12-info.mdx
tujoworker: Solved.

tujoworker:
The type and backwards compatibility should now be taken care of.

langz:
It does not look like we support `status={true}` any more, is that intentional?
Could that be breaking?

packages/dnb-design-system-portal/src/docs/uilib/components/date-picker/Examples.tsx
```diff
@@ -199,7 +199,7 @@ export const DatePickerErrorMessage = () => (
       date="2019-05-05"
       showInput
       showSubmitButton
-      status={
+      statusMessage={
```
langz, line 202: Will this still now work using `status`, or will one have to refactor from `status` to `statusMessage` for this to still be working as expected?

tujoworker: `status` does still work. ✅

tujoworker:
> It does not look like we support `status={true}` any more, is that intentional? Could that be breaking?

No. We could add support for that. But did we document it as supported like this?

packages/dnb-design-system-portal/src/docs/uilib/components/date-picker/Examples.tsx
langz: Great, thanks. It did not work when I tried to just revert this example from using statusMessage and back status as it initially was. The output/component did not look the same.

langz:
> > It does not look like we support `status={true}` any more, is that intentional? Could that be breaking?
> 
> 
> 
> No. We could add support for that. But did we document it as supported like this?

I can try to verify more precisely, but I do remember that status did support the Boolean type in our docs. 
And also there was multiple usage/reference in our code examples using status={true}

packages/dnb-design-system-portal/src/docs/uilib/components/date-picker/Examples.tsx
tujoworker: Yeah, right, React.Node needs support too. Fixed now.

langz: Great, thanks. Doesn't look like the type accepts a React.Node now.

packages/dnb-eufemia/src/components/input/InputDocs.ts
```diff
@@ -26,15 +26,26 @@ export const InputProperties: PropertiesTableProps = {
     type: 'string',
     status: 'optional',
   },
+  statusMessage: {
+    doc: 'Text with a status message. Use `status` to set the visual style.',
+    type: ['string', 'React.ReactNode'],
+    status: 'optional',
+  },
   status: {
-    doc: 'Text with a status message. The style defaults to an error message. You can use `true` to only get the status color, without a message.',
```
langz, line 30: `true` was documented here at least 🙏

langz:
Looks great 👍 
My only two concerns now is that:
- `status={true}` is silently broken
- `status` prop type is narrower now than the legacy type, so I guess builds depending on ts would break for our consumers

packages/dnb-eufemia/src/components/input/InputDocs.ts
tujoworker: Update the v12 docs about this and added support for backwards compatibility.

tujoworker:
> Looks great 👍 My only two concerns now is that:
> 
>     * `status={true}` is silently broken
> 
>     * `status` prop type is narrower now than the legacy type, so I guess builds depending on ts would break for our consumers

Both should work now.

langz:
I've not looked specifically in to this, but one thought I have is that it would be nice to make it easy to actually understand which code that should be removed in v12. Not sure if it's any possibility of making this more understandable. 

It could be that this is done perfectly as of now, as this was just a thought I had and have not looked into the code yet.

packages/dnb-eufemia/src/components/slider/types.ts
```diff
@@ -134,6 +198,13 @@ export type SliderContextValue = {
   thumbIndex: RefObject<number>
   showStatus: boolean
   showButtons: boolean
+  effectiveStatus:
```
langz, line 201:
```suggestion
  effectiveStatus?:
```

Not sure if this is correct, but I see the other defined effectiveStatus in `StepItemButtonProps` is optional

packages/dnb-eufemia/src/components/toggle-button/ToggleButtonGroup.tsx
langz: Okay, thanks, no problem then.

I just see that all other components gate their classname using `effectiveStatus`, so I thought that pattern applied here as well 🙏

packages/dnb-eufemia/src/components/input/InputDocs.ts
```diff
@@ -26,15 +26,26 @@ export const InputProperties: PropertiesTableProps = {
...
-    type: ['"error"', '"information"', 'boolean'],
+    doc: 'Visual status of the component. When set, applies the corresponding visual styling (e.g. red border for error) even without a status message. When set to `error`, `aria-invalid` is automatically applied. When used without a `statusMessage`, add an `aria-label` or `aria-labelledby` to convey the status to assistive technologies.',
```
langz, line 35: I tried only using status now, and for me it did not work 🤔 

Image
Image

langz:
As we discussed today, we are not entirely sure about the API here.
I guess there's multiple options to consider. As we could make it work similarly like Eufemia Forms by using properties like `warning`, `error`, `marketing`, `information`, etc..

Or we could:

Keep the Forms-layer vocabulary (`error` / `warning` / `info`) so we stay consistent, but nest it under **one** prop:

```jsx
<Input status="This field is required" />          {/* shorthand: error + message */}
<Input status={{ warning: 'Double-check this' }} />
<Input status={{ info: 'We never share this' }} />
<Input status={{ error: true }} />                 {/* state only, no message → aria-invalid */}
```

Why this one:

- **"One message" is enforced by the shape** — it's a single prop with one severity key, so you literally can't set two at once. No precedence rules, no runtime guard to discover.
- **Consistent vocabulary** — same `error`/`warning`/`info` words as the Forms fields, just inside a container that means "pick one."
- **Nothing changes meaning** — `status` becomes a new shape rather than silently reinterpreting old values. The old props can be deprecated with a dev-only warning, e.g. `[Eufemia] <Input>: `statusState` is deprecated, use `status={{ error: 'message' }}`.`

The only cost is `status={{ warning: '…' }}` being a touch more verbose than a flat `warning="…"` — but that wrapping is what signals "one slot, not three toggles." Curious what others think.

But again, does it make sense that this feature/functionality works different and has different APIs in Eufemia and Eufemia forms?

tujoworker:
In EF we support showing several statuses at once. All carefully wired together for best accessibility.

I think we may rather remove status support in most components at some point of time in favor of EF. 
This way we can have focus on the component itself than the dependency complexity.

PR #8366, opened 2026-06-02
feat(Filter): add new filter component

tujoworker:
Preview: https://feat-filter-field.eufemia-e25.pages.dev/uilib/components/filter

Motivation: 
- https://dnb-it.slack.com/archives/CMXABCHEY/p1762773916178429
- https://dnb-it.slack.com/archives/CMXABCHEY/p1701694759544969
- https://dnb-it.slack.com/archives/CMXABCHEY/p1741264946283029


langz:
In Autocomplete, we highlight the search words/results, In Filter we do not:
Screenshot 2026-06-02 at 11 41 39

Screenshot 2026-06-02 at 11 42 58


Have that been considered?

langz:
Just a minor nitpick, animation together with datepicker seems a bit off:


https://github.com/user-attachments/assets/c87bfb7c-3dd3-4f75-909a-594b0c8cb71a

langz:
When removing filters set through checkboxes, the checkboxes is reset.
But when removing filters set through date picker, the date picker is not reset:


https://github.com/user-attachments/assets/32024e6b-429f-4cde-aaae-1dfc69372dad

langz:
Based on the placeholder in the example, I assumed I was able to actually search for amount as well, but it did not seem like that's possible:


https://github.com/user-attachments/assets/b89b3e16-7499-43df-a0ba-3c0edcd8d56c

packages/dnb-eufemia/src/components/filter/style/dnb-filter.scss
```diff
@@ -0,0 +1,122 @@
...
+
+  &__toolbar {
+    display: flex;
+    align-items: flex-end;
+    justify-content: space-between;
+    flex-wrap: wrap;
+    gap: 1rem;
+  }
+
+  &__quick-filter {
```
langz: Is this meant to be used?
Does not look like it's used as of right now 👍

langz:
Some places we do `filterKey="type"` and in others we do `filterKey="/type"`, which is correct?
And what's the difference for a user? Initially when I read `"/"` I think about eufemia forms path or a URL path

langz:
In manual mode, if I type the search is applied immediate I do not need to confirm/commit or press a search button for the search to fire.

Also, if I select a filter without commiting/confirming it, then text searching, the filter is activated/used, even without commiting/confirming it. And the filter is not displayed as a tag, like "aktive filtre".

https://github.com/user-attachments/assets/f11dbf4e-40f3-4246-a8ae-b2d4b0325080

langz:
Sould we support some kind of `debounce` prop to avoid excessive network requests, so that it does not fire on every single key press?

langz:
I often find autosuggest/autocomplete to be very annoying. It could of course be helpful in certain scenarios. I think it’s by default off in autocomplete, and not sure if we can enable it?
Anyhow, I think by default autosuggestion in search fields is not good:


https://github.com/user-attachments/assets/b4ad80c9-3bc3-4321-bcdc-5047f1107a44

langz:
Inline Datepicker filter on my phone looks a bit strange. Not sure if it’s possible to fix:

image

langz:
Some different borders in darkmode in the following examples:
- https://feat-filter-field.eufemia-e25.pages.dev/uilib/components/filter/#multi-selection-filter-with-grid-layout
- https://feat-filter-field.eufemia-e25.pages.dev/uilib/components/filter/#decoupled-hook

Screenshot 2026-06-02 at 14 23 34

tujoworker:
> Inline Datepicker filter on my phone looks a bit strange. Not sure if it’s possible to fix:

It works better with no additional padding like we have in the example boxes:

Screenshot 2026-06-03 at 08 07 56

packages/dnb-eufemia/src/components/filter/style/dnb-filter.scss
tujoworker: Its used now. 🙏

tujoworker (commented):
Thanks for the great feedback so far 🙏 Everything should be covered now.

langz:
> Thanks for the great feedback so far 🙏 Everything should be covered now. 

Great, I'll take a round of testing now :shipit:

packages/dnb-eufemia/src/components/filter/hooks/useReactRouter.ts
```diff
@@ -0,0 +1,124 @@
...
+  const readFromUrl = useCallback(() => {
+    const urlSearch = excludeSearch
+      ? ''
+      : (searchParams.get(searchParam) ?? '')
+    let urlFilters: Record<string, FilterValue> = {}
+
+    const filtersRaw = searchParams.get(filtersParam)
+    if (filtersRaw) {
+      try {
+        urlFilters = JSON.parse(filtersRaw)
```
langz: In all places where we do `urlFilters = JSON.parse(filtersRaw)`, should we have some kind of validation or such, to prevent users to injecting potential dangerous values?
As I assume they can write a lot of different scary things in a query param ?scary=__proto__something

packages/dnb-eufemia/src/components/filter/hooks/useQueryLocator.ts
```diff
@@ -0,0 +1,125 @@
...
+      }
+
+      const filterKeys = Object.keys(currentFilters)
+      if (filterKeys.length > 0) {
+        url.searchParams.set(filtersParam, JSON.stringify(currentFilters))
+      } else {
+        url.searchParams.delete(filtersParam)
+      }
+
+      window.history.pushState({}, '', url.toString())
```
langz: Do we change the query param on every key stroke that the user enters?
If so, the typing "brukskonto" will probably result in many many changes in the URL, and then bacwards navigation will be tedious and take time.

Example of searching "brukskonto":
- http://localhost:8000/?query=b
- http://localhost:8000/?query=br
- http://localhost:8000/?query=bru
- http://localhost:8000/?query=bruk
- http://localhost:8000/?query=bruks
- http://localhost:8000/?query=bruksk
- http://localhost:8000/?query=bruksko
- http://localhost:8000/?query=brukskon
- http://localhost:8000/?query=brukskont
- http://localhost:8000/?query=brukskonto

So if I want to navigate back to the initial state, I will have to press back button 10 times.
Perhaps this could be solved by having a debounce by default.

packages/dnb-eufemia/src/components/filter/FilterSearch.tsx
```diff
@@ -0,0 +1,134 @@
...
+        'dnb-filter__search',
+        showIndicator && 'dnb-filter__search--show-indicator',
+        className
+      )}
+    >
+      <Input
+        {...rest}
+        label={label}
+        labelDirection="vertical"
+        value={isManual ? localValue : context.state.search}
```
langz, line 116: When it's manual, I'm not sure if we reset the value(localValue) when calling `resetFilters()`

packages/dnb-eufemia/src/components/filter/hooks/useQueryLocator.ts
tujoworker: This is resolved by replacing history instead of adding.

packages/dnb-eufemia/src/components/filter/FilterSearch.tsx
tujoworker: This is resolved as well. Nice you did catch this!

langz:
Sorting, can that be implemented and realized in the given state of the code?
Or does that need more code from our point of view?

Perhaps nothing for now, but a future feature request?

langz:
After testing the [quick filter example](https://feat-filter-field.eufemia-e25.pages.dev/uilib/components/filter/demos/#quick-filters), I was left with the feeling that this filter works different than in other examples.


[In the quick filter example](https://feat-filter-field.eufemia-e25.pages.dev/uilib/components/filter/demos/#quick-filters), selecting card and transfer will try to find transactions with card AND transfer (resulting in 0 matches):


https://github.com/user-attachments/assets/736cac5b-533a-44df-bf32-5251ac61a142




In the [basic example](https://feat-filter-field.eufemia-e25.pages.dev/uilib/components/filter/demos/#basic-usage), selecting card and transfer will try to find transactions with card OR transfer (resulting in multiple matches):

https://github.com/user-attachments/assets/8e6b72f4-1609-414b-a7db-c8a82fbd05d3




I guess how this works and is implemented is entirely decided by the application, so it does not matter that the examples behave a bit different.

tujoworker:
> After testing the [quick filter example](https://feat-filter-field.eufemia-e25.pages.dev/uilib/components/filter/demos/#quick-filters), I was left with the feeling that this filter works different than in other examples.

> I guess how this works and is implemented is entirely decided by the application, so it does not matter that the examples behave a bit different.

Yeah, I think the same. Nice you catch such differences!

packages/dnb-eufemia/src/components/filter/FilterDocs.ts
```diff
@@ -0,0 +1,384 @@
...
+  },
+}
+
+export const SearchProperties: PropertiesTableProps = {
+  label: {
+    doc: 'Label for the search input.',
+    type: 'string',
+    status: 'required',
+  },
+  variant: {
```
langz: Could this also be `behavior`?
Or does it make sense that this is `variant` while we use `behavior` in a different place?

Personally, I like behavior better than variant, as with variant I sort of mostly expect a change in visual styling, and not "functionality".

Could also be `mode` for both `behavior` and `variant`? Just thoughts...

tujoworker:
> Sorting, can that be implemented and realized in the given state of the code? Or does that need more code from our point of view?
> 
> Perhaps nothing for now, but a future feature request?

Yeah, the data structure will be very different from case to case, so as of now, I think we are fine.

packages/dnb-eufemia/src/components/filter/FilterDocs.ts
```diff
@@ -0,0 +1,384 @@
...
+    doc: 'Label for the selection filter.',
+    type: 'string',
+    status: 'required',
+  },
+  filterKey: {
+    doc: 'Unique key for this filter in the state.',
+    type: 'string',
+    status: 'required',
+  },
+  options: {
```
langz: In DrawerList components we call it `data` and not `options`.
I'm fine with calling it `options` as long as we've named it like this intentionally, because there's a difference 🤝

packages/dnb-eufemia/src/components/filter/FilterDocs.ts
tujoworker: Yeah, I was about to change it to `submitBehavior`.

packages/dnb-eufemia/src/components/filter/FilterDocs.ts
tujoworker: Good catch! I think we should change it to `data`.

langz:
> > Inline Datepicker filter on my phone looks a bit strange. Not sure if it’s possible to fix:
> 
> It works better with no additional padding like we have in the example boxes:
> 
> Screenshot 2026-06-03 at 08 07 56

Aha, thanks :)

But I guess this will end up being a problem down the road either way, as we can't control how much padding other apps will have around this filter component in their app, so it will most likely happen on smaller devices there I guess.

tujoworker:
> In Autocomplete, we highlight the search words/results, In Filter we do not: 
> Have that been considered?

Yes. But I think we can add this later as a enhancement. Same with `Field.MultiSelection`.

langz (approved):

packages/dnb-eufemia/src/components/filter/FilterDocs.ts
```diff
@@ -0,0 +1,425 @@
...
+}
+
+export const SearchProperties: PropertiesTableProps = {
+  label: {
+    doc: 'Label for the search input.',
+    type: 'string',
+    status: 'required',
+  },
+  submitBehavior: {
+    doc: 'When set to `"onSubmit"`, the search state is only updated when the user presses Enter or clicks the submit button. Requires `type="search"` for the submit button to appear.',
```
langz: Could `type="search"` be auto applied if/when `submitBehavior="onSubmit"`, or would that be confusing for the user?

packages/dnb-eufemia/src/components/filter/FilterMultiSelection.tsx
```diff
@@ -0,0 +1,95 @@
...
+      label={label}
+      filterKey={filterKey}
+      defaultOpen={defaultOpen}
+    >
+      <MultiSelection
+        label={label}
+        labelSrOnly
+        data={data}
+        value={currentValue}
+        onChange={handleChange as (value: unknown) => void}
```
langz:
```suggestion
        onChange={handleChange as FieldMultiSelectionProps['onChange'] => void}
```

packages/dnb-eufemia/src/components/filter/FilterDocs.ts
langz: So now one is named `behavior`, and the other one named `submitBehavior`.
Does it make sense that the property name is not identical? Should they be the same?

packages/dnb-eufemia/src/components/filter/FilterContext.ts
```diff
@@ -0,0 +1,60 @@
+import { createContext } from 'react'
+import type { RefObject } from 'react'
+
+export type FilterValue = {
+  value: string | number | boolean | Record<string, unknown>
+  label: string
+  filterLabel?: string
```
langz: Could this be named something else?
`groupLabel` or `categoryLabel`?

Reason for suggesting a change is because to me everything inside `FilterValue` here is already a "filter", so I find it slightly confusing when setting a label of a filter, and setting a filterLabel of a filter.

packages/dnb-eufemia/src/components/filter/FilterExport.ts
```diff
@@ -0,0 +1,28 @@
...
+export { default as QuickFilters } from './FilterQuickFilters'
+export { default as Highlighting } from './FilterHighlighting'
+export { default as SortButton } from './FilterSortButton'
+export { useFilter } from './useFilter'
+export { useFilterAsync } from './useFilter'
+export { useFilterContext } from './useFilter'
+export type { FilterAsyncOptions } from './useFilter'
+export { default as useQueryLocator } from './hooks/useQueryLocator'
+export { default as useReactRouter } from './hooks/useReactRouter'
+export { default as useNextRouter } from './hooks/useNextRouter'
```
langz, line 24: Should these hooks also be prefixed with Filter, such as the other hooks?
`useQueryLocator`, `useReactRouter` and `useNextRouter` seems a bit generic.
Perhaps `useFilterQueryLocator`, `useFilterReactRouter` and `useFilterNextRouter`?

tujoworker: They are documented as `Filter.useQueryLocator` 👍

packages/dnb-eufemia/src/components/filter/FilterContext.ts
tujoworker: categoryLabel 👍

packages/dnb-eufemia/src/shared/locales/en-GB.ts
```diff
@@ -197,5 +197,18 @@ export default {
...
+      activeFiltersLabel: 'Active filters',
+      panelButtonLabel: 'Filter',
+      hideFilterLabel: 'Hide filter',
+      applyFilterLabel: 'Apply filter',
+      cancelFilterLabel: 'Cancel',
+      dateLabel: 'Date',
+      sortButtonLabel: 'Sort',
+      noResultsMessage:
+        'No results found. Try changing or removing some filters.',
+      resultCountMessage: '%s result(s)',
```
langz, line 212: Just a minor cosmetic nitpick is that we could improve translations for pluralization.


Suggestion: 
```
// en-GB
resultCountMessage: '%s results',
resultCountMessageSingular: '%s result',

// nb-NO — "treff" is already invariant, but for consistency:
resultCountMessage: '%s treff',
resultCountMessageSingular: '%s treff',

// da-DK
resultCountMessage: '%s resultater',
resultCountMessageSingular: '%s resultat',

// sv-SE
resultCountMessage: '%s resultat',
resultCountMessageSingular: '%s resultat',
```

langz:
The new Datepicker on small screens looks great 😄 

Looks to be some animation improvements:

https://github.com/user-attachments/assets/18f61be9-1f19-40b2-921d-b8c8c9d8371f

langz:
Should I take one more round and look through this?

tujoworker:
yes. Added some commits. Here is a StackBlitz using an external API:

https://stackblitz.com/edit/zsvkba4w?file=src%2FApp.tsx

langz:
Nice nice, I like the new `4 treff`, etc 🚀 

Not sure if it's possible to not show it when removing filter/search back to showing all results?
It temporarily displays, before being animated away. No biggie:

It briefly displays "3 treff":


https://github.com/user-attachments/assets/9c0e940c-039b-4f76-8803-040026e53729

langz:
No problem, but showing both `0 treff `and `Ingen treff. Prøv å endre eller fjerne noen filtre.` seems a bit redundant:

Screenshot 2026-06-05 at 14 45 52

langz:
I really like the "clear filters" button, how does one display this? Is it always displayed? Could it always be displayed? Even though if only one filter is selected?
Screenshot 2026-06-05 at 14 48 06

I can not see it [here](https://feat-filter-field.eufemia-e25.pages.dev/uilib/components/filter/demos/#basic-usage) when applying filters:

 
image

Personally, I really like having this "remove all filters" available at all times, even though if I've only selected 1 item.

langz:
> yes. Added some commits. Here is a StackBlitz using an external API:
> 
> https://stackblitz.com/edit/zsvkba4w?file=src%2FApp.tsx

Works great 🚀

langz:
> The new Datepicker on small screens looks great 😄
> 
> Looks to be some animation improvements:
> 
>  trim.17A40D1A-6BF7-4AA6-B598-0C936D0FF7AA.MOV

This is working great now ❤️

langz (approved):
💃

tujoworker:
> Not sure if it's possible to not show it when removing filter/search back to showing all results?

yeah, its now like UX has defined it. But we have a prop `alwaysVisible` 👍

tujoworker:
> Personally, I really like having this "remove all filters" available at all times, even though if I've only selected 1 item.

yeah, UX has defined it like this. 🤷‍♂️
There are reasons why this is good I guess.

langz:
> > Personally, I really like having this "remove all filters" available at all times, even though if I've only selected 1 item.
> 
> 
> 
> yeah, UX has defined it like this. 🤷‍♂️
> 
> There are reasons why this is good I guess.

I guess I would find it confusing if it "sometimes" show and "other times" do not show when I'm using filters in the same session. 

Let's say if it only shows if there's more than f filters selected, it's nothing that I as a user will naturally understand.

tujoworker:
> Let's say if it only shows if there's more than f filters selected, it's nothing that I as a user will naturally understand.

Strong argument.

langz:
The "5 treff" info is acting strange in manual mode:

It displays "5 treff" before I've applied/confirmed the filter. It's also "wrong", and when applying the filter it's actually only "3 treff":

https://github.com/user-attachments/assets/696e064f-9dcf-4384-be06-dcaba2061ade

langz:
"Fjern alle" in manual mode doesn't seem to work correctly:

https://github.com/user-attachments/assets/25697cff-0e19-481c-a10b-54920bd2759f

langz:
Manual mode is still a bit difficult to get right.
If I add a filter as a draft (not commited it by pushing use filter-button), and then removing the existing filter by using the tag, then suddenly the drafted filter which was not commited is now an active filter:



https://github.com/user-attachments/assets/40baa321-8b35-4543-b5bb-5d809f39b480

langz:
> "Fjern alle" in manual mode doesn't seem to work correctly:
> 
>  Screen.Recording.2026-06-08.at.09.21.13.mov

Still able to reproduce this btw.

langz:
> The "5 treff" info is acting strange in manual mode:
> 
> It displays "5 treff" before I've applied/confirmed the filter. It's also "wrong", and when applying the filter it's actually only "3 treff":
> 
>  Screen.Recording.2026-06-08.at.09.20.40.mov

This looks better now 🤝 
Thanks!

However, it looked to also affect the "live searching" in manual mode. Now there's no "X treff" displayed when using the live search in "manual mode". Manual mode is a bit difficult, as not every filter and/or search follows the same "draft and confirm/commit pattern":


https://github.com/user-attachments/assets/7e6f1cea-655c-4c15-b09c-c9823b6de694

For comparison, it displays "X treff" in "normal mode" (non manual mode):

https://github.com/user-attachments/assets/9e7c927c-adfc-4c45-8e72-096d4c1d1d56

packages/dnb-eufemia/src/components/filter/FilterDocs.ts
```diff
@@ -0,0 +1,483 @@
...
+  },
+  error: {
+    doc: 'The error thrown by the fetcher, if any. Reset to `undefined` on each new fetch.',
+    type: ['Error', 'undefined'],
+    status: 'required',
+  },
+}
+
+export const RootEvents: PropertiesTableProps = {
+  onChange: {
```
langz, line 475: This did not fire in manual mode, not sure if that's expected or not.

packages/dnb-eufemia/src/components/filter/FilterHighlighting.tsx
langz, line 1: We could consider documenting this component in the properties docs as well

packages/dnb-eufemia/src/components/filter/FilterDocs.ts
tujoworker: It should and does. Update the description.

packages/dnb-eufemia/src/components/filter/FilterHighlighting.tsx
tujoworker: Yeah, I think so too. Its added now.

tujoworker:
Thank you @langz for the fantastic good feedback and review. I think all feedback has been looked into.

packages/dnb-eufemia/src/components/filter/FilterDocs.ts
langz: > It should and does. Update the description.

Good, thanks. I don't think it did so when searching only in manual mode. 
Will check again.

langz:
> Manual mode is still a bit difficult to get right. If I add a filter as a draft (not commited it by pushing use filter-button), and then removing the existing filter by using the tag, then suddenly the drafted filter which was not commited is now an active filter:
> 
>  Screen.Recording.2026-06-08.at.14.35.26.mov

I'm still experiencing this.

To reproduce visit the following [example](https://feat-filter-field.eufemia-e25.pages.dev/uilib/components/filter/#manual-behavior):
- Select active filter
- Press "bruk filter"
- Select inactive filter
- Press the x-button of the tag for the active filter.
- Now suddenly the inactive filter has become "active" (in use), but I've never pressed "bruk filter" to confirm that we should use inactive filter. All I did was press the x-button of the tag for the active filter.

langz:
A minor nitpick is that one can briefly see «treff 4» when no filters is applied after one have removed/cleared existing filters in the quick filters example. Reason for me mentioning this is because we do greatly improved the same scenario in other examples some time last week:



https://github.com/user-attachments/assets/6a907b08-36a8-4fc5-9fe2-24d73ac277f1

packages/dnb-eufemia/src/components/filter/FilterMultiSelection.tsx
```diff
@@ -0,0 +1,98 @@
+import { useCallback, useContext, useMemo, useRef } from 'react'
+import MultiSelection from '../../extensions/forms/Field/MultiSelection/MultiSelection'
```
langz, line 2: Can this be problematic?
That a regular "eufemia component" imports a "eufemia forms" component?

langz (approved):
🎉 Excellent work on this Filter component!

Thanks for looking into all my feedback 😫 🙏 

This is going to be a valuable addition to Eufemia. Thanks for the solid work! 👏

packages/dnb-eufemia/src/components/filter/FilterMultiSelection.tsx
tujoworker: It should be fine. Because its actually only shipped in the bundle if `Filter.MultiSelection` is used.

PR #8367, opened 2026-06-02
fix(PaymentCard): inherit text color from card instead of typography token

langz:
Motivation: In v11, the P element started setting an explicit color via var(--token-color-text-neutral), which overrides the card's own text color in light mode. Add color: inherit to the card number and card type text selectors so they respect the card's design color.

Because of the existing css styling that exists, I think it's just as nice to add css instead of `doing surface="dark"` or such


packages/dnb-eufemia/src/extensions/payment-card/__tests__/__image_snapshots__/PaymentCard-have-to-match-a-card-with-status-1.png
karini6, line 1: These overlays with card statuses should have a white color as well :)

langz: Yeah, great, I just noticed as well :) 
Should be updated in https://github.com/dnbexperience/eufemia/pull/8367/commits/9a13001225f2070567d78de47bab2467289a51ff

tujoworker (approved):

PR #8371, opened 2026-06-02
fix: align Tailwind breakpoint variables with allAbove mixin

tujoworker:
The `allAbove` SCSS mixin adds a `0.00625em` offset to breakpoint values, but the generated `--breakpoint-*` CSS variables in `properties-tailwind.css` used round values (e.g. `40em`). This caused Tailwind breakpoints and Grid spans/mixins to trigger at different viewport widths, leading to inconsistent layout behavior.

This applies the same offset during the Tailwind CSS variable generation so that values like `--breakpoint-small: 40.00625em` match what the mixin actually uses.


joakbjerk (approved):

langz (approved):
Image

PR #8374, opened 2026-06-02
feat(PortalRoot.Provider): add `translate` property

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1780388191576599

Consumers who experience Google Translate crashing their app can now wrap their app in `<PortalRoot.Provider translate='no'>` to prevent translation tools from mutating DOM inside portals.

This is opt-in rather than a default, since I believe the crash depends on the apps code, like React version and specific translation tool behavior.

One could also add a `translate` prop or a `portalRootProps` prop to each component using a portalroot, but that could be excessive? Many translate properties in many components:
- Autocomplete
- Dropdown
- DatePicker
- Dialog
- Modal
- Popover
- Tooltip
- Drawer


langz:
https://github.com/dnbexperience/eufemia/pull/8422

PR #8382, opened 2026-06-02
chore: update Yarn from 4.6.0 to 4.16.0

### LINKED ISSUE #8373: Update Yarn
## ~~🚀 Feature Proposal~~ Update encouragement 🔒

Update yarn to more recent version. Current pinned version is: 4.6.0 from December 2024.

## Motivation

Fixed and improved security features in the package manager.

## Example

4.10.0+ add `npmMinimalAgeGate` and staying on a 1,5 year old yarn version might not be recommended.
4.14.0+ Makes `enableScripts: false` the default
4.15.0+ Makes `npmMinimalAgeGate: 1d` the default

Recommend updating to 4.15.0 (or 4.16.0).

langz:
Update Yarn to the latest stable version (4.16.0) for improved security features including npmMinimalAgeGate and enableScripts defaults.

Closes #8373


.yarnrc.yml
```diff
@@ -1,11 +1,18 @@
...
 defaultSemverRangePrefix: ""
 
 enableGlobalCache: false
 
+enableScripts: true
+
 nodeLinker: node-modules
 
+npmMinimalAgeGate: 0
+
```
langz: Let's discuss these settings 🏵️

tujoworker (approved):

.yarnrc.yml
```diff
@@ -4,8 +4,12 @@ defaultSemverRangePrefix: ""
 
 enableGlobalCache: false
 
+enableScripts: true
```
langz:
```suggestion
enableScripts: false
```

.yarnrc.yml
langz: I've now set `enableScripts: false`

langz: And `npmMinimalAgeGate: 3d`

langz: We can always change this if we experience that it does not work out in our project. But for now it's a good baseline to start off with I guess.

PR #8386, opened 2026-06-02
fix(Input): keep submit button icon at default size for medium inputs

tujoworker:
The submit button icon in medium-sized inputs was incorrectly set to `medium` (24px) when it should stay at `default` (16px). Only large inputs should upscale the icon.

Previously the condition was `size === 'medium' || size === 'large'` → `'medium'`, now only `size === 'large'` triggers the medium icon size.


langz (approved):

langz:
We'll most likely have to update some screenshot tests 🤞

PR #8400, opened 2026-06-03
fix(Autocomplete, Dropdown): scroll selected item to top when opening

tujoworker:
Preview: https://fix-drawer-list-scroll-selec.eufemia-e25.pages.dev/uilib/components/autocomplete/demos

When a DrawerList opens with a pre-selected item far down the list, the selected item was placed at the bottom of the visible area because `scrollIntoView` used `block: 'nearest'`. This changes the initial scroll to use `block: 'start'` so the selected item appears at the top, making it immediately visible. Keyboard navigation still uses `block: 'nearest'` for minimal scrolling.

Screenshot 2026-06-03 at 14 06 24


langz (approved):

langz (approved):
Works good on my iPhone 📲

PR #8401, opened 2026-06-03
fix(Dropdown): correct portal positioning for right-aligned dropdowns

tujoworker:
Preview: https://fix-dropdown-right-align-por.eufemia-e25.pages.dev/uilib/components/dropdown/demos


langz (approved):
Do we use popover in this component? If so, does this fix work as expected together with that and edge cases such as positioned in the outer corners of the view and what then happens if the portal is right aligned.

tujoworker:
We do not use Popover. It has its own logic for that.

PR #8413, opened 2026-06-04
fix(Switch): prevent visual toggle when preventDefault is called in onClick

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1780566122012469
Issue reprod: https://stackblitz.com/edit/u6dlg7ok-bke62smr?file=src%2FApp.tsx
Fix reprod: https://stackblitz.com/edit/u6dlg7ok-98jq1e2i?file=src%2FApp.tsx

The preventDefault callback in onClick was mutating internal checked state (isCheckedRef) and calling forceUpdate, causing the switch to visually flip even though the controlled checked prop hadn't changed. The flag was also never reset, permanently blocking subsequent onChange events.

Now preventDefault only sets a flag, and event.preventDefault() is called after the user's onClick returns. The flag is reset at the start of each click and in onChangeHandler as a safety net.

### DIFF
--- a/packages/dnb-eufemia/src/components/switch/Switch.tsx
+++ b/packages/dnb-eufemia/src/components/switch/Switch.tsx
@@ -184,6 +184,7 @@ function Switch(props: SwitchProps) {
   const onChangeHandler = useCallback(
     (event) => {
       if (preventChangeRef.current) {
+        preventChangeRef.current = false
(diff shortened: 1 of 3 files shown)


tujoworker (approved):
Nice 👍

PR #8416, opened 2026-06-04
chore(Portal): add ErrorBoundary to the portal

tujoworker:
The portal had no ErrorBoundary — an unhandled render error would crash the entire app with a blank screen.

This adds an `ErrorBoundary` that catches render errors and shows a `GlobalError` page with the error message, giving users a meaningful error screen instead.

Also updates `GlobalError` to skip rendering the "Error code:" text when `statusCode` is null or empty, since a client-side render crash has no HTTP status code.


langz (approved):

packages/dnb-design-system-portal/src/core/ErrorBoundary.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  static getDerivedStateFromError(error: Error): ErrorBoundaryState {
+    return { error }
+  }
+
+  render() {
+    if (this.state.error) {
+      return (
+        <GlobalError
+          statusCode={null}
```
langz, line 28: Should we update the type of `statusCode` to support `null`, or perhaps it's all good.

langz: I guess this will also result in the css class`dnb-global-error--null`, which I guess is fine

PR #8422, opened 2026-06-04
feat(PortalRoot): forward HTML attributes to portal elements

tujoworker:
Preview: https://feat-portal-root-forward-htm.eufemia-e25.pages.dev/uilib/components/portal-root#forward-html-attributes-to-portal-content
Closes #8374

### DIFF
--- a/packages/dnb-eufemia/src/components/portal-root/PortalRoot.tsx
+++ b/packages/dnb-eufemia/src/components/portal-root/PortalRoot.tsx
@@ -47,22 +47,35 @@ export type PortalRootProps = {
 } & SelectorOptions &
   Omit<HTMLProps<HTMLElement>, 'ref' | 'id'>
 
-type PortalRootContextValue = SelectorOptions
+type PortalRootContextValue = SelectorOptions & {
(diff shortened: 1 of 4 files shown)


langz:
I guess we should close this then?
https://github.com/dnbexperience/eufemia/pull/8374

Even though the approach is quite similar.

tujoworker:
Yeah, it's similar. But I want to remove all Google Translate docs from this PR before merge.

langz (approved):
🤝

packages/dnb-eufemia/src/components/portal-root/PortalRoot.tsx
```diff
@@ -154,6 +167,7 @@ function PortalRootInstance(props: PortalRootProps = {}): ReactNode {
           className
         )}
         style={{ ...scopeStyle, ...style }}
+        {...selectorContext?.portalAttributes}
```
langz, line 170: Not a big problem, that will not likely happen, but now that we support passing attributes, one could now potentiall pass style and className which would override the existing style and className set here as props, and would hence override internally set things.

PR #8423, opened 2026-06-04
fix(DrawerList): add padding to scroll when list is on top

snorrekim:
A fix for a bug caused by this PR: https://github.com/dnbexperience/eufemia/pull/8400


snorrekim:
This was happening:
image
should be like this:
image

tujoworker (approved):

langz (approved):

PR #8424, opened 2026-06-04
chore: update ESLint to v10 and related packages

langz:


packages/dnb-design-system-portal/eslint.config.mjs
```diff
@@ -39,6 +40,9 @@ export default [
           ],
         },
       ],
+      // The portal is a documentation site, not shipped to browsers with
+      // restricted compat requirements – disable compat checking.
+      'compat/compat': 'off',
```
langz, line 45: I guess we should remove this

packages/dnb-eufemia/eslint.config.mjs
```diff
@@ -116,6 +101,14 @@ export default [
     },
     rules: {
       ...securityRecommendedRules,
+
+      // New rules added to eslint:recommended in ESLint 10 – set to warn
+      // so they don't block CI while the codebase is incrementally updated.
```
tujoworker: 👍

tujoworker (approved):

packages/dnb-design-system-portal/eslint.config.mjs
tujoworker: Will many APIs fail then? I guess no, so when we can remove it, yes.

langz: I'll keep this, as it's only for portal files for now.

PR #8427, opened 2026-06-04
chore: add descriptive context to error/warning messages across components

langz:
Replace bare warn(e) and console.error(error) calls with descriptive messages that identify which component failed and what operation was attempted. This helps developers debug issues faster.


tujoworker (approved):
Nice!

PR #8429, opened 2026-06-05
feat(BrowserTranslate): add shared helper to prevent browser translation on form components

tujoworker:
Depends on #8422
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1780471641364799?thread_ts=1780388191.576599&cid=CMXABCHEY
Preview: https://feat-google-translate-helper.eufemia-e25.pages.dev/uilib/components/portal-root/demos/#browsertranslate

Adds a `BrowserTranslate` shared helper component that prevents browser translation tools (e.g. Google Translate) from modifying form component content.

## What it does

`BrowserTranslate` combines the Eufemia Provider's `formElement` context with `PortalRoot.Provider`, so both the visible component (e.g. a dropdown button) and its portal-rendered content (e.g. the dropdown list) receive `translate="no"`.

```tsx
import { BrowserTranslate } from '@dnb/eufemia/shared'

<BrowserTranslate off>
  <Dropdown data={['Rød', 'Blå', 'Grønn']} />
</BrowserTranslate>
```

## Feedback

- Any thoughts on the placement?
- And the docs?

### DIFF
--- a/packages/dnb-eufemia/src/shared/GoogleTranslate.tsx
+++ b/packages/dnb-eufemia/src/shared/GoogleTranslate.tsx
@@ -1,0 +1,28 @@
+import type { JSX, ReactNode } from 'react'
+import Provider from './Provider'
+import { PortalRootProvider } from '../components/portal-root/PortalRoot'
+
+export type GoogleTranslateProps = {
+  /**
+   * Set to `true` to disable browser translation (e.g. Google Translate)
+   * on all form components and their portal content within this scope.
+   */
+  off?: boolean
+
+  children: ReactNode
+}
+
+export default function GoogleTranslate({
+  off,
+  children,
+}: GoogleTranslateProps): JSX.Element {
+  if (!off) {
+    return <>{children}</>
+  }
+
+  return (
+    <Provider formElement={{ translate: 'no' }}>
+      <PortalRootProvider translate="no">{children}</PortalRootProvider>
+    </Provider>
+  )
+}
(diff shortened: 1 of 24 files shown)


packages/dnb-eufemia/src/shared/GoogleTranslate.tsx
```diff
@@ -0,0 +1,28 @@
...
+  /**
+   * Set to `true` to disable browser translation (e.g. Google Translate)
+   * on all form components and their portal content within this scope.
+   */
+  off?: boolean
+
+  children: ReactNode
+}
+
+export default function GoogleTranslate({
```
kambo: **thought:** GoogleTranslate is probably the most common translation tool the attribute is not specific to Google Translate. Maybe something like `AutomatedTranslation` or `Translators` or maybe just `Translate`

kambo (approved):

packages/dnb-eufemia/src/shared/GoogleTranslate.tsx
tujoworker: Its renamed to `AutoTranslate` for now.

langz:
There's a a lot of references left saying `AutoTranslate` and not `BrowserTranslate` 🙏

langz:
Could we drop the `off` prop entirely? And name it like `NoTranslate`, or `BrowserTranslateOff`.
And having the component always apply `translate="no"`. Users who don't want it simply don't use the component. That could lead to a simpler API.

Just a thoughts, not important, I think the existing API as of now works as well, and @snorrekim liked the existing API as well 🤝

langz (approved):
Looks good to me, a great addition 👏 
I added a few comments, but nothing really critical, perhaps only remember to change from `AutoTranslate` to `BrowserTranslate` 🤞

langz (approved):

PR #8434, opened 2026-06-05
chore(SkipContent, Popover): add type="button" to native button elements

langz:
Native <button> elements without type default to "submit", which can cause accidental form submissions when these components are used inside a form. The Eufemia Button component already defaults to type="button", but these raw <button> elements were missing the attribute.


tujoworker (approved):
A test in Popover would be nice too.

langz:
> A test in Popover would be nice too.

Thanks, I've done so in the following commit https://github.com/dnbexperience/eufemia/pull/8434/changes/88b71368b99bc32c11390ab92d5e9e3742a65931

PR #8436, opened 2026-06-05
fix(DatePicker): range mode showing same month in both calendar views

langz:
When startDate and endDate are in the same month, the right calendar
view incorrectly displayed the same month as the left. It now correctly
shows the next month.

For the following example:
`<DatePicker range startDate="2024-10-01" endDate="2024-10-24" />`

From:
Screenshot 2026-06-05 at 10 04 04

To:
Screenshot 2026-06-05 at 10 04 20


joakbjerk (approved):
🔥

PR #8443, opened 2026-06-05
fix(DatePicker): clear hide timeout on unmount

tujoworker:
To prevent CI flake, clear the timeout on unmount.


langz (approved):

langz (commented):
Is it correct to place a "test fix" in the actual component code?

tujoworker:
> Is it correct to place a "test fix" in the actual component code?

Yes, because it's a runtime bug. That's why the tests sometimes have failed.

PR #8452, opened 2026-06-07
style(Card): unify design

joakbjerk:
Both theme files were identical, so styles were moved to dnb-card.scss, so no screenshots needed updating, or visual styles changed for either ui or sbanken.


langz (approved):
💳

PR #8459, opened 2026-06-08
feat(Form.SubmitIndicator): refine with scale animation

tujoworker:
This PR makes the pending submit indicator easier to read visually by using styled dot elements with a scale animation while keeping the left-to-right timing intact.

It also updates the related tests so they verify the rendered dot elements instead of relying on text punctuation.


packages/dnb-eufemia/src/extensions/forms/Form/SubmitIndicator/SubmitIndicator.tsx
```diff
@@ -77,7 +77,7 @@ function SubmitIndicator(props: FormSubmitIndicatorProps) {
           'aria-hidden': true,
         }
 
-  const dot = <b>.</b>
+  const dot = <b />
```
langz, line 80: The dot is no longer an actual dot (.) ? 😱 😄

langz (approved):
I'm fine with this approach, as long as it visually still looks nice (the same) :)

packages/dnb-eufemia/src/extensions/forms/Form/SubmitIndicator/SubmitIndicator.tsx
tujoworker: yeah, for the scaling animation, its better without.

joakbjerk (approved):

PR #8483, opened 2026-06-08
chore(DatePicker): remove `DatePickerPortal`

joakbjerk:
The debounced resize/scroll handler was recreated on every alignment change due to useCallback dependencies, defeating the debounce. Replaced with a stable useRef so the debounce instance is created once and reused.


langz (approved):
Technically I guess it looks decent, can one experience this?
What is it actually improving for our users?

joakbjerk:
> Technically I guess it looks decent, can one experience this? What is it actually improving for our users?

I guess its more of a performance improvement than user experience one? 🤔

tujoworker (changes requested):
I think the file can be deleted instead. We don't use it actually.

joakbjerk:
> I think the file can be deleted instead. We don't use it actually.

Verified that removal is possible 🫡

tujoworker (approved):

PR #8490, opened 2026-06-08
chore: remove unused CSS custom properties

langz:
Remove the following unused CSS custom properties from all themes:
- --sb-layout-x-large (80em) - sbanken specific
- --sb-layout-xx-large (90em) - sbanken specific


tujoworker:
We should check if `--shadow-sharp-blur-*` is maybe documented? I think I have seen this. And what about the layout width? no documentation about it?

langz:
> We should check if `--shadow-sharp-blur-*` is maybe documented? I think I have seen this. And what about the layout width? no documentation about it?

I'll do a round of investigation, thanks 😃

langz:
> > We should check if `--shadow-sharp-blur-*` is maybe documented? I think I have seen this. And what about the layout width? no documentation about it?
> 
> I'll do a round of investigation, thanks 😃

I did one, and I only remove the unused variables that's not documented now.
There's actually quite a few unused variables that's documented, but will not actually do anything.
A future task for v12

PR #8496, opened 2026-06-08
chore(DrawerList): respect dropdown z-index variable

tujoworker:
## Why

`--z-index-dropdown` is documented as the shared z-index token for inline dropdowns and drawer lists. Wiring DrawerList to the token keeps the public styling surface aligned with the implementation.

⚠️ Not sure about using "drop-down" for "drawer-list" purpose. 🤔


joakbjerk (approved):
Z

langz (approved):

PR #8497, opened 2026-06-08
chore(Forms): prevent CI timer leak

tujoworker:
A Forms submit-status timer can survive test teardown after an abort state, which lets React schedule an update after jsdom has already removed `window`. This caused an unhandled `ReferenceError: window is not defined` in CI even though the failing test itself had completed.

CI job where it happened: https://github.com/dnbexperience/eufemia/actions/runs/27156945621/job/80162209706


langz (approved):
⏲️

PR #8498, opened 2026-06-08
chore(Form.Appearance): clarify size example typing

tujoworker:
Tightens the Form.Appearance size example so the documented default option maps cleanly to the component API. This keeps the example type-safe while preserving the intended runtime behavior.

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Appearance/Examples.tsx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Appearance/Examples.tsx
@@ -1,14 +1,20 @@
 import ComponentBox from '../../../../../../shared/tags/ComponentBox'
 import { Form, Field } from '@dnb/eufemia/src/extensions/forms'
 import { Flex } from '@dnb/eufemia/src'
+import type { FormAppearanceProps } from '@dnb/eufemia/src/extensions/forms/Form/Appearance'
 
 export const Size = () => {
   return (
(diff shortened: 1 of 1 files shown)


packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Appearance/Examples.tsx
```diff
@@ -1,14 +1,20 @@
...
 import { Form, Field } from '@dnb/eufemia/src/extensions/forms'
 import { Flex } from '@dnb/eufemia/src'
+import type { FormAppearanceProps } from '@dnb/eufemia/src/extensions/forms/Form/Appearance'
 
 export const Size = () => {
   return (
     <ComponentBox data-visual-test="form-appearance-size">
       {() => {
+        type AppearanceData = {
+          size: 'default' | NonNullable<FormAppearanceProps['size']>
```
langz, line 11:
```suggestion
          size: 'default' | FormAppearanceProps['size']
```

Not sure if NonNullable is important for the example.
Also, I cant remember using it too much in our codebase either 😅

langz (approved):

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Appearance/Examples.tsx
tujoworker: Yeah, it's not needed in our case. But when using stricter typing config, then it is.

PR #8501, opened 2026-06-09
chore(Portal): restore hash scrolling after route changes

tujoworker:
## Summary

Portal hash links can miss their target after client-side route changes because the layout stays mounted while the new page content is rendered. This reruns the hash scroll behavior when the pathname or hash changes so route-change anchors scroll after the target content exists.


joakbjerk (approved):
#️⃣

PR #8512, opened 2026-06-09
chore(DatePicker): remove internal-only exports

joakbjerk:


langz (approved):
Good good, let's hope no consumers have magically found and used these for anything 😂 🤞

PR #8514, opened 2026-06-09
style(Input): unify styling (Sbanken)

joakbjerk:
All inputs should be 2rem in height based on figma

Screenshot tests of Inputs that have a dropdown menu will look a little off in this pr, since they need to be fixed in their own respective prs 😅


langz (approved):
I guess it was true once upon a time that input fields in sbanked should have different dimentions, but am assuming that does not longer apply 🤝

PR #8520, opened 2026-06-10
fix(Forms): skip onBlurValidator on submit when already validated

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1781073525433709
Issue reprod: https://codesandbox.io/p/sandbox/elastic-heyrovsky-forked-9kyw2x
Fix reprod: https://stackblitz.com/edit/44ggkpfq-ajezdn4e?file=src%2FApp.tsx

When a field's onBlurValidator has already run on blur with the current value, skip re-running it during form submit. The validator still runs on submit if the field was never blurred or if the value changed after the last blur.

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/hooks/__tests__/useFieldProps.test.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/hooks/__tests__/useFieldProps.test.tsx
@@ -6982,6 +6982,206 @@ describe('useFieldProps', () => {
         document.querySelector('.dnb-form-status')
       ).not.toBeInTheDocument()
     })
+
+    it('should not re-call onBlurValidator on submit when it already ran on blur with the same value', async () => {
+      const onBlurValidator = vi.fn(() => undefined)
+
+      render(
+        <Form.Handler>
+          <Field.String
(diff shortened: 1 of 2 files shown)


packages/dnb-eufemia/src/extensions/forms/hooks/__tests__/useFieldProps.test.tsx
```diff
@@ -6982,6 +6982,135 @@ describe('useFieldProps', () => {
...
+      // Submit without blur
+      fireEvent.submit(document.querySelector('form'))
+
+      await waitFor(() => {
+        expect(onBlurValidator).toHaveBeenCalledTimes(1)
+      })
+    })
+
+    it('should re-call onBlurValidator on submit when value changed after blur', async () => {
+      const onBlurValidator = vi.fn(() => undefined)
```
tujoworker:
```suggestion
      const onBlurValidator = vi.fn()
```

I don't think `() => undefined` is needed, as it is the same as without.

langz:
Closed this, as we discussed that this behavior by default can be scary

PR #8521, opened 2026-06-10
fix(Modal): fix controlled `open` prop not opening dialog

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1781087024326319
Issue reprod: https://stackblitz.com/edit/44ggkpfq-9bmvcvjy?file=src%2FApp.tsx
Fix reprod: https://stackblitz.com/edit/44ggkpfq-d5xg7pxh?file=package.json

The useEffect that calls toggleOpenClose relied on the 'hide' state to gate the open check, but 'hide' was stale (still true) when 'open' changed from false to true in the same render cycle.

Split the condition: when 'open' actually changed, use the prop directly instead of the stale 'hide' state. When only parent props changed, keep the existing hide-based checks since hide is stable in that case.

### DIFF
--- a/packages/dnb-eufemia/src/components/dialog/__tests__/Dialog.test.tsx
+++ b/packages/dnb-eufemia/src/components/dialog/__tests__/Dialog.test.tsx
@@ -674,6 +674,45 @@ describe('Dialog', () => {
     // Verify that focus is NOT on the submit button
     expect(document.activeElement).not.toBe(submitButton)
   })
+
+  it('should open controlled dialog without noAnimation', async () => {
+    const originalIsTest = window['IS_TEST']
+    window['IS_TEST'] = false
+
+    const TestComponent = () => {
+      const [showDialog, setShowDialog] = useState(false)
+
+      return (
+        <>
+          <button
+            data-testid="trigger"
+            onClick={() => setShowDialog(true)}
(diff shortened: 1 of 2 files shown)


packages/dnb-eufemia/src/components/dialog/__tests__/Dialog.test.tsx
```diff
@@ -674,6 +674,66 @@ describe('Dialog', () => {
...
+
+      // Reopen
+      await userEvent.click(screen.getByTestId('trigger'))
+
+      await waitFor(() => {
+        expect(document.querySelector('.dnb-dialog')).toBeInTheDocument()
+      })
+    } finally {
+      window['IS_TEST'] = originalIsTest
+    }
```
tujoworker, line 738: Why do we need to do that try/finally in that test?

tujoworker (approved):

packages/dnb-eufemia/src/components/dialog/__tests__/Dialog.test.tsx
langz: I'll look into it. If I remember rcorrwctly, it's because when is_test we automatically set noAnimation.

PR #8523, opened 2026-06-10
fix(DatePicker): prevent shortcuts from widening picker

tujoworker:
Preview: https://chore-date-picker-addon-wrap.eufemia-e25.pages.dev/uilib/components/date-picker/demos

The DatePicker addon could make the picker wider than the calendar when shortcut buttons had enough content to fit on one long row. This keeps shortcut content wrapping within the picker width so the calendar remains the sizing anchor.

Verified with the focused DatePicker SCSS Vitest block.

Screenshot 2026-06-10 at 13 52 08


joakbjerk (approved):

langz (approved):
🦁

PR #8524, opened 2026-06-10
chore(Portal): make sidebar resizable

tujoworker:
I had the need sometimes to just quickly adjust the left side. But;

- Maybe a toggle button would do the trick instead (would replace full screen button in the Tab bar).
- Maybe this feature is nice, but the width should be remembered in session storage so it makes more sense? I don't know.
- Double-click resets the position.

Preview: https://chore-make-portal-sidebar-re.eufemia-e25.pages.dev/uilib/


langz (approved):
Cool features!

PR #8531, opened 2026-06-11
chore: unify SASS/SCSS terminology across docs and scripts

langz:
Standardize on "SCSS" when referring to the file format/syntax (.scss files with curly braces) and "Sass" for the preprocessor tool.

- Replace "SASS" with "SCSS" in all documentation prose
- Replace "SASS (SCSS)" and "SCSS (SASS)" with just "SCSS"
- Fix "SASS function" to "SCSS mixin" where @mixin is referenced
- Update build script comments to say "SCSS to CSS"
- Update test descriptions to say "SCSS to CSS"
- Keep "Sass" (correct casing) for preprocessor tool references
- Keep `sass` in code identifiers (npm package, API calls)

PR #8532, opened 2026-06-11
chore: remove outdated IE reference in before-started docs

langz:
The "(to support IE)" note is no longer accurate — Eufemia does not target IE.


tujoworker (approved):

langz:
This was solved in https://github.com/dnbexperience/eufemia/pull/8531

PR #8533, opened 2026-06-11
style(HelpButton): unify design (Sbanken)

joakbjerk:


packages/dnb-eufemia/src/style/themes/sbanken/sbanken-theme-components.scss
langz, line 23: I find it a bit strange that the file `sbanken-theme-components.scss‎` contains references to both `theme-sbanken.scss` files and `theme-ui.scss`.

I guess it's correct and intentional, but I would guess all files should be references to `theme-sbanken.scss` 😂

langz (approved):

packages/dnb-eufemia/src/style/themes/sbanken/sbanken-theme-components.scss
joakbjerk: It's really strange, but I think they work as a sort of fallback, and they don't contain any color styles I think, just sizing and spacing, so It works, even though it looks weird 😅

PR #8534, opened 2026-06-11
feat(Forms): add validator options for submit control (`withValidatorOptions` including `runOnSubmit`)

tujoworker:
This adds a validator-level way to control whether custom field validators run during form submit.

The change is useful for request-backed validators where rerunning on submit can cause duplicate calls or unnecessary latency after the validator already ran from field interaction. `withValidatorOptions` keeps that behavior attached to the validator while `runOnSubmit` supports the explicit states `always`, `never`, and `when-changed`.

Validation coverage was added for both `onChangeValidator` and `onBlurValidator`, including the `when-changed` behavior, and the forms documentation was updated with usage guidance.


packages/dnb-eufemia/src/extensions/forms/hooks/useFieldProps.ts
```diff
@@ -627,6 +632,27 @@ export default function useFieldProps<Value, EmptyValue, Props>(
...
+  const isValidatorValueChanged = useCallback(
+    (validatorName: ValidatorName) => {
+      const cachedValue =
+        validatedValidatorValuesRef.current[validatorName]
+
+      if (!cachedValue) {
+        return true
+      }
+
+      return !Object.is(cachedValue.value, valueRef.current)
```
langz, line 651: I guess this works perfect for primitive fields.

What about fields returning objects, arrays, etc?
I guess they will always be considered "changed/updated". Perhaps it's only Field.Upload? 🤔

packages/dnb-eufemia/src/extensions/forms/DataContext/Provider/__tests__/Provider.test.tsx
```diff
@@ -1213,6 +1214,232 @@ describe('DataContext.Provider', { retry: isCI ? 5 : 0 }, () => {
           'onSubmit',
         ])
       })
+
+      it('should not rerun onBlurValidator during submit when disabled', async () => {
```
langz:
```suggestion
      it('should not rerun onBlurValidator during submit when runOnSubmit is 'never'', async () => {
```

packages/dnb-eufemia/src/extensions/forms/DataContext/Provider/__tests__/Provider.test.tsx
```diff
@@ -1213,6 +1214,232 @@ describe('DataContext.Provider', { retry: isCI ? 5 : 0 }, () => {
...
+
+        await userEvent.click(button)
+
+        await waitFor(() => {
+          expect(onSubmit).toHaveBeenCalledTimes(1)
+        })
+        expect(onBlurValidator).toHaveBeenCalledTimes(1)
+      })
+
+      it('should not rerun onChangeValidator during submit when disabled', async () => {
```
langz:
```suggestion
      it('should not rerun onChangeValidator during submit when runOnSubmit is 'never'', async () => {
```

packages/dnb-eufemia/src/extensions/forms/DataContext/Provider/__tests__/Provider.test.tsx
```diff
@@ -1213,6 +1214,232 @@ describe('DataContext.Provider', { retry: isCI ? 5 : 0 }, () => {
...
+
+        await userEvent.click(button)
+
+        await waitFor(() => {
+          expect(onSubmit).toHaveBeenCalledTimes(1)
+        })
+        expect(onChangeValidator).toHaveBeenCalledTimes(1)
+      })
+
+      it('should still run remaining validators during submit when one is disabled', async () => {
```
langz:
```suggestion
      it('should still run remaining validators during submit when one runOnSubmit is 'never'', async () => {
```

packages/dnb-eufemia/src/extensions/forms/DataContext/Provider/__tests__/Provider.test.tsx
```diff
@@ -1213,6 +1214,232 @@ describe('DataContext.Provider', { retry: isCI ? 5 : 0 }, () => {
...
+        await userEvent.click(button)
+
+        await waitFor(() => {
+          expect(onSubmit).toHaveBeenCalledTimes(1)
+        })
+        expect(onChangeValidator).toHaveBeenCalledTimes(1)
+        expect(onBlurValidator).toHaveBeenCalledTimes(0)
+      })
+
+      it('should skip custom validators during submit when disabled', async () => {
```
langz:
```suggestion
      it('should skip custom validators during submit when runOnSubmit is 'never'', async () => {
```

langz (approved):

packages/dnb-eufemia/src/extensions/forms/hooks/useFieldProps.ts
```diff
@@ -6,10 +6,11 @@ import {
   useReducer,
   useRef,
 } from 'react'
+import isEqual from 'lodash.isequal'
```
langz: Great 👍 
I think we need to move this to be a `dependency`, as of right now I think it's only a `devDependencies`

packages/dnb-eufemia/src/types/lodash.isequal.d.ts
langz, line 1: I guess we can remove this custom type declaration if we change lodash.isequal to being a`dependency`, as of right now I think it's only a `devDependencies`

langz (approved):

langz:
Looks like it should be rebased with latest of main before merging 🚀 🙏

packages/dnb-eufemia/src/types/lodash.isequal.d.ts
tujoworker: I hope we can avoid comparing objects deep.

packages/dnb-eufemia/src/extensions/forms/hooks/useFieldProps.ts
tujoworker: I hope we can avoid comparing objects deep. So no lodash atm.

packages/dnb-eufemia/src/extensions/forms/hooks/useFieldProps.ts
tujoworker: I hope we can avoid comparing objects deep. For now we compare arrays and objects by reference. The docs are updated on that limitation.

tujoworker: But it should be ok for most cases. And it rather will run the validator several times, if not, which is not really an critical issue either.

langz (approved):

PR #8535, opened 2026-06-11
design(DatePicker): remove picker arrow

tujoworker:
The DatePicker popup should appear without the Popover arrow while keeping its existing alignment behavior and spacing from the trigger.

This hides the Popover arrow for DatePicker and keeps the trigger offset so the picker still opens with the intended visual separation. A regression test verifies that the arrow is not rendered when the picker opens.

Preview: https://chore-date-picker-hide-arrow.eufemia-e25.pages.dev/uilib/components/date-picker/demos


langz (approved):

langz:
I would guess we would need to update screenshot tests 🤔

joakbjerk (approved):

langz (approved):

PR #8538, opened 2026-06-11
fix(MultiSelection): add accessible label to search input

langz:
Deploy preview https://fix-multi-selection-search-a.eufemia-e25.pages.dev/uilib/extensions/forms/base-fields/MultiSelection/demos#with-search
Latest release: https://eufemia.dnb.no/uilib/extensions/forms/base-fields/MultiSelection/demos/#with-search

- Add proper accessible label to MultiSelectionSearch input using labelSrOnly
- Add axe accessibility test for Filter.MultiSelection

The search input was missing an accessible label which violated WCAG 4.1.2. Now uses the placeholder text as a screen-reader-only label.


joakbjerk (approved):

langz:
I'll try to use type="search" instead.

tujoworker:
As tested with VoiceOver, we do not need this extra label.

PR #8539, opened 2026-06-11
chore(Filter): add axe accessibility tests for Filter components

langz:


joakbjerk (approved):
🪓

packages/dnb-eufemia/src/components/filter/__tests__/FilterContent.test.tsx
```diff
@@ -346,3 +347,29 @@ describe('Filter.Content aria-live', () => {
...
+      </FilterRoot>
+    )
+
+    expect(await axeComponent(container)).toHaveNoViolations()
+  })
+
+  it('has no axe violations when loading', async () => {
+    const { container } = render(
+      <FilterRoot id="content-loading-a11y-test" resultLoading>
+        <FilterContent connectedTo="content-loading-a11y-test">
```
tujoworker: When its possible to inherit the context, we do not need an id. For performance reasons we may not want it in tests.

langz: Great observation 👍 
This is only set because of the existing pattern in the other tests 🤝

tujoworker (approved):

langz:
Thanks for the improvements @tujoworker ❤️

PR #8540, opened 2026-06-11
chore(Filter.Search): add aria-label to search submit button in manual mode

langz:
When FilterSearch uses submitBehavior="manual", the Input renders a submit button with a loupe icon but no accessible name. Screen readers cannot identify the button's purpose.

Add a searchSubmitLabel translation key to nb-NO, en-GB, and sv-SE locales, and pass it as submitButtonTitle to the Input component. Consumers can still override via the submitButtonTitle prop.

Deploy preview: https://fix-filter-search-submit-but.eufemia-e25.pages.dev/uilib/components/filter#search-only
Main preview: https://main.eufemia-e25.pages.dev/uilib/components/filter#search-only


joakbjerk (approved):

langz:
Rather fixing this in the Input component: fix(Input): change submitButtonTitle to Search/Søk

PR #8541, opened 2026-06-11
fix(Tabs): prevent vertical wheel scroll trapping on tab strip

langz:
Chromium translates vertical wheel events into horizontal scrolling on the tab strip, trapping page scroll. Add a wheel listener that detects vertical-dominant gestures, calls preventDefault, and forwards the scroll to the nearest scrollable ancestor.

Intentional horizontal scrolling (deltaX-dominant or Shift+wheel) is left untouched. The handler is a no-op when the strip has no overflow.

Closes #6827

### DIFF
--- a/packages/dnb-eufemia/src/components/tabs/Tabs.tsx
+++ b/packages/dnb-eufemia/src/components/tabs/Tabs.tsx
@@ -902,10 +902,64 @@ function TabsComponent(ownProps: TabsProps) {
       resizeObserver.observe(tabsRef.current)
     }
 
+    // Prevent vertical wheel events from being translated into horizontal
+    // scrolling of the tab strip (Chromium behavior). Instead, forward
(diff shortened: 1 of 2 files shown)


packages/dnb-eufemia/src/components/tabs/Tabs.tsx
```diff
@@ -902,10 +902,64 @@ function TabsComponent(ownProps: TabsProps) {
...
+      })
+    }
+
     return () => {
       isMounted = false
       sharedStateRef.current = null
       resizeObserver?.disconnect()
+      if (tablistEl) {
+        tablistEl.removeEventListener('wheel', handleWheel)
+      }
```
langz, line 962: Wow, that was a lot of code. Can't believe we need all of this, hehe.

PR #8548, opened 2026-06-11
refactor(Input): show search icon instead of submit button for `type="search"`

tujoworker:
From:
Screenshot 2026-06-11 at 21 29 04

To:
Screenshot 2026-06-11 at 21 29 16


langz (approved):
Great work, nice change 🚀 

But it also sort of feels breaking to me, soft breaking change? 😅 
Especially since we add a new prop for this, `showSubmitButton`.

It feels breaking since existing consumer code, left untouched, produces a different result, and recovering the old result requires a code change, using `showSubmitButton`. 

If we would not support `showSubmitButton` I feel it would be different, as then we say that there's no way we want any apps to have a search/submit button. So then we would ONLY support the loupe-by-default look.

tujoworker:
I would not call it breaking. It changes user interaction/behavior. But that is something we should be able to do.

PR #8522, opened 2026-06-10
chore(eufemia-starter): downgrade Vite 8 to 6 for StackBlitz compatibility

langz:
Motivation: Today I've seen several reported stackblitz's that fail like [this one](https://stackblitz.com/edit/u6dlg7ok-wyemna4c?file=src%2FApp.tsx):

```
> eufemia-starter@1.0.0 dev
> vite

[rolldown] Downloading @rolldown/binding-wasm32-wasi@1.0.2 on WebContainer...
Packages: +7
+++++++
Progress: resolved 7, reused 0, downloaded 7, added 7, done

dependencies:
+ @rolldown/binding-wasm32-wasi 1.0.2 (1.1.0 is available)

Done in 6.3s
(node:12) ExperimentalWarning: WASI is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
(node:12) ExperimentalWarning: WASI is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
worker (tid = 43) sent an error! Invalid atomic access index
RangeError: Invalid atomic access index
    at Object.storeSizeTypeValue (/tmp/rolldown-1.0.2/node_modules/.pnpm/@emnapi+core@1.10.0/node_modules/@emnapi/core/dist/emnapi-core.cjs.js:1218:29)
    at Object.pushQueue (/tmp/rolldown-1.0.2/node_modules/.pnpm/@emnapi+core@1.10.0/node_modules/@emnapi/core/dist/emnapi-core.cjs.js:990:28)
    at eval (/tmp/rolldown-1.0.2/node_modules/.pnpm/@emnapi+core@1.10.0/node_modules/@emnapi/core/dist/emnapi-core.cjs.js:1040:36)
    at Object.execute (/tmp/rolldown-1.0.2/node_modules/.pnpm/@emnapi+core@1.10.0/node_modules/@emnapi/core/dist/emnapi-core.cjs.js:1094:36)
    at Object.push (/tmp/rolldown-1.0.2/node_modules/.pnpm/@emnapi+core@1.10.0/node_modules/@emnapi/core/dist/emnapi-core.cjs.js:1029:33)
    at napi_call_threadsafe_function (/tmp/rolldown-1.0.2/node_modules/.pnpm/@emnapi+core@1.10.0/node_modules/@emnapi/core/dist/emnapi-core.cjs.js:1612:31)

Node.js v22.22.0
```

image

[...]

PR #8549, opened 2026-06-12
chore(eufemia-starter): keep eufemia-starter on Vite 8 tooling

tujoworker:
Updates the starter template tooling so it stays aligned with the Vite version used across the workspace.

Verified with:
- `yarn workspace eufemia-starter build`

Note: the build passes with existing Vite 8 warnings about `optimizeDeps.esbuildOptions` deprecation and chunk size.

### DIFF
--- a/packages/eufemia-starter/package.json
+++ b/packages/eufemia-starter/package.json
@@ -31,7 +31,7 @@
     "@eslint/js": "10.0.1",
     "@types/react": "19.2.14",
     "@types/react-dom": "19.2.3",
-    "@vitejs/plugin-react": "4.7.0",
+    "@vitejs/plugin-react": "6.0.2",
     "eslint": "10.4.1",
     "eslint-plugin-react-hooks": "7.1.1",
     "eslint-plugin-react-refresh": "0.5.2",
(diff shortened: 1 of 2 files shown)


langz (approved):
Fine by me, but will not this fail in Chrome, as we discussed earlier?
Perhaps it's fine and the tradeoff of not using same vite version as portal is worse 🤔

tujoworker:
> Fine by me, but will not this fail in Chrome, as we discussed earlier? 

I tested it, and it does not fail in Chrome. Can you test it as well?

langz:
> > Fine by me, but will not this fail in Chrome, as we discussed earlier?
> 
> I tested it, and it does not fail in Chrome. Can you test it as well?

Does not work here, the error is now different than previosly as reported in https://github.com/dnbexperience/eufemia/pull/8522.
Now I get the error:

image

Tested both in production and in this PR, and I get the same error as above.
When I used main it works in this Chrome (Version 149.0.7827.103 (Official Build) (x86_64))

tujoworker:
But that's again on your old mac, right? I think there must be "another" issue, maybe related to CPU architecture? Works fine here.

langz:
> But that's again on your old mac, right? I think there must be "another" issue, maybe related to CPU architecture? Works fine here.

Yes, this is on my "old" mac, the report from this https://github.com/dnbexperience/eufemia/pull/8522 was on my "new mac". I'm fine with trying out this approach suggested in this PR. I don't know how many of our users/reporters have this or that cpu 😂  Hopefully it's only "my machines" who has not worked 🙏

PR #8550, opened 2026-06-12
fix(Form.useTranslation): cache in sync

tujoworker:
Caches the merged forms translation base without letting same-locale translation prop changes go stale.

This keeps large forms from deep-cloning locale data for every field while preserving the expected behavior when a provider receives a new translations object.

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/hooks/__tests__/useTranslation.test.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/hooks/__tests__/useTranslation.test.tsx
@@ -251,6 +251,58 @@ describe('Form.useTranslation', () => {
         )
       )
     })
+
+    it('should update custom translation when translations object changes', () => {
+      type CustomTranslation = {
+        Custom: {
+          translation: string
+        }
+      }
+
+      const initialTranslations = {
+        'nb-NO': {
+          Custom: {
+            translation: 'Initial translation with a {myKey}',
+          },
+        },
+      }
+      const updatedTranslations = {
+        'nb-NO': {
+          Custom: {
+            translation: 'Updated translation with a {myKey}',
+          },
+        },
+      }
+
+      const MockComponent = () => {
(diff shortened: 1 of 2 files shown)


langz (approved):
Looks good to me 💯 
Will this work with nested providers? Or perhaps we do not need it, could be complex?

```
const nb = { 'nb-NO': { Custom: { t: 'A' } } }
const en = { 'en-GB': { Custom: { t: 'B' } } }

// <Field/> = anything that calls Form.useTranslation()
render(
  <Provider locale="nb-NO" translations={nb}>
    <Field />                            {/* no cache exists → caching = (nb-NO) */}
    <Provider locale="en-GB" translations={en}>
      <Field />                          {/* no cache exists → caching = (en-GB) */}
    </Provider>
    <Field />                            {/* no chache exists→ caching = (nb-NO) again */}
  </Provider>
)
```

Perhaps it could be improved by using a WeakMap? 💭

tujoworker:
> Looks good to me 💯 Will this work with nested providers? Or perhaps we do not need it, could be complex?
> 
> ```
> const nb = { 'nb-NO': { Custom: { t: 'A' } } }
> const en = { 'en-GB': { Custom: { t: 'B' } } }
> 
> // <Field/> = anything that calls Form.useTranslation()
> render(
>   <Provider locale="nb-NO" translations={nb}>
>     <Field />                            {/* no cache exists → caching = (nb-NO) */}
>     <Provider locale="en-GB" translations={en}>
>       <Field />                          {/* no cache exists → caching = (en-GB) */}
>     </Provider>
>     <Field />                            {/* no chache exists→ caching = (nb-NO) again */}
>   </Provider>
> )
> ```
> 
> Perhaps it could be improved by using a WeakMap? 💭



The cache is single-entry, so an inner provider with another locale or translation object can overwrite the cached value. When a later outer provider field renders, the locale/object identity check fails and the base is recomputed. That may cost an extra recompute in mixed nested-provider trees, but it should not return stale or cross-provider translations.

PR #8552, opened 2026-06-12
fix(Forms): improve shared attachment render cascade

tujoworker:
Fixes the Form.Appearance demo slowdown caused by reactive shared attachment updates after #8009.

The Provider does not need to subscribe to the id-based attachment store it owns. Keeping that path imperative avoids repeated synchronous re-renders while still notifying external subscribers when validation state changes.

Closes #8342


langz (approved):
Great finding/work ❤️🙌

langz:
I also did a quick verification, and as far as I can see it works as expected 🐎

langz (approved):

PR #8554, opened 2026-06-12
feat(MultiSelection): highlight filtered options to be easier to scan

tujoworker:
MultiSelection filtering narrows the option list, but users still have to visually parse each remaining item to find the matching text. Highlighting the matched part makes the search results quicker to scan and aligns the experience with Filter and Autocomplete.

Preview: https://chore-multi-selection-search.eufemia-e25.pages.dev/uilib/extensions/forms/base-fields/MultiSelection/demos#with-search


langz (approved):
Nice, looks great 🥇 
Nothing we need to do now, but in the future it would be cool if we would not duplicate the highlight logic everywhere we have highlighting (Filter, Autocomplete, MultiSelection=, but rather make something shared. Perhaps we already have it?

joakbjerk (approved):

tujoworker:
yes, thought the same. Even Autocomplete is way more complex. They have some code in common.

PR #8555, opened 2026-06-12
fix(MultiSelection): keep search input literal

tujoworker:
MultiSelection uses its search field for filtering exact option text, so browser text assistance can get in the way by changing or annotating what the user typed. Matching the Filter search input keeps that behavior predictable across searchable controls and keeps the visual baselines aligned with the updated input state.

Preview: https://chore-multi-selection-search-oqp9.eufemia-e25.pages.dev/uilib/extensions/forms/base-fields/MultiSelection/demos#with-search


packages/dnb-eufemia/src/extensions/forms/Field/MultiSelection/MultiSelectionSearch.tsx
```diff
@@ -32,8 +32,11 @@ export function MultiSelectionSearch({
         onChange={(e) => onSearchChange(e.value)}
         disabled={disabled}
         stretch
-        showClearButton={value.length > 0}
+        showClearButton
```
langz, line 35: Clear button always visible, fine fine, we perhaps have it like that in the Filter component as well now 🤞

langz (approved):

joakbjerk (approved):

packages/dnb-eufemia/src/extensions/forms/Field/MultiSelection/MultiSelectionSearch.tsx
tujoworker: We do. The thing is; value.length > 0 we do anyway in the component. So this was redundant from before.

PR #8562, opened 2026-06-12
fix(Filter): let search input own manual icon

### LINKED ISSUE #8560: Filter search only example is a bit strange with multiple search icons.
As discussed with @tujoworker, the following Filter search only example is a bit strange with multiple search icons.
https://main.eufemia-e25.pages.dev/uilib/components/filter/demos#search-only

image

Not sure how this will look and work with async search/data

_Originally posted by @langz in https://github.com/dnbexperience/eufemia/issues/8440#issuecomment-4680882019_

tujoworker:
This keeps `Filter.Search` aligned with the `Input` search behavior by letting `type="search"` provide the manual submit icon instead of forcing a separate loupe icon.

This is fine because async cannot be used with `submitBehavior="manual"`, so the manual submit button can rely on the default search icon while the loading indicator remains scoped to typing/loading state.

Closes #8560

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/components/filter/Examples.tsx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/components/filter/Examples.tsx
@@ -2,7 +2,7 @@
  * Filter component examples for the portal docs.
  */
 
-import { useState } from 'react'
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react'
 import ComponentBox from '../../../../shared/tags/ComponentBox'
 import {
   Button,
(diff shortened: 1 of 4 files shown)


langz:
Is it correct that _async can't be used with manual submit_?
I would guess it could be fired an async request when pressing the manual search button in [this example](https://chore-filter-search-manual-i.eufemia-e25.pages.dev/uilib/components/filter/demos#search-only)?

joakbjerk (approved):

tujoworker:
> Is it correct that _async can't be used with manual submit_? I would guess it could be fired an async request when pressing the manual search button in [this example](https://chore-filter-search-manual-i.eufemia-e25.pages.dev/uilib/components/filter/demos#search-only)?

Yes, that is true. But then its wrong to show the indicator on the left side. I would then rather assume we show a skeleton. We can adjust the example we have.

langz (approved):

langz:
> > Is it correct that _async can't be used with manual submit_? I would guess it could be fired an async request when pressing the manual search button in [this example](https://chore-filter-search-manual-i.eufemia-e25.pages.dev/uilib/components/filter/demos#search-only)?
> 
> 
> 
> Yes, that is true. But then its wrong to show the indicator on the left side. I would then rather assume we show a skeleton. We can adjust the example we have.

Thanks 🙏

PR #8564, opened 2026-06-12
chore: remove empty theme-extensions and theme-forms files [Draft]

joakbjerk:
Testing out if its easily done, and breaking to much 😅


packages/dnb-design-system-portal/src/docs/uilib/extensions.mdx
```diff
@@ -25,14 +25,6 @@ There are several great reasons behind opting for having extensions separated:
...
 ```
 
-### Import extensions styles
-
-The styles for extensions are not a part of the default styles, so you have to import them explicitly.
-
-```js
-import '@dnb/eufemia/style/themes/ui/extensions'
-```
-
```
langz, line 35: Let's discuss if this is "safe" to remove now or if we should wait until v12 🤔 
Could this PR be breaking for any of our consumers who do `import '@dnb/eufemia/style/themes/ui/extensions'`? If so, I guess we could still keep the "export" (even though it will not to anything) with a TODO comment for remember to removing it in v12.

packages/dnb-eufemia/scripts/prebuild/tasks/__tests__/makeMainStyle.test.ts
```diff
@@ -75,11 +75,9 @@ if (isCI) {
       expect(global.components[0]).toContain(
         'ATTENTION: This file is auto generated'
       )
-      expect(global.components[0]).toContain('@charset "UTF-8";')
       expect(global.components[0]).not.toContain(';--')
       expect(global.components[0]).not.toContain('}:root{--')
 
-      expect(global.components[1]).toContain('@charset "UTF-8";')
```
langz, line 82: Hmm, looks a bit suspicious 🤔

packages/dnb-design-system-portal/src/docs/uilib/extensions.mdx
joakbjerk: Yeah, that might be a bit breaking 😄

packages/dnb-eufemia/scripts/prebuild/tasks/__tests__/makeMainStyle.test.ts
joakbjerk: it was to circumvent importing empty files, since there weren't any UTF-8 characters in there when empty, but probably not the best way to solve this no 🤔

joakbjerk:
Closing this for now — deferring to **v12**.

On re-checking against current `main`, this isn't a "remove empty files" cleanup: the affected theme files aren't empty (they `@use` barrels that pull in real styles), and the change is **breaking**.

Verified trace:
- The main style entry (`src/style/index.ts` / `index.scss`) imports `themes/ui/ui-theme-components.scss`.
- `dnb-ui-components.scss` contains **no** forms styles on its own.
- Forms field styles reach the main `@dnb/eufemia/style` bundle **only** via `@use './ui-theme-forms.scss'` in `ui-theme-components.scss` — which this PR removes.

Net effect:
- **Forms field styles** would be dropped from the default `@dnb/eufemia/style` import (they'd remain only in `@dnb/eufemia/style/extensions`). 🔴 breaking regression for consumers of the main style bundle.
- `payment-card` and `vipps-wallet-button` are dual-delivered (extensions bundle + each component's own `style` entry), so those are unaffected.

Also, as @langz noted, removing the `.../style/themes/ui/extensions` (and `.../forms`) import paths is breaking for any consumer importing them directly.

This should be revisited as part of **v12**, with an explicit decision on whether forms belongs only in the extensions bundle, and a deprecation/no-op-export path for the removed entry points. Closing the draft to reopen as a v12-scoped task.

PR #8566, opened 2026-06-13
fix(Tabs): restore page scrolling over tablist

tujoworker:
Scrolling the page while hovering the Tabs tablist could be trapped by the horizontal tab scroller in Chromium-based browsers, and Firefox could stop page scrolling in the same area.

This keeps horizontal tab scrolling available while letting vertical wheel intent continue to scroll the page, so users do not get stuck when the pointer is over the tablist.

Closes #6827


langz (approved):
Nice, a bit simpler and easier approach than the few ones I've explored.

PR #8570, opened 2026-06-13
feat(Forms): reduce rerenders for path-based fields

### LINKED ISSUE #5453: Optimize Form Rendering: Granular Updates for Eufemia Fields
## 🚀 Feature Proposal

Implement granular form state updates in Eufemia forms so that only the field/component whose value changes is re-rendered, rather than all components in the current form step. This would improve performance and responsiveness, especially in complex forms.

## Motivation

Currently, when using the `path` attribute for fields (e.g. `<Field.String path="/lastName" />`), every component in the form step re-renders on each keystroke. This is due to the global update mechanism in [`Provider.tsx`](src/extensions/forms/DataContext/Provider/Provider.tsx), where `setData` and `updateDataValue` trigger a `forceUpdate()` for all consumers. This leads to noticeable lag and degraded user experience in large forms.

## Example

**Current behavior:**
```tsx
<Field.String path="/lastName" />
<Field.String path="/firstName" />
// Typing in "lastName" causes both fields to re-render on every keypress.
```

**Proposed behavior:**
```tsx
<Field.String path="/lastName" />
<Field.String path="/firstName" />
// Typing in "lastName" only re-renders the "lastName" field/component.
```

## Pitch

This feature belongs in "@dnb/eufemia" because it directly improves the performance and scalability of forms built with the design system. By reducing unnecessary re-renders, it enhances user experience and makes Eufemia more suitable for complex, enterprise-grade applications. The change would be implemented in the form context provider, aligning with Eufemia’s commitment to efficient and accessible UI components.

tujoworker:
Motivation: https://dnb.enterprise.slack.com/archives/CMXABCHEY/p1756195234567449

Update Form.Handler data propagation so form fields and derived values subscribe to the data paths they depend on instead of rerendering from unrelated form data changes.

This improves responsiveness in larger forms while keeping the public Forms API and field behavior unchanged.

Also updates Forms internals that read secondary data paths, such as selection data sources, currency paths, visibility dependencies, Iterate helpers, and custom value/field hooks, so they stay reactive with the new path-scoped update model.

Closes #5453

After:

https://github.com/user-attachments/assets/130e1e9a-9756-4822-b489-9ba7255b0225

Before:

https://github.com/user-attachments/assets/637d1fc0-e326-4726-846d-a2850381cbf2


packages/dnb-eufemia/src/extensions/forms/DataContext/At/At.tsx
```diff
@@ -68,17 +192,33 @@ function At(props: DataContextAtProps) {
...
+  if (path === '/') {
+    return data
+  }
+
+  return data && pointer.has(data, path)
+    ? pointer.get(data, path)
+    : undefined
+}
+
+function makePath(path: Path, changePath: Path): Path {
```
langz: As we already have a `makePath` in `usePath`, perhaps we could call this `joinPath`,  `concatPath` or `appendPath`?

packages/dnb-eufemia/src/extensions/forms/DataContext/At/At.tsx
```diff
@@ -68,17 +192,33 @@ function At(props: DataContextAtProps) {
...
-        handlePathChange,
-      }}
-    >
-      {children}
-    </Context>
-  )
+  return <Context value={scopedDataContext}>{children}</Context>
+}
+
+function getValue(data: unknown, path: Path) {
```
langz:
```suggestion
function getValueByPath(data: unknown, path: Path) {
```

packages/dnb-eufemia/src/extensions/forms/DataContext/At/At.tsx
langz:
```suggestion
function joinPath(path: Path, changePath: Path): Path {
```

langz (approved):
🐎

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/DataContext/Context.mdx
```diff
@@ -18,4 +18,4 @@ breadcrumb:
 
 ## Description
 
-The main context for [DataContext.Provider](/uilib/extensions/forms/DataContext/Provider) which the [Field](/uilib/extensions/forms/base-fields/) and [Value](/uilib/extensions/forms/Value/) components connect to (optional) for sources and callbacks when it is present. It can be used for creating custom components in similar ways.
+The main context for [DataContext.Provider](/uilib/extensions/forms/DataContext/Provider) which the [Field](/uilib/extensions/forms/base-fields/) and [Value](/uilib/extensions/forms/Value/) components connect to (optional) for sources and callbacks when it is present. For custom field and value components, prefer [useFieldProps](/uilib/extensions/forms/create-component/useFieldProps/), [useValueProps](/uilib/extensions/forms/create-component/useValueProps/) or [Form.useData](/uilib/extensions/forms/Form/useData/) instead of consuming `DataContext.Context` directly. Direct reads from `data` or `internalDataRef.current` do not subscribe the component to data updates.
```
langz, line 21: _Direct reads from `data` or `internalDataRef.current` do not subscribe the component to data updates._, is that a change in this PR? Or has it always been like this?

If it's a change, we could consider making this change a bit more visible, like displayed in the release log, etc.

tujoworker: It should be fine, because we did not document the usage of `data` or `internalDataRef`.

PR #8573, opened 2026-06-15
fix(Forms): show Wizard required errors during async validation

### LINKED ISSUE #6048: Async `onChangeValidator` in Wizard required issue
## 🐛 Bug Report

Handling required fields in a wizard is not behaving consistently.

In the first video, I get the correct required error message when pressing the next button after not entering any data into either fields. So this works as expected:

https://github.com/user-attachments/assets/2e066aad-c65f-4dff-bbcb-1ea68a278173

In the second video, when I enter any data into the non-required field with async validator, then I do not get the required error message for the 1st field. It's like if there's any active async validators it ignores/forget about the required field:

https://github.com/user-attachments/assets/7bcdddf1-782f-49a2-9a97-06fc09c17331

I think I would expect to get the required error message, and not been able to proceed with the form, or? 🤔

tujoworker:
Required fields in a Wizard can stay hidden when another field's async validator is pending, allowing users to leave the step without correcting the missing required value.

This keeps Wizard fallback navigation aligned with the form error registry so pending async validation does not bypass known errors.

Fixes #6048

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Wizard/Container/WizardContainer.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Wizard/Container/WizardContainer.tsx
@@ -133,6 +133,7 @@ function WizardContainer(props: WizardContainerProps) {
     setShowAllErrors,
     setSubmitState,
     hasFieldState,
+    hasErrors,
   } = dataContext
 
   const id = useId(idProp)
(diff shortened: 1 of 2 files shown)


langz (approved):

langz:
Works great :)

PR #8579, opened 2026-06-15
feat(DatePicker): add `onlyMonth="without-label"` variant to hide month label

langz:
Deploy preview: https://feat-date-picker-hide-month.eufemia-e25.pages.dev/uilib/components/date-picker#show-days-in-a-specific-month-without-month-label

Not sure if we actually need a demo/example for this? 🤔

### DIFF
--- a/packages/dnb-eufemia/src/components/date-picker/DatePicker.tsx
+++ b/packages/dnb-eufemia/src/components/date-picker/DatePicker.tsx
@@ -152,9 +152,9 @@ export type DatePickerProps = {
    */
   hideDays?: boolean
   /**
-   * If set to `true`, the calendar only displays days belonging to the currently displayed month, and month navigation via buttons and keyboard is disabled. The displayed month is determined by the `month` or `startMonth` prop, and ultimately defaults to the current month. Defaults to `false`.
+   * If set to `true`, the calendar only displays days belonging to the currently displayed month, and month navigation via buttons and keyboard is disabled. The displayed month is determined by the `month` or `startMonth` prop, and ultimately defaults to the current month. Use `'without-label'` to also hide the month label. Defaults to `false`.
    */
-  onlyMonth?: boolean
+  onlyMonth?: boolean | 'without-label'
   /**
    * Use `true` to only show the last week in the current month if it needs to be shown. The result is that mainly five (5) weeks (rows) will be shown instead of six (6). Defaults to `false`.
    */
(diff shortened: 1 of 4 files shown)


kimroen:
> but remains accessible to screen readers

Maybe I misunderstood the use case from Slack, but it seemed to me like what's actually desired here is a kind of "day of month"-picker, so having "February" present here at all seems like something that could be confusing (since it's just a hack to get a calendar with 28 days, which itself might cause issues in a leap year?).

Maybe this is something that should be supported explicitly instead? I have seen other solutions to this where "last day of the month" is available as an option too, for example.

Maybe a calendar isn't the right choice at all?

langz:
> > but remains accessible to screen readers
> 
> Maybe I misunderstood the use case from Slack, but it seemed to me like what's actually desired here is a kind of "day of month"-picker, so having "February" present here at all seems like something that could be confusing (since it's just a hack to get a calendar with 28 days, which itself might cause issues in a leap year?).
> 
> Maybe this is something that should be supported explicitly instead? I have seen other solutions to this where "last day of the month" is available as an option too, for example.
> 
> Maybe a calendar isn't the right choice at all?

Thanks @kimroen, it definitely does not make sense to still have the month label readable for screen readers.
It's been adjusted now.

Yeah, I do also think a better approach would be to provide a way for users to select this kind of "date" (installment date), and have a user experience tailored to that need. We are indeed a design system for a bank 🏦 😂 

Field.InstallmentDate, or such. 

Great feedback @kimroen  ❤️

packages/dnb-design-system-portal/src/docs/uilib/components/date-picker/demos.mdx
```diff
@@ -64,6 +65,12 @@ import enUS from '@dnb/eufemia/shared/locales/en-US'
   <DatePickerMonthOnly />
 </VisibleWhenNotVisualTest>
 
+### Show days in a specific month without month label
+
+<VisibleWhenNotVisualTest>
```
tujoworker: This makes not sense to use I would argue, unless the picker is open by default.

tujoworker (approved):
I don't think we need an extra example for that. It's very edge case. We could rather add the new prop with an additional button to the other existing example. Then devs will be able to remove it when they see they need the month label.

tujoworker:
Since hiding the label only makes sense together with onlyMonth, maybe we can expand onlyMonth from a boolean to a variant prop, e.g. onlyMonth="without-label". That keeps the API tied to the relevant mode without adding a prop that only works in combination with another.

`onlyMonth="without-label"`

packages/dnb-design-system-portal/src/docs/uilib/components/date-picker/demos.mdx
langz: Removed :)

langz:
> I don't think we need an extra example for that. It's very edge case. We could rather add the new prop with an additional button to the other existing example. Then devs will be able to remove it when they see they need the month label.

Removed :)

langz:
> Since hiding the label only makes sense together with onlyMonth, maybe we can expand onlyMonth from a boolean to a variant prop, e.g. onlyMonth="without-label". That keeps the API tied to the relevant mode without adding a prop that only works in combination with another.
> 
> `onlyMonth="without-label"`

I've adjusted it to be `onlyMonth="without-label"` now :)

PR #8600, opened 2026-06-16
feat(Forms): add variant='tiles' to Field.InstallmentDay [DRAFT]

langz:
Deploy preview: https://feat-field-installment-date-v2n5.eufemia-e25.pages.dev/uilib/extensions/forms/feature-fields/InstallmentDay/demos#tiles-variant


tujoworker:
I think this variant is a bit challenging from an accessibility perspective.

Keyboard users would need to tab through every option, which can become quite heavy. Screen reader users may also need to go through the full list before getting a clear overview of what is currently selected.

It also takes quite a lot of space in forms, especially on smaller screens.

PR #8604, opened 2026-06-17
fix(Stat.Currency): support custom currency codes

tujoworker:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1781683191563499
Preview: https://chore-stat-currency-code.eufemia-e25.pages.dev/uilib/components/stat/properties#statcurrency

Stat.Currency is expected to support ISO currency codes in the same way as the underlying number-format currency API, but explicit codes were being overwritten by the component default. This meant values could still render with the default NOK currency even when applications passed EUR, USD, or another code.

This makes the prop behavior match the intended API and keeps the generated docs table aligned so consumers can discover and use non-default currencies confidently.


snorrekim (approved):
I was debating if we should force `currency` to be `true` even if user says `false`, as `Stat.Percent` does not allow the user to turn it off either.  But we allow it in `NumberFormat.Currency`, and not in `NumberFormat.Percent`, so we should just keep the same behaviour. And it's an edge case that doesn't prevent the intended use case.

Also, those screenshots are telling 😀 We really did intent currency to work like this.

langz (approved):

PR #8607, opened 2026-06-17
feat(Forms): add `useDataWithoutSubscription` hook for non-subscribing data access

tujoworker:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1781704782737049?thread_ts=1756195234.567449&cid=CMXABCHEY
Preview: https://chore-use-data-without-subsc.eufemia-e25.pages.dev/uilib/extensions/forms/Form/useData

Adds a Forms data hook for custom fields and integrations that need imperative access to form data helpers without subscribing components to data updates.

This reduces avoidable rerenders while keeping the familiar `Form.useData` helper API available for non-rendering reads and updates. The docs clarify when to use the non-subscribing variant so render-dependent UI still uses reactive data access.


packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/useData/info.mdx
```diff
@@ -52,6 +55,30 @@ render(
 - `filterData` will filter the data based on your own logic.
 - `reduceToVisibleFields` will reduce the given data set to only contain the visible fields (mounted fields).
 
+### Without subscription
+
+`Form.useData` subscribes the component to form data changes. This is useful when rendered output depends on the data.
+
+For custom fields and other internal integrations that only need imperative access to the same helpers, use `Form.useDataWithoutSubscription`. It returns the same API and accepts the same optional `id` and `initialData` arguments as `Form.useData`, but does not subscribe the component to data updates.
```
langz: I guess we could also have done:
`Form.useData(id, initialData, { subscribe: false })`

Or called the hook `useDataLatest` or `useDataRef` (just suggestions/alternatives)

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/useData/info.mdx
```diff
@@ -52,6 +55,30 @@ render(
...
+  const handleClick = () => {
+    const value = getValue('/foo')
+    update('/bar', value)
+  }
+
+  return <button onClick={handleClick}>Update</button>
+}
+```
+
+The `data` property is the value from the current render. Use `getValue` inside callbacks when you need the latest value without subscribing. The component can still re-render for other reasons, such as parent renders or other context changes.
```
langz: Does it make sense to return `data`, as if I understand correctly `data` will be out of date after the form changes.
I guess a lot of users will likely read data as current, and could be a trap 🤔

langz:
```suggestion
The `data` property is the value from the current render. Use `getValue` inside callbacks when you need the latest value without subscribing. The component can still re-render for other reasons, such as parent renders or other context changes. `data` will be out of date after the form changes — use getValue for the current value).
```

langz (approved):

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/useData/info.mdx
tujoworker: True! `data` is removed now and we give instead `getData`.

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/useData/info.mdx
tujoworker: I considered `{ subscribe: false }`, but it becomes a different internal mode anyway because context-based `useData()` would otherwise still rerender through `useContext`.

I also considered `useDataRef`, but it does not return a React ref, so that felt misleading. `useDataWithoutSubscription` is longer, but it states the important behavior directly, especially now that it returns imperative helpers like `getData`/`getValue` and not `data`.

PR #8608, opened 2026-06-18
feat(Forms): add `Form.useDataValue('/path')` hook for path-scoped data value

tujoworker:
Rendering with `Form.useData` subscribes consumers to the whole form data object, while imperative reads do not create subscriptions. This adds a path-scoped data value hook so components inside `Form.Handler` can render one form value without rerendering on unrelated data or context updates.

The docs also clarify when to use the path-scoped hook, the non-subscribing data helper, and field/value hooks so custom integrations avoid stale direct context reads.

Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1781762369791639?thread_ts=1756195234.567449&cid=CMXABCHEY
Preview: https://chore-forms-data-value-subsc.eufemia-e25.pages.dev/uilib/extensions/forms/Form/useDataValue


langz (approved):
I see `Form.useDataWithoutSubscription` and `Form.useData` both accepts `id` to target a form outside its handler. But I guess it's intention that the `Form.useDataValue` does not support an `id`?

joakbjerk (approved):

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/getting-started.mdx
```diff
@@ -192,6 +192,8 @@ function MyForm() {
 }
 
 function MyComponent() {
+  const companyName = Form.useDataValue('/companyName')
```
langz, line 195: Doesn't look like `companyName` is used in this example 🤔

tujoworker:
> I see `Form.useDataWithoutSubscription` and `Form.useData` both accepts `id` to target a form outside its handler. But I guess it's intention that the `Form.useDataValue` does not support an `id`?

Yes, its for usage within fields.

PR #8613, opened 2026-06-18
fix: change line-height sizes to corresponding font-size and design

snorrekim:
- `--line-height-x-small`: 1.125rem -> 1.25rem
- Line-height variables have been refactored to correctly match font-size i.e. `--font-size-large` is now used with `--line-height-large` and not `--line-height-medium`.
  - `--line-height-medium`: 2rem -> 1.5rem
  - `--line-height-large`: 2.5rem -> 2rem
  - `--line-height-x-large`: 3.5rem -> 2.5rem


langz:
I guess we could now update this doc?
https://github.com/dnbexperience/eufemia/blob/3439a01db112507f29e1df81c9fa675232597806/packages/dnb-design-system-portal/src/docs/uilib/typography/line-height.mdx?plain=1#L26

As it still labels xx-large as "Same as x-large"

langz:
I see Avatar use `var(--line-height-large)`, but wasn't updated the way Dialog/Drawer were, I guess it's intended, or did we miss changing there?

langz (approved):
Image

snorrekim:
> I see Avatar use `var(--line-height-large)`, but wasn't updated the way Dialog/Drawer were, I guess it's intended, or did we miss changing there?

Line-height doesn't actually seem to affect anything in that code.

langz:
> > I see Avatar use `var(--line-height-large)`, but wasn't updated the way Dialog/Drawer were, I guess it's intended, or did we miss changing there?
> 
> Line-height doesn't actually seem to affect anything in that code.

Great, thanks for checking it out 👍

tujoworker (approved):

PR #8614, opened 2026-06-18
style: Sbanken typography values to be same as other brands

snorrekim:
Sbanken now shares the same base font-sizes and line-heights as the other brands.

- Affects all base font-size and line-height css variables: `--font-size-large`, `line-height-medium`, etc.
- Affects all typography elements: `Heading`, `P`, `Span`, `Lead`, etc.
- Basis font size is now `18px`, not `16px`. Affects all components.

**Deprecated css variables:**
- Deprecate `--sb-font-size-*`, use `--font-size-*` instead
- Deprecate `--sb-font-size-medium--plus`, use `--font-size-medium` instead
- Deprecate `--sb-line-height-*`, use `--line-height-*` instead

Based on the line-height change PR: https://github.com/dnbexperience/eufemia/pull/8613

PR #8615, opened 2026-06-18
fix(Heading): add `proseMaxWidth`

snorrekim:
- add `proseMaxWidth` to `<Heading>` same as `<P>` and `<H1>, <H2>...`
- can be controlled by `<Typography.Provider>` in the same way

Based on the Sbanken typography PR: https://github.com/dnbexperience/eufemia/pull/8614
and on the line-height change PR: https://github.com/dnbexperience/eufemia/pull/8613


langz (approved):
Looks good to me :)
We could perhaps add a unit test for the new prop, but I guess it's not too important.

langz:
I see that Typography.tsx lets `proseMaxWidth` win over an explicit `style`, but Heading.tsx lets an explicit `style` win over `proseMaxWidth`. Not sure if it's intentional for this to behave different or if we should "standardize" it.

tujoworker (approved):
We need an integration test as well. I will add one.

joakbjerk (approved):

snorrekim:
> I see that Typography.tsx lets `proseMaxWidth` win over an explicit `style`, but Heading.tsx lets an explicit `style` win over `proseMaxWidth`. Not sure if it's intentional for this to behave different or if we should "standardize" it.

That seems like an old bug, I'll fix it. And add a test.

langz:
> > I see that Typography.tsx lets `proseMaxWidth` win over an explicit `style`, but Heading.tsx lets an explicit `style` win over `proseMaxWidth`. Not sure if it's intentional for this to behave different or if we should "standardize" it.
> 
> 
> 
> That seems like an old bug, I'll fix it. And add a test.

Thanks a lot, can also be done in a follow-up PR when we get time. 😊

snorrekim:
> > > I see that Typography.tsx lets `proseMaxWidth` win over an explicit `style`, but Heading.tsx lets an explicit `style` win over `proseMaxWidth`. Not sure if it's intentional for this to behave different or if we should "standardize" it.
> > 
> > 
> > That seems like an old bug, I'll fix it. And add a test.
> 
> Thanks a lot, can also be done in a follow-up PR when we get time. 😊


I think I should just improve the use of Typography context in this PR as well. Create a `useTypography` of some kind that handles all that logic and just returns the correct props/attributes. For now, just the proseMaxWidth and the new responsive props, but later, in another PR we can add the rest of the typography stuff, like `size` or `weight`

langz:
> > > > I see that Typography.tsx lets `proseMaxWidth` win over an explicit `style`, but Heading.tsx lets an explicit `style` win over `proseMaxWidth`. Not sure if it's intentional for this to behave different or if we should "standardize" it.
> 
> > > 
> 
> > > 
> 
> > > That seems like an old bug, I'll fix it. And add a test.
> 
> > 
> 
> > Thanks a lot, can also be done in a follow-up PR when we get time. 😊
> 
> 
> 
> 
> 
> I think I should just improve the use of Typography context in this PR as well. Create a `useTypography` of some kind that handles all that logic and just returns the correct props/attributes. For now, just the proseMaxWidth and the new responsive props, but later, in another PR we can add the rest of the typography stuff, like `size` or `weight`

Sounds clever 😎

langz (approved):

PR #8616, opened 2026-06-18
fix(Stat.Currency): fix missing space between currency and suffix in flex layout

joakbjerk:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1781770298695429

### DIFF
--- a/packages/dnb-eufemia/src/components/stat/Amount.tsx
+++ b/packages/dnb-eufemia/src/components/stat/Amount.tsx
@@ -76,10 +76,6 @@ const renderAffix = (
     resolved = resolved()
   }
 
-  if (typeof resolved === 'string' && resolved.startsWith(' ')) {
-    resolved = `\u00A0${resolved.slice(1)}`
-  }
-
   return <span className={className}>{resolved as ReactNode}</span>
 }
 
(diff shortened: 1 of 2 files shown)


langz:
Can the same issue occur using `prefix`?

langz:
Doesn't really seem like the docs is true/correct on this one https://github.com/dnbexperience/eufemia/blob/3439a01db112507f29e1df81c9fa675232597806/packages/dnb-eufemia/src/components/number-format/NumberFormatDocs.ts#L49

As if I do as it says, there's no spacing between:

image


Could be a separate issue, not entireyl sure :)

tujoworker (approved):

tujoworker:
> Doesn't really seem like the docs is true/correct on this one

Yeah, maybe we need to evaluate this and actually add a space as long as not / is given as the first character.

joakbjerk:
> > Doesn't really seem like the docs is true/correct on this one
> 
> Yeah, maybe we need to evaluate this and actually add a space as long as not / is given as the first character.

Should this be fixed before merge @tujoworker ?

joakbjerk:
> Doesn't really seem like the docs is true/correct on this one
> 
> https://github.com/dnbexperience/eufemia/blob/3439a01db112507f29e1df81c9fa675232597806/packages/dnb-eufemia/src/components/number-format/NumberFormatDocs.ts#L49
> 
> As if I do as it says, there's no spacing between:
> image
> 
> Could be a separate issue, not entireyl sure :)

The current fix only allows for the user to add a space at the beginning of the prop value, based on a discussion we had yesterday, but we could also make it so the space is added automatically before we merge this? 🤷‍♂️

tujoworker:
> but we could also make it so the space is added automatically before we merge this? 🤷‍♂️

Yes, I think we rather should do that so we are aligned on how `NumberFormat` does handle this.

tujoworker:
The same should be applied for prefix as well.

PR #8617, opened 2026-06-19
fix(NumberFormat): preserve formatted part boundaries

tujoworker:
This improves how formatted number output is split into semantic parts for consumers that style the number, currency, and percent pieces separately.

The previous fallback parser had to infer boundaries from the final display string. That works for common spaced output, but it is fragile for compact notation where the compact suffix and currency can be adjacent or locale-dependent. The formatter pipeline now keeps Intl display parts available through an internal opt-in path, so `useNumberFormatWithParts` can use the same formatted output with more reliable part boundaries while regular formatter calls keep their existing public shape and cost profile.


tujoworker:
Seems not needed for now.

PR #8619, opened 2026-06-19
chore: replace lodash.isequal/isequalwith with node:util isDeepStrictEqual

langz:
Remove deprecated [lodash.isequal](https://www.npmjs.com/package/lodash.isequal) (4.5.0) and lodash.isequalwith (4.4.0) in favor of the built-in isDeepStrictEqual from node:util.

These packages were only used in the Figma doc-helpers script.


tujoworker (approved):
Nice :D

PR #7767, opened 2026-04-24
style(Dropdown, Autocomplete): unify design across brands

snorrekim:
**Design changes**:
- New design for `DrawerList`, `Dropdown`, and `Autocomplete`.
- Added optional `noDivider` prop to remove divider lines between items.
- Deprecated prop `arrowPosition` as there is no longer an arrow in the new design.

**Fixes:**
- checkmark for selected item no longer appears above text
- improved scroll behaviour
  - only scroll if the item is out of view. Not every time the user keyboard navigates. 
  - only scroll until the item is visible. Not so the active item is always on the top.
  - prevent browser focus from scrolling the list. Different browsers behaved differently.

---

Deploy preview for Dropdown: https://chore-dropdown-autocomplete.eufemia-e25.pages.dev/uilib/components/dropdown/demos
Deploy preview for Autocomplete: https://chore-dropdown-autocomplete.eufemia-e25.pages.dev/uilib/components/autocomplete/demos

PR #8622, opened 2026-06-19
chore(DrawerList): do not forward deprecated arrowPosition prop to the DOM

langz:
Follow-up to https://github.com/dnbexperience/eufemia/pull/7767

The deprecated arrowPosition prop was no longer destructured out of the props, so it leaked onto the DOM element and triggered a React 'does not recognize the prop' warning. Strip it before spreading attributes.

### DIFF
--- a/packages/dnb-eufemia/src/fragments/drawer-list/DrawerList.tsx
+++ b/packages/dnb-eufemia/src/fragments/drawer-list/DrawerList.tsx
@@ -422,6 +422,7 @@ const DrawerListInstance = memo(function DrawerListInstance(
     ignoreEvents,
     optionsRender,
     className,
+    arrowPosition: _arrowPosition,
     cacheHash: _cacheHash,
     wrapperElement: _wrapperElement,
     direction: _direction,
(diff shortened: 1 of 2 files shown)


tujoworker (approved):

packages/dnb-eufemia/src/fragments/drawer-list/DrawerList.tsx
```diff
@@ -152,7 +152,7 @@ export type DrawerListProps = {
   cacheHash?: string
   /**
    * Position of the arrow on the popup drawer. Set to `left` or `right`. Defaults to `left` if not set.
-   * @deprecated does nothing as there is no longer any arrow.
+   * @deprecated does nothing as there is no longer any arrow, and will be removed in v12.
```
tujoworker, line 155: Even next breaking will be v13 ;) Not sure if we should ensure we already should take this number into account.

langz: Yeah, makes sense. I can rename it in an upcoming PR.

langz: https://github.com/dnbexperience/eufemia/pull/8633

PR #8629, opened 2026-06-19
fix(NumberFormat): preserve formatted part boundaries

tujoworker:
NumberFormat's parts hook relied on parsing the finished display string, which made compact suffixes, currency placement, and signs fragile across locales.

This PR carries semantic formatter parts through the standard Eufemia number, currency, and percent formatters so consumers can style number, currency, and percent segments without losing the formatter boundaries. Custom formatter objects still fall back to display-string parsing, keeping the existing hook contract and public return shape intact.


langz:
Does the semantic-parts work correctly for negative Norwegian numbers?

I got an agent running the branch and it noticed that for `no`/`nb`/`nn` locales, negative values silently fall back to the old string-parsing path instead of using the new semantic parts:

| input | path |
| --- | --- |
| `formatCurrency(1234, { locale: 'nb-NO' })` | semantic ✅ |
| `formatCurrency(-1234, { locale: 'nb-NO' })` | fallback ⚠️ |
| `formatNumber(-1234, { locale: 'nb-NO' })` | fallback ⚠️ |
| de-CH / en-GB / percent negatives | semantic ✅ |

It says the cause is `prepareMinus` rewriting the minus glyph (`−` U+2212 → `-` U+002D), so the `display === formatted.number` guard fails and the parts get dropped. Personally, I'm not 100% convinced. 
Output is still correct via the fallback, but the new code never runs for what's probably the most common case (negative NOK).

I guess we could sync `prepareMinus` onto the `minusSign` part (or normalising before the equality check) so negatives keep the semantic path?

langz (approved):

PR #8630, opened 2026-06-19
chore(deps): improve Dependabot monorepo update coverage

tujoworker:
## Why
Dependabot's npm configuration did not fully reflect this monorepo's workspace structure, which can lead to updates that are harder to apply cleanly across the repository.

## Context
The repository uses Yarn workspaces spread across root, packages, plugin subfolders, and tools. Dependabot needs matching directory coverage and strategy settings so dependency maintenance aligns with how the repo is actually organized.

## User impact
This reduces maintenance friction by making dependency update behavior more predictable for this monorepo, while keeping scheduled update noise low and preserving room for grouped updates when limits are changed later.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -59,7 +59,6 @@ updates:
     open-pull-requests-limit: 5
     labels:
       - 'dependencies'
-      - 'ci'
     commit-message:
       prefix: 'chore(ci)'
       include: 'scope'
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,18 +5,65 @@
 
 version: 2
 updates:
-  - package-ecosystem: 'npm' # See documentation for possible values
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -5,18 +5,65 @@
...
+
+  - package-ecosystem: 'github-actions'
+    directory: '/'
     schedule:
       interval: 'monthly'
-    open-pull-requests-limit: 0 # disable dependeabit deps updates because of too many PRs (for each dependency)
+    open-pull-requests-limit: 5
+    labels:
+      - 'dependencies'
+      - 'ci'
```
langz:
```suggestion
```

Or we can of course create the ci label here in GitHub, but I guess `dependencies` is good enough, and it already exists 👍

langz (approved):

PR #8647, opened 2026-06-19
feat(Forms): add `Form.Bridge` component to render form fields outside the `Form.Handler` subtree

tujoworker:
Form.Bridge makes it possible to place form fields outside the Form.Handler subtree while keeping them connected to the same shared data, validation, and submit flow.

This supports layouts where form controls need to live in separate UI regions without splitting the underlying form state, and adds documentation so the pattern is easier to discover and apply consistently.


tujoworker:
I had it named first "outlet" and still like it better. Any thoughts?

tujoworker:
Closing this in favor of #8648

PR #8648, opened 2026-06-19
feat: add `Form.Outlet` component to render form fields outside the `Form.Handler` subtree

tujoworker:
Form.Outlet makes it possible to place form fields outside the Form.Handler subtree while keeping them connected to the same shared data, validation, and submit flow.

This supports layouts where form controls need to live in separate UI regions without splitting the underlying form state, and adds documentation so the pattern is easier to discover and apply consistently.

Preview: https://feat-forms-outlet.eufemia-e25.pages.dev/uilib/extensions/forms/Form/Outlet


packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/changelog.mdx
```diff
@@ -17,6 +17,7 @@ Get more [details about releases](/uilib/about-the-lib/releases) or have a look
 
 ## v11.7.0
 
+- Added [Form.Bridge](/uilib/extensions/forms/Form/Bridge/) to connect fields and submit controls outside a `Form.Handler` subtree to the same form state.
```
langz: This line should be updated 🤞

langz (approved):
Looks good to me, the only thing I can think about in regards to Outlet is [React Router Outlet](https://reactrouter.com/api/components/Outlet), but I think it works good :)

PR #8649, opened 2026-06-20
feat(Forms): add type-checked paths support

tujoworker:
## Why

Forms field `path` props are plain strings, so path typos only surface at runtime. This adds opt-in, type-only path checking: register your form data type once via the `Register` interface, then use the pre-typed `RegisteredField`, `RegisteredValue`, `RegisteredForm` and `RegisteredIterate` namespaces to get autocomplete and compile-time errors on the `path` prop.

Fully opt-in and erased at build time, so there is no runtime or bundle-size cost and existing code is unaffected.

Docs preview: https://feat-forms-typed-paths.eufemia-e25.pages.dev/uilib/extensions/forms/typed-paths/


langz (approved):
Looks good to me :) 
Not sure if there's a way to improve so that we do not need to use the `as unknown as`-pattern.

tujoworker:
Yeah, I checked out the possibility already. AI was not capable of finding a nice solution there yet.

PR #8656, opened 2026-06-22
fix(DrawerList): keep keyboard focus ring visible when option is also hovered

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1782124381303239
Deploy preview: https://fix-drawer-list-keyboard-foc.eufemia-e25.pages.dev/uilib/components/autocomplete/demos#default-autocomplete

The hover style set --drawer-list-option-color-border to an undefined variable (--drawer-list-option-color-border--hover). Because the hover selector has higher specificity than the keyboard focus rule, hovering a keyboard-focused option wiped the focus ring, making the active item indistinguishable from a hovered one (hover bg equals focus bg by design). Remove the dead border override from the hover block so the keyboard focus border persists.


joakbjerk (approved):
Looks good 🧘

PR #8666, opened 2026-06-22
chore(Radio): remove references to undefined CSS custom properties

langz:
The disabled state set --radio-border-inset and --radio-border-width to undefined variables (--radio-border-inset--disabled and --radio-border-width--disabled). While other state variants (default, active, hover, focus) are all defined, the disabled variants were never added.

Because var() with an undefined custom property produces a guaranteed-invalid value, the border-inset and border-width of disabled radios fell back to the initial value instead of retaining the default inset border style, which could cause the disabled radio border to render inconsistently.

### DIFF
--- a/packages/dnb-eufemia/src/components/radio/style/dnb-radio.scss
+++ b/packages/dnb-eufemia/src/components/radio/style/dnb-radio.scss
@@ -394,8 +394,6 @@
 
   // General
   &__input[disabled] ~ &__button {
-    --radio-border-inset: var(--radio-border-inset--disabled);
-    --radio-border-width: var(--radio-border-width--disabled);
     --radio-color-border: var(--radio-color-border--disabled);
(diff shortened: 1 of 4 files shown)


packages/dnb-eufemia/src/components/radio/style/dnb-radio.scss
```diff
@@ -394,8 +394,6 @@
 
   // General
   &__input[disabled] ~ &__button {
-    --radio-border-inset: var(--radio-border-inset--disabled);
-    --radio-border-width: var(--radio-border-width--disabled);
```
langz, line 398: It's safe to remove, as `var(--radio-border-inset--disabled)` and `var(--radio-border-width--disabled)` is `undefined` but could perhaps the intention be to use some other values? Personally, I don't know.

PR #8668, opened 2026-06-22
fix(PaymentCard): add missing CSS custom property definitions

langz:
The card design variants business-no-visa and business-with-visa referenced --dnb-payment-bg-business-no-visa and
--dnb-payment-bg-business-with-visa, but these custom properties were never defined. All other card background variables (default, pluss, white, gold, etc.) are defined at the top of the file.

Because var() with an undefined custom property produces a guaranteed-invalid value, the background-color declaration was silently invalid for these card types, causing them to render with a transparent background instead of the intended card color.


langz:
I believe these styles are actually unused, hence closing this PR and opening a new one removing them.

langz:
https://github.com/dnbexperience/eufemia/pull/8672

PR #8669, opened 2026-06-22
chore(MultiSelection): remove references to undefined CSS custom property

langz:
The focus-visible outline-offset was set to var(--focus-ring-offset), but this custom property is never defined anywhere in the codebase (neither in SCSS nor in JavaScript). The related --focus-ring-width IS defined in the theme, but --focus-ring-offset was never added.

Because var() with an undefined custom property produces a guaranteed-invalid value, the outline-offset declaration was silently invalid, causing it to fall back to the initial value (0). While the result happened to be correct by accident, the dead reference adds confusion and could mask issues if the outline behavior is changed in the future.

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/MultiSelection/style/dnb-multi-selection.scss
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/MultiSelection/style/dnb-multi-selection.scss
@@ -27,7 +27,6 @@
     &:focus-visible {
       outline: var(--focus-ring-width) solid
         var(--token-color-stroke-action-focus);
-      outline-offset: var(--focus-ring-offset);
     }
   }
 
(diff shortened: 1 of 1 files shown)


joakbjerk (approved):

packages/dnb-eufemia/src/extensions/forms/Field/MultiSelection/style/dnb-multi-selection.scss
```diff
@@ -192,7 +191,6 @@
     &:focus-visible {
       outline: var(--focus-ring-width) solid
         var(--token-color-stroke-action-focus);
-      outline-offset: var(--focus-ring-offset);
```
langz, line 195: It's safe to remove, as `var(--focus-ring-offset)` is `undefined`, but could perhaps the intention be to use some other values? Personally, I don't know.

Originally introduced in https://github.com/dnbexperience/eufemia/commit/c0801e8c8900a37bdc6a8dda7a0236532519e9d5#diff-bb9e1d862bebd104819745980a63f40c2ff5a9533591061c60f9d70ed3daedd0R189

PR #8670, opened 2026-06-22
chore(GlobalResponsiveTypography): add global responsive typography

snorrekim:
Adds a hidden extension `<GlobalResponsiveTypography>` that makes all wrapped children typography responsive.


tujoworker:
Closing this PR for now. We may open it up again when we actually have the need to provide this. For now we can use StackBlitz to create prototypes using e.g. `https://pkg.pr.new/@dnb/eufemia@8670.tgz`.

PR #8682, opened 2026-06-23
chore(Portal): fix broken spacing image on the Spacing page

langz:
Import the UX layout spacing image and render it via the Img component instead of markdown relative image syntax, which the portal build does not bundle (resulting in a broken asset URL).

Deploy preview: https://fix-docs-relative-image-rend.eufemia-e25.pages.dev/uilib/layout/spacing


joakbjerk (approved):
𖦂

PR #8701, opened 2026-06-24
docs(Forms): clarify how non-finite numbers (Infinity, NaN) are validated

tujoworker:
Documents how Eufemia Forms treats non-finite numbers (`Infinity`, `-Infinity`, `NaN`), so developers find it from the symptom (a number field blocking on an impossible result).

- `Field.Number`: renders non-finite as an empty input, but validation fails with a clear, localized message and blocks.
- Schema validation: a schema you supply also rejects non-finite values (Zod's `z.number()`); store an empty value (e.g. `undefined`) for impossible results.

Docs only. Complements #8699.

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/schema-validation/info.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/schema-validation/info.mdx
@@ -12,6 +12,7 @@ showTabs: true
 - [About schemas](#about-schemas)
 - [Using schema with DataContext](#using-schema-with-datacontext)
 - [Fields which are disabled or read-only](#fields-which-are-disabled-or-read-only)
+- [Non-finite numbers (Infinity and NaN)](#non-finite-numbers-infinity-and-nan)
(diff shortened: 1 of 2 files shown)


langz (approved):

packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/schema-validation/info.mdx
```diff
@@ -136,6 +137,21 @@ Also, note you can describe the schema without using the `type` property, as the
 
 Fields which have the `disabled` property or the `readOnly` property, will skip validation.
 
+## Non-finite numbers (Infinity and NaN)
```
langz, line 140: Based on your comment in Slack, I got the impression that this was based on which version of did that was being used. 

Should we mention something about version of zod in these docs? Or perhaps we already do somewhere in the docs here already.

tujoworker: No, v4 is the default. Should be fine.

PR #8703, opened 2026-06-25
feat(Provider): add missing component types to ContextComponents

langz:
Add Autocomplete, Radio, RadioGroup, ToggleButton and ToggleButtonGroup to the ContextComponents type so their default props can be set through the shared Provider with full type support.

Also remove the now-redundant 'as Record<string, unknown>' casts that were used to read these defaults (and formElement) from context in Radio, RadioGroup, ToggleButton and ToggleButtonGroup.


tujoworker (approved):
👣

PR #8716, opened 2026-06-25
refactor(Button): control status spacing via FormStatus shellSpace

tujoworker:
## Why

The vertical spacing between a `Button` and its trailing `FormStatus` was applied through a global CSS rule (`.dnb-button + .dnb-form-status { margin-top: 0.5rem }`). Relying on a global adjacent-sibling selector for spacing is fragile: it leaks into any layout where a form status follows a button, and it lives outside the component spacing system, making the spacing hard to reason about and easy to break.

This moves the responsibility to the component itself by letting `Button` render its `FormStatus` with `shellSpace={{ top: 'small' }}`, so the spacing comes from the established spacing system rather than an ad-hoc margin.

In the Dialog actions, the flex `gap` is narrowed to `column-gap` so the row gap no longer adds extra vertical space now that the status spacing is owned by the form status shell.

## Impact

- No visual change to the default spacing (small shell space matches the previous 0.5rem margin).
- Removes a global selector that could affect unrelated button + form-status combinations.
- Affected component snapshots are regenerated to reflect the removed global rule and the dialog `column-gap`.


langz (approved):
🌄

PR #8720, opened 2026-06-25
fix(Forms): stabilize mounted fields to prevent field re-render loop on mount

tujoworker:
When a linked `Form.Outlet` (and Form.Appearance) re-creates its data context on every validation version bump, the data context's `setMountedFieldState` function gets a new identity on each render. That identity change caused a field's lifecycle effects to re-run, tearing the field down and re-registering it on mount, which produced a re-render feedback loop.

Wrapping the mutation in a stable callback keeps the field registered across context identity changes, so mount/unmount effects only run when the field itself actually changes. This stops the loop without altering field behavior.

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/hooks/__tests__/useFieldProps.test.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/hooks/__tests__/useFieldProps.test.tsx
@@ -7222,6 +7222,49 @@ describe('useFieldProps', () => {
       })
     })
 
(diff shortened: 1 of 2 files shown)


langz (approved):
Could the same happen to:
- `updateDataValue`
- `setFieldError`
- `setFieldConnection`
- `revealError`

Or perhaps it's not applicable

tujoworker:
> Could the same happen to:
> - `updateDataValue`
> - `setFieldError`
> - `setFieldConnection`
> - `revealError`
> 
> Or perhaps it's not applicable

No, all other hooks depend on this hook.

PR #8736, opened 2026-06-26
fix(Typography): make Sbanken typography responsive by default again

snorrekim:
A "just-in-case" PR if it turns out making Sbanken typography responsive by default again is needed.


tujoworker:
Motivation: https://dnb-it.slack.com/archives/C04NL73S2Q5/p1782480681208589?thread_ts=1782477816.086399&cid=C04NL73S2Q5

tujoworker:
We can open this PR when we need it.

PR #8737, opened 2026-06-26
chore: render portal MDX components and resolve Field.Date tables in LLM docs

tujoworker:
The LLM metadata generator left 124 docs files with unhandled standalone MDX components and silently dropped the `Field.Date` property/event tables, lowering the quality of the generated `*.md` used for LLM consumption.

- Render the portal-global `RelatedComponents`, `ListComponentsOverview` and `MenuCard` into markdown instead of leaving raw JSX in the output.
- Extract `datePickerPropKeys` into its own lightweight module so the docs generator can evaluate `Field.Date` docs without loading the full component (which previously failed and dropped the tables).

No runtime behavior change in `@dnb/eufemia` – `datePickerPropKeys` is still exported from `./Date`.


langz (approved):
🤩🤯

langz (approved):
We should probably release this today or what do you think?

tujoworker:
No hurry 😉 but nice of you can look at the feiling tests.

PR #8760, opened 2026-06-30
chore: standardize dependency caching to skip installs across workflows

langz:
e2e, visual-regression and preview previously ran 'yarn install' on every run (visual-regression cached only .yarn/cache; preview used setup-node's yarn cache). They now cache the full node_modules tree and skip install on an exact cache hit, matching verify.yml.

All jobs share one cache key, so the cached paths are now identical everywhere: node_modules + packages/*/node_modules + tools/*/node_modules (a superset that also covers mcp-lambda's tools/mcp-lambda/node_modules). The already-conditional jobs (dev, release, icons-lib, mcp-worker, mcp-lambda) were widened from a narrower path to this same set so a job populating the cache first can never leave it incomplete.

Note: visual-regression now caches node_modules (macOS) instead of .yarn/cache; exact hits skip install entirely, while a cache miss re-downloads during install.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -1,0 +1,52 @@
+name: Setup dependencies
+description: >-
+  Set up Node.js and restore dependencies. Caches the full node_modules tree so
+  that "yarn install" is skipped on an exact cache hit, and caches the yarn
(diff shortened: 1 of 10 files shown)


joakbjerk (approved):
h🙏

PR #8768, opened 2026-07-01
chore: shard unit and visual regression --run-all

langz:
Includes both https://github.com/dnbexperience/eufemia/pull/8767 and https://github.com/dnbexperience/eufemia/pull/8750

Just want to see the CI build times


tujoworker:
We may look into this at a later time.

PR #8781, opened 2026-07-02
chore(deps): align eufemia-starter to vite 8 / plugin-react 6

langz:
## What

Aligns `eufemia-starter` to the same **vite 8.0.16** and **@vitejs/plugin-react 6.0.1** used by the other workspaces (it was on vite 6.4.3 / plugin-react 4.7.0).

## Why

Pure hygiene / version coordination — reduces duplicate `vite` (and plugin) installs across the monorepo and keeps the starter template current with the rest of the repo. Not a security fix (vite 6.4.3 is already patched).

## Verification

- `yarn workspace eufemia-starter build` succeeds (✓ built in ~7s) with the standard `vite.config.ts` unchanged.
- Only `packages/eufemia-starter/package.json` and `yarn.lock` change.

Kept separate from the security PR (#8775) so the 2-major vite bump can be reviewed on its own.

### DIFF
--- a/packages/eufemia-starter/package.json
+++ b/packages/eufemia-starter/package.json
@@ -31,7 +31,7 @@
     "@eslint/js": "10.0.1",
     "@types/react": "19.2.14",
     "@types/react-dom": "19.2.3",
-    "@vitejs/plugin-react": "4.7.0",
+    "@vitejs/plugin-react": "6.0.1",
     "eslint": "10.4.1",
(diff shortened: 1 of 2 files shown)


langz:
I tested in Chrome if the StackBlitz works now with this version, and it seems to work now :) 
Chrome: Version 149.0.7827.201 (Official Build) (arm64)

PR #8807, opened 2026-07-03
chore(security): add vulnerability-management routine + team ownership

langz:
## What

Docs-only. Brings Eufemia's security process in line with DNB's internal standards (SECC _Vulnerability and Patch Management_, SECARC, SI-02 Flaw Remediation).

- **`SECURITY.md`** — how to report privately (GitHub advisory first, email fallback), response targets (Critical 3d / High 7d / Moderate 30d / Low 60d, calendar days from confirmation), supported versions (`11.x` all severities, `10.x` **critical-only**), and the Security Champion / risk-owner split.
- **`docs/vulnerability-management.md`** — the maintainer routine: separate _what ships_ from _what builds_, triage decision tree, severity × exposure response matrix, the native `yarn npm audit` gate, and review cadence.
- **`CONTRIBUTING.md`** — a "before adding a dependency" checklist and a Security section linking both.

## Notes

Every factual claim was checked against the actual workflows, manifests, and live repo settings — audit gate is green. Companion to #8808 (native audit gate, merged), #8812, and #8930.

**Sources** (DNB-internal, SSO): [SECC/2032337941](https://dnb-asa.atlassian.net/wiki/spaces/SECC/pages/2032337941) · [SECARC/135177479](https://dnb-asa.atlassian.net/wiki/spaces/SECARC/pages/135177479) · [EDS/1326776462](https://dnb-asa.atlassian.net/wiki/spaces/EDS/pages/1326776462)

PR #8812, opened 2026-07-03
chore(security): SBOM per release + Dependabot cooldown

langz:
Two supply-chain controls from the DNB internal-standards review (EDS-823). Neither relaxes an existing Eufemia gate.

**SBOM + vulnerability report per release** — after publish, `release.yml` generates a CycloneDX 1.6 SBOM of `@dnb/eufemia`'s production dependencies plus a production-scope vulnerability report, attaches both to the GitHub Release, and uploads them as a workflow artifact. Runs with `continue-on-error`, so it can never block a release — the pre-publish audit stays the gate. `verify.yml` runs the same generators as a non-gating PR-CI smoke test, so a broken plugin, flag, or `yarn npm audit --json` shape is caught on the PR.

**Dependabot cooldown** — a cool-off period so freshly published versions settle before adoption (npm 7d / 3d patch, forward-looking behind `open-pull-requests-limit: 0`; github-actions 7d, active). Version updates only; never delays security updates.

Also documents the SBOM in `docs/sbom.md` and gitignores the generated files. Companion to #8807.

---

**Sources** — DNB internal security standards (SSO-gated; DNB staff only): [SECARC/135177479](https://dnb-asa.atlassian.net/wiki/spaces/SECARC/pages/135177479) · [SECC/2032337941](https://dnb-asa.atlassian.net/wiki/spaces/SECC/pages/2032337941) · [SECC/1907851772](https://dnb-asa.atlassian.net/wiki/spaces/SECC/pages/1907851772)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,8 +18,7 @@ permissions:
   contents: write # Required for semantic-release to push tags and changelog
   issues: write # Required for semantic-release to comment on issues
   pull-requests: write # Required for semantic-release to comment on PRs
-  id-token: write # Required for npm provenance and SBOM attestation signing
-  attestations: write # Required to record the release SBOM attestation
+  id-token: write # Required for npm provenance
 
 env:
   GH_EMAIL: ${{ secrets.GH_EMAIL }}
@@ -38,6 +37,10 @@ jobs:
     name: Run release and deploy
 
     runs-on: ubuntu-latest
+
+    outputs:
+      released: ${{ steps.release-info.outputs.released }}
+      tarball: ${{ steps.pack.outputs.tarball }}
 
     steps:
       - name: Git checkout
@@ -90,13 +93,11 @@ jobs:
           endsWith(github.ref, '.x'))
         run: yarn workspace @dnb/eufemia publish:ci
 
-      # Release supply-chain artifacts: a CycloneDX SBOM of the published
-      # library's production dependencies plus a vulnerability report.
-      # Non-blocking (continue-on-error) — the "Audit dependencies" step above
-      # is the release gate, not this. See docs/sbom.md for what these files
-      # are, where they are stored, and why.
-      - name: Generate SBOM and vulnerability report
-        id: sbom
+      # Capture the published version so the least-privilege SBOM job below can
+      # target the matching GitHub Release and stamp the SBOM. Empty or the
+      # placeholder version means nothing was published.
+      - name: Capture released version
+        id: release-info
         continue-on-error: true
         if: github.repository == 'dnbexperience/eufemia' &&
           (github.ref == 'refs/heads/release' ||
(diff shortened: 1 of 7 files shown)


tujoworker:
The SBOM is generated from packages/dnb-eufemia, but the package actually published by semantic-release is packages/dnb-eufemia/build. The source package also has the fixed version 0.0.0-development, while semantic-release sets the real version in the build package.

This likely makes the SBOM identify the released component as @dnb/eufemia@0.0.0-development rather than the version that was published. Could we generate it from the prepared/published package, or otherwise explicitly set and verify the released component version?

.github/workflows/release.yml
```diff
@@ -89,3 +89,41 @@ jobs:
...
+          github.ref == 'refs/heads/next' ||
+          endsWith(github.ref, '.x'))
+        working-directory: packages/dnb-eufemia
+        run: |
+          yarn dlx -q @cyclonedx/yarn-plugin-cyclonedx \
+            --output-format JSON --spec-version 1.6 \
+            --production --output-file sbom.cdx.json
+          # Accompanying vulnerability report (production scope). Non-fatal: the
+          # "Audit dependencies" step above is the actual release gate.
+          yarn npm audit --environment production --json > vulnerability-report.json || true
```
tujoworker: This fallback can turn an audit execution failure into a report saying "No known vulnerabilities". For example, a registry/network failure may produce no stdout, after which the empty file is replaced with the success-looking fallback.

Could we capture the exit status separately and distinguish:

* audit completed and found no vulnerabilities;
* audit completed and found vulnerabilities;
* audit could not be completed?

The step can remain non-blocking, but the artifact should not report an audit failure as a clean result.

langz: Good catch — fixed in 62e192d. The step now captures the audit exit code and separates the three cases: exit 0 → clean summary; non-zero **with** a report → keep the findings; non-zero **with no** output → record `{"error":"audit could not be completed"}` instead of a clean-looking result. Still non-blocking. Thanks :)

langz:
> …identify the released component as `@dnb/eufemia@0.0.0-development`…

Good catch — fixed in 62e192d. Generating straight from `build/` isn't feasible (the CycloneDX Yarn plugin needs a Yarn project, and the published `build/` is a plain package), so I went with explicitly setting it: the step reads the real version semantic-release writes to `build/package.json` and stamps it onto the SBOM component (version + purl), with a warning if it's still `0.0.0-development`. Thanks :)

tujoworker (changes requested):
Two release-security issues need addressing:

1. The verification examples use the predicate type https://cyclonedx.org/bom/v1.6, while actions/attest with sbom-path produces the CycloneDX predicate type https://cyclonedx.org/bom. The documented commands will not find the attestation.

2. The release job downloads and executes the CycloneDX generator with yarn dlx while the job has OIDC and attestation-writing permissions. Clearing named environment secrets does not remove those job credentials. Please use a lockfile-pinned dependency or move generation into a separate least-privilege job and pass the resulting artifact to the attestation job.

langz:
Thanks @tujoworker — both points are addressed in the latest push:

1. **Predicate type** — the `gh attestation verify` examples in `docs/sbom.md` now use `--predicate-type https://cyclonedx.org/bom`. That's the version-agnostic type `actions/attest` emits for CycloneDX SBOMs (only SPDX carries the spec version), so the documented commands now match the attestation.

2. **Untrusted generator vs. job credentials** — the CycloneDX generator (`yarn dlx`) now runs in a dedicated `sbom` job with `permissions: contents: read` and no OIDC/attestation credentials, with the workflow secrets blanked. It uploads the SBOM + vulnerability report as an artifact, and a separate `attest` job (holding `id-token`/`attestations: write`) downloads it and runs `actions/attest`. `attestations: write` is off the release job entirely, so the untrusted code no longer shares a job with those credentials. Generation stays non-blocking.

tujoworker (approved):
Both requested release-security changes are addressed: the documentation uses the predicate type produced by actions/attest, and SBOM generation now runs in a separate least-privilege job before the privileged attestation job. Checks are green; no remaining blocking findings.

PR #8824, opened 2026-07-03
chore: revert remove deprecated validateDOMAttributes helper (#8759)

langz:
Reverts the changes from #8759 (`chore: remove deprecated validateDOMAttributes helper`).

## Why

#8759 removed the shared `validateDOMAttributes` helper and migrated ~52 components to explicit helpers. The migration did not fully reconstruct one of the helper's jobs — merging the public `attributes` prop — on the main input elements of **Input**, **Checkbox**, **Radio** and **Textarea**.

As a result, passing `attributes={{ 'data-foo': 'bar' }}` to those components:

- **silently dropped** the intended attribute, and
- **leaked the raw object** onto the DOM as `attributes="[object Object]"`.

React emits no warning for this, so the existing test suite did not catch it (the guard tests added in #8759 only asserted spacing-prop non-leakage). For Radio, `labelDirection` could also leak onto the `<input>`.

## Why revert (vs. a targeted fix)

Reverting restores the full `validateDOMAttributes` safety net in one step — this covers the `attributes` merge, the `labelDirection` leak, and any other invalid-attribute scrubbing gaps that the migration may have left in components not individually audited. It returns these files to the known-good v11.8.1 behavior.

A **targeted alternative** exists in **#8823**, which keeps the #8759 refactor and only re-adds the `attributes` merge (+ Radio `labelDirection`) on the four affected components, with regression tests. **These two PRs are mutually exclusive — pick one:**

- This revert: safest/most complete, but undoes the deliberate refactor and reinstates the deprecated helper.
- #8823: minimal/surgical, preserves the refactor, but only fixes the components that were audited.

## Verification

- Revert applies cleanly onto latest `main` (git 3-way auto-merge; no conflicts).
- `yarn test:types` — 0 errors.
- Component suites green (Input, Checkbox, Radio, Textarea, Switch, Autocomplete, Button, Dropdown).

## Notes

- 64 files changed (restores the helper, its call sites, and its unit tests; removes the #8759 guard tests).
- Later type-only commits that touched these files were auto-merged on top and still type-check.

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/contribute/getting-started/making-changes.mdx
+++ b/packages/dnb-design-system-portal/src/docs/contribute/getting-started/making-changes.mdx
@@ -229,7 +229,10 @@ How you make [spacing](/uilib/layout/space) support available varies from case t
 ```tsx
 import { Context } from '../../shared'
 import { clsx } from 'clsx'
-import { extendPropsWithContext } from '../../shared/component-helper'
+import {
+  validateDOMAttributes,
+  extendPropsWithContext,
+} from '../../shared/component-helper'
 import { useSpacing } from '../space/SpacingUtils'
 
 import type { SpacingProps } from '../../shared/types'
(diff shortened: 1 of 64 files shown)


langz:
When trying to migrate next time, I should look into adding such tests as well https://github.com/dnbexperience/eufemia/pull/8823/changes

PR #8825, opened 2026-07-03
release of v11.8.3

langz:
Eufemia's next release 🌞 

Deploy preview: https://main.eufemia-e25.pages.dev/

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/components/list/Examples.tsx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/components/list/Examples.tsx
@@ -374,6 +374,44 @@ export const PendingState = () => {
   )
 }
 
+export const DisabledState = () => {
+  return (
+    <ComponentBox data-visual-test="list-disabled" scope={{ fish_medium }}>
+      <List.Container>
+        <List.Item.Action
+          icon={fish_medium}
+          title="Navigate to details 1"
+          onClick={() => console.log('Clicked')}
+        >
+          <List.Cell.End>
+            <NumberFormat.Currency value={1234} />
+          </List.Cell.End>
+        </List.Item.Action>
+
+        <List.Item.Action
+          icon={fish_medium}
+          title="Navigate to details 2"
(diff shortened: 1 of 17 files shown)


langz:
Is this something you would want to consider releasing this week @joakbjerk?
If so, just let me know and I’ll press the button :)

joakbjerk:
> Is this something you would want to consider releasing this week @joakbjerk? If so, just let me know and I’ll press the button :)

I guess we could 😄 But I thought there was a code freeze? @langz

langz:
> > Is this something you would want to consider releasing this week @joakbjerk? If so, just let me know and I’ll press the button :)
> 
> 
> 
> I guess we could 😄 But I thought there was a code freeze? @langz 

Perhaps, I'm not entirely sure how we treat it. 
But since all applications has code freeze, it means that none of our releases made during the code freeze will ever be used in an application now during code freeze. 
Also, we have many many versions, so they can always pick the previously working. 

There's also quite some value in that we are delivering releases which devs will now already use in their test environment, etc. I guess some are already working on their next/upcoming release for right after the summer freeze. 

Perhaps we can deliver bug fix releases, but wait with features or such.

joakbjerk:
> > > Is this something you would want to consider releasing this week @joakbjerk? If so, just let me know and I’ll press the button :)
> > 
> > 
> > I guess we could 😄 But I thought there was a code freeze? @langz
> 
> Perhaps, I'm not entirely sure how we treat it. But since all applications has code freeze, it means that none of our releases made during the code freeze will ever be used in an application now during code freeze. Also, we have many many versions, so they can always pick the previously working.
> 
> There's also quite some value in that we are delivering releases which devs will now already use in their test environment, etc. I guess some are already working on their next/upcoming release for right after the summer freeze.
> 
> Perhaps we can deliver bug fix releases, but wait with features or such.

Great point 👍 A fix release shouldn't really cause any issues?

joakbjerk (approved):

PR #8832, opened 2026-07-28
feat(PostalCodeAndCity): add validation to reject 0000 as an invalid postal code

joakbjerk:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1783941182114489


langz (commented):
Looks nice 😊 
Should we add some docs for this bevahior? Is this a validator that's exposed so that users can extend/build upon this validator?

joakbjerk:
> Looks nice 😊 Should we add some docs for this bevahior? Is this a validator that's exposed so that users can extend/build upon this validator?

Add docs, and also exposed the validator 🙏

langz (approved):

PR #8839, opened 2026-08-03
fix(Flex): remove trailing horizontal spacing

tujoworker:
Removes trailing spacing from the final horizontal Flex child.

Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1783588408764629


langz (approved):

tujoworker:
For containers using span, the established gutter behavior is intentionally preserved because percentage-based column sizing relies on it. So the fix now avoids the unwanted trailing spacing without breaking spanned layouts.

PR #8842, opened 2026-08-03
docs(Portal): add command+k support for search

joakbjerk:
## Summary

Adds Cmd/Ctrl+K support to the portal search by focusing the existing header `Autocomplete`. The search input also shows a platform-aware shortcut hint when empty.

## Changes

- Adds a global Cmd/Ctrl+K keyboard listener that prevents the browser default and focuses `#portal-search`.
- Exposes the supported shortcuts through `aria-keyshortcuts="Meta+K Control+K"`.
- Shows `⌘ K` on macOS and `Ctrl K` on other platforms inside the search input.
- Omits the platform-specific hint during server rendering, then resolves it on the client.
- Hides the hint when the search has a value and restores it when the input is cleared.
- Adds styling for the shortcut hint and adjusts the existing autocomplete spacing.
- Adds tests for rendering, shortcut metadata, Cmd/Ctrl+K focus behavior, server rendering, platform-specific labels, and hiding the hint when typing.

## Testing

All PR checks pass, including tests, type checks, lint, build tests, end-to-end tests, and visual regression tests.

## Breaking changes

None. This only changes the portal search behavior and presentation.

### DIFF
--- a/packages/dnb-design-system-portal/src/shared/menu/SearchBar.tsx
+++ b/packages/dnb-design-system-portal/src/shared/menu/SearchBar.tsx
@@ -17,7 +17,6 @@ import {
   searchButtonContentStyle,
   searchButtonShortcutStyle,
   dialogAutocompleteStyle,
-  dialogClassStyle,
   drawerClassStyle,
 } from './SearchBar.module.scss'
 import { scrollToAnimation } from '../parts/layout-utils'
@@ -166,7 +165,6 @@ export const SearchBarInput = () => {
         open={openState}
         onClose={closeDialog}
         focusSelector=".dnb-input__input"
-        className={dialogClassStyle}
       >
         <Autocomplete
           id="portal-search-dialog"
@@ -176,8 +174,7 @@ export const SearchBarInput = () => {
           noScrollAnimation
           preventSelection
           disableFilter
-          skipPortal
-          keepOpen
+          fixedPosition
           stretch
           size="medium"
           placeholder="Search ..."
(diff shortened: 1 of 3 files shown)


packages/dnb-design-system-portal/src/shared/menu/SearchBar.tsx
```diff
@@ -26,6 +29,40 @@ import { formatSearchResultMarkdown } from './SearchBarMarkdown'
 export const SearchBarInput = () => {
   const searchIndex = useRef(null)
   const [status, setStatus] = useState(null)
+  const [openState, setOpenState] = useState(false)
+  const [isMac, setIsMac] = useState(false)
```
tujoworker: Here we should use the `isMac()` helper function we provide, instead.

tujoworker (approved):
Looks very good!

packages/dnb-design-system-portal/src/shared/menu/SearchBar.tsx
joakbjerk: Should be changed now 🫡

tujoworker:
I think we should display the same shortcut as we before in the button. Just shown as long as not the clear button is shown. Or maybe even then. I guess we can use the submitElement for that.

Screenshot 2026-08-04 at 13 17 39

joakbjerk:
> I think we should display the same shortcut as we before in the button. Just shown as long as not the clear button is shown. Or maybe even then. I guess we can use the submitElement for that.
> Screenshot 2026-08-04 at 13 17 39

This should be fixified now!

joakbjerk:
Such beauty!
Screenshot 2026-08-04 at 14 00 21

tujoworker (approved):

tujoworker:
I think `useSyncExternalStore` is cleaner than `isMounted`, so, pushed a commit.

langz (approved):

PR #8844, opened 2026-08-04
fix(Autocomplete): rank results by word coverage instead of single-word repetition

joakbjerk:
## Summary

Fixes Autocomplete result ranking so that an item matching **more of the distinct search words** always outranks an item that merely **repeats a single search word** many times.

### The bug

When searching `the godfather`, the item _The Lord of the Rings: The Return of the King_ ranked above _The Godfather_ — even though it does not contain the word "godfather" at all. It won purely because it repeats "the" four times, and every occurrence was added to the score.

### The fix

Each matched search word now contributes to the score only **once** (presence, not frequency):

```ts
const wordOccurrences = (contentChunk.match(scoreRegex) || []).length
wordScore += Math.min(wordOccurrences, 1)
```

Word coverage (the number of distinct query words matched) therefore dominates ranking, while the existing first-word-start bonus and overall sort behaviour are preserved.

### Before / After

Query: `the godfather`

| Before | After |
| --- | ---|
| autocomplete-ranking-before | autocomplete-ranking-after|

Before: _The Lord of the Rings…_ ranks #1 above the exact match _The Godfather_.
After: _The Godfather_ ranks #1.

### Tests

- Added a test asserting `the godfather` ranks _The Godfather_ first.
- Full Autocomplete suite passes (125/125), no snapshot changes.


langz (approved):

tujoworker:
Good fix. One edge case remains: the first-word bonus can still outweigh word coverage – for `foo bar`, `foo` ranks above `x foo bar`. I'll provide a commit that prioritizes distinct word coverage and keeps the current score as a tie-breaker.

tujoworker (approved):

langz (approved):

PR #8878, opened 2026-08-05
fix(ToggleButton.Group): remove invalid aria-labelledby from role-less shell

langz:
## Summary

Every `ToggleButton.Group` **with a `label`** rendered its inner shell `<span class="dnb-toggle-button-group__shell" aria-labelledby="…">` **without a role**. A generic element (implicit role `generic`) is prohibited from having a naming attribute such as `aria-labelledby`, so axe flagged `aria-prohibited-attr` (serious) on the ToggleButton documentation page.

## Root cause

`aria-labelledby` was added to the spread `params` object and applied to the role-less shell `<span>`. The group is **already** correctly labelled on the `<fieldset role="group|radiogroup">` (with its `<legend>`), so the copy on the shell was both redundant and invalid.

## Fix

Removed the `aria-labelledby` assignment from `params` (and the now-unused `combineLabelledBy` import). The group keeps its accessible name via the `<fieldset>` + `<legend>`; the shell no longer carries an invalid naming attribute.

## Tests

- Added a test asserting the shell has no `aria-labelledby` while the `<fieldset>` is still labelled by the `<legend>`.
- Added an axe assertion for a labelled group.
- Existing tests for the fieldset's `aria-labelledby`/`role` and the role-less shell continue to pass.

## How to verify

- **Before:** on https://eufemia.dnb.no/uilib/components/toggle-button the group shells expose `aria-labelledby` with no role.
- **After:** the shell has no `aria-labelledby`; the group is still announced with its label via the fieldset/legend, and axe reports no `aria-prohibited-attr` violation.

### DIFF
--- a/packages/dnb-eufemia/src/components/toggle-button/ToggleButtonGroup.tsx
+++ b/packages/dnb-eufemia/src/components/toggle-button/ToggleButtonGroup.tsx
@@ -12,7 +12,6 @@ import {
   validateDOMAttributes,
   getStatusState,
   combineDescribedBy,
-  combineLabelledBy,
   dispatchCustomElementEvent,
   removeUndefinedProps,
 } from '../../shared/component-helper'
@@ -212,9 +211,6 @@ function ToggleButtonGroup(ownProps: ToggleButtonGroupProps) {
       showStatus ? id + '-status' : null,
       suffix ? id + '-suffix' : null
     )
-  }
-  if (label) {
-    params['aria-labelledby'] = combineLabelledBy(params, id + '-label')
   }
 
   // also used for code markup simulation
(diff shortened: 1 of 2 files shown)


tujoworker (approved):

tujoworker:
Looks good! Removing `aria-labelledby` from the role-less shell while keeping the `fieldset` labelled is the correct fix. I noticed that `RadioGroup` appears to have the same pattern—could we fix that here as well, or create a follow-up?

langz:
> Looks good! Removing `aria-labelledby` from the role-less shell while keeping the `fieldset` labelled is the correct fix. I noticed that `RadioGroup` appears to have the same pattern—could we fix that here as well, or create a follow-up?
> 
> 

I'll look into it :)

Here's a PR https://github.com/dnbexperience/eufemia/pull/8892

langz:
Looked into it — `Radio.Group` had the exact same pattern, so I've opened #8892 as a follow-up. Same fix there: removed the invalid `aria-labelledby` from the role-less `.dnb-radio-group__shell` (plus the now-unused `combineLabelledBy` import), keeping the accessible name on the `fieldset` + `legend`. Added a test asserting the shell has no `aria-labelledby` while the fieldset stays labelled by the legend. Thanks for catching it :)

PR #8881, opened 2026-08-05
fix(Tabs): omit aria-controls on the selected tab when no tabpanel exists

langz:
## Description

The selected tab always set `aria-controls="{id}-content"`, even when the `Tabs` component rendered no tabpanel — e.g. when it is used for layout only (with `data` for the tab strip but no content, and no external `<Tabs.Content>`). Since `ContentWrapper` returns `null` when it has no children, no panel with that id exists, leaving a **dangling `aria-controls` reference**.

Confirmed on the live docs with axe-core as `aria-valid-attr-value` (critical, WCAG 4.1.2) on three examples (`unique-tabs-row`, `unique-tabs-max-width`, and a content-less `Tabs`).

## Fix

Set `aria-controls` on the selected tab only when a tabpanel actually exists:

- **Internal content** (`data[].content` / `children`) — known synchronously via `getContent`, so `aria-controls` is present in the prerendered/SSR output too.
- **External `<Tabs.Content id="...">`** — detected in the DOM after render (layout effect).
- **No content** (layout-only tabs) — no panel, so `aria-controls` is omitted.

## Tests

Added a `Tabs aria-controls` describe block:

- keeps `aria-controls` when internal content is provided (regression guard),
- keeps `aria-controls` when content is provided by an external `Tabs.Content` (regression guard),
- omits `aria-controls` when no tabpanel is rendered (the bug — this test fails before the change).

All 38 Tabs tests pass (including the existing ARIA-rules and snapshot tests). Verified live: axe `aria-valid-attr-value` went from 3 → 0, content-ful tabs (including the external-content example) keep a valid `aria-controls`, and layout-only tabs no longer emit one.

### DIFF
--- a/packages/dnb-eufemia/src/components/tabs/Tabs.tsx
+++ b/packages/dnb-eufemia/src/components/tabs/Tabs.tsx
@@ -56,6 +56,7 @@ import CustomContent from './TabsCustomContent'
 import ContentWrapper from './TabsContentWrapper'
 import {
   createSharedState,
+  useSharedState,
   type SharedStateReturn,
 } from '../../shared/helpers/useSharedState'
 import type { ButtonProps } from '../Button'
@@ -410,8 +411,13 @@ function TabsComponent(ownProps: TabsProps) {
   const [isFirst, setIsFirst] = useState<boolean | undefined>(undefined)
   const [isLast, setIsLast] = useState<boolean | undefined>(undefined)
 
-  // True once an external <Tabs.Content> tabpanel is detected in the DOM.
-  const [hasTabPanel, setHasTabPanel] = useState(false)
+  // Reactively reflects whether an external <Tabs.Content> tabpanel is mounted.
+  // ContentWrapper registers/unregisters its presence in this shared state, so
+  // the selected tab links to it only while it actually exists.
+  const { data: tabPanelPresence } = useSharedState<{ count?: number }>(
+    `${_id}-tabpanel-presence`
+  )
+  const hasTabPanel = Boolean(tabPanelPresence?.count)
 
   // Track previous props for getDerivedStateFromProps equivalent
   const [prevDataSource, setPrevDataSource] = useState(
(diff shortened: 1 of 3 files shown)


joakbjerk (approved):

tujoworker (approved):

tujoworker:
Nice fix! I see two potential issues:

1. `hasTabPanel` is only updated when `_id`, `selectedKey`, or `data` changes. If an external `<Tabs.Content>` is conditionally mounted or removed while the Tabs props remain stable, `aria-controls` becomes stale. Could `ContentWrapper` explicitly register/unregister its presence through the shared state instead? A test that toggles the external content would cover this.

2. `getContent(currentSelectedKey)` is now called both to determine `aria-controls` and to render the content. This evaluates function-based content twice during the initial render and SSR. Could we avoid resolving the content solely to check whether it exists?

langz:
Great points, thanks — both are addressed in 8e99275:

**1. Stale `aria-controls` on toggle** — spot on. The `getElementById` layout effect only re-ran on `[_id, selectedKey, data]`, so it went stale whenever an external `<Tabs.Content>` mounted/unmounted while the `Tabs` props stayed put. I've dropped the DOM query entirely: `ContentWrapper` now registers/unregisters its tabpanel presence through a dedicated shared-state channel (`${id}-tabpanel-presence`, ref-counted so it's StrictMode-safe), and `Tabs` reads that reactively via `useSharedState`. The internal panel is flagged `nested` so it skips self-registration — `Tabs` already knows about its own content synchronously. Added a regression test that toggles an external `Tabs.Content` and asserts `aria-controls` appears when it mounts and is removed again when it unmounts.

**2. Resolving the content twice** — also fixed. The selected key's content is now resolved once per render (`selectedContent`) and reused by both the tabpanel rendering and the `aria-controls` decision. I went with resolve-once-and-reuse rather than a non-invoking `hasContent()` check, since a render-prop child can legitimately return `null` for a given key — so reusing the actual result stays accurate while only evaluating once (this also keeps SSR/first paint correct for internal content). Added a test that uses the `render` prop as a per-render counter to assert the content function runs exactly once per render.

All 40 Tabs tests pass. Thanks for the thorough review :)

PR #8889, opened 2026-08-05
style(StepIndicator): progress bar and redesign

snorrekim:


snorrekim:
Looks like there have been som changes to the design, so this PR will have to wait until those are finalized

tujoworker:
We will pick this up when UX is ready.

PR #8900, opened 2026-08-06
chore(Forms): register event listeners before layout effects

langz:
## Problem

`useEventListener` registered listeners as a side effect of `useMemo`. Mutating the listener registry during render is unsafe because React may abandon that render without running cleanup.

Moving registration to a normal `useEffect` is also insufficient for this event system. Descendant layout and passive effects run before an ancestor's corresponding effect, so a child calling `Form.useSubmit()` while mounting could submit before `SubmitConfirmation` or `WizardContainer` had registered its listener.

## Fix

Register and clean up listeners with `useInsertionEffect`.

This removes the render-time side effect while ensuring the internal listener registry is ready before descendant layout and passive effects can dispatch. The insertion effect only mutates the provider's ref-backed registry; it does not update React state or access the DOM.

## Verification

Added regression coverage to the existing `SubmitConfirmation` test suite for submission from both a child `useEffect` and `useLayoutEffect`. The tests verify that confirmation reaches `readyToBeSubmitted` and that the final form `onSubmit` is not called.

Verified with:

- 180 focused Forms tests covering `SubmitConfirmation`, `Isolation`, `WizardContainer`, and `Tools.Errors`
- TypeScript
- ESLint
- Prettier


tujoworker:
This change introduces a timing regression when submission is triggered while the tree is mounting.

`SubmitConfirmation` registers its `onSubmit` listener in a parent `useEffect`, but React runs descendant effects first. A child using the public `Form.useSubmit()` API from either `useEffect` or `useLayoutEffect` can therefore submit before the listener exists. The real `Form.Handler` `onSubmit` is called and confirmation is bypassed. `WizardContainer` uses the same listener mechanism and has the same exposure.

I reproduced this with regression tests covering both child effect types:

- PR implementation (`useEffect` registration): both tests fail; `onSubmit` is called once
- previous render-time registration: both tests pass
- `useLayoutEffect` registration: child `useEffect` passes, but child `useLayoutEffect` still fails

I committed the reproduction locally as `553ebe48112` (`test(Form.SubmitConfirmation): cover mount-time submission`). The current `useMemo` side effect should be replaced, but a passive/layout effect is not equivalent to the existing timing contract. I think this needs a subscription redesign, or another mechanism that guarantees registration before descendants can dispatch, before merging.

tujoworker:
Implemented and pushed a fix for the timing regression.

The listener now registers with `useInsertionEffect`. This removes the render-time side effect while ensuring registration completes before descendant layout and passive effects can dispatch. The registry operation is safe in this phase because it only mutates the provider's internal ref and does not update React state or access the DOM.

Added regression coverage for `Form.useSubmit()` called from a child `useEffect` and `useLayoutEffect`. Both cases now reach `SubmitConfirmation`'s `readyToBeSubmitted` state and do not call the final form `onSubmit`.

Verification:
- 180 focused Forms tests passed (`SubmitConfirmation`, `Isolation`, `Tools.Errors`, and `WizardContainer`)
- TypeScript passed
- ESLint passed
- Prettier passed

Pushed commits:
- `553ebe48112` — timing regression tests
- `4314f8ba4b2` — insertion-phase listener registration

langz (commented):
Great, thanks @tujoworker. 
We should look into if it's possible to actually add a unit test documenting this timing, to prevent regressions in the future or such.

tujoworker:
Yes, the added tests document this special timing requirement: this internal listener must ideally be registered before descendant layout effects can dispatch. That's why we use `useInsertionEffect` rather than the usual `useEffect`.

tujoworker (approved):

langz:
> Yes, the added tests document this special timing requirement: this internal listener must ideally be registered before descendant layout effects can dispatch. That's why we use `useInsertionEffect` rather than the usual `useEffect`.

Thanks, you are the best 😍

PR #8905, opened 2026-08-06
chore(InteractionInvalidation): rename shadowed selector param for clarity

langz:
## Context

A code-quality finding flagged that the `bypassSelectors.some(...)` callback parameter `selector` shadows the outer `selector` variable (the full built CSS selector), suggesting `querySelector` used the wrong value.

## Assessment

This is **not a behavioral bug**: due to shadowing, every `selector` reference inside the callback resolves to the **parameter** (the correct per-iteration value), so `querySelector` already used the right selector. Verified by the existing bypass-selector tests (direct `.bypass-invalid` and wildcard `.bypass *`), which pass unchanged before and after.

## Change

Pure clarity refactor: rename the callback parameter to `bypassSelector` to remove the confusing shadowing. No behavior change.

## Verification

`InteractionInvalidation` suite: 13/13 passing, identical before and after the rename.


langz:
@/var/folders/8v/ht0ljtts4fg9kzngp34kqsbc0000gn/T//c8905.md

PR #8916, opened 2026-08-07
chore(analytics): add S3 + Athena analytics storage tool

joakbjerk:
## Motivation

[EDS-937](https://dnb-asa.atlassian.net/browse/EDS-937) — we want a simple, low-maintenance place to collect analytics records and query them flexibly (group-by / trends) for a future dashboard, without standing up and operating a database.

## Summary

Adds `tools/analytics/` — internal AWS tooling to store and retrieve analytics records over an HTTP API (`POST /records`, `GET /records`, `GET /healthz`). Records are written to S3 (date-partitioned) and read back via Athena over a Glue table with partition projection. Internal tooling only; not part of the published `@dnb/eufemia` package.

## Why this way

S3 + Athena suits flexible group-by/trend queries far better than DynamoDB, and reusing the `tools/mcp-lambda` deploy pattern (pre-created role per ADR 0004, shared S3 state backend, OIDC deploy, `nodejs22.x`) keeps a single, consistent way to ship AWS tooling in this repo.

### DIFF
--- a/.github/workflows/analytics-lambda.yml
+++ b/.github/workflows/analytics-lambda.yml
@@ -8,7 +8,7 @@ name: Analytics Lambda
 #
 # Triggers:
 # - `release` branch and `v*` / `v*.*.*` tags — deploys automatically.
-# - Any branch named `**/analytics` (e.g. `feat/analytics`).
+# - Any branch under the `analytics/` prefix (e.g. `analytics/new-field`).
 # - `workflow_dispatch` — manual deploy from the Actions tab.
 #
 # See tools/analytics/README.md for deploy details.
@@ -17,7 +17,7 @@ on:
   push:
     branches:
       - release
-      - '**/analytics'
+      - 'analytics/**'
     tags:
       - 'v*'
       - 'v*.*.*'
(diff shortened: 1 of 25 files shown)


langz (approved):
Looks good to me!

Image

langz:
isAuthorized returns true when API_TOKEN is unset, and the Terraform api_token defaults to "". With no authorizer on the HTTP API, that means a deploy without the token set is a publicly readable and writable endpoint. Is that intended for a write route with no edge in front of it?

langz:
The MCP pipeline audits its production deps at deploy time (audit:ci), but this workflow doesn't, and the package has no audit:ci script. Given the EDS-823 supply-chain work, why is the audit gate left out for this new dependency tree?

joakbjerk:
> isAuthorized returns true when API_TOKEN is unset, and the Terraform api_token defaults to "". With no authorizer on the HTTP API, that means a deploy without the token set is a publicly readable and writable endpoint. Is that intended for a write route with no edge in front of it?

No it should have a token, just haven't gotten there yet for this draft 😄

joakbjerk:
> The MCP pipeline audits its production deps at deploy time (audit:ci), but this workflow doesn't, and the package has no audit:ci script. Given the EDS-823 supply-chain work, why is the audit gate left out for this new dependency tree?

Have not gotten there yet, but will implement 😄

PR #8919, opened 2026-08-07
feat(Form.Section.EditContainer): prevent navigation with unconfirmed changes

tujoworker:
## Description

Adds `preventUncommittedChanges` to `Form.Section.EditContainer`. When enabled, both `Form.Handler` submission and Wizard navigation are blocked while the section is in edit mode, until Done or Cancel is selected—even when no values have changed.

The parent `Form.Section` must have a `path`; using the property without one throws an error to avoid guarding or restoring the whole form.

Includes tests covering `Form.Handler` submission and Wizard navigation, translations, documentation, and a Wizard example.

Preview: https://feat-form-section-prevent-un.eufemia-e25.pages.dev/uilib/extensions/forms/Form/Section/EditContainer#prevent-uncommitted-changes

Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1786088982839399

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Section/EditContainer/Examples.tsx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Section/EditContainer/Examples.tsx
@@ -1,5 +1,10 @@
 import ComponentBox from '../../../../../../../shared/tags/ComponentBox'
-import { Field, Form, Value } from '@dnb/eufemia/src/extensions/forms'
+import {
+  Field,
+  Form,
+  Value,
+  Wizard,
+} from '@dnb/eufemia/src/extensions/forms'
 
 export const ViewAndEditContainer = () => {
   return (
@@ -49,6 +54,35 @@ export const ViewAndEditContainer = () => {
           </Form.Handler>
         )
       }}
+    </ComponentBox>
+  )
+}
+
+export const PreventUncommittedChanges = () => {
+  return (
+    <ComponentBox>
+      <Form.Handler>
+        <Wizard.Container>
+          <Wizard.Step title="Profile">
+            <Form.Section path="/profile" containerMode="edit">
+              <Form.Section.EditContainer preventUncommittedChanges>
+                <Field.Name.First path="/firstName" />
+              </Form.Section.EditContainer>
+
+              <Form.Section.ViewContainer>
+                <Value.Name.First path="/firstName" />
+              </Form.Section.ViewContainer>
+            </Form.Section>
+
+            <Wizard.Buttons />
+          </Wizard.Step>
+
+          <Wizard.Step title="Summary">
+            <Value.Name.First path="/profile/firstName" />
+            <Wizard.Buttons />
+          </Wizard.Step>
+        </Wizard.Container>
+      </Form.Handler>
     </ComponentBox>
   )
 }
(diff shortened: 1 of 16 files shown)


langz:
I've been testing `preventUncommittedChanges` and hit one issue worth sorting out before merge.

### When `Form.Section` has no `path`, the guard tracks the whole form instead of the section

`useContainerDataStore` subscribes to and snapshots the data at `'/'`, and that goes through the section-aware path helper. With `<Form.Section path="/profile">` it resolves to `/profile` (correct), but with a plain `<Form.Section>` it resolves to the form root.

**What it causes:** any edit anywhere else in the form marks the section as having uncommitted changes. Submit and Wizard navigation get blocked, and the section renders _"Du må enten bekrefte eller forkaste endringene."_ — asking the user to confirm or discard changes they never made in that section. The only way out is pressing Done or Cancel on a section they never touched.

https://github.com/user-attachments/assets/8e9f321b-9e14-47a9-9c82-c700e92a6680




### Reproduce it by hand

Drop this into a demo. Type something into **Outside the section**, leave the section alone, then press Send — nothing submits and the section complains:

```tsx
  <Form.Handler onSubmit={(data) => console.log('onSubmit', data)}>
    <Form.Section containerMode="edit">
      <Form.Section.EditContainer preventUncommittedChanges>
        <Field.String path="/insideSection" label="Inside the section" />
      </Form.Section.EditContainer>

      <Form.Section.ViewContainer>view</Form.Section.ViewContainer>
    </Form.Section>

    <Field.String path="/outsideSection" label="Outside the section" />

    <Form.SubmitButton />
  </Form.Handler>
```

Add `path="/somewhere"` to the `Form.Section` and it behaves correctly.

### Same thing as a test

Fails on 5357454:

```tsx
it('should not block on edits outside the section', async () => {
  const onSubmit = vi.fn()

  render(
    <Form.Handler onSubmit={onSubmit}>
      <Form.Section containerMode="edit">
        <Form.Section.EditContainer preventUncommittedChanges>
          <Field.String path="/insideSection" />
        </Form.Section.EditContainer>
        <Form.Section.ViewContainer>view</Form.Section.ViewContainer>
      </Form.Section>

      <Field.String path="/outsideSection" />

      <Form.SubmitButton />
    </Form.Handler>
  )

  // type only in the field that lives outside the section
  await userEvent.type(document.querySelectorAll('input')[1], 'hello')
  await userEvent.click(document.querySelector('.dnb-forms-submit-button'))

  expect(onSubmit).toHaveBeenCalledTimes(1) // gets 0
})
```

The tests added here can't catch it: the three that use a pathless section have the section as the form's only content, so "the whole form" and "the section" are the same thing in those cases.

### Not a regression, to be clear

The same scoping already decides what `Cancel` restores. On both this branch and its base, cancelling a pathless section reverts a field that lives *outside* it — no `preventUncommittedChanges` involved:

```
defaultData={{ inside: 'in-orig', outside: 'out-orig' }}
// edit only /outside, then Cancel the section
// -> {"inside":"in-orig","outside":"out-orig"}   identical on base and head
```


https://github.com/user-attachments/assets/b086f097-cebe-47b2-9c91-22812961c41a



So it's a pre-existing scope issue that the new prop turns from a silent over-reach into a visible dead end for the user. Fixing the scope resolution would take care of both.

Two directions, your call: either require a `path` on the section when `preventUncommittedChanges` is used (with a dev warning when it's missing, rather than silently guarding everything), or scope the tracking to the fields actually mounted inside the container. The latter is the semantically right answer and handles the pathless case properly.

langz:
Will this work in a `Form.Isolation`? 🤔 
Or perhaps that's just overcomplicating stuff.
I'll try to think about it 🧠

langz (approved):
👏

PR #8922, opened 2026-08-07
feat: add Vite CSS optimizer plugin (beta)

tujoworker:
## How it works

1. **When Eufemia is built:** A manifest is generated from the Eufemia source. It maps each component to the styles and other components it depends on.
2. **When an application is built:** The Vite plugin scans the application's `src` directory for Eufemia imports.
3. **While Vite processes CSS:** The plugin keeps the detected component styles and their dependencies, then removes unused component CSS with PurgeCSS.

The manifest is shipped at `@dnb/eufemia/style/style-manifest.json`. The optimizer only changes Eufemia's aggregate component CSS; application CSS, core styles, and theme foundations are left untouched.

## Preview

[Removing unused CSS (beta)](https://feat-css-optimizer.eufemia-e25.pages.dev/uilib/usage/customisation/styling/consume-styles#removing-unused-css-beta
)
## Motivation

[Slack thread](https://dnb-it.slack.com/archives/CMXABCHEY/p1782287633022819)


langz (approved):

langz:
**Nice — the block-level unit of removal and the safety-first bias are the right calls.** One thing worth fixing before beta graduates, as it quietly guts the savings for two of the most-used components.

## `Dropdown` / `Autocomplete` keep the entire `dnb-forms` namespace

Importing just a `Dropdown` keeps **44 of 62 components** — the whole `dnb-forms-*` stylesheet included (every field type, plus `Table`, `Upload`, `DatePicker`, `Autocomplete`, …).

Root cause is one import in `components/country-flag/CountryFlag.tsx`:

```ts
import useCountries from '../../extensions/forms/Field/SelectCountry/useCountries'
```

`useCountries` is a pure data hook — no JSX, no `dnb-forms-*` classes. But `resolveRuntimeImports` treats any runtime import resolving inside a component dir as a CSS dependency, so `country-flag` gets a false edge to the whole `forms` entry, propagating `country-flag → drawer-list → dropdown → autocomplete`. Removing just that one edge flips `forms` from kept → dropped everywhere:

| component | with edge | edge removed |
|---|---|---|
| `<CountryFlag />` | 44 | 4 |
| `<Dropdown />` | 44 | 26 |
| `<Autocomplete />` | 44 | 27 |

Not a correctness bug — over-keeping is the safe direction and styling stays intact — but the optimizer does little for the components people reach for most.

**Fix:** move `useCountries` (and the `countries` constant) out of `extensions/forms/Field/SelectCountry/` into `shared/`. Removes the false edge and untangles a base `components/` → `extensions/` layering inversion. `filter → forms` is legit (renders a Forms field), so leave that.

tujoworker:
Fixed the false dependency from `CountryFlag` to Forms reported above.

### How

- Moved `useCountries`, the country data, and the country locale data into `shared`.
- Updated all consumers to use the new paths.
- Added a manifest regression test to ensure `CountryFlag`, `Dropdown`, and `Autocomplete` no longer retain Forms.

### Result

| Component | Dependency count |
| --- | ---: |
| `CountryFlag` | 4 |
| `Dropdown` | 26 |
| `Autocomplete` | 27 |

langz:
> Fixed the false dependency from CountryFlag to Forms reported above.\n\nMoved useCountries, the country data, and country locale data into shared, then updated all consumers. Added a manifest regression test ensuring CountryFlag, Dropdown, and Autocomplete no longer retain Forms.\n\nGenerated dependency counts now match the expected result:\n- CountryFlag: 4\n- Dropdown: 26\n- Autocomplete: 27

Thanks!

langz:
Confirmed the relocation does the trick — `country-flag` no longer pulls in Forms, so `Dropdown`/`Autocomplete` drop from 44 → 26/27 kept components, and `runtimeCssContract` stays green (no real edge lost). Nice touch fixing the `currencies`/`allCurrencies` sibling imports too. Thanks for the quick turnaround! 🙌

PR #8939, opened 2026-08-10
feat(HeightAnimation): support hidden until found (`hidden="until-found"`)

tujoworker:
## Summary

- add opt-in `hidden="until-found"` support to HeightAnimation
- coordinate find-in-page behavior with opening and closing animations
- document accessibility guidance and add an interactive example
- add regression tests for lifecycle and `beforematch` handling

## Preview

- https://feat-height-animation-until.eufemia-e25.pages.dev/uilib/components/height-animation/#find-collapsed-content

### DIFF
--- a/packages/dnb-eufemia/src/components/height-animation/HeightAnimation.tsx
+++ b/packages/dnb-eufemia/src/components/height-animation/HeightAnimation.tsx
@@ -1,5 +1,5 @@
 import withComponentMarkers from '../../shared/helpers/withComponentMarkers'
-import { useEffect, useRef } from 'react'
+import { useEffect, useRef, useState } from 'react'
 import type { HTMLProps, RefObject } from 'react'
 import { clsx } from 'clsx'
 import type { UseHeightAnimationOptions } from './useHeightAnimation'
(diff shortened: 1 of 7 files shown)


snorrekim:
Bug: content is not findable after user closes it again. I believe the added `style` attribute causes the bug:

```tsx
<div
  class="dnb-space dnb-height-animation dnb-height-animation--hidden"
  aria-hidden="true" 
  id="until-found-content" 
  style="height: 0px; visibility: hidden; overflow-y: clip;" // added after user closes
  hidden="until-found">
```

snorrekim:
Interesting side note: `<details>` has findability built in. If you have a 
```
<details>
  <summary>Title</sumary>
  Findable text
</details>
```

It will expand when found with cmd+f (sets `<details open>`). Unless the content has `visibility: hidden`.

packages/dnb-eufemia/src/components/height-animation/HeightAnimation.tsx
```diff
@@ -7,13 +7,24 @@ import { useHeightAnimation } from './useHeightAnimation'
...
 
+  /**
+   * Set to `true` to keep closed content available to the browser find-in-page feature with `hidden="until-found"`. This implies `keepInDOM`. Defaults to `false`.
+   */
+  untilFound?: boolean
+
+  /**
+   * Is called when the browser finds matching content inside a closed animation using `untilFound`.
+   */
+  onBeforeMatch?: (event: Event) => void
```
langz, line 26: I guess `onBeforeMatch` is mandatory in practice but documented as optional, when `untilFound` is used?

Omit it and find-in-page reveals nothing at all: the browser removes hidden, display: none comes straight back, and React never re-renders. Worth a development-time warning when `untilFound` is set without it, and clearer wording in the props table than "optional".

The mechanism, if it helps: collapsed content sits at `.dnb-height-animation--hidden { display: none }`, and the only thing flipping it to display: block is the `&[hidden='until-found']` override. On a match, both the browser and our own `beforematch` handler drop the hidden attribute — so the override stops matching and it falls straight back to display: none. And since `open` stays false, nothing re-renders to bring the findable state back.

Worth noting it also undoes the native behavior: a plain `hidden="until-found"` reveals fine on its own; here the extra display: none means the section only shows once `open` flips true, which is the bit `onBeforeMatch` is meant to do.

Confirmed in Chromium with a text-fragment reveal: wire up `onBeforeMatch` to set `open` and it expands and scrolls into view; leave it out and the element ends up display: none, height 0, and the page never scrolls to it.

Aside from the dev warning, maybe the component could just reveal itself on `beforematch` (drop the `--hidden` state internally), so `onBeforeMatch` can stay a genuinely optional callback instead of being load-bearing?

langz (approved):

PR #8940, opened 2026-08-10
feat(List): support element prop on List.Item to allow non-li rows

langz:
Motivation: @karini6 

## Summary

A consumer asked to reuse the `List` component's row/cell layout inside a `Dropdown` (and, by extension, `Autocomplete`) option. Those options are already an `<li role="option">` whose content is wrapped in a `<span>`, and `List.Item.Basic` renders its own `<li class="dnb-list__item">` with `<div>` cells — so the naive composition produces both invalid nested `<li>` elements and block elements inside a `<span>`.

This PR makes the escape hatch explicit: `List.Item.Basic` (via `ItemContent`) and the presentational cells (`List.Cell.*`) take an `element` prop. Set `element="span"` on the row and its cells to reuse the List row layout inside markup that already provides the list item, keeping the markup valid phrasing content.

**Note on prior behaviour:** `element` was already reachable on the row, because `ItemContentProps` extends `FlexContainerAllProps` and the spread of the remaining props came after the hardcoded `element="li"` — so a consumer-supplied `element` already won, and the type already accepted it. What this PR changes:

- The prop is now intentional, typed with JSDoc, documented and tested, instead of an accident of prop ordering — on both the row (`List.Item.Basic`) and the presentational cells (`List.Cell.Start`/`Center`/`Title`/`End`).
- It fixes a real bug: `element={undefined}` — what you get when forwarding an optional prop — used to fall through to the Flex container's own `'div'` default and render a `<div>` instead of an `<li>`. With the explicit `element = 'li'` default it now renders an `<li>`. Covered by a test.
- The `title` and `icon` convenience props on `List.Item.Basic` now follow the row `element`, so `<List.Item.Basic element="span" title="…" />` renders a `<span>` title (not a `<div>`) rather than silently reintroducing the invalid nesting the prop exists to avoid.

## What changed

- **`List.Item` `element` prop** — `ItemContent` forwards an `element` prop (default `li`) to the underlying Flex container. `List.Item.Basic` also forwards it to its auto-rendered icon and title so they stay valid phrasing content inside a `span` row. Fully backwards compatible.
- **`List.Cell.*` `element` prop** — the presentational cells (`Start`, `Center`, `Title`, `End`) take a typed, documented `element` prop (default `div`). Set it to `span` inside an option.
- **Types & docs** — typed with JSDoc, documented in the List properties tables, plus a new "Rendering a row outside a `List.Container`" section in the List docs covering what the row does and does not get, the width requirement, and the plain-text props needed for value and search.
- **Dropdown** — new "Dropdown with List item content" demo + integration tests (assert no nested `<li>`, no `<div>`, cells render, axe clean).
- **Autocomplete** — the same pattern works out of the box because `Dropdown` and `Autocomplete` share the `DrawerList` engine and data model (`AutocompleteData = DrawerListData`). Added a demo + integration tests, including one asserting the input value after selecting. No source change was needed for Autocomplete.

## Using it in an option — three things to get right

1. **Give the row width.** A List row keeps the end cell at content width and the title in a flexible column, so a short row squeezes the title. In the default 256px drawer the title column collapsed to 46px and "Accounts" wrapped as "Acc/oun/ts". Both demos now set a width (22rem) with the documented custom-width pattern. This is not caused by `element="span"` — a row inside a `List.Container` behaves the same when squeezed.
2. **Don't re-apply the list surface.** Outside a `List.Container` the row gets the layout and cell styling but no row height, padding, background or border — which is what you want, since the option supplies its own. Setting `variant` to bring them back paints a light background on top of the option, and the white text of a selected option then sits on white.
3. **Provide plain text.** Text passed through the `title` prop is invisible to the value and search text that `Dropdown`/`Autocomplete` derive from `content`, so provide `selectedValue` (and `searchContent` for `Autocomplete`).

## Why this layer

The `element` prop lives on the **List** side (the reusable content), not on the shared `DrawerList` fragment. `DrawerList` powers both `Dropdown` and `Autocomplete` and owns strict `listbox`/`option` ARIA semantics and a grid-based option layout, so it is intentionally left untouched. The composition approach solves the need for any host that already provides an `<li>`.

## Example

Dropdown:

```jsx
<Dropdown
  data={[
    {
      selectedKey: 'accounts',
      selectedValue: 'Accounts',
      content: (
        <List.Item.Basic element="span">
          <List.Cell.Title element="span">Accounts</List.Cell.Title>
          <List.Cell.End element="span">Bills, Savings</List.Cell.End>
        </List.Item.Basic>
      ),
    },
  ]}
/>
```

Autocomplete (provide `selectedValue` and `searchContent` so the input value and filtering keep working with rich content):

```jsx
<Autocomplete
  data={[
    {
      selectedKey: 'accounts',
      selectedValue: 'Accounts',
      searchContent: 'Accounts Bills Savings',
      content: (
        <List.Item.Basic element="span">
          <List.Cell.Title element="span">Accounts</List.Cell.Title>
          <List.Cell.End element="span">Bills, Savings</List.Cell.End>
        </List.Item.Basic>
      ),
    },
  ]}
  label="Label"
/>
```

## Accessibility note

`Dropdown`/`Autocomplete` are `listbox`/`combobox`; options must stay flat `<li role="option">` with phrasing content. Only reuse the presentational `List.Cell.*` this way — do not place interactive items (`List.Item.Action`, `List.Item.Accordion`) or a nested `List.Container` inside an option.

## Follow-up

Both demos carry a `data-visual-test` marker but no screenshot test: the baselines have to be generated in an environment that reproduces the committed ones, which this branch does not include.

## Breaking changes

None. The `element` prop defaults to `li` on the row and `div` on the cells, so all existing usage renders identically.

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/components/autocomplete/Examples.tsx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/components/autocomplete/Examples.tsx
@@ -725,7 +725,9 @@ export const AutocompleteWithListItemContent = () => (
         const data = [
           {
             selectedKey: 'accounts',
+            // selectedValue is the plain text shown in the input once selected;
             // searchContent keeps typing/filtering working with rich content
+            selectedValue: 'Accounts',
             searchContent: 'Accounts Bills Savings',
             content: (
               <List.Item.Basic element="div" title="Accounts">
@@ -735,6 +737,7 @@ export const AutocompleteWithListItemContent = () => (
           },
           {
             selectedKey: 'loans',
+            selectedValue: 'Loans',
             searchContent: 'Loans Mortgage Car',
             content: (
               <List.Item.Basic element="div" title="Loans">
@@ -744,6 +747,7 @@ export const AutocompleteWithListItemContent = () => (
           },
           {
             selectedKey: 'cards',
+            selectedValue: 'Cards',
             searchContent: 'Cards Visa Mastercard',
             content: (
               <List.Item.Basic element="div" title="Cards">
--- a/packages/dnb-design-system-portal/src/docs/uilib/components/autocomplete/Examples.tsx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/components/autocomplete/Examples.tsx
@@ -720,8 +720,21 @@ export const AutocompleteContentDecoupledExample = () => (
 
 export const AutocompleteWithListItemContent = () => (
   <Wrapper>
-    <ComponentBox scope={{ List }}>
+    <ComponentBox
+      scope={{ List }}
+      data-visual-test="autocomplete-list-item-content"
+    >
       {() => {
+        // A List row keeps its end cell at content width, so the row needs
+        // horizontal room. Give the Autocomplete a width of its own, or the
+        // title column will be squeezed by the end cell.
+        const AccountAutocomplete = styled(Autocomplete)`
+          .dnb-autocomplete__shell,
+          .dnb-drawer-list__root {
+            width: 22rem;
+          }
+        `
+
         const data = [
           {
             selectedKey: 'accounts',
@@ -757,7 +770,7 @@ export const AutocompleteWithListItemContent = () => (
           },
         ]
 
-        return <Autocomplete data={data} label="Label" />
+        return <AccountAutocomplete data={data} label="Label" />
       }}
     </ComponentBox>
   </Wrapper>
--- a/packages/dnb-design-system-portal/src/docs/uilib/components/autocomplete/Examples.tsx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/components/autocomplete/Examples.tsx
@@ -17,6 +17,7 @@ import {
   Flex,
   IconPrimary,
   NumberFormat,
+  List,
 } from '@dnb/eufemia/src'
 
 const Wrapper = styled.div`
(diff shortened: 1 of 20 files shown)


tujoworker:
I think we should use `span` so we don't divs in a span. Yes. A <span> may contain only phrasing content, while <div> is flow content. Browsers render it anyway, but the markup is non-conforming HTML and axe won’t flag it.

Screenshot 2026-08-11 at 07 36 40

Made a commit.

Preview:

- https://feat-list-item-element-prop.eufemia-e25.pages.dev/uilib/components/dropdown/#dropdown-with-list-item-content
- https://feat-list-item-element-prop.eufemia-e25.pages.dev/uilib/components/autocomplete#autocomplete-with-list-item-content

tujoworker (approved):

langz:
> I think we should use `span` so we don't divs in a span. Yes. A may contain only phrasing content, while
> 
> is flow content. Browsers render it anyway, but the markup is non-conforming HTML and axe won’t flag it.
> Screenshot 2026-08-11 at 07 36 40
> Made a commit.
> 
> Preview:
> 
> * https://feat-list-item-element-prop.eufemia-e25.pages.dev/uilib/components/dropdown/#dropdown-with-list-item-content
> * https://feat-list-item-element-prop.eufemia-e25.pages.dev/uilib/components/autocomplete#autocomplete-with-list-item-content

Thanks, great improvement 👏

langz:
I've asked @karini6, which came with this great feature request, to also take a look at this PR 🙏

PR #8942, opened 2026-08-10
feat(Flex): add opt-in CSS gap layout engine

tujoworker:
## Summary

Introduces a native CSS gap engine for Flex through `layoutEngine="css"`.

The legacy engine remains the default for backwards compatibility. Making CSS gap the default now would change child rendering, spacing geometry, and existing layouts. Migration guidance and the planned breaking changes are documented in `v13-info`.

## Next breaking change

For v13:

1. Make the CSS engine the default.
2. Deprecate the legacy child-inspection APIs.
3. Migrate internal layouts and later remove the legacy engine.

Enabling it by default would be a breaking change because it changes:

- Rendered DOM structure and generated wrappers
- Spacing and divider behavior
- Layout geometry
- Behavior of `_supportsSpacingProps`, `Flex.withChildren`, and `wrapChildrenInSpace`

Therefore, existing layouts keep the legacy engine by default. CSS gap is available through `layoutEngine="css"` for gradual migration, then becomes the default in v13.

## Next step

Opt selected layouts into `layoutEngine="css"`, verify their visuals, and remove legacy-only wrappers and markers where possible.

## How to test

- Run the Flex and affected component unit tests.
- Run visual regression tests for spacing, wrapping, dividers, and form alignment.
- Verify both UI themes.

## Motivation

https://dnb-it.slack.com/archives/CMXABCHEY/p1785921129466409


packages/dnb-design-system-portal/src/docs/uilib/about-the-lib/releases/eufemia/v13-info.mdx
```diff
@@ -49,6 +50,27 @@ $ yarn add @dnb/eufemia@13
...
+
+If a wrapper's children need their own Flex layout, render an explicit `Flex.Container` inside the wrapper instead of using `Flex.withChildren`.
+
+The previous child-inspection engine remains temporarily available during migration:
+
+```tsx
+<Flex.Container layoutEngine="legacy">...</Flex.Container>
+```
+
+Do not use the deprecated APIs for new integrations. See the [Flex.Container migration guidance](/uilib/layout/flex/container/#legacy-compatibility-mode) for details.
```
langz: Seems like this URL is not pointing to any location that exists (404)

langz: Perhpas it should be?

```suggestion
Do not use the deprecated APIs for new integrations. See the [Flex.Container migration guidance](/uilib/layout/flex/container/#backwards-compatibility) for details.
```

tujoworker: Good catch. Updated the link to the `#backwards-compatibility` section in a571043cd5e.

langz:
Is it safe to remove the divider elements, could that cause changes in the accessibility tree?
If so, perhaps we should mentioned in the migration notes.

tujoworker:
An inserted `<hr>` would become a flex item and affect gaps, wrapping, sizing, and alignment. Negative margins may work for simple vertical layouts, but not reliably for horizontal or wrapped layouts. CSS-painted dividers avoid changing the layout; semantic separators should instead be added explicitly with `<Hr />`.

langz (approved):

langz:
Following up on the divider work — I think the horizontal wrap-detection in `FlexDividers` can silently drop dividers under any cross-axis alignment other than `flex-start`/`stretch`. Wrapping is inferred from the difference in `top` between adjacent items:

```ts
const isWrapped = isHorizontal
  ? Math.abs(rect.top - previousRect.top) > 0.5
  : rect.top <= previousRect.top
```

In a `row` container `top` is the *cross* axis, so it legitimately varies for same-row items — under non-`flex-start`/`stretch` alignment, or with a block-start margin. Those items then read as "wrapped" and their divider is skipped. I checked it against the deployed branch preview — forcing a real `divider="line"` CSS container to `direction: row` — plus a standalone repro of the algorithm, and they line up:

| horizontal config | dividers painted |
|---|---|
| `align="flex-start"` (default) / `align="stretch"` | all ✅ |
| `align="center"`, uneven item heights | none ❌ |
| `align="flex-end"`, uneven item heights | none ❌ |
| `align="baseline"`, differing baselines (e.g. mixed font sizes) | none ❌ |
| per-item `alignSelf="center"`/`"flex-end"` on an item | the dividers on *both* sides of that item ❌ |

`divider="line"` and `divider="line-framed"` share the exact same detection code (`FlexDividers` doesn't branch on the divider type — the framing is CSS-only), so both are affected. It only surfaces when adjacent items actually end up at different `top`: `center`/`flex-end` are safe with equal-height items, whereas `baseline` needs matching baselines (equal box heights isn't enough). A vertically-centred row of unequal-height items separated by vertical rules feels like a fairly ordinary pattern, so I don't think it's purely theoretical.

The vertical branch keys off the main axis, which alignment doesn't touch — could we detect the horizontal row break from the main-axis/inline position instead of `top` (or otherwise guard the non-`flex-start` case)? If we go the inline-position route it'd want a guard for `dir="rtl"`, where the inline axis runs the other way. Happy for this to be a follow-up rather than a blocker, since it only affects the opt-in CSS engine and the default `flex-start` is safe.

tujoworker:
Thanks — this was valid. I changed horizontal wrap detection to use main-axis progression instead of cross-axis `top`, added RTL handling, and added regression coverage for `center`, `flex-end`, `baseline`, per-item `alignSelf`, actual wrapping, and RTL. Fixed in 46d0f0d1a8f.

langz:
> Thanks — this was valid. I changed horizontal wrap detection to use main-axis progression instead of cross-axis `top`, added RTL handling, and added regression coverage for `center`, `flex-end`, `baseline`, per-item `alignSelf`, actual wrapping, and RTL. Fixed in 46d0f0d1a8f.

Thanks @tujoworker ✊

PR #8944, opened 2026-08-10
feat(List): add striped rows

tujoworker:
Adds an opt-in `striped` prop to `List.Container`

Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1786365892411519


langz (approved):
Looks good to me, as long as ux or us see the need for it, then I guess it's good 😊

PR #8946, opened 2026-08-11
docs(Flex): v13 migration snippets + fix keep CSS dividers under non-stretch alignment

langz:
## Summary

Follow-up to #8942, targeting its branch (`refactor/flex-css-gap`). Two related changes to the CSS-gap Flex work: concrete migration docs, plus a fix for a divider bug found while reviewing.

## 1. Docs — concrete v13 migration snippets (`v13-info.mdx`)

The Flex section explained *what* to migrate but gave no code. Added worked **before/after** snippets for the two most common cases:

- **Custom components** — `_supportsSpacingProps` -> `useSpacing` on the rendered root. Clarifies the minimal case: for a component that already applies spacing to its root, deleting the marker is the only change; spacing props stay additive to the container gap in CSS mode.
- **Wrapper components** — `Flex.withChildren` -> an explicit nested `Flex.Container`. Documents the behavior change flagged in the #8942 review: the legacy engine *hoisted* a wrapper's children into the parent's spacing line, whereas the CSS engine treats the wrapper root as one flex item and lays its children out as a **nested group**. Transparent wrappers (Fragment/provider) need no inner container.

Import path and Before/After format match the existing `v11-info` notes and the shipped `useSpacing` example.

## 2. Fix — CSS dividers dropped under non-stretch alignment (`FlexDividers`)

Also from the #8942 review: the divider engine inferred flex-line wrapping from the raw `top` delta between adjacent items. Under any cross-axis alignment other than `stretch`/`flex-start` (`center`, `flex-end`, `baseline`, or a per-item `alignSelf`), same-line items legitimately sit at different `top` values, so they were misread as *wrapped* and their dividers were **silently dropped** — for both `divider="line"` and `divider="line-framed"`.

**Fix:** detect wrapping from the cross axis instead. Items on the same flex line always overlap on the cross axis regardless of alignment or writing direction, whereas a wrapped item is pushed entirely past the previous item. The vertical branch keeps its main-axis `top` comparison (alignment never affects it).

- Extracted the decision into a pure, exported `isItemWrapped` helper.
- Added unit tests covering stretch/flex-start/center/flex-end/alignSelf (painted) and real wraps in both directions (dropped).
- Default `flex-start` and `stretch` layouts are unchanged, so existing divider screenshots are unaffected.

### Verification

- New + existing Flex tests pass locally (`FlexDividers` 9, `Container` 61, `Item` 21).
- `prettier --check` clean on all changed files.
- Full geometry is exercised by the existing browser screenshot suite / branch preview; the unit test locks the decision boundary that regressed.

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/about-the-lib/releases/eufemia/v13-info.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/about-the-lib/releases/eufemia/v13-info.mdx
@@ -98,7 +98,7 @@ function MyComponent({ className, ...props }) {
 }
 ```
 
-For a component that already applied spacing to its root, deleting the marker is the only change. The rendered `<div>` is now a flex item automatically, and `useSpacing` keeps the component's own `top`, `right`, `bottom`, and `left` props working — in CSS mode those values are added to the container gap. If a component did not previously apply spacing to its root, adopt `useSpacing` as shown so authors can keep adjusting spacing per instance. A component that never needs per-instance spacing can simply drop the marker and receive the container gap.
+For a component that already applied spacing to its root, deleting the marker is the only change. The rendered `<div>` is now a flex item automatically, and `useSpacing` keeps the component's own `top`, `right`, `bottom`, and `left` props working — in CSS mode those values participate in pairwise spacing and may replace the adjacent container gap. If a component did not previously apply spacing to its root, adopt `useSpacing` as shown so authors can keep adjusting spacing per instance. A component that never needs per-instance spacing can simply drop the marker and receive the container gap.
 
 #### Wrapper components
 
--- a/packages/dnb-design-system-portal/src/docs/uilib/about-the-lib/releases/eufemia/v13-info.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/about-the-lib/releases/eufemia/v13-info.mdx
@@ -61,7 +61,7 @@ Flex now uses native CSS gap and renders React children unchanged. The following
 
 #### Custom components
 
-The `_supportsSpacingProps` marker is gone. A custom component becomes a flex item through its rendered root, so no marker is needed to participate in the layout. When the component should accept [spacing properties](/uilib/layout/space/), apply them to its root with `useSpacing` so an author can still adjust the spacing per instance.
+The `_supportsSpacingProps` marker is no longer needed. A custom component becomes a flex item through its rendered root, so nothing needs to opt it in to the layout. When the component should accept [spacing properties](/uilib/layout/space/), apply them to its root with `useSpacing` so an author can still adjust the spacing per instance.
 
 **Before (v11):**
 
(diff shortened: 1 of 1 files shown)


tujoworker (changes requested):
Thanks for the follow-up. The migration examples are useful, but I found one blocker in the divider fix: cross-axis rectangle overlap does not reliably identify flex-line membership.

packages/dnb-eufemia/src/components/flex/FlexDividers.tsx
```diff
@@ -0,0 +1,178 @@
...
+ * For vertical layouts the cross axis is horizontal and never affects `top`, so
+ * the main-axis `top` comparison remains correct.
+ */
+export function isItemWrapped(
+  rect: WrapRect,
+  previousRect: WrapRect,
+  isHorizontal: boolean
+): boolean {
+  return isHorizontal
+    ? rect.top >= previousRect.bottom - 0.5
```
tujoworker: This still misclassifies valid same-line layouts. Cross-axis rectangles do not always overlap: e.g. a row with `align="flex-start"`, one short item, and the next item using `alignSelf="flex-end"` can produce `previous.bottom === current.top`, so this returns `true` and drops the divider even though neither item wrapped. Auto cross-axis margins can create the same separation. The unit case named `alignSelf` only uses a previous item that fills the whole line, so it misses this. Please derive the break from main-axis progression (with RTL handling), or otherwise use actual flex-line bounds, and add a regression where adjacent same-line items occupy disjoint cross-axis bands.

packages/dnb-design-system-portal/src/docs/uilib/about-the-lib/releases/eufemia/v13-info.mdx
```diff
@@ -49,6 +50,120 @@ $ yarn add @dnb/eufemia@13
...
+
+  return <div {...params} />
+}
+```
+
+For a component that already applied spacing to its root, deleting the marker is the only change. The rendered `<div>` is now a flex item automatically, and `useSpacing` keeps the component's own `top`, `right`, `bottom`, and `left` props working — in CSS mode those values are added to the container gap. If a component did not previously apply spacing to its root, adopt `useSpacing` as shown so authors can keep adjusting spacing per instance. A component that never needs per-instance spacing can simply drop the marker and receive the container gap.
+
+#### Wrapper components
+
+`Flex.withChildren` is removed. In the legacy engine the HOC hoisted a wrapper's children into the parent container's spacing calculation, so items rendered inside the wrapper were spaced as if they were direct children of the outer `Flex.Container`. To lay out a wrapper's children in v13, render an explicit `Flex.Container` inside the wrapper.
```
tujoworker: The legacy behavior is described incorrectly here. `Flex.withChildren` / `_supportsSpacingProps = "children"` did not hoist the wrapper children into the outer flex line. `Flex.Container.wrapChildren` recreated the wrapper with a nested `FlexContainer`, so the DOM was already outer container → wrapper root → inner container (the existing legacy snapshots assert exactly that). The explicit nested container is still the right replacement, but it should be explained as replacing the container that the HOC previously injected automatically; the claimed behavior difference on line 155 should be removed as well.

langz:
Thanks for the feedback :) 
I'll try to adjust it then🙏

langz:
I've now addressed your great feedback @tujoworker :)

Please consider taking a new look at this PR, or we could of course also close this if it's not needed in https://github.com/dnbexperience/eufemia/pull/8942

tujoworker (approved):

tujoworker (approved):

PR #8949, opened 2026-08-11
chore(security): harden npm release (least-privilege secrets, package validation, consumer smoke tests, OIDC)

langz:
> [!IMPORTANT]
> **This must land before the next release.** The long-lived `NPM_TOKEN` has already been deleted from the repository secrets, but `main`'s release workflow still expects it (`NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}` plus a `registry-url` `.npmrc`) — so a release from `main` will now fail to authenticate against npm. The Trusted Publishing (OIDC) cutover in this PR is what restores a working publish path.

Motivation: https://dnb-asa.atlassian.net/browse/EDS-951 (part of EDS-823).

Hardens the `@dnb/eufemia` npm release against supply-chain attacks and accidental package regressions, and moves publishing off a long-lived npm token — which now fails under npm's 2FA/OTP enforcement (`EOTP`) — onto short-lived OIDC.

- **npm Trusted Publishing (OIDC)** — the release authenticates with a short-lived, GitHub-issued `id-token` instead of a stored publish token, and publishes with provenance. Drops the token and the `registry-url` from `actions/setup-node`.
- **Least-privilege jobs** — `release.yml` splits into a build/validate job with only `contents: read` and a separate publish job that alone holds npm OIDC and GitHub write. Each secret is injected only into the step that uses it (unused `FIGMA_*` removed); `verify.yml` scopes its `pull-requests`/`statuses` writes to the `build-check` job only.
- **Isolated, tamper-resistant publish** — the publish job installs locked tooling with dependency lifecycle scripts disabled, restores the validated build artifact, re-materialises the release config from the trusted checkout, and publishes with npm lifecycle scripts disabled — so a compromised build job cannot change how the release is published.
- **Package-content validation** — `validate:package` runs `npm pack --dry-run` and blocks publish on files that must never ship (tests, stories, `.env`, OS junk like `Thumbs.db`), a missing declared entry point or CSS bundle, or a suspicious file count/size. Wired into `publish:prepare`. (`npm` itself always strips `.npmrc`, `node_modules` and `.DS_Store` from a pack, so those never reach the tarball; the security-relevant `.npmrc` is instead rejected by a filesystem check in the credentialed publish job — `npm publish` would otherwise apply a project-level `.npmrc`'s proxy/TLS/registry settings that `npm pack` hides from content validation.)
- **Consumer smoke tests** — pack `build/` once and production-build minimal Vite + Next.js consumers against that exact tarball (pinned toolchains via committed lockfiles + `npm ci`), including a tree-shaking budget on the Vite build. New `consumer-smoke.yml` (`contents: read`).
- **Supply chain & docs** — pin `pkg-pr-new`; add a "Release security" section to `contribute/deploy.mdx`.

### Release prerequisites

- [x] **npm Trusted Publisher for `@dnb/eufemia` configured** — `semantic-release` now authenticates via OIDC with a short-lived, GitHub-issued token. Setup + token-cleanup notes: https://github.com/dnbexperience/eufemia/pull/8949#issuecomment-5341888057
- [x] **Long-lived npm publish token (`NPM_TOKEN`) removed** from the repository secrets — confirmed done, so there is no post-merge token cleanup left. Until this PR lands, `main` has no npm credential at all: its release workflow still reads the deleted secret, so the next release from `main` fails at the publish step. Merge this before the next release.


langz:
> [!NOTE]
> **Superseded — this plan changed.** npm Trusted Publishing (OIDC) is **now implemented in this PR**, not a follow-up: the cutover (drops `NODE_AUTH_TOKEN`, adds `id-token: write`, `@semantic-release/npm` OIDC + provenance, removes `registry-url`) is part of #8949. The one manual, registry-side prerequisite and the post-merge token cleanup are tracked in the current checklist comment: https://github.com/dnbexperience/eufemia/pull/8949#issuecomment-5341888057
>
> The original "separate PR" note is kept below for history.

<details>
<summary>Original note (outdated)</summary>

## Next step: npm Trusted Publishing (OIDC) — separate PR

Publishing via **npm Trusted Publishing (OIDC)** with no long-lived `NPM_TOKEN` is **not part of this PR**; it will be delivered in its own PR (implementation + docs together). It has a registry-side prerequisite that can't be done in code, and flipping it prematurely would break releases.

### Why it needs its own PR
`npm publish` uses OIDC only when (a) the package has a **Trusted Publisher** configured on npmjs.com and (b) no auth token is present. Removing `NODE_AUTH_TOKEN` before the registry side is set up would fail the next release.

### Step 1 — one-time, npm package owner (on npmjs.com)
Configure a Trusted Publisher for `@dnb/eufemia`:
- Package settings → **Publishing access** → add **Trusted Publisher** → GitHub Actions
- Organization / repository: `dnbexperience/eufemia`
- Workflow filename: `release.yml`
- Environment: leave blank (unless we introduce one)

### Step 2 — workflow cutover (in the separate PR, after Step 1)
1. Ensure the runner uses npm 11.5+ (OIDC support).
2. Remove `NODE_AUTH_TOKEN` / `NPM_TOKEN` from the `Release` step — with `id-token: write` already set and no token present, `npm publish` exchanges the OIDC token automatically.
3. Verify `@semantic-release/npm` publishes via OIDC (bump the plugin if its `verifyConditions` still requires a token).
4. Once version tagging is decoupled from publish, move the release job toward `contents: read` + `id-token: write`.

### Verify after cutover
- A release publishes with **no** `NPM_TOKEN` in use.
- The published version shows **provenance** on npmjs.com (already enabled via `publishConfig.provenance: true`).

Ping me when Step 1 is done and I'll open the OIDC PR (cutover + docs) and validate it.


</details>

tujoworker:
The “Build once” section overstates the current guarantee. The smoke workflow and release workflow build independently, validate:package only runs npm pack --dry-run, and semantic-release later creates the published tarball. Could we reword this to avoid claiming artifact identity? Publishing the exact validated tarball is correctly listed as a follow-up.

tujoworker:
What happened here? 😅

Screenshot 2026-08-12 at 11 47 52

.github/workflows/consumer-smoke.yml
```diff
@@ -0,0 +1,48 @@
...
+# against the packed tarball to catch broken exports, types or CSS imports before
+# a release reaches consumers. Needs no secrets, so it runs with least privilege.
+
+on:
+  pull_request:
+    paths:
+      - 'packages/dnb-eufemia/**'
+      - 'smoke/**'
+      - '.github/workflows/consumer-smoke.yml'
+  workflow_dispatch:
```
tujoworker, line 15: This path filter misses several inputs to the publishable build, including the root `package.json`, `yarn.lock`, the shared setup action, and tools used by `build:ci`, such as `eufemia-llm-metadata`. Changes there can alter or break the packed artifact without running this check. Could we run this for every PR, or include all transitive build inputs?

packages/dnb-eufemia/scripts/postbuild/validatePackageContents.mjs
```diff
@@ -0,0 +1,145 @@
...
+ * (packageDir defaults to "build").
+ */
+
+import { execFileSync } from 'node:child_process'
+import path from 'node:path'
+import { pathToFileURL } from 'node:url'
+
+// Files that must always be present in the published package.
+export const REQUIRED_FILES = ['package.json']
+
```
tujoworker, line 29: `package.json` is effectively guaranteed by a successful `npm pack`, so this currently does not guard any usable entry point. A broken package with ten unrelated files would pass even if `index.js`, `index.d.ts`, or the main style bundles were missing. Could we validate at least `main`, `module`, `types`, and the concrete export targets from the built `package.json`, plus the essential CSS files?

packages/dnb-design-system-portal/src/docs/contribute/deploy.mdx
```diff
@@ -88,3 +88,39 @@ Run the steps and prepare the git branches as above, but before you push to orig
...
+- type checks (`test:types`)
+- dependency audit and lockfile checks (`audit:ci`, `yarn dedupe --check`)
+
+At release time, `release.yml` additionally enforces the dependency audit and the package-content validation before `semantic-release` publishes. Protect the release branches (`release`, `beta`, `alpha`, `next`, `*.x`) and require the `verify` checks so a publish cannot happen from an unchecked state.
+
+### Permissions and trigger
+
+The release job runs with `contents: write` (so `semantic-release` can push the changelog commit and version tag), `issues: write` and `pull-requests: write` (release comments), and `id-token: write` (provenance). Releases run only from the protected release branches on a clean CI checkout — never from a developer machine.
+
+### Build once, validate, smoke-test, publish
```
tujoworker: This is not quite true with the current workflows: the consumer tests build and pack during the PR, while the release workflow later performs a separate checkout/build and semantic-release packs again. The smoke-tested artifact is therefore not the exact artifact that gets published. Could we adjust this wording until the “publish the exact packed tarball” follow-up is implemented?

tujoworker (commented):
I found three places where the new release safeguards do not yet fully match their intended coverage or documentation.

langz:
Thanks for looking into this draft @tujoworker, I'll look into the comments provided :) 
As said earlier in the morning today, I would prefer to focus on https://github.com/dnbexperience/eufemia/pull/8812 and https://github.com/dnbexperience/eufemia/pull/8807 before continuing working with this draft #8949

.github/workflows/consumer-smoke.yml
langz: Good catch — dropped the `paths:` filter so this runs on every PR. `build:ci` also pulls in `tools/*` (eufemia-llm-metadata via build:docs, eufemia-css-optimizer via build:style-manifest) plus the root deps and setup action, so a narrow filter would miss those inputs. Now consistent with `verify.yml`.

packages/dnb-design-system-portal/src/docs/contribute/deploy.mdx
langz: Agreed — reworded to drop the artifact-identity claim. It now describes per-run build + validate/smoke and notes the smoke-tested tarball is not byte-for-byte the published one, with the exact-tarball publish called out as a follow-up.

langz:
Thanks @tujoworker — two from your top-level notes:

- **Build once wording** — reworded to avoid claiming the smoke-tested tarball is the published one (details in the inline thread).
- **Blank “Branch Preview URL (stable)”** — that comment is produced by `preview.yml`, which parses the `wrangler pages deploy` output (the stable alias comes from the “Deployment alias URL:” line). This PR does not touch `preview.yml`, so it is not caused by these changes — the alias line just was not emitted/parsed for that run. Happy to look into `preview.yml` separately if it keeps happening.

packages/dnb-eufemia/scripts/postbuild/validatePackageContents.mjs
langz: Good catch. Fixed: `validate:package` now checks that the `main`/`module`/`types`/`exports` targets declared in the built `package.json`, plus the essential CSS bundles, are all present in the packed tarball. Added unit tests too.

langz:
## Enabling npm Trusted Publishing (OIDC)

This PR switches the npm publish of `@dnb/eufemia` from a long-lived `NPM_TOKEN` to **Trusted Publishing (OIDC)**. The repo side is ready; one manual step on npm is required before the next release, plus a cleanup afterwards.

### 1. Required — configure the Trusted Publisher on npmjs.com
An **owner/admin of `@dnb/eufemia` on npm** needs to add a GitHub Actions trusted publisher for the package:

- npmjs.com → the `@dnb/eufemia` package → **Settings** → **Trusted Publisher** → **GitHub Actions**, with:
  - **Organization/owner:** `dnbexperience`
  - **Repository:** `eufemia`
  - **Workflow filename:** `release.yml`
  - **Environment:** leave **empty** (the release job does not use a GitHub Environment)

Do this **before merging / before the next release runs**. Until the trusted publisher exists, `semantic-release` has no npm credential and the publish step will fail.

### 2. After the first successful OIDC publish — cleanup
- Delete the now-unused `NPM_TOKEN` repo secret (GitHub → **Settings** → **Secrets and variables** → **Actions**). No long-lived publish token is needed anymore.

### Already handled in this PR (no action needed)
- `id-token: write` on the release job.
- `@semantic-release/npm@13.1.5` (OIDC support since 13.1.0) running on Node 24 / npm ≥ 11.5.1.
- `publishConfig`: `access: public`, `provenance: true` (provenance is auto-generated with trusted publishing).
- Removed `registry-url` from `actions/setup-node` — otherwise it writes an `.npmrc` with an empty `NODE_AUTH_TOKEN` that conflicts with OIDC (`EINVALIDNPMTOKEN`).

langz:
> ## Enabling npm Trusted Publishing (OIDC)
> This PR switches the npm publish of `@dnb/eufemia` from a long-lived `NPM_TOKEN` to **Trusted Publishing (OIDC)**. The repo side is ready; one manual step on npm is required before the next release, plus a cleanup afterwards.
> 
> ### 1. Required — configure the Trusted Publisher on npmjs.com
> An **owner/admin of `@dnb/eufemia` on npm** needs to add a GitHub Actions trusted publisher for the package:
> 
> * npmjs.com → the `@dnb/eufemia` package → **Settings** → **Trusted Publisher** → **GitHub Actions**, with:
>   
>   * **Organization/owner:** `dnbexperience`
>   * **Repository:** `eufemia`
>   * **Workflow filename:** `release.yml`
>   * **Environment:** leave **empty** (the release job does not use a GitHub Environment)
> 
> Do this **before merging / before the next release runs**. Until the trusted publisher exists, `semantic-release` has no npm credential and the publish step will fail.
> 
> ### 2. After the first successful OIDC publish — cleanup
> * Delete the now-unused `NPM_TOKEN` repo secret (GitHub → **Settings** → **Secrets and variables** → **Actions**). No long-lived publish token is needed anymore.
> 
> ### Already handled in this PR (no action needed)
> * `id-token: write` on the release job.
> * `@semantic-release/npm@13.1.5` (OIDC support since 13.1.0) running on Node 24 / npm ≥ 11.5.1.
> * `publishConfig`: `access: public`, `provenance: true` (provenance is auto-generated with trusted publishing).
> * Removed `registry-url` from `actions/setup-node` — otherwise it writes an `.npmrc` with an empty `NODE_AUTH_TOKEN` that conflicts with OIDC (`EINVALIDNPMTOKEN`).

@tujoworker, do you have time to try this out?
Then we can can re-run the action/job in this PR https://github.com/dnbexperience/eufemia/pull/9131 and see if it works or not? 🤞

langz:
> > ## Enabling npm Trusted Publishing (OIDC)
> > This PR switches the npm publish of `@dnb/eufemia` from a long-lived `NPM_TOKEN` to **Trusted Publishing (OIDC)**. The repo side is ready; one manual step on npm is required before the next release, plus a cleanup afterwards.
> > ### 1. Required — configure the Trusted Publisher on npmjs.com
> > An **owner/admin of `@dnb/eufemia` on npm** needs to add a GitHub Actions trusted publisher for the package:
> > 
> > * npmjs.com → the `@dnb/eufemia` package → **Settings** → **Trusted Publisher** → **GitHub Actions**, with:
> >   
> >   * **Organization/owner:** `dnbexperience`
> >   * **Repository:** `eufemia`
> >   * **Workflow filename:** `release.yml`
> >   * **Environment:** leave **empty** (the release job does not use a GitHub Environment)
> > 
> > Do this **before merging / before the next release runs**. Until the trusted publisher exists, `semantic-release` has no npm credential and the publish step will fail.
> > ### 2. After the first successful OIDC publish — cleanup
> > 
> > * Delete the now-unused `NPM_TOKEN` repo secret (GitHub → **Settings** → **Secrets and variables** → **Actions**). No long-lived publish token is needed anymore.
> > 
> > ### Already handled in this PR (no action needed)
> > 
> > * `id-token: write` on the release job.
> > * `@semantic-release/npm@13.1.5` (OIDC support since 13.1.0) running on Node 24 / npm ≥ 11.5.1.
> > * `publishConfig`: `access: public`, `provenance: true` (provenance is auto-generated with trusted publishing).
> > * Removed `registry-url` from `actions/setup-node` — otherwise it writes an `.npmrc` with an empty `NODE_AUTH_TOKEN` that conflicts with OIDC (`EINVALIDNPMTOKEN`).
> 
> @tujoworker, do you have time to try this out? Then we can can re-run the action/job in this PR #9131 and see if it works or not? 🤞

OIDC is now configured :)

langz (commented):
@langz — AI-agent review of current head `bd5ed39c`.

Two blocking gaps remain in the untrusted-artifact guard:

- `build/package.json.repository` is not compared with the trusted manifest. I reproduced `writeReleaseConfig` accepting an attacker URL and semantic-release sending a synthetic `GH_TOKEN` to that host after a Basic-auth challenge.
- Top-level `build/package.json.tag` is not rejected. I reproduced pinned `libnpmpublish` using `"tag":"latest"` instead of semantic-release’s trusted `alpha` tag, promoting a prerelease to `latest`.

Please validate these fields against the trusted source and set `repositoryUrl` explicitly. Focused tests pass (92/92); the `.env` and release-queue fixes look sound.

— RAIWork AI agent

langz:
Follow-up on this review — I verified both findings against the source and a local control.

## 1. `repository` not validated → token exfiltration — confirmed, and fixed

The chain holds: the release config did not pin `repositoryUrl`, semantic-release derives it from `package.json`'s `repository` before falling back to the git remote (`get-config.js`: `pkgRepoUrl(...) || repoUrl(...)`), and it embeds the token into the authenticated push URL (`get-git-auth-url.js` → `formatAuthUrl`). Because `writeReleaseConfig` did not assert `repository`, a tampered artifact could redirect the changelog/tag push and leak `GH_TOKEN` after a Basic-auth challenge — which breaks the "a compromised build job cannot change how the release is published" guarantee.

Fixed in 9465b19:
- **Pinned `repositoryUrl`** in the release config. `options.repositoryUrl` outranks `package.json`'s `repository`, so a tampered field can no longer redirect the push.
- **Added a `repository` assertion** in `writeReleaseConfig` against the trusted source (defence in depth, mirroring the existing `scripts`/`publishConfig` checks), plus unit + integration coverage.

## 2. Top-level `tag` → prerelease promoted to `latest` — could not reproduce

Two independent reasons:
- **`@semantic-release/npm` 13.1.5 does not use `libnpmpublish`.** Its `publish.js` shells out to `npm publish … --tag <distTag>`, where `distTag = getChannel(nextRelease.channel)` — derived from the branch/channel (trusted; the artifact can't change it) and passed **explicitly** on the CLI, which overrides any manifest value.
- **npm ignores a top-level `tag` field.** Local control: a package with top-level `"tag":"beta"` still publishes to `latest`, and an explicit `--tag alpha` wins. The only manifest field npm reads for the dist-tag is `publishConfig.tag`, which `writeReleaseConfig` already validates.

So there doesn't appear to be an artifact-controlled path to the dist-tag on the 13.1.5 publish path. Happy to revisit with a concrete repro against that path if you have one.

— RAIWork AI agent

langz:
@langz — reviewed the three follow-up commits (whole-manifest guard, fail-closed `is-main` + drift consolidation, non-regular-entry rejection). No critical issues.

I checked the two changes that could have rejected a *legit* release, empirically against a real ~27k-file build:
- **Whole-manifest guard:** accepts a faithful artifact — `findManifestMismatch` reports 0 differences, and the drift test pins `expectedReleaseManifest` to the real `prepareForRelease` output (95/95 in `writeReleaseConfig.test.ts`).
- **Non-regular-entry guard:** 0 findings on a real build, and the walk completes in ~260 ms.
- **`realpath` `is-main` fix:** correct and consistent in both scripts.

Good to merge once CI is green and it has an approving review.

— RAIWork AI agent

langz:
Final review — no blocking issues from my side.

The drift guard (now pinned to the real producer via the `prepareForRelease` run in the test), the non-regular-entry rejection, and the fail-closed self-invocation all check out and are well covered by tests. I verified the one false-positive risk directly: real builds carry zero symlinks/non-regular entries, so the new check only trips on a tampered artifact.

Only standing item before the first release: confirm the npm Trusted Publisher (OIDC) is configured, now that `NPM_TOKEN` is gone — safest to cut the first release on a prerelease channel (alpha/next).

langz:
Final review of this branch (3e226ea) — no blockers found.

Verified end-to-end rather than by reading: the manifest a real build produces still passes the publish guard, and tampered variants are refused (dist-tag, `private`, name, repository, re-added scripts, publishConfig, symlinked `CHANGELOG.md`). 118 postbuild tests pass, and CI is green apart from visual regression still running. Also checked that no build step creates symlinks, so the new entry-kind check should not block a release.

One note: #9222 is still open, so the artifact-root case is not in this branch yet.

tujoworker (approved):

PR #8957, opened 2026-08-12
chore(MCP): share docs server core with Lambda

tujoworker:
## Summary

- make the shared MCP docs core the single source of truth for local and Lambda serving
- remove the duplicated Lambda docs server and docs-source implementations
- retain the existing MCP SDK v1 protocol behavior
- carry the Lambda filesystem containment and compatibility fixes into the shared core


langz (approved):

tools/mcp-lambda/src/docs-server.ts
```diff
@@ -1,893 +0,0 @@
...
-}
-
-function createDocsContext(source: DocsSource) {
-  let cachedMarkdownFiles: string[] | null = null
-  let cachedAt = 0
-  const ttlMs = 30_000
-  const maxContentCacheEntries = 256
-  const contentCache = new Map<string, string | null>()
-
-  async function readCached(filePath: string): Promise<string | null> {
```
langz, line 238: Not sure if this cache functionality is something we should consider still having?

tools/mcp-lambda/src/docs-server.ts
```diff
@@ -1,893 +0,0 @@
...
-    {
-      title: 'Search docs',
-      description:
-        'Search across all markdown and MDX documentation using a free-text query, returning a JSON array of ranked matches with relevance scores and text snippets. Use this when you know what you are looking for conceptually (for example a component, feature, or concept name), but you do not know the exact file path yet. Prefer this after you have called the docs entry tool so you understand how the docs are structured. In particular, use this to find and read the appropriate getting started or first-steps documentation before you rely on any specific examples or code snippets.',
-      inputSchema: DocsSearchInput.shape,
-    },
-    async ({ query, limit, prefix }) => {
-      const hits = await context.searchInMarkdown(query, limit, prefix, {
-        concurrency: 12,
-        timeoutMs: 5000,
```
langz, line 773: Not sure if this timeout functionality is something we should consider still having?

tools/mcp-lambda/src/docs-server.ts
tujoworker: Good catch. I restored the bounded content cache in the shared core, so Lambda, local Node, and Worker adapters now use the same 256-entry LRU cache. It also caches missing-file results and is cleared when the markdown file list refreshes. I added tests for repeated reads and LRU eviction.

tools/mcp-lambda/src/docs-server.ts
tujoworker: This is still retained in the shared core. Search uses a deadline with bounded concurrency; the shared `docsSearch` handler currently passes `timeoutMs: 2000` (and the implementation clamps configured values to 200–15,000 ms). No further change was needed here.

PR #8960, opened 2026-08-12
ci(release): adopt npm Trusted Publishing (OIDC)

langz:
Adopts **npm Trusted Publishing (OIDC)** for `@dnb/eufemia` releases and documents it. Split out of #8949.

The `Release` step no longer uses a long-lived `NPM_TOKEN`: `@semantic-release/npm` exchanges the workflow's short-lived `id-token` for a scoped npm token at publish time, and every package is published with provenance.

## ⚠️ Merge gate (npm package owner)
**Do not merge or release until a Trusted Publisher for `@dnb/eufemia` is configured on npmjs.com** (repo `dnbexperience/eufemia`, workflow `.github/workflows/release.yml`). Until that exists an OIDC publish fails, so this PR is kept as a **draft**. Once configured, the next release from a protected branch publishes over OIDC with no stored secret.

## What's here
- **`release.yml`** — removes `NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}` from the `Release` step; keeps `id-token: write` (OIDC auth + provenance) and `GH_*` (GitHub release + changelog/tag push).
- **`deploy.mdx`** — a concise "Publishing method" section plus reconciled least-privilege/permission notes so the docs match the workflow.

## Why the tooling is ready
- `@semantic-release/npm` added Trusted Publishing in **v13.1.0**; the repo is on **13.1.5**, so the plugin performs the OIDC token exchange itself (passes the GitHub `id-token` as a bearer token) — the local npm CLI version is not the gating factor. Node is pinned to **24.16.0** (npm 11.x).
- `contents: write` stays on the release job because `@semantic-release/git` pushes the changelog commit and version tag; OIDC only removes the npm auth secret.

Base: `refactor/ci-release-least-privilege-secrets` (stacks on #8949); retarget to `main` after that merges.

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/contribute/deploy.mdx
+++ b/packages/dnb-design-system-portal/src/docs/contribute/deploy.mdx
@@ -100,7 +100,7 @@ The release pipeline (`.github/workflows/release.yml`) and this process are owne
 ### Controls in place
 
 - **Provenance** — packages are published with npm provenance (`publishConfig.provenance: true` plus `id-token: write` on the release job), so each release has a verifiable, signed link back to the source commit and workflow.
-- **Least-privilege secrets** — `release.yml` has no job-wide `env:`; each secret is injected only into the step that needs it (Algolia config for the portal build and search push; npm and GitHub credentials only for the publish step). Unused tokens are never exposed to the install, prebuild or build steps.
+- **Least-privilege secrets** — `release.yml` has no job-wide `env:`; each secret is injected only into the step that needs it (Algolia config for the portal build and search push; the GitHub token only for the publish step). npm needs no stored publish secret at all — it authenticates with a short-lived OIDC token (see [Publishing method](#publishing-method)). Unused tokens are never exposed to the install, prebuild or build steps.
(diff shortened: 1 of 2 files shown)


packages/dnb-design-system-portal/src/docs/contribute/deploy.mdx
```diff
@@ -124,3 +124,32 @@ The release job runs with `contents: write` (so `semantic-release` can push the
 ### Build once, validate, smoke-test, publish
 
 The library is built once into `build/`; the same built package is what `validate:package` checks, what the consumer smoke tests install, and what `semantic-release` publishes.
+
+### Publishing method: Trusted Publishing (OIDC)
+
+npm publishing is moving to **Trusted Publishing** (OpenID Connect) to remove the long-lived `NPM_TOKEN`. The release job requests a short-lived OIDC token from GitHub (`id-token: write`) and npm exchanges it for publish access — the same `id-token` mechanism that backs provenance — leaving no static publish secret to store, rotate or leak.
```
tujoworker: hmm, it don't feels right to read the docs mentioning a transition state. Does it give actually a value mentioning such details? I think for most devs it will not.

tujoworker (approved):
Have some thoughts about what we document and for "who" we do it. See my comment.

packages/dnb-design-system-portal/src/docs/contribute/deploy.mdx
langz: Good point — removed the transition framing, now just a concise steady-state note.

langz: Update: landed the actual `release.yml` cutover on this branch too — dropped `NPM_TOKEN` so the docs and the workflow now match. Keeping it as a draft until the npm Trusted Publisher is configured on npmjs.com.

PR #8962, opened 2026-08-12
chore(TabBar): make fullscreen control a real button so Space activates it

langz:
Issue: Pressing **Space** on the fullscreen toggle in the docs portal's `TabBar` did nothing — only **Enter** worked. For a control that looks and reads like a button, both keys should activate it.


tujoworker (approved):
In the upcoming portal designs, the whole button is gone. We may consider to remove it  and its functionality already now. I think the focus mode is a valid replacement for it.

langz:
It's an anchor today, and anchor's do not support space, so I guess I'll just keep it as it is.
However, it's feels more like a button than an anchor, anyways, no point making changes if this will be removed quite soon

PR #8963, opened 2026-08-12
chore(AGENTS.md): improve coding guidelines

tujoworker:
Clarifies expectations for coding, validation, and pull request workflows so AI-assisted changes stay focused and preserve user edits.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -11,7 +11,7 @@
 
 - Write as little code as possible to achieve the best result. Prefer simple solutions and remove unnecessary complexity.
 - Keep changes focused on the requested outcome. Preserve existing behavior and avoid unrelated refactors or scope creep.
-- Prefer extending or improving existing code over creating a parallel implementation. Introduce a new abstraction or system only when the existing code cannot reasonably support the requested outcome.
+- Find and understand the relevant existing code before implementing a solution. Extend or improve it when possible; otherwise replace it rather than creating a duplicate or parallel implementation.
 - Follow established patterns in the surrounding code unless the task requires changing them.
 - Consider correctness, security, performance, and accessibility for every change. Do not expose sensitive data, weaken safeguards, or add avoidable work to frequently used code paths.
 - Review the complete diff before handing off work or creating a pull request. Check for unintended changes and reassess correctness, readability, security, performance, accessibility, and test coverage.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -2,45 +2,75 @@
 
 ## Accuracy and Documentation
 
-- Do not hallucinate - Use the documentation exactly as provided.
-- Write concise code and add comments only when they clarify intent.
-- Gather all required information from the documentation before using it as a reference.
-- Do not make assumptions or infer missing details unless explicitly instructed to do so.
-- Do NOT update changelog files (such as `EUFEMIA_CHANGELOG.mdx` or `changelog.mdx`) unless the user explicitly asks you to. Changelog entries should be written by humans.
+- Base decisions on the repository source and its documentation. Do not invent APIs, behavior, or requirements.
+- Gather the necessary context before making changes. Ask for clarification when required information is missing.
+- Add comments only when they clarify intent that the code cannot express on its own.
+- Do not update changelog files, such as `EUFEMIA_CHANGELOG.mdx` or `changelog.mdx`, unless the user explicitly asks, and preserve changelog changes already made by the user.
 
-## Code Style Rules
+## Working on Code
 
-- Write code in logical blocks.
-- Keep blank lines between logical blocks to ensure better readability.
-- Follow existing Prettier and ESLint configurations, and run them on files after making changes.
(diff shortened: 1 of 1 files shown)


langz (approved):
Interesting, looks good, let's give it a go 🚀 

When reading this, both the existing one and the new one, I also think a lot of this information is to be find in our other docs. I'm wondering how important it is to have information listed in both docs and in AGENTS.MD file. One would guess that an agent finds the information in both places. Or perhaps it prefers AGENTS.MD file 🤖

AGENTS.md
```diff
@@ -2,45 +2,75 @@
...
+- Follow the documented unit guidance and use `rem` by default.
+- In component SCSS, use `rem` values directly instead of `var(--spacing-*)`. Spacing custom properties are reserved for the spacing infrastructure in `space/`, `flex/`, and `grid/`.
+- Spacing reference: `xx-small` = `0.25rem`, `x-small` = `0.5rem`, `small` = `1rem`, `medium` = `1.5rem`, `large` = `2rem`, `x-large` = `3rem`, and `xx-large` = `3.5rem`.
+- Always use braces for control-flow statements, including single-line bodies.
+- An early return that breaks control flow may use `// stop here` when the comment clarifies intent.
+
+## Tests and Validation
+
+- Write tests for new functionality and bug fixes. Tests are a contract for the intended behavior.
+- For a bug fix, write a failing regression test before implementing the fix.
```
joakbjerk: Maybe add a line about verifying that the test fails without the fix? My agent has skipped that sometimes, not running the test after creating it 😅

AGENTS.md
```diff
@@ -2,45 +2,75 @@
...
-- Tests act as a contract to verify the intention of the functionality.
-- Write tests before implementing a fix - this ensures the test captures the expected behavior and verifies the fix works correctly.
-- Prefer document.querySelector over screen from @testing-library/react.
+### Creating Pull Requests
 
-## Commit Messages
+- Before creating a pull request, check whether one already exists for the branch and review the complete diff against `main`.
+- Format the title according to the [Git convention style guide](packages/dnb-design-system-portal/src/docs/contribute/style-guides/git.mdx). Use a Conventional Commit decorator and, when targeting a component, a PascalCase scope.
+- Examples: `fix(Button): prevent double click submission` and `feat(DatePicker): add month-only mode`.
+- For extensions/forms, use the compound name, such as `feat(Field.Date): ...` or `fix(Forms): ...`.
```
langz:
```suggestion
- For extensions/forms, use the compound name, such as feat(Field.Date): ... or fix(Form.Section): .... Use Forms as the scope for changes that span the whole forms extension.
```

Just a suggestion.
Personally, I find it a bit difficult from time to time to understand when to use `x(Forms)` instead of a compound name like `x(Form.Section)`

langz (approved):

AGENTS.md
```diff
@@ -2,45 +2,75 @@
...
+
+## Git and Pull Requests
+
+- Never commit directly to `main` or release branches such as `release`. Create a branch and open a pull request.
+
+### Commits
 
-### Test Requirements
+- Write concise commit messages in the imperative mood.
+- Do not add decorators such as `fix:`, `feat:`, or `chore:` to individual commits. Use a Conventional Commit decorator only in the pull request title.
```
langz, line 51: I'm fine with this. But why wouldn't we want to follow the same standard for commit messages in individual commits vs the commit message that get merged?

The following scenarios is not a problem but if one merge branch a to branch b which has an open pr to main, then branch b will now contain a conventional commit

joakbjerk (approved):
🤖

AGENTS.md
tujoworker: I don't see the value for individual commits having a decorator. At the end, when squashing, they are anyway not a part of main. 

Do you find it useful?

langz: Not particularly, but I don't see why we should use multiple standards for commit messages instead of having one that's always used.

AGENTS.md
tujoworker: Good point. Updated the guideline to require running the regression test before the fix and verifying that it fails for the expected reason.

tujoworker: Added:

> and verify that it fails for the expected reason.

snorrekim:
One issue I've been having is how the AI sometimes ends up reinventing the wheel. Right now, working on table of contents, I have been fighting tooth and nail to get the AI to improve the existing code, instead of making a whole new system.

```
- Write as little code as possible to achieve the best result. Prefer simple solutions and remove unnecessary complexity.
- Keep changes focused on the requested outcome. Preserve existing behavior and avoid unrelated refactors or scope creep.
- Follow established patterns in the surrounding code unless the task requires changing them.
```

These items indirectly covers the issue, but I would like if it was more explicit. Though I am unsure about the best wording.

tujoworker:
Good point. I made this explicit: prefer extending or improving existing code over creating a parallel implementation, and only introduce a new abstraction or system when the existing code cannot reasonably support the requested outcome.

PR #8977, opened 2026-08-13
docs(HeightAnimation): document untilFound browser-support limitation

langz:
## Motivation

Alternative to #8976.

`untilFound` relies on the native `hidden="until-found"` behavior, which is only supported in some browsers (Chromium 102+, Firefox 148+, with partial support in Safari 26.2+). Where it is unsupported, a closed `untilFound` section is **not reliably hidden** — the collapsed content can remain visible — because the hiding depends on the browser applying `content-visibility` for `hidden="until-found"`.

#8976 **fixes** this in code (feature-detect support and fall back to `keepInDOM`). This PR is the **document-instead-of-fix** alternative: it keeps the current behavior and instead documents the limitation — in the component guide and the `untilFound` prop description — so consumers can opt in consciously.

Merging either one makes the other redundant. This lets you decide whether to fix the behavior or document the limitation.


tujoworker (approved):
yeah, I have tested it in Safari. It did work fine, except that the found text was not highlighted/colorized. 🌈

langz:
> yeah, I have tested it in Safari. It did work fine, except that the found text was not highlighted/colorized. 🌈

Good good, in older versions of Safari or v26?

tujoworker:
v26 – I did read this before I went for it. The partial support was scary, but luckly it was "working" ok. I'm fine with the extra docs. ✅

PR #8979, opened 2026-08-13
refactor(Autocomplete): extract text highlighting into a shared engine

joakbjerk:
Motivation: https://dnb-asa.atlassian.net/jira/software/c/projects/EDS/boards/734?assignee=712020%3A11a274b8-3b2a-4c7f-9968-07cf788ff7dd&selectedIssue=EDS-926

Autocomplete, Filter and `Field.MultiSelection` each had their own text-highlighting implementation. This meant duplicated logic and inconsistent behaviour — for example, only Autocomplete supported multi-word matching, while Filter and MultiSelect were single-term and carried their own regex-escaping and a fragile stateful-regex pattern.

This extracts the highlighting into a single shared internal engine (a pure `highlightText` util plus a `useHighlightText` hook that owns the render cache), so the behaviour is consistent across all three and there is one place to maintain it. Filter and MultiSelect now get multi-word highlighting for free.

Each component keeps its existing markup (`<span>`/`<mark>` and class names) in this PR; unifying on a single `<mark>` is left as a follow-up.

### DIFF
--- a/packages/dnb-eufemia/src/components/autocomplete/Autocomplete.tsx
+++ b/packages/dnb-eufemia/src/components/autocomplete/Autocomplete.tsx
@@ -52,6 +52,7 @@ import {
   getClosestParent,
 } from '../../shared/component-helper'
 import { IS_MAC, debounce, hasSelectedText } from '../../shared/helpers'
+import { highlightText } from '../../shared/helpers/highlightText'
 import useId from '../../shared/helpers/useId'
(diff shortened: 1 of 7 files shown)


langz (approved):
Wow, great, I really like this, less duplicated code, one place to maintain, and strong test coverage 🚀 

Perhaps @tujoworker has a few pointers?

tujoworker (approved):

Issue #4989, opened 2025-04-23
Conditional required Field.Upload does not prevent navigation to next step


torryt:
## 🐛 Bug Report

Sometimes, when a conditional required Field.Upload is not filled, it still does not prevent to the next step in the form.

## To Reproduce

https://github.com/user-attachments/assets/55c9e362-e676-4a42-a798-4b9179d5fc0c

I've only been able to consistently reproduce it with exactly these steps. I haven't been able to reproduce it in a CSB.

Here's a [link to the source code](https://gitlab.tech.dnb.no/dnb/dcf/dnb-web-codd-apps/-/blob/main/apps/dkt-as/src/pages/Form/Step3/OwnershipAndControl/OwnershipAndControl.tsx?ref_type=heads)

## Expected behavior

I'd expect the validation to prevent navigation to the next page. Sidenote: should the page also be scrolled to closest form error when navigation is prevented and the error is not in view?

## Run `npx envinfo`

Paste the results here:

```bash
  System:
    OS: macOS 15.3.2
    CPU: (11) arm64 Apple M3 Pro
    Memory: 71.91 MB / 18.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.13.1 - ~/.nvm/versions/node/v22.13.1/bin/node
    Yarn: 1.22.22 - ~/.nvm/versions/node/v22.13.1/bin/yarn
    npm: 10.9.2 - ~/.nvm/versions/node/v22.13.1/bin/npm
    pnpm: 9.14.2 - ~/.nvm/versions/node/v22.13.1/bin/pnpm
  Managers:
    Homebrew: 4.4.32 - /opt/homebrew/bin/brew
    pip3: 24.2 - ~/miniconda3/bin/pip3
    RubyGems: 3.0.3.1 - /usr/bin/gem
  Utilities:
    Make: 3.81 - /usr/bin/make
    GCC: 17.0.0 - /usr/bin/gcc
    Git: 2.45.1 - /opt/homebrew/bin/git
    Clang: 17.0.0 - /usr/bin/clang
    FFmpeg: 7.1.1 - /opt/homebrew/bin/ffmpeg
    Curl: 8.7.1 - /usr/bin/curl
    OpenSSL: 3.0.15 - /Users/torrytufteland/miniconda3/bin/openssl
  Servers:
    Apache: 2.4.62 - /usr/sbin/apachectl
  Virtualization:
[...]

PR #8980, opened 2026-08-13
fix(Field.Upload): reveal required error on submit when validateInitially={false}

langz:
When a required `Field.Upload` (or any required field) uses `validateInitially={false}` inside a Wizard, clicking "Next" correctly blocked navigation but showed no error message — the button appeared to do nothing.

The one-time reveal skip for `validateInitially={false}` relied on an initial reveal (e.g. from an ancestor `validateInitially`) to consume it, which is why it worked inside `Form.Section`. In a bare Wizard/Form there is no initial reveal, so the first submit-driven reveal was swallowed.

The reveal now bypasses the skip when it originates from an explicit submit (`showAllErrors`/`showBoundaryErrors` hold a timestamp) but not from the initial `validateInitially` (a boolean), so the functionality matches the behavior without `validateInitially={false}`.

**Trade-off:** since `showAllErrors`/`showBoundaryErrors` remain set after a submit attempt, a `validateInitially={false}` field that is mounted or made visible _after_ that attempt now reveals its error immediately — consistent with fields that don't set `validateInitially={false}`. A dedicated test documents this.

Refs #4989 — in the reproduced case navigation was already prevented, so this restores the missing required-error message; leaving the issue open to confirm the reporter's intermittent navigation case.

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Form/Visibility/__tests__/Visibility.test.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Form/Visibility/__tests__/Visibility.test.tsx
@@ -1233,4 +1233,57 @@ describe('Visibility', () => {
 
     expect(container.innerHTML).toMatchInlineSnapshot(`"Child"`)
   })
+
+  describe('validateInitially={false} with show-all-errors after submit', () => {
+    // Pins the behavior introduced by the #4989 fix: once the form has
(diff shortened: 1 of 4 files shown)


tujoworker (approved):
Reviewed the implementation and surrounding validation paths. The focused regression tests pass, and I found no actionable issues.

PR #8985, opened 2026-08-13
fix(Dialog): use full body height when locking scroll

tujoworker:
Opening a Modal currently forces the page body to `height: auto`. In Safari, this can change the page geometry for legacy documents and cause the modal layout to cover only part of the viewport.

Use `height: 100%` while the body is scroll-locked so the modal retains full-page geometry in these documents.

Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1786363232385539
Dialog: https://fix-modal-preserve-body-heig.eufemia-e25.pages.dev/uilib/components/dialog/demos


langz (approved):

langz:
I tested a bit on iOS Safari. Looks good

PR #8986, opened 2026-08-13
feat(SidebarMenu): add new menu component as an Eufemia Extension

tujoworker:
Adds a reusable nested SidebarMenu extension for accessible, persistent navigation with linked accordions and scroll handling.

## The strongest UX features:

- **Active-page awareness:** automatically opens the correct hierarchy and marks the current page.
- **Smart scrolling:** brings an off-screen active item into view without an initial animated jump.
- **State persistence:** remembers expanded groups and scroll position while navigating.
- **Collapsible active paths:** users can close an active branch; a dot indicates it still contains the current page.
- **Linked accordions:** the label navigates while a separate chevron expands or collapses—no ambiguous click behavior.
- **Smooth transitions:** coordinated arrow, text, and accordion animations with reduced-motion support.
- **Find-in-page support:** searching can reveal content inside collapsed accordions.
- **Static categories:** related items can have clear headings without unnecessary nested accordions.
- **Responsive behavior:** works in a constrained `ScrollView` and as a mobile navigation menu.
- **Consistent alignment:** icons, text, badges, wrapped labels, and active arrows remain visually aligned.
- **Section switching:** supports platform-like views such as Web, iOS, and Android.
- **Accessible interaction:** semantic navigation, native controls, keyboard support, current-page announcements, and clear expansion states.

## All SidebarMenu features

- Declarative compound API:
  - `Container`
  - `Item`
  - `Accordion`
  - `Group`
  - `Section`
  - `Header`
  - `Divider`
- Recursive data and section APIs.
- Deeply nested accordions.
- Controlled and uncontrolled open state.
- Collapsible and permanently expanded groups.
- Static titled categories without accordion behavior.
- Optional linked group titles.
- Leaf items rendered as links or buttons.
- Linked accordions with separate navigation and expansion controls.
- Custom router-link elements through `element` and `to`.
- Active/current-page state with selection arrow.
- Automatic opening of the active item’s ancestors.
- Active ancestors can still be manually collapsed, with an indicator showing they contain the current page.
- Controlled and uncontrolled selection.
- Switchable top-level sections.
- Controlled and uncontrolled section state.
- Automatic activation of the section containing the selected item.
- Optional item and accordion icons.
- Badges and custom suffix content.
- Disabled items and accordions.
- Dividers before data items or declarative `Divider` elements.
- Fully custom data entries through `type: "custom"`.
- Open-state persistence in session or local storage.
- Scroll-position persistence in session or local storage.
- Initial instant scroll restoration without visible animation.
- Automatic scrolling of an off-screen selected item into view.
- Support for the browser viewport and nested `ScrollView`.
- Waits for opening animations before positioning the selected item.
- Initial prerender scroll restoration support.
- Browser find-in-page support through `hidden="until-found"`.
- Option to disable `until-found`.
- Spacing props on the container.
- Wrapped labels without splitting words.
- Animated expansion, selection arrow, and text transitions.
- Reduced-motion support.
- Responsive usage.
- Tree-shakeable namespace API.

## Accessibility

- `nav` landmark.
- Semantic nested lists.
- Native links and buttons.
- `aria-current="page"` for the selected destination.
- `aria-expanded` and `aria-controls` on accordions.
- Accessible section toggle group.
- Correct disabled-link and disabled-button behavior.
- Keyboard operation with Tab, Shift+Tab, Enter, and Space.
- Decorative selection arrow hidden from screen readers.
- Automated axe coverage.

You can preview the menu used in the Portal in this PR: https://github.com/dnbexperience/eufemia/pull/8987
Here are the docs for the menu itself: https://feat-sidebar-menu-extension.eufemia-e25.pages.dev/uilib/extensions/sidebar-menu


langz:
I added a few thoughts in https://github.com/dnbexperience/eufemia/pull/8993, about this PR/sidebarmenu.

packages/dnb-eufemia/src/extensions/sidebar-menu/SidebarMenuAccordion.tsx
```diff
@@ -0,0 +1,253 @@
...
+  )
+
+  const itemStyle = {
+    '--sidebar-menu-indent': `${context.indent}rem`,
+  } as React.CSSProperties
+  const currentIndicator = containsSelectedItem && !isOpen && (
+    <span
+      className="dnb-sidebar-menu__accordion__current-indicator"
+      role="img"
+      aria-label="Contains current page"
```
joakbjerk: Should this text come from translations?

langz:
Just so you don't go hunting: I don't think the failing visual regression check is yours.

43 diffs, and 27 of them are List (26) and Timeline (1) — files this PR doesn't touch at all. The sizes are 1–24 px (0.001–0.018%), and our tolerance is 0, so a single pixel fails.

I'm fairly sure it's fallout from the icon refactor in #9328. It nudged path coordinates on ~400 icons by fractions of a pixel (`5.355` → `5.356` and similar) and only renewed the Icon component's own snapshots.

Why it lands on you: this PR touches a global style file, so the runner logs *"Global visual impact detected (2 triggers). Running all screenshot tests"* — which makes it the first full run since that refactor. The branches that went green right after logged *"No visual source changes detected. Skipping screenshot tests"*, so they never ran List at all.

All 16 SidebarMenu diffs are exactly 1 px, so I think your baselines just need renewing once main is sorted 👍

langz:
The two storage writes are not guarded, while all the reads are. If `setItem` throws — Safari private mode, or a full `localStorage` — it takes the menu down with it.

I ran it with a throwing `setItem`:

- with `openItemsStorageKey` set it throws already on mount, from the layout effect at `SidebarMenuContainer.tsx:206` → `writeStoredOpenState` (`:836`)
- with `scrollPositionStorageKey` set it throws on every scroll event, in `persistPosition` (`:241`)

The read path already does the right thing (`try/catch` around `getItem`, and `getStorage()` guards the access itself), so I think it's only those two writes that need the same treatment.

This is the one I'd like to see fixed before merge 🙂

langz:
I think `openItems` isn't really controlled when the selected item sits inside the accordion.

With `openItems={[]}` and `selectedItem="cards"` nested under `products`:

- it renders `aria-expanded="true"` even though the prop is empty
- clicking the trigger collapses it, and it stays collapsed even when the parent re-renders with the same `openItems={[]}`
- `onOpenItemsChange` fires with `[]`, so the parent can't even see that something changed

If the selected item is *not* nested in that accordion it behaves as expected and the prop wins. So it's specific to the "auto-open the selected ancestors" path and the `closedSelectionPath` state behind it.

Not sure what you intended here. If it's by design I think it deserves a line in the docs, since `openItems` is documented as "Controlled array of open accordion ids" — as it stands a controlled parent has no way to know about the change, or to say no to it.

langz:
`SidebarMenu.Header` renders as a plain list item:

```html
<li class="dnb-sidebar-menu__header">Category A</li>
```

So it ends up a flat sibling of the items it heads, with no heading role and nothing tying those items back to it. Sighted users get the grouping from the styling, but on a screen reader it's just a list where one of the items is a bit of loose text.

`SidebarMenu.Group` already solves this nicely with `<ul aria-labelledby>` — could `Header` do something similar?

langz:
The auto scrolling always animates, also for people who have asked for less motion — `SidebarMenuContainer.tsx:360` and `:382` both pass `behavior: 'smooth'`.

We do have the global guard in `dnb-ui-core.scss` that sets `scroll-behavior: auto !important` under `prefers-reduced-motion`, but per the CSSOM spec that's only consulted when `behavior` is `auto` or left out. An explicit `'smooth'` from JS wins over the CSS, so the global guard doesn't reach these two calls.

I think it just needs a `matchMedia('(prefers-reduced-motion: reduce)')` check. Everything CSS-driven in the menu is fine, it's only the scrolling 🙂

langz:
What is this guard fixing? `adjustTo` already bails out on exactly the same condition:

```ts
if (fromHeight === toHeight) {
  return // no need to adjust
}
```

`HeightAnimationInstance.ts:373` — and the only thing running before it is `getOptions()`, which is just `return this.opts`. So as far as I can tell this is a no-op.

Since `useHeightAnimation` is shared with Tabs, Accordion, Forms, List, Upload and more, I'd rather we either drop it, or keep it with a test for the case it's meant to cover — so nobody later assumes it's load-bearing.

langz (approved):
Menu looks really great 🤩 Approving so this isn't blocked on me.

I left a few separate comments. The storage one I think should be fixed before merge — it can crash the menu on mount. The rest are fine as follow-ups if you prefer.

Also: the failing visual regression check doesn't look related to this PR, see my comment about it.

langz:
I think the legacy sidebar styles broke with the scoping change here 🤔

In `SidebarMenu.module.scss` the parent selector became `.portal-sidebar-menu__legacy-item.dnb-sidebar-menu`, and Sass appends `&__item` to the *last* class in the pair, so it compiles to:

```css
.portal-sidebar-menu__legacy-item.dnb-sidebar-menu__item
```

But `portal-sidebar-menu__legacy-item` sits on the `li` (line 473) and `dnb-sidebar-menu__item` on the child `div` (line 487), so nothing ever matches both and the rule is dead. Same story for `&__expand-button`.

I compared the branch preview against prod: the item rows compute to `display: block` instead of `flex`, so the +/- button wraps down onto its own line and the rows grow from 56px to 80px. It hits all the expandable rows.

Guess writing them as descendants instead would fix it, so `.dnb-sidebar-menu__item { ... }` inside the same block.

langz:
We read `sessionStorage` during render in `SidebarMenuContainer.tsx` (`readStoredOpenState`, ~line 69), so the server and client HTML don't match. React logs a hydration mismatch and says it won't patch it up.

For a returning user on a prerendered page that means the group renders closed while the state thinks it's open, so the first click does nothing and you have to click twice. I get `afterHydrate=false`, `click1=false`, `click2=true`. With no storage key, or with empty storage, there's no mismatch at all, so it's the render-time read that does it.

I think the load effect further down (~line 160) was meant to cover this, but `loadedOpenItemsStorageIdRef` is initialised with the same id on line 87, so the guard returns early and it never runs. Moving the initial read into that effect should sort it.

The Portal prerenders and #9296 sets `openItemsStorageKey`, so we would hit this.

langz:
`getStorage()` only guards *getting* the storage object, not the write itself.

`writeStoredOpenState` runs from a layout effect on mount, so if `setItem` throws (quota, restricted storage) it throws out of render and the mount fails — no user interaction needed. I reproduced it by making `setItem` throw: one call on mount, and the client mount dies with `QuotaExceededError`.

`Tabs.tsx` wraps every storage call in its own try/catch with `warn()`, maybe we should do the same here?

langz:
Coming back to my localization comment from #8993 🙂

`aria-label="Contains current page"` in `SidebarMenuAccordion.tsx` is hardcoded and has no prop, so it can't be overridden at all. `sectionLabel` also defaults to `'Menu section'`.

Our default locale is `nb-NO`, so Norwegian screen reader users get English here. `Breadcrumb` handles this nicely with `extendPropsWithContext` plus an entry in `nb-NO.ts` (`navText` etc.) if we want a pattern to follow.

langz:
Think the scroll-into-view ignores `prefers-reduced-motion`.

Our global `scroll-behavior: auto !important` doesn't cover it, since an explicit `behavior: 'smooth'` in JS overrides the CSS property. I tested it in Chrome with reduced motion enabled and it still animates all the way.

`SidebarMenuContainer.tsx` line 360 and 379-383. A `matchMedia('(prefers-reduced-motion: reduce)')` check before choosing `smooth` would sort it.

langz (approved):
Menu looks really great 🤩 Impressive amount of work, and nicely tested.

Approving so this isn't blocked. I'd like the legacy sidebar styling one sorted before merge though, since it affects the Portal as it stands today. The rest can follow up.

packages/dnb-eufemia/src/components/height-animation/useHeightAnimation.ts
```diff
@@ -298,7 +298,9 @@ function useAdjust({ children, instRef, isInitialRenderRef, targetRef }) {
        */
       const toHeight = instRef.current.getHeight()
 
-      instRef.current.adjustTo(fromHeight.current, toHeight)
+      if (fromHeight.current !== toHeight) {
+        instRef.current.adjustTo(fromHeight.current, toHeight)
+      }
```
joakbjerk: This guard looks redundant: `adjustTo` already early-returns when `fromHeight === toHeight` (in `HeightAnimationInstance.ts`) before any state update or callback runs. Since this hook is shared by Accordion/Dropdown/Tabs/GlobalStatus etc., consider dropping the wrapper to keep the shared path minimal — or add a one-line comment explaining a case the inner guard does not cover, so it does not read as dead code later.

packages/dnb-eufemia/src/extensions/sidebar-menu/SidebarMenuContainer.tsx
```diff
@@ -0,0 +1,857 @@
...
+    ids: initialOpenState.closedItems,
+  }
+  const initialOpenItems =
+    openItems ??
+    addOpenItem(
+      initialOpenState.openItems,
+      initialSelection?.selectedAccordionId
+    )
+  const [internalOpenItems, setInternalOpenItems] =
+    useState(initialOpenItems)
```
joakbjerk: These `initial*` values run recursive tree scans (`findActiveDataItem`, `findActiveDeclarativeItem`, `findSelection`, `readStoredOpenState`) on every render, but are only consumed by the `useState(...)` initializers on first render. Consider lazy state initializers (`useState(() => ...)`) so the scans do not repeat for large menus. The live `selection`/`resolvedOpenItems` computations below are genuinely needed per render and can stay as-is.

packages/dnb-eufemia/src/extensions/sidebar-menu/__tests__/SidebarMenu.test.tsx
```diff
@@ -0,0 +1,2007 @@
...
+        <SidebarMenu.Accordion id="products" text="Products">
+          <SidebarMenu.Item id="cards" text="Cards" />
+        </SidebarMenu.Accordion>
+      </SidebarMenu.Container>
+    )
+
+    expect(sessionStorage.getItem(storageKey)).toBeNull()
+  })
+
+  it('has no automated accessibility violations', async () => {
```
joakbjerk: This axe test covers Item/Accordion/Divider/disabled. Consider extending it to the `sections` Dropdown, a `Group` (with `aria-labelledby`), a linked accordion, and the selected-state current indicator (`role="img"`) — those introduce the more complex ARIA relationships this component adds and are the most valuable to guard.

joakbjerk (approved):

langz:
Coming back to this while reviewing the rest of the stack — an icon-only `SidebarMenu.Item` ends up with no accessible name, and there's no supported way to give it one.

`SidebarMenuItemProps` does `Omit<HTMLAttributes<HTMLElement>, 'title' | ...>`, so `title` is gone, and everything else in `...rest` is spread onto the `<li>` rather than the `<a>`/`<button>`. Rendering `<SidebarMenu.Item id="x" icon={person} href="/x" aria-label="Profile" />` gives:

```html
<li aria-label="Profile" data-sidebar-menu-id="x" class="dnb-sidebar-menu__item">
  <a href="/x" class="… dnb-sidebar-menu__item__action">…</a>
```

The `aria-label` lands on the `li`; the link itself has `aria-label: null` and empty text content. `text` is optional and `Icon` is `aria-hidden` unless `alt` is passed, so that's a nameless link.

Forwarding `aria-label`/`aria-labelledby`/`title` to the action element would fix it. It's a props-shape change, so cheaper now than after this ships.

langz:
With `disableUntilFound`, every collapsed accordion trigger has a dangling `aria-controls`.

`aria-controls={`${id}-content`}` is emitted unconditionally, but the `<ul id={`${id}-content`}>` lives inside `HeightAnimation`, which returns `null` when closed unless the content is kept in the DOM. Rendered both ways:

```
disableUntilFound=false → aria-controls="acc-content", target in DOM: true
disableUntilFound=true  → aria-controls="acc-content", target in DOM: false
```

The default is fine, since `hidden="until-found"` keeps the content mounted. Emitting `aria-controls` only when the content is in the DOM would cover both cases.

The `axeComponent` test doesn't include `disableUntilFound`, which is why this passes — adding it there would catch this class of thing.

langz:
`SidebarMenuAccordion` passes `untilFound={useUntilFound}` to `HeightAnimation`, and that prop is marked `@deprecated Use openOnFind instead`.

The public container prop is `disableUntilFound` too — an inverted flag named after the deprecated term, where the equivalent public prop elsewhere is `openOnFind` (Accordion).

Since this API hasn't shipped yet, something like `openOnFind?: boolean` defaulting to `true` would match the rest of the library and costs nothing to change now. Later it's breaking.

langz:
Thanks for having a go at this 🙂 Dropping `dnb-sidebar-menu` from the `li` is the right call — that removes the class clash with the new component. But I don't think the styling is fixed yet, and I think one more rule broke on the way.

The parent selector is now `.portal-sidebar-menu__legacy-item`, while the children in `SidebarMenu.tsx` still use the old stem. So the three `&`-suffix rules compile to:

```css
.portal-sidebar-menu__legacy-item__item
.portal-sidebar-menu__legacy-item__expand-button
.portal-sidebar-menu__legacy-item--accordion
```

None of those exist in the markup — the elements are still `dnb-sidebar-menu__item`, `dnb-sidebar-menu__expand-button` and `dnb-sidebar-menu--accordion`. And `&--accordion` *did* match before this push (the compound selector happened to line up), so the expanded-state block is newly dead — that's the `border-right` and active-path backgrounds.

On the preview the rows are still `display: block` instead of `flex`, 80px instead of 56px, with the +/- button wrapped onto its own line. Same as before the fix.

Think just spelling them out with the real class names would do it:

```scss
.dnb-sidebar-menu__item { … }
.dnb-sidebar-menu__expand-button { … }
&.dnb-sidebar-menu--accordion { … }
```

langz:
Small correction to my visual regression comment above — "all 16 SidebarMenu diffs are exactly 1 px" isn't true any more after the section chevron change.

On the last full run it splits in two:

- 12 snapshots at 1 px (0.001%) — the icon-refactor noise from main
- 4 snapshots at 23 px (0.012%) — `matches-the-open-section-selector` and `matches-a-keyboard-focused-section-option`, both themes

Those four are exactly the states where the new chevron-up renders, so they're a real change of yours rather than noise, and no baselines were renewed in that push. Probably worth eyeballing those four diffs before renewing, rather than renewing everything blind 🙂

(The chevron itself works fine as far as I can tell — closed and open paths do switch on the preview.)

langz:
Storage guards, hydration, reduced motion and the controlled `openItems` all look sorted now 👍 Verified those. One thing the new `animate` flag introduced though.

`animate` starts as `!openItemsStorageKey`, and the only place it's flipped back to `true` is the storage-load effect — which returns early when `openItems` is controlled. So if someone passes both `openItems` and `openItemsStorageKey`, `animate` stays `false` for the lifetime of the menu and nothing ever animates.

I ran all four combinations:

| `openItems` | `openItemsStorageKey` | `animate` |
| --- | --- | --- |
| uncontrolled | – | true |
| uncontrolled | set | true |
| controlled | – | true |
| controlled | set | **false** |

Only that last one. I guess the `requestAnimationFrame` that flips it could run regardless of whether `openItems` is provided, or `animate` shouldn't key off the storage key when controlled.

langz:
The `aria-label` forwarding on `Item` looks good 🙂 But I think `Accordion` and `Group` need the same treatment — they still land it on the `li`:

| | `aria-label` ends up on | action's accessible name |
| --- | --- | --- |
| `Item` | the `<a>` ✅ | "Profile" |
| `Accordion` | the `<li>` | none |
| `Accordion` + `href` | the `<li>` | none |
| `Group` + `href` | the `<li>` | none |

`text` is optional on all three, so an icon-only accordion or group gives a nameless button/link in exactly the way `Item` did before the fix.

They also still carry `'title'` in their `Omit<HTMLAttributes<…>, …>` (`types.ts:239` and `:287`), so `title` isn't accepted as a fallback there either.

langz:
Wrapping `Section`s in a fragment renders an empty menu, and nothing warns about it:

```jsx
<SidebarMenu.Container>
  <>
    <SidebarMenu.Section id="a" text="A">…</SidebarMenu.Section>
    <SidebarMenu.Section id="b" text="B">…</SidebarMenu.Section>
  </>
</SidebarMenu.Container>
```

| children | section dropdown | items rendered |
| --- | --- | --- |
| fragment | no | 0 |
| `.map()` | yes | 1 |

`Children.toArray` keeps the fragment as a single child, so the `child.type === SidebarMenuSection` check never matches. It then falls through to `sectionContent = children`, and since `Section` renders `null` the `<ul>` comes out completely empty — `innerHTML` is `""`.

Any wrapper does this, not just fragments. A dev-only warning when `children` contains no recognised items would make it a lot easier to spot 🙂

langz:
`SidebarMenu.Group` without `text` gives its nested list an empty accessible name:

```html
<div id="g-title" class="dnb-sidebar-menu__group__title"></div>
<ul class="… dnb-sidebar-menu__group__list" aria-labelledby="g-title">
```

The `aria-labelledby` target is there, but its text content is `""`. axe doesn't flag it, so it passes quietly.

Nice thing is the docs already say what I think you intended — `SidebarMenuDocs.ts` marks `text` as **required**, while `types.ts` has it as `text?: ReactNode`. So either tightening the type to match the docs, or skipping `aria-labelledby` when there's no title, would close it.

langz:
In the `data` API, `active` only opens the hierarchy when the data is present on the first render:

| `data` supplied | ancestor `aria-expanded` | child `aria-current` |
| --- | --- | --- |
| on first render | `true` | `page` |
| on a later render | **`false`** | `page` |

So for a menu whose items are fetched, the current page gets marked but sits inside a collapsed group.

`selectedItem` doesn't behave this way — passing it later does open the ancestors, there's an effect for that. It's only the initial `useState` scan of `data`/`children` that is first-render-only.

Very happy to call this by design if `active` is meant as "initial" — in that case the "automatically opens the correct hierarchy" bullet in the description is just a touch generous 🙂

langz:
Tiny one from the last push 🙂 Dropping `redirect_from` from `sidebar-menu.mdx` leaves the changelog link dangling.

`EUFEMIA_CHANGELOG.mdx:16` still points at `/uilib/components/sidebar-menu`, and that path only resolved through the `redirect_from` route that the portal-pages plugin generated. The page itself lives at `/uilib/extensions/sidebar-menu`.

langz:
Small one on the properties page: the `SidebarMenu.Data` table documents a `children` prop that `Data` doesn't accept.

`SidebarMenuDataProperties` is built as `{ data, sections, ...SidebarMenuRootProperties }`, and `SidebarMenuRootProperties` leads with `children` — so the Data table inherits it, described as "Declarative menu content composed with SidebarMenu.Item, …". But `SidebarMenuDataProps` is `Omit<SidebarMenuRootProps, 'children'>`, so following the table doesn't compile:

```
error TS2322: Type '{ children: Element; "aria-label": string; data: … }' is not assignable to
type 'IntrinsicAttributes & Omit<SidebarMenuRootProps, "children"> & { data?: …; sections?: … }'
```

The same shape on `Root` compiles fine, so it's specific to `Data`. At runtime the children are simply dropped — `<SidebarMenu.Data data={[{ id: 'a', text: 'A' }]}><SidebarMenu.Item id="b" text="B" /></SidebarMenu.Data>` renders just `A`. The inherited wording is a bit misleading too, since `Data` is the non-declarative half of the API.

Destructuring `children` out before the spread should be all it needs. Nice contrast: `events.mdx` already handles the same overlap neatly with "`SidebarMenu.Data` supports the same events." 🙂

langz:
One more, and I think it ties two of the earlier ones together 🙂

Compound parts are found by reading `_sidebarMenuRole` off `child.type`, so the role is lost through any wrapper — including `React.memo`, which is a natural thing to reach for on a big menu.

A memoized `Item` selected inside a section:

| | section shown | items visible | current page in DOM | "contains current page" dot |
| --- | --- | --- | --- | --- |
| `SidebarMenu.Item` | Section B | Group B, B1 | yes | yes |
| `React.memo(SidebarMenu.Item)` | **Section A** | A1 only | **no** | no |

So you land on B1 and the menu shows a different section, with the current page nowhere on screen. A plain wrapper (`const MyItem = (props) => <SidebarMenu.Item {...props} />`) does exactly the same. Nothing warns.

Putting this next to the [fragment one](https://github.com/dnbexperience/eufemia/pull/8986#issuecomment-5663260899) and the [`data.active` one](https://github.com/dnbexperience/eufemia/pull/8986#issuecomment-5663261536): all three end up as *"current page is marked, but hidden or in the wrong section"*. The traversal is identity-based, so it can't see through a wrapper, and it's first-render-only, so it can't see data that arrives later. Might be worth one look at `findDeclarativeSelection` rather than three separate patches.

Worth saying this one isn't new to the refactor — `child.type === SidebarMenuItem` had the same blind spot. And `Root` importing none of the sibling parts is a genuinely neat win 🙂

langz:
One naming thing while I'm here: `active` carries two different meanings in `types.ts`.

On `SidebarMenuItemData` it's documented as "Marks a leaf item as the current page, or a section as initially active", and on `SidebarMenuSectionData` as "Marks this section as initially active". So the same prop name is a current-state assertion for items and an initial-state default for sections — and it's the second one that's likely to trip people up, since there's no `defaultActive` to reach for instead.

`active` itself is fine and widely used (16 props outside this extension), so this is only about the two meanings sharing one name inside this extension. Naming the section one `defaultActive`, mirroring `defaultActiveSection` on `Root`, would make the intent obvious at the call site.

Separate and much smaller: `SidebarMenu.Data` as a *component* is a genuine first — no other Eufemia namespace has a `Data` member, and none of the libraries I checked (React Aria, Mantine, MUI, Radix) expose the data-driven variant as a separate component; it's always a prop (`items` / `data` / `options`). The bundle split is a good reason to diverge, though — `Root` on its own bundles to ~38 KB against ~235 KB for `Data` — so I'd keep it and just add a sentence in the docs saying why, since people will reasonably expect `<SidebarMenu.Root data={…}>` to work 🙂

langz:
Looks to be something strange now, when changing between Personal and Business, I have to press twice. 
It only happens on my iPhone, using safari

https://feat-sidebar-menu-extension.eufemia-e25.pages.dev/uilib/extensions/sidebar-menu/demos/

https://github.com/user-attachments/assets/e2dfa493-783a-4269-8c13-0cf6cfeb73e2

langz:
When hovering over one of the non-selected dropdown item like Business in the following scenario, the items of the sidebar menu underneath/below changes color and gets some kind of hover effect or such:

https://github.com/user-attachments/assets/65cf8563-aa9a-491d-8757-373014a8e453

langz:
It looks like the menu has been updated/changed in this PR.

The following video is from my phone, but can also be seen in desktop https://feat-sidebar-menu-extension.eufemia-e25.pages.dev/uilib/extensions/sidebar-menu/demos/

https://github.com/user-attachments/assets/384b4b51-0b2f-4cc5-850f-9ab8e77d58a4

---

**✅ Settled — re-checked.** Correcting the basis I first gave here: the layout re-measurement below was taken at 12:57 on the 15th, before the hamburger icon was split later that evening, so it never covered the icon. The icon has since been reverted in `c99047e9`, and the asset now differs from `main` only by a trailing newline — the path data (`M1 14h14M1 8h14M1 2h14`) and the generated `hamburger.tsx` are byte-identical to `main`. So nothing visual is left outstanding.

langz:
I tested a bit in https://chore-portal-home-mdx.eufemia-e25.pages.dev/uilib/components/fragments
And I really like that the menu items renders anchors, as I really like to cmd+click, to open new tabs, and that works for most items.
It makes sense that it does not work for items that is only working as an accordion.
But for the ones that act as both a link and an accordion, should cmd+click also open in a new tab? Like for Fragments which is both an accordion and link?

---

**✅ Answered below.** The one exception it turned up — a linked accordion that *is* the current page — is fixed, and the new `preserves native %s activation on a selected page accordion` test fails 3/3 without the modifier guard, so it genuinely pins it.

langz:
Still seeing this one on the latest preview 🙂 Measured the current build against prod on `/uilib/components/`:

| | prod | preview |
| --- | --- | --- |
| row `display` | `flex` (161/161 rows) | `block` (162/162) |
| "About the Library" row | 56px | 80px |
| the other 9 expandable rows | 40px | 64px |
| expand button | inline, `align-self: center` | wrapped onto its own line |

10 expandable rows affected on that page. The three `&`-suffix rules still compile against the new stem — `.portal-sidebar-menu__legacy-item__item`, `__expand-button` and `--accordion` — while the markup still carries `dnb-sidebar-menu__item`, `dnb-sidebar-menu__expand-button` and `dnb-sidebar-menu--accordion`.

Two things that make me want it fixed here rather than later: the Portal has no screenshot coverage of its own sidebar, so CI stays green on it, and #9296 is where this markup gets replaced — so if this lands without it, the breakage rides along to the live Portal on the next release. Writing those three against the old class names should be enough — `__item` and `__expand-button` as descendants, and `--accordion` as `&.dnb-sidebar-menu--accordion`, since that one sits on the same `li`.

*(Edited twice, sorry: the row count said 40/40, but it's every row — 161 on prod and 162 on the preview, the extra one being this PR's own docs page. And I'd written that merging ships this to the live Portal; production deploys from `release`, not `main`, so it lands there on the next release instead. Every other number re-measured identical.)*

langz:
Good catch 🙂 It also gets stuck — after switching section with the mouse the whole menu stays at `opacity: 0.5`, and nothing clears it.

Going Business → Personal on the demos page:

| how I switched | menu afterwards |
| --- | --- |
| mouse | dimmed, and it stays dimmed |
| keyboard | clean |

Moving the mouse away, hovering the menu, or switching again doesn't clear it. Personal → Business is fine, so it depends on which option ends up under the cursor.

What I think happens: as the list realigns while closing, a `mouseover` fires on the *other* option and sets `hoveredSection` again after `onChange` has already cleared it. The cleanup is a `mouseleave` on `.dnb-drawer-list`, and that element unmounts — removing an element never fires `mouseleave`, so it never runs.

`does not dim a newly selected section when it is hovered` stays green because it hovers the already-selected option in jsdom, where there's no realignment. Gating the dimming on the dropdown being open would cover the whole class 🙂

---

**✅ Fixed — verified.** `onItemMouseEnter` now returns early unless `sectionSelectorOpenRef.current` is true, and both `onChange` and `onClose` clear it, so the late `mouseover` during realignment can no longer re-set `hoveredSection`. I confirmed the gate is what does it: with the gate removed the menu dims again, with it in place it stays clean.

One caveat raised separately — the accompanying test passes with the gate removed too, so it doesn't currently pin the fix.

langz:
The late `active` data fix works 👍 but I think it took `defaultSelectedItem` with it.

The new effect pushes the `active` item into the selection and only guards on `selectedItem`, so when `defaultSelectedItem` and an `active` item are both set, `active` wins on the first render. I disabled just that effect to check:

| | late `active` data opens ancestors | `defaultSelectedItem` respected |
| --- | --- | --- |
| this branch | yes | no — selection lands on the `active` item |
| that effect disabled | no | yes |

So it's a straight trade between the two right now. Skipping it when `defaultSelectedItem` was supplied should keep both 🙂

---

**✅ Fixed — verified.** The effect now guards `defaultSelectedItem === undefined` as well, so `defaultSelectedItem` wins while late `active` data still opens ancestors. `keeps defaultSelectedItem authoritative over active data` fails when I remove that guard, so it discriminates correctly.

langz:
Two small docs leftovers from the rename 🙂

`SidebarMenu.Group`'s `text` still renders as **(required)** on the properties page, but it's `text?: ReactNode` in `types.ts` and the code now deliberately handles it being absent — that's the fix for the empty `aria-labelledby`, so it's the old mismatch just flipped the other way.

And a section-level `active` doesn't exist any more, but two places still point at it:

| where | says |
| --- | --- |
| `defaultActiveSection` — on the properties page, in both the Root and Data tables | "Defaults to the section marked active, then the first section" |
| `SidebarMenuItemData.active` — `types.ts:45` | "Marks a leaf item as the current page, or a section as initially active" |

`SidebarMenuSectionData` has `defaultActive` and `SidebarMenu.Section` has no such prop at all, so both just want `defaultActive` instead.

---

**✅ Fixed — verified.** `SidebarMenuGroupProperties.text` is now `status: 'optional'`, matching `text?: ReactNode` — and `Section.text` is genuinely required in `types.ts`, so that one is right to stay as it is. The `defaultActiveSection` doc now says "marked `defaultActive`" — and since `SidebarMenuDataProperties` spreads Root minus `children`, that single entry covers both tables. `SidebarMenuItemData.active` now reads just "Marks a leaf item as the current page." — the section clause was dropped rather than reworded. No occurrences of the old wording remain.

langz:
Two small housekeeping ones 🙂

**`badgeProps` renders as a broken type on the properties page.**

```ts
type: ['Omit<BadgeProps, "content"', '"children">'],
```

`PropertiesTable` joins array elements with `<br/>`, so the cell comes out as two lines — and the second one gets the red string colour, because `isString()` sees the leading quote:

```
Omit<BadgeProps, "content"
"children">
```

Three tables: `Item`, `Accordion` and `Group`.

I had the cause wrong in my first version of this comment — it isn't hand-written, our own `docs-types/warn-supported-types` autofix produces it. It splits on a plain `trimmedValue.split('|')`, and `isSimpleUnionType` only exempts `Array<…>` and values starting with `(`, `{` or `[`, so every other generic gets cut in half:

| written | after `eslint --fix` |
| --- | --- |
| `Omit<BadgeProps, "content" \| "children">` | `['Omit<BadgeProps, "content"', '"children">']` |
| `Pick<Foo, "one" \| "two">` | `['Pick<Foo, "one"', '"two">']` |
| `Record<string, Alpha \| Beta>` | `['Record<string, Alpha', 'Beta>']` |
| `Promise<Alpha \| undefined>` | `['Promise<Alpha', 'undefined>']` |
| `Array<Alpha \| Beta>` | left alone |

It's a `warn` and `yarn lint` runs `eslint --quiet`, so CI never shows it, but `--fix` still applies it. That also means simply restoring the pipe here won't hold — the next `--fix` re-splits it. The rule wants to skip pipes inside `<>`; that part is pre-existing and not really yours, these three are just the only places in `src` it has actually bitten so far.

**`SidebarMenu.Data` accepts both `data` and `sections`, and silently ignores `data`.** `SidebarMenuData.tsx:20` branches on `sections?.length`, with no warning either way:

| passed | rendered |
| --- | --- |
| `data` + `sections` | sections only |
| `data` + `sections={[]}` | data |

Still think it's worth tightening while this is unreleased, but correcting myself here too: I pointed at `SidebarMenuGroupProps` as an exclusivity precedent and it isn't one — its first branch allows `href` and `to` together. And a plain two-branch union would reject `<SidebarMenu.Data />` with neither prop, which is legal today, so it needs the third branch:

```ts
| { data: SidebarMenuItemData[]; sections?: never }
| { sections: SidebarMenuSectionData[]; data?: never }
| { data?: never; sections?: never }
```

*(Edited: I had the cause of the `badgeProps` one wrong, and the `Group` comparison wrong. Both corrected above — the rendering itself measures the same.)*

---

**✅ Fixed — verified, including the part that would have undone it.** The type is a single string again in all three tables, and the rule no longer re-splits it: `splitTopLevelUnion` tracks bracket depth, so all five cases from my table survive intact while a genuine top-level union still splits. I ran `eslint --fix` on the file for real — exit 0, file unchanged.

The `Data` union has all three branches, so `<SidebarMenu.Data />` with neither prop stays legal.

langz:
Answering my own question 🙂 It does work — except for one case: a linked accordion that *is* the current page.

Measured `defaultPrevented` on the accordion label:

| entry | cmd+click |
| --- | --- |
| linked `Accordion`, is the current page | **prevented** — no new tab, it toggles instead |
| linked `Accordion`, not the current page | opens a new tab |
| plain `Item`, is the current page | opens a new tab |
| linked `Group`, is the current page | opens a new tab |

`handleLinkClick` calls `event.preventDefault()` and toggles when `isSelected && collapsible`, and there's no modifier check before it — cmd, ctrl and shift all come out `defaultPrevented=true` on that entry, which is why it works everywhere else. An early return for those should be enough 🙂

Confirmed on the preview you linked, so it's not just a synthetic case: `Fragments` there is `<a href="/uilib/components/fragments">` with `aria-current="page"` and `aria-expanded="true"`, and cmd+click comes out `defaultPrevented=true` on it. That's the real portal navigation.

One thing to note if you go looking on the demos page — the latest push dropped `href` from the Products accordion in both examples, so there's no linked accordion left there to reproduce it on. The component behaviour is unchanged; `href` is still a documented `Accordion` prop, it just no longer has an example.

---

**✅ Fixed — verified.** `handleLinkClick` returns early on meta/ctrl/shift/alt while still calling `onClick`, so cmd+click is no longer prevented on a current-page linked accordion. Removing that early return fails all three `it.each` variants.

langz:
Final pass from me. Everything else I've raised is optional polish, so the one thing I'd still hold this for is the legacy Portal sidebar styling: #8986 (comment) https://github.com/dnbexperience/eufemia/pull/8986#issuecomment-5677109181

I wanted to be sure it deserves that label, so I tested the cause rather than the symptom — injecting the two rules written as descendants into the live preview:

| | preview as-is | with the rules injected |
| --- | --- | --- |
| row `display` | `block` (162/162) | `flex` (162/162) |
| the 10 expandable rows | 80px, then 64px × 9 | 56px, then 40px × 9 |

Those are prod's numbers exactly, so the three dead `&`-suffix rules really are the cause and spelling them out against the real class names is the whole fix. It's every `/uilib/*` section page (162 rows each, plus 13 on `/quickguide-designer/`), and the menu ends up 280px taller. The part you actually see is the expand button: it leaves the right edge and drops onto its own line under the label.

Two corrections to my own earlier notes while I'm here. The expanded-accordion **background is not lost** — it measures identical on both, so only the 4px right border goes, and only on pages that have an accordion open. And merging doesn't change the live Portal, since production deploys from `release` — it would ride along on the next release instead.

The rest (docs `type` rendering, `Data` accepting both props, `defaultSelectedItem`, the section dimming, cmd+click on a linked accordion that is the current page, and the rename leftovers) are all fine as follow-ups as far as I'm concerned 🙂

langz:
Think I've found this one, and it explains the "twice" exactly 🙂

Hovering a non-selected option calls `setHoveredSection`, the Dropdown re-renders, and **every option `<li>` is replaced** — the element under your finger gets detached mid-gesture:

| hover over `Business` | option nodes replaced | hovered node still in the DOM |
| --- | --- | --- |
| 1st | yes | **no — detached** |
| 2nd | no | yes |
| 3rd | no | yes |

That's the two taps: the first tap's hover tears the node out from under the click so it never completes, and the second works because `hoveredSection` is already that section — the update is a no-op, React bails out, and the node survives. With a mouse the hover and the click are separate movements, so the remount finishes long before you press, which is why it only shows on touch. (I couldn't reproduce the lost tap itself in WebKit emulation, so that last step is inference from the measurements above rather than something I've seen on a device.)

The replacement itself is `DrawerList` rather than anything here: `GroupItems`, `Items` and `ItemsRendered` are all declared inside the render (`DrawerList.tsx:681-720`), so their function identity changes on every render and React remounts the whole option subtree — the stable `key={hash}` can't help. That's been there a long time; the new hover state is just the first thing that re-renders a Dropdown while a finger is still down.

So the fix is either not re-rendering the Dropdown on hover, or hoisting those three out of `DrawerList`'s render. I haven't tried either yet — the second would help every Dropdown in the library, and it'd take the stuck dimming with it.

langz:
> Looks to be something strange now, when changing between Personal and Business, I have to press twice. It only happens on my iPhone, using safari
> 
> https://feat-sidebar-menu-extension.eufemia-e25.pages.dev/uilib/extensions/sidebar-menu/demos/
> 
>  trim.C9A4F550-037F-4044-8A94-72687DA38B0B.MOV

I'm still experiencing this one 🤔

langz:
> It looks like the menu has been updated/changed in this PR.
> 
> The following video is from my phone, but can also be seen in desktop https://feat-sidebar-menu-extension.eufemia-e25.pages.dev/uilib/extensions/sidebar-menu/demos/
> 
>  trim.E96AC889-D4E2-4C92-BD39-56025443A46A.MOV

I'm still experiencing this one 🤔

langz:
The layout fix works 👍 Re-measured the new preview against prod and they're identical now:

| | prod | preview |
| --- | --- | --- |
| row `display` | `flex` | `flex` |
| the 10 expandable rows | 56px, then 40px × 9 | same |
| expand button | `align-self: center` | same |
| expanded accordion border | 4px | 4px |

Checked eight pages (`/uilib/components/`, `/elements/`, `/extensions/`, `/layout/`, `/usage/`, `/quickguide-designer/`, a single component page and `/fragments/`) and they all match, with no dead `legacy-item__*` selectors left in the compiled stylesheet. Only difference left is one extra row, which is this PR's own docs page. Thanks for turning that around so quickly 🙂 I've hidden that comment now.

Still open from earlier, none of which I'd hold the PR for:

- `badgeProps` rendering as a split type, and `Data` accepting both `data` and `sections` — https://github.com/dnbexperience/eufemia/pull/8986#issuecomment-5678703050
- `defaultSelectedItem` losing to the late `active` effect — https://github.com/dnbexperience/eufemia/pull/8986#issuecomment-5677240814
- the rename leftovers on the properties page — https://github.com/dnbexperience/eufemia/pull/8986#issuecomment-5677324197
- cmd+click on a linked accordion that is the current page — https://github.com/dnbexperience/eufemia/pull/8986#issuecomment-5678713719
- the section dimming getting stuck — https://github.com/dnbexperience/eufemia/pull/8986#issuecomment-5677240512

On that last one, just so it's measured rather than assumed: `Personal → Business` is clean, but `Business → Personal` still leaves the menu at `opacity: 0.5`, and neither moving the mouse away nor hovering the menu clears it. I nearly reported it fixed by only testing the first direction.

*(That one had been marked resolved, so I've unhidden it — the control on the same page load showed all 162 rows as `flex`, so it was measured on the new build, not a stale one.)*

---

**✅ Settled.** Hiding this thread.

langz (approved):
The legacy sidebar styling is sorted 🎉 Thanks for bearing with me through all of this — and for taking `&.dnb-sidebar-menu--accordion` as a compound selector, that was the fiddly one of the three.

Approving on the back of the measurements just above. Two things I checked that aren't covered there: it holds at mobile width too, and every row resolves through a `portal-sidebar-menu__legacy-item` ancestor, so nothing is left orphaned by the switch to descendant selectors. All 18 checks green.

That was the only thing I wanted this held for, so it's a yes from me 👍

Everything still open is optional polish and absolutely fine as follow-ups — they're listed in the comment above. I'm still poking at the iPhone double-tap on my end and will follow up on that separately rather than hold this up.

Really nice piece of work — thanks for all the effort you've put into it 🤩

langz:
> > Looks to be something strange now, when changing between Personal and Business, I have to press twice. It only happens on my iPhone, using safari
> 
> > 
> 
> > https://feat-sidebar-menu-extension.eufemia-e25.pages.dev/uilib/extensions/sidebar-menu/demos/
> 
> > 
> 
> >  trim.C9A4F550-037F-4044-8A94-72687DA38B0B.MOV
> 
> 
> 
> I'm still experiencing this one 🤔 

Still experiencing this one

langz:
> > > Looks to be something strange now, when changing between Personal and Business, I have to press twice. It only happens on my iPhone, using safari
> > 
> > > 
> > 
> > > https://feat-sidebar-menu-extension.eufemia-e25.pages.dev/uilib/extensions/sidebar-menu/demos/
> > 
> > > 
> > 
> > >  trim.C9A4F550-037F-4044-8A94-72687DA38B0B.MOV
> > 
> > 
> > 
> > I'm still experiencing this one 🤔 
> 
> Still experiencing this one

Thanks, works great now.

langz (approved):

langz:
One last nitpick I have is that when I change between the two alternatives Personal and Business, I can see that the spacing between the menu items is not the same in both "menus", like for instance Overview which is the first menu items in both scenarios is not placed in the same position (I guess it's related to Overview in Personal having a badge?), and spacing between overview and the next menu item, Products in Personal and Payments in Business, is also not the same between the two scenarios. 

Not important, and it's possible that we can't enforce this because of different type of menu items and its content, like badge, etc. But worth a consideration. 

https://github.com/user-attachments/assets/8c6bb3f9-1339-4455-9c0d-ba8a703ea8b1

langz:
Sometimes when changing values in the dropdown, the icons used inside of the open dropdown shifts a bit and jumps a tad bit. It happens when opening, and could perhaps be related to the animation, not sure:

https://github.com/user-attachments/assets/53d2eb77-63dd-4b66-ac8f-a2490fcdc215

langz (approved):
All of it fixed, and properly — thank you 🙂 I verified each one against the new preview and locally rather than taking the diff at face value:

| what | how I checked | result |
| --- | --- | --- |
| legacy Portal sidebar layout | 8 pages vs prod | identical — rows `flex`, 56px/40px, button centred, 4px border back |
| `badgeProps` type | the properties page | one line, full `Omit<BadgeProps, "content" \| "children">`, all three tables |
| the `docs-types` autofix | ran `--fix` over 11 generic shapes | leaves `Omit`/`Pick`/`Record`/`Extract`/`Promise`/nested alone, still splits real top-level unions |
| `Data` exclusivity | typecheck probe | both together rejected, each alone fine, and `{}` still legal |
| `defaultSelectedItem` | with a control | now authoritative, and late `active` data still opens ancestors |
| docs leftovers | the properties page | `text` no longer required, both tables say `defaultActive` |
| cmd+click | linked accordion that is the current page | cmd/ctrl/shift pass through, plain click still toggles |
| section dimming | 8 open/close cycles, both directions | clears every time, and still dims while hovering |

I especially liked that you went for the root cause on two of them rather than the symptom — the depth-aware union split in the lint rule, and hoisting `Items`/`ItemsRendered` out of `DrawerList`'s render instead of just suppressing the hover. I re-ran the consumers for that second one since it's shared: DrawerList 78, Dropdown 78, Autocomplete 166 and the four Forms fields 312, all green, plus lint, stylelint and typecheck clean.

Two tiny things, neither worth another round:

- `splitTopLevelUnion` counts the `>` in `=>` as a closing bracket, so a union sitting after an arrow inside a generic still gets split — `Array<() => void \| null>` and `Array<(x: string) => boolean \| undefined>` both come back halved. I checked the blast radius with the real fixer rather than a grep: `eslint --fix-dry-run` over `src` and `scripts` would rewrite **0 files**, and there are no fixable pipe-union warnings left anywhere. So nothing is broken today — but `React.RefObject<() => void>` already appears in `AccordionDocs.ts` and `TableDocs.ts`, so the shape is one union away from tripping it.
- passing `defaultSelectedItem` while marking a *different* item `active` renders two `aria-current="page"`. Worth being precise about how small this is: axe reports no violations on it, and the ARIA spec only *should*s it ("only one element in a set should have `aria-current` set to a value other than false"), so it's guidance rather than a failure. It's also not from your change — `SidebarMenuItem` derives the attribute from `isSelected || active`, so a controlled `selectedItem` and two `active` items both produce it already, and the effect you guarded is bypassed entirely in those cases. Only worth a look if you want the combination to resolve to one.

Approving 👍 I'm leaving my own two reports open, since those are mine to confirm on device: the iPhone double-tap, and the dropdown icons shifting when opening. On that last one I couldn't reproduce it by measuring — the icons hold the exact same position inside their row across 3328 samples over 8 open/close cycles, so if it's real it's more likely a paint artefact during the animation than a layout shift.

Re-checked everything above on `c99047e916` too (the dimming-test strengthening and the icon revert) — 82 sidebar tests green and both notes above still measure the same. Nice catch making the dimming test hover the *other* option; that's the direction that was actually broken.

langz:
The gate in `onItemMouseEnter` is the right fix — I confirmed the menu dims again with it removed.

The new test doesn't pin it, though. `does not reapply dimming while the section selector closes` passes with the gate removed too, because it re-hovers the option that was just selected — and hovering the now-active option never dims, which is the same reason the older `does not dim a newly selected section when it is hovered` stayed green. Hovering the *other* option after the click reproduces the original sequence, and that version does fail without the gate.

---

**✅ Fixed in `725184d8` — verified.** Hovering `Personal` instead of the just-selected option is exactly it. The test now discriminates: it passes with the gate in place and fails with `!sectionSelectorOpenRef.current` removed, so it genuinely pins the fix.

langz:
Reproduced this one in WebKit, but I don't think it's ours 🙂

Sampling every animation frame, on the first open after a page load the option list paints one frame ~51px too high and then snaps down — while it's fully opaque, visible and on screen. It's the first open only (opens #2 and #3 move ~5px/frame, smoothly), and Chromium never shows it.

The same happens on a plain `Dropdown` — including on **production**, which has none of this PR's changes. One open per fresh browser, with every other dropdown removed from the page first so only one list can open:

| | first open, biggest visible jump (WebKit, 6 fresh loads) |
| --- | --- |
| plain `Dropdown` on production | 3/6 loads, ~40px |
| plain `Dropdown` on this preview | 4/6 loads, ~40px |
| SidebarMenu section selector | 3/6 loads, ~51px |

Same rate and shape on pre-PR code, so this looks like existing DrawerList positioning behaviour that the sidebar just makes more noticeable — the extra ~11px tracks the taller rows. Probably worth its own issue rather than this PR.

langz:
Small one on the hamburger icon 🙂

`hamburger.svg` and the generated `hamburger.tsx` now draw the three bars as three separate `<path>`s instead of one. Visually it's a no-op: at 1x the two render byte-identically, and on retina 9 of 1024 subpixels differ by at most 2/255 — anti-aliasing noise rather than a visible change.

Two things though: nothing in here targets the bars individually, and `hamburger_medium.svg` still has the combined `d="M.75 21h22.5M.75 12h22.5M.75 3h22.5"` — one `<path>` against the small icon's three. So the two variants of the same icon now differ structurally, in a published asset every consumer gets. Was the split needed for something, or can it come out?

langz (approved):
Re-reviewed at `792ccff`, and re-checked the test change in `725184d` — adding to the review above rather than repeating it, since it covers the fixes thoroughly already.

What I checked that wasn't in it: eight of the new tests fail against the previous head and pass here. For the dimming one, failing-against-the-previous-head isn't the right criterion, because other unfixed code there masks it — so I mutation-tested that guard instead. Removing just the `sectionSelectorOpenRef` check makes `does not reapply dimming while the section selector closes` fail at this head. Worth saying that the version of that test *before* "Strengthen SidebarMenu dimming regression" did **not** catch that mutation, so the strengthening did exactly what it claims — I'd have got this backwards if I'd only compared against the previous head.

The hamburger icon split is a visual no-op (byte-identical at 1x; on retina 9 of 1024 subpixels differ by at most 2/255), though `hamburger_medium.svg` was left unsplit.

On the dropdown icon shift I did reproduce it — details below. It's first-open-only, which is why repeated open/close cycles inside one page load miss it, and production shows the same thing, so it isn't this PR's.

Approving 👍

langz:
Following up on my own note above, since it contradicts what I said in the review 🙂

The difference is *what* got measured, not how many times. The earlier check looked at each icon's position **inside its row**, and that offset is rock solid — a constant 20px, varying by at most 0.3px through the whole open. But the jump is the entire list repositioning, so a row-relative measurement can't see it by construction.

Measuring the icon's absolute position instead, one open per fresh browser: ~51–62px in a single frame, on the first open after a page load only, while the list is fully opaque and on screen. So it is a layout shift rather than a paint artefact — of the list, not of the icons within it, which is why they appear to jump while keeping their spacing.

The conclusion doesn't change: production shows the same thing, so it still isn't this PR's.

langz:
Worth retesting this one on your phone — the likely cause landed after you last checked.

The `DrawerList` refactor in `e46b0751` (19:39 on the 15th, after your 15:07 note here) replaced the components that were being declared inside the render with plain arrays and a `Fragment`. That was what replaced the option node under your finger between the two taps. `keeps section option nodes mounted while previewing both directions` fails if I swap that file back to its previous version and passes on the current one, so node identity is genuinely fixed now.

On the current preview in WebKit with iPhone 13 emulation I couldn't reproduce the double-tap at all — 8 single taps across two fresh loads, both directions, and every one switched section first time:

| | Personal → Business | Business → Personal |
| --- | --- | --- |
| load 1 | one tap, one tap | one tap, one tap |
| load 2 | one tap, one tap | one tap, one tap |

That's an emulator and not your device, though, and the original only showed up on real iOS Safari — so I'd rather you confirm on the phone before we call it closed.

langz (approved):

langz (approved):
Final pass at `c99047e9` — and from my side this is done. Every single thing I raised across this review is now fixed, and fixed properly rather than papered over. Thank you for the patience, this was a long list and you worked through all of it 🙂

No new commits since my last look, so this round was re-verification rather than new ground. What I checked, and how:

| what | how I checked | result |
| --- | --- | --- |
| the `DrawerList` refactor doesn't regress shared consumers | ran `DrawerList` (61), `Dropdown` (78) and `SidebarMenu` (82) locally at `c99047e9` | 221 passed |
| option node identity (the iOS double-tap root cause) | `keeps section option nodes mounted while previewing both directions` | holds |
| the dimming gate is actually pinned | `does not reapply dimming while the section selector closes`, now hovering the *other* option | discriminates correctly |
| `hamburger.tsx` really is back to `main` | diffed the generated icon against `origin/main` | zero diff |
| localisation | `sectionLabel` + `containsCurrentPage` present in all four locales | ✅ |
| the legacy Portal sidebar can't be restyled by the new component | the only class the two still share is `dnb-sidebar-menu__item`, and the extension sets just `list-style: none` on it — which lands on a `<div>` in the legacy markup | no-op, no leakage |
| `badgeProps` renders as one type | the new `splitTopLevelUnion` is depth-aware; the three real doc types that mix `=>` with `\|` all survive it | ✅ |

Two leftovers, both tiny, neither worth holding the merge for.

**1. `hamburger.svg` kept one byte from the split round-trip.** The restore got the generated `hamburger.tsx` exactly back to `main`, but the asset is 192 bytes against `main`'s 191 — a trailing newline. The markup is byte-identical otherwise. It's the only one of the 594 files in `assets/icons/dnb/` that ends in a newline; the other 593, `hamburger_medium.svg` included, end at `>`. `optimizeSVG()` in `scripts/figma/tasks/assetsExtractors.ts` writes the SVGO output verbatim, so the next Figma sync will strip it again and produce an unrelated one-line diff. Worth truncating the newline so the file matches `main` exactly.

**2. `aria-label` on a non-linked `Group` goes nowhere.** Following on from the labelling thread earlier: on a *linked* group it correctly lands on the `<a>`, but on a non-linked one it's destructured out of `rest` and never applied, so it doesn't reach the `<li>` or the `<ul>` either. I rendered `<SidebarMenu.Group aria-label="Shortcuts">` and the attribute appears zero times in the output. Since an untitled group's list is deliberately left unlabelled, that currently leaves no way to name it. Putting it on the `<ul>` in the no-link case covers it:

```jsx
aria-label={!hasLink ? ariaLabel : undefined}
aria-labelledby={text ? titleId : !hasLink ? ariaLabelledBy : undefined}
```

One thing deliberately left open rather than fixed here: the first-open list jump in WebKit. I reproduced it on **production** at the same rate and shape, so it's existing `DrawerList` positioning behaviour that the taller sidebar rows just make easier to see — not this PR's, and better off as its own issue. I've collapsed that thread along with the other resolved ones so the conversation isn't cluttered, but the measurements are still there if you want them for the follow-up.

Approving. Nice work — the menu is in good shape 🎉

PR #8993, opened 2026-08-14
chore(Portal): remove duplicate mobile Portal Tools trigger

tujoworker:
The mobile sidebar duplicates the Portal Tools button that already exists in the portal header. Removing only the sidebar button leaves a single, consistent entry point.

### DIFF
--- a/packages/dnb-design-system-portal/src/e2e/responsiveness.spec.ts
+++ b/packages/dnb-design-system-portal/src/e2e/responsiveness.spec.ts
@@ -27,6 +27,10 @@ test.describe('Responsiveness', () => {
     )
     await menuButton.click()
 
+    await expect(
+      page.locator('#portal-sidebar-menu').getByText('Portal Tools')
+    ).toHaveCount(0)
+
     const sidebar = page.locator('nav#portal-sidebar-menu')
     await expect(sidebar).toHaveCSS('overflow-y', 'auto')
 
--- a/packages/dnb-design-system-portal/src/shared/menu/SidebarMenu.tsx
+++ b/packages/dnb-design-system-portal/src/shared/menu/SidebarMenu.tsx
@@ -28,7 +28,6 @@ import {
   setPageFocusElement,
   applyPageFocus,
 } from '@dnb/eufemia/src/shared/helpers'
-import PortalToolsMenu from './PortalToolsMenu'
 import { navStyle, resizeHandleStyle } from './SidebarMenu.module.scss'
 import { defaultTabsValue } from '../tags/defaultValues'
 
@@ -175,18 +174,6 @@ export default function SidebarLayout({
         )}
         ref={scrollRef}
       >
-        <PortalToolsMenu
-          triggerProps={{
-            left: 'large',
-            top: 'large',
-            bottom: 'large',
-            text: 'Portal Tools',
-            icon: 'chevron_right',
-            iconPosition: 'right',
-          }}
-          tooltipPosition="bottom"
-          hideWhenMediaLarge
-        />
         <ul className="dev-grid">{navItems}</ul>
       </nav>
 


langz:
It's perhaps a bit early to add the web/ios/android switch/toggle, if we do not have any content under ios or andrioid?
Screenshot 2026-08-14 at 10 45 29

langz:
Menu looks cool and great 🤩

langz:
Sometimes when pressing a link/menu point that can be expanded, it opens/changes the page and other times not.
Like these two menu points work differently, but look the same:
Screenshot 2026-08-14 at 10 49 33
Screenshot 2026-08-14 at 10 49 48


Just a thought, no worries, I think it's good to be flexible and support both. Just that my mind like that stuff always do the same.

packages/dnb-eufemia/src/extensions/sidebar-menu/SidebarMenuAccordion.tsx
```diff
@@ -0,0 +1,281 @@
...
+          >
+            {linkContent}
+          </Anchor>
+          {collapsible && (
+            <button
+              type="button"
+              className="dnb-sidebar-menu__accordion__toggle"
+              aria-expanded={isOpen}
+              aria-controls={`${id}-content`}
+              aria-label={`${isOpen ? 'Collapse' : 'Expand'} ${label}`}
```
langz: We could add localization.
Not sure if we need to do it now, as I guess the component is still only internal.

packages/dnb-eufemia/src/extensions/sidebar-menu/SidebarMenuAccordion.tsx
```diff
@@ -0,0 +1,281 @@
...
+
+  const itemStyle = {
+    '--sidebar-menu-indent': `${context.level * 2}rem`,
+  } as React.CSSProperties
+  const label = typeof text === 'string' ? text : 'section'
+  const currentIndicator = containsSelectedItem && !isOpen && (
+    <span
+      className="dnb-sidebar-menu__accordion__current-indicator"
+      role="img"
+      aria-label="Contains current page"
```
langz: We could add localization.
Not sure if we need to do it now, as I guess the component is still only internal.

packages/dnb-eufemia/src/extensions/sidebar-menu/SidebarMenuDocs.ts
```diff
@@ -0,0 +1,357 @@
...
+  },
+  defaultActiveSection: {
+    doc: 'Id of the initially visible section. Defaults to the section marked active, then the first section.',
+    type: 'string',
+    status: 'optional',
+  },
+  sectionLabel: {
+    doc: 'Accessible label for the section toggle group.',
+    type: 'React.ReactNode',
+    defaultValue: '"Menu section"',
```
langz: We could add localization.
Not sure if we need to do it now, as I guess the component is still only internal.

packages/dnb-eufemia/src/extensions/sidebar-menu/SidebarMenuDocs.ts
```diff
@@ -0,0 +1,357 @@
...
+    status: 'optional',
+  },
+  onSelectedItemChange: {
+    doc: 'Called when a leaf item is selected.',
+    type: '(itemId: string) => void',
+    status: 'optional',
+  },
+}
+
+export const SidebarMenuItemProperties: PropertiesTableProps = {
```
langz: We could also document `suffix`

packages/dnb-eufemia/src/extensions/sidebar-menu/SidebarMenuDocs.ts
```diff
@@ -0,0 +1,357 @@
...
+
+export const SidebarMenuItemEvents: PropertiesTableProps = {
+  onClick: {
+    doc: 'Called when the item is activated after the container selection state is updated.',
+    type: '(event: React.MouseEvent<HTMLElement>) => void',
+    status: 'optional',
+  },
+}
+
+export const SidebarMenuAccordionProperties: PropertiesTableProps = {
```
langz: We could also document `suffix`

langz:
Not sure if this is a real world scenario, but at times I like and expect the open accordions to close when selecting/activating/navigating to a new page. To prevent having such a long open menu. Not sure what's the best practice here. 
Here's a small video where I've visited a few pages in a session, and am opening the menu again. 
In our previous menu, I think I would be scrolled to the active menu item or such. Now I have to manually scroll a long way to "get back to where I last was in the menu".

https://github.com/user-attachments/assets/c65e6c75-51e7-41b3-9ee8-4c5d6aa38f27

langz (approved):

PR #8999, opened 2026-08-14
refactor(HeightAnimation): deprecate untilFound in favor of openOnFind

tujoworker:
`openOnFind` better communicates that collapsed content opens when browser find-in-page locates a match. This keeps `untilFound` as a deprecated compatibility alias while guiding consumers toward the clearer API.

Thanks @kimroen for the good input.


langz:
Could be included in a quick bugfix release, together with the text alignment of action items in the list component.

langz (approved):

PR #9006, opened 2026-08-17
deps: bump @navikt/fnrvalidator from 2.1.5 to 2.2.1 - add support for fødselsnummer from 2032

langz:
Routine maintenance bump of the `@navikt/fnrvalidator` runtime dependency (used for Norwegian national identity number validation, shipped to consumers) from 2.1.5 to 2.2.1 to stay current with upstream fixes.


tujoworker:
For some dependencies, I think it's good to include them in the release log. What about a deps decoration ([PR](https://github.com/dnbexperience/eufemia/pull/9007))? I don't think we ever need CI changes to be part of the release log.

tujoworker (approved):

PR #9007, opened 2026-08-17
chore: replace ci commit type with deps

tujoworker:
The dependency-related Conventional Commit category is currently named `ci`, which is narrower than its intended use. Use `deps` consistently for PR title validation, generated dependency updates, contributor guidance, and release notes.


langz (approved):
weee

PR #9020, opened 2026-08-17
chore: add screenshot color scheme option

tujoworker:
Screenshot tests currently need custom page mutations to capture dark mode, which can leak color-scheme state into later cases. Add a first-class `colorScheme` option to `setupPageScreenshot` and `makeScreenshot`, applying it before portal bootstrap and resetting it between navigations.


langz (approved):

langz:
Many more screenshot test incoming now? 😂 All tests also in all colorSchemes? ⏰

joakbjerk (approved):

tujoworker:
> Many more screenshot test incoming now? 😂 All tests also in all colorSchemes? ⏰ 

I hope not. I think using it granularly where it makes sense is the way to go. Same applies to brand theming when using tokens.

We may maybe "allow" one screenshot for dark mode per component or so. 

But I don't think this example usage is good either. Just wanted to have at least one.

packages/dnb-eufemia/src/core/vitest-screenshots/commands/screenshotEngine.ts
```diff
@@ -404,6 +437,7 @@ const ensureSession = async (
   await page.addInitScript(() => {
     ;(globalThis as Record<string, unknown>).IS_TEST = true
   })
+  await page.addInitScript(applyScreenshotColorScheme)
```
snorrekim, line 440:
```suggestion
  await page.addInitScript(() => {
    ;(globalThis as Record<string, unknown>).IS_TEST = true
    applyScreenshotColorScheme
  })
```

Could we perhaps do it in the same call and perhaps save a little bit of time?

tujoworker: It could be combined, but not quite as suggested: `applyScreenshotColorScheme` would only be referenced here, and it would not be available in the browser context once Playwright serializes the callback. We would need to merge both behaviors into one self-contained initializer. Since this runs once per session and would only save one protocol call, I think keeping the two concerns separate is clearer.

PR #9021, opened 2026-08-17
feat(List): support even striped rows

tujoworker:
Allows striped lists next to neutral backgrounds to start with an uncolored item, avoiding adjacent subtle background rows. Existing `striped` behavior remains unchanged, while `striped="even"` shifts the pattern to even items.

Motivation: https://dnb.enterprise.slack.com/archives/CMXABCHEY/p1786975403794129?thread_ts=1786365892.411519&channel=CMXABCHEY&message_ts=1786975403.794129

### DIFF
--- a/packages/dnb-eufemia/src/components/list/__tests__/Container.test.tsx
+++ b/packages/dnb-eufemia/src/components/list/__tests__/Container.test.tsx
@@ -66,13 +66,17 @@ describe('List Container', () => {
     expect(element.classList).toContain('dnb-list--separated')
   })
 
-  it('adds striped modifier class when striped is true', () => {
-    render(<Container striped>Content</Container>)
+  it.each([true, 'odd'] as const)(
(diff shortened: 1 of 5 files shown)


langz (approved):
Looks good to me :)
I guess we could have a visual test for the even example, but all good 🙏

packages/dnb-eufemia/src/components/list/__tests__/Container.test.tsx
```diff
@@ -74,6 +74,15 @@ describe('List Container', () => {
     expect(element.classList).toContain('dnb-list--striped')
   })
 
+  it('adds even modifier class when striped is even', () => {
```
langz, line 81: Could also add a test for `"odd",` just to verify that it works the same as `true`

tujoworker: Thanks for the recommendation ❤️

PR #9042, opened 2026-08-20
fix(FormStatus): align icon with a Button child

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1787223501615979

When a `Button` is used as the content of a `FormStatus`, the icon is misaligned — it stays at the top of the box instead of aligning with the button. Plain text and inline links align correctly because they flow within the text line, but a `Button` is taller than a single line, so the icon ends up too high.

Reported by a user via Slack:

```tsx
<FormStatus state="warning" variant="outlined">
  <Button variant="tertiary" text="Breaking the Icon styling" />
</FormStatus>
```

The shell intentionally aligns the icon to the first text line so multi-line messages keep the icon at the top. This centers the icon only when the content is a `Button`, leaving text alignment unchanged.

Deploy preview: https://8a2ef9d9.eufemia-e25.pages.dev/uilib/components/form-status/demos#with-a-button-as-content


langz:
A quick note on why I kept this button-specific rather than making it general.

The underlying cause is broader: the shell top-aligns the icon so it lines up with the *first line* of the message, which is what we want for multi-line text. Anything taller than a single text line, like a `Button`, then leaves the icon sitting too high.

I did look at solving it generically (center the icon whenever the content is a single element), but I don't think that's safe here:

- Our inline custom-content case (text + `Anchor`) is trickier than it looks. In CSS the anchor still matches `:only-child`, since text nodes don't count as element siblings, so a general `:has(> :only-child)` rule would catch it too and mis-center the icon as soon as the message wraps to two lines.
- There's no reliable static CSS signal to tell "a single tall control" apart from "flowing text that might wrap".

To be clear, the `.dnb-button` rule leans on the same single-line assumption — a mixed `text + <Button>` that wraps would center the icon too. The difference is it doesn't regress anything: no existing text or inline-link pattern changes behavior, whereas a general rule would.

So targeting `.dnb-button` felt like the right boundary: it fixes the reported case (and all button variants) without touching plain text or inline links. If we later need it for other controls like `ToggleButton`, we can extend the selector to a small allow-list. I'd rather do that when there's a real case than over-reach now.

tujoworker:
A tertiary button should ideally include an icon. Also, should we add some text before/after just to make it more clear that this is just a part of the content? also, can we move the example after the warning example we have further down?

Screenshot 2026-08-20 at 13 47 45

langz:
> A tertiary button should ideally include an icon. Also, should we add some text before/after just to make it more clear that this is just a part of the content? also, can we move the example after the warning example we have further down?
> 
> Screenshot 2026-08-20 at 13 47 45

Good points, thanks! Updated the demo: the tertiary button now has a `bell` icon, added some leading text so it's clear the button is just part of the content, and moved it below the warning example.

tujoworker (approved):

PR #9046, opened 2026-08-20
feat(Autocomplete): support inline results

joakbjerk:
## Motivation

https://dnb-asa.atlassian.net/jira/software/c/projects/EDS/boards/734?assignee=712020%3A11a274b8-3b2a-4c7f-9968-07cf788ff7dd&selectedIssue=EDS-939

Autocomplete only renders results as a floating overlay (portal). Some layouts need them in normal document flow instead — e.g. when the overlay is clipped, fights surrounding scroll, or should be a permanent part of the page.

## Solution

New opt-in boolean `inline` prop: results render persistently open in document flow (no portal), flat, with the toggle button hidden. Default overlay behaviour is unchanged. It reuses DrawerList + `skipPortal`, so filtering, selection, keyboard and ARIA match the overlay mode.

**Why:** boolean fits Autocomplete's existing API; persistently open avoids a layout jump on every toggle; reusing DrawerList keeps behaviour/a11y consistent.

**Alternatives:** `variant="inline"` enum (kept the boolean API); open/close in flow (layout shift); matching `Field.MultiSelection` inline exactly with `max-height: none`/no dividers (kept bounded/scrollable + dividers since results can be large).


tujoworker:
Preview: https://feat-autocomplete-inline-res.eufemia-e25.pages.dev/uilib/components/autocomplete#inline-results

tujoworker:
I found and fixed issues with inline keyboard navigation, Provider/runtime state synchronization, initial input filtering, and disabled option interaction. The affected Autocomplete and DrawerList suites pass (207 tests), along with type checking and lint.

tujoworker:
Refined the inline presentation: removed the list background and surrounding outline, disabled inline open/height/scroll animations, and applied the max height immediately.

packages/dnb-eufemia/src/components/autocomplete/Autocomplete.tsx
```diff
@@ -2515,9 +2554,11 @@ function AutocompleteComponent(ownProps: AutocompleteAllProps) {
               defaultValue={defaultValue}
               scrollable={scrollable}
               focusable={focusable}
-              noAnimation={noAnimation}
-              noScrollAnimation={noScrollAnimation}
+              noAnimation={inline || noAnimation}
+              noScrollAnimation={inline || noScrollAnimation}
               skipPortal={skipPortal}
+              inline={inline}
+              ignoreEvents={disabled}
```
langz: Not entirely sure if this is correct or not.
What happens if user themselves set `ignoreEvents`?
Also, I guess this line potentially adds `ignoreEvents` to all different modes of autocomplete.

Could it be breaking for some of our consumers? 🤔

tujoworker:
Addressed the inline Autocomplete follow-ups: fixed context/default handling, filtering and runtime toggling; restored keyboard navigation, native Tab order and initial focus; disabled interaction when disabled; removed card styling/animation, capped height, added spacing, reset list indentation and preserved stretch; added SSR rendering and updated unit, CSS and visual snapshots.

packages/dnb-eufemia/src/components/autocomplete/Autocomplete.tsx
tujoworker: Good catch. I scoped the automatic event suppression to `inline && disabled`, so non-inline Autocomplete keeps its previous behavior. Added regression coverage for both cases.

langz: Thanks, great :)

tujoworker (approved):
Looks good.

PR #9049, opened 2026-08-21
refactor(HelpButton): use CSS gap in inline content

tujoworker:
Moves the internal inline-help content layout to native CSS gap while preserving the existing public API and visual spacing. This reduces reliance on the legacy Flex layout engine ahead of its removal in the next major version.


langz (approved):

langz:
I see a few visual tests are failing for this one 🤔

PR #9062, opened 2026-08-21
chore(visual-regression): surface the diff report in the run summary and host it

langz:
## Why

When a visual screenshot test fails on CI, contributors have to open the run, find the **visual-test-artifact**, download the ZIP, unzip it and open `index.html` locally just to see _which_ screenshots differ and _why_. That friction makes triage slow.

## What

Make the existing diff report available without leaving the browser:

- **Run summary table** — the failed screenshots are now rendered directly on the GitHub Actions run **Summary** page (test name, `file:line`, `data-visual-test` id and the diff message).
- **Hosted report** — on failure the self-contained report is deployed to Cloudflare Pages (reusing the same credentials/project as the branch preview deploy), and each summary row embeds a diff thumbnail linking to the full before/after/diff view.
- **Graceful fallback** — when hosting is unavailable (e.g. a forked PR without secrets), the summary still lists the failures and points to the downloadable artifact, which is unchanged.

The reporter now writes a small structured `report.json` next to `index.html` (image paths kept in sync with the HTML via a shared helper) that the workflow reads to build the summary.

No new repository secrets or permissions are required.

### DIFF
--- a/.github/workflows/visual-regression.yml
+++ b/.github/workflows/visual-regression.yml
@@ -177,6 +177,85 @@ jobs:
           path: |
             ./packages/dnb-eufemia/visual-diff-report/**
 
-      - name: Run visual tests info
+      # The visual diff report is a self-contained folder (index.html +
+      # images/). Only proceed when it actually exists (a non-screenshot
+      # failure produces none), and detect whether Cloudflare credentials are
+      # available so the deploy is skipped gracefully on forked PRs (no
+      # secrets) or when the token is not configured. The secret comparison
+      # yields only a boolean; the secret value itself is never emitted.
+      - name: Detect report hosting capability
+        id: report_host
         if: failure()
-        run: echo '\n\n👉 Download the diff files as a ZIP file. \nIt is called "visual-test-artifact" and you find it in the test "Summary" under "Artifacts".\n\n\n'
+        run: |
+          if [ -f packages/dnb-eufemia/visual-diff-report/report.json ]; then
+            echo "has_report=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "has_report=false" >> "$GITHUB_OUTPUT"
+          fi
+          echo "enabled=${{ secrets.CLOUDFLARE_API_TOKEN != '' }}" >> "$GITHUB_OUTPUT"
+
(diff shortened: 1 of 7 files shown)


langz:
## See it live

The companion PR #9066 intentionally breaks one Badge screenshot baseline to trigger this pipeline, so you can see exactly how a failing visual-regression run looks with these changes:

- **Run summary** — the failed-screenshot table with before/after/diff **thumbnails**: https://github.com/dnbexperience/eufemia/actions/runs/32484610617
- **Failing job** — the `Deploy visual report to Cloudflare Pages` → `Publish visual report to job summary` → `Run visual tests info` steps, with **clickable links in the log**: https://github.com/dnbexperience/eufemia/actions/runs/32484610617/job/96778228259
- **Hosted report** — the self-contained, browsable report (expected / actual / diff): https://12563603.eufemia-e25.pages.dev

Notes: the hosted URL is an immutable per-run Cloudflare Pages preview deployment (may be pruned eventually). #9066 is a throwaway demo — **not for merge**.

tujoworker (changes requested):
The report is now deployed to a public Cloudflare URL, but index.html still interpolates test-derived values into markup without HTML escaping. A crafted test name, file path, visual-test id, or failure message can therefore become executable markup in the hosted report. Please escape every untrusted value used by the HTML renderer before publishing it.

langz:
Thanks, good catch — addressed in 9607150.

`renderHtml` now HTML-escapes every test-derived value it interpolates — test name, `file:line` path, `data-visual-test` id, failure message, and the image `href`/`src` paths — via a new `escapeHtml` helper (`& < > " '`). The failure message is escaped before newlines are turned into `<br />`.

I also removed the inline `onclick` copy handler (the JS-string injection point) in favour of a `data-clipboard-text` attribute read by a single delegated click listener, so nothing is evaluated as JS.

Added tests asserting a crafted payload is escaped everywhere and that no inline `onclick` is emitted.

tujoworker (approved):
The hosted HTML report now escapes all test-derived text and attribute values, removes the inline JavaScript interpolation point, and includes focused regression coverage. No remaining blocking findings.

PR #9065, opened 2026-08-21
chore(ci): show diffs for screenshot size mismatches

tujoworker:
Visual test reports currently omit the diff image when expected and actual screenshots have different dimensions. Generate a shared-canvas diff so added or removed regions are visible alongside the existing expected and actual images.

Estimated mismatch-only overhead: ~1.7 ms per 376×296 image; passing tests are unaffected.


langz (approved):
Looks good to me, got an ai agent to make a dry run / smoke test:
_Ran it against a realistic case (120×80 reference vs 120×100 actual, with a changed box inside the shared area). It produced exactly 3000 red pixels = 600 for the changed box + 2400 for the 20 added rows, and the rendered diff shows a clean red box top-left plus a red bar for the added region. Nothing over- or under-highlighted._

And it looked to work great 🚀

PR #9067, opened 2026-08-21
fix(Modal): correct dialogTitle fallback casing and docs

langz:
## Motivation

Follow-up from a review of generic fallback copy in Eufemia. `Modal.dialogTitle` is the `aria-label` applied to a `Modal` / `Dialog` / `Drawer` only when neither `title` nor `labelledBy` is provided. Two small correctness issues:

1. **Casing inconsistency (nb-NO).** `nb-NO` used title-case **"Separat Vindu"**, while `da-DK` ("Separat vindue") and `sv-SE` ("Separat fönster") use sentence case. Aligned nb-NO to **"Separat vindu"** (correct Bokmål sentence case).
2. **Incorrect documentation.** The `dialogTitle` doc (in both `ModalDocs` and the `ModalContentProps` JSDoc) claimed it *"Defaults to `Vindu`"* — but the actual fallback is the localized string (e.g. "Separat vindu"). Corrected the text and added guidance that a meaningful `title` or `labelledBy` is preferred for accessibility, since `dialogTitle` is only a generic fallback.

No visual impact: `dialogTitle` is an `aria-label`, not rendered text. Only two tests that used the fallback string as a value assertion were updated.

## Deliberately out of scope

The more substantive accessibility improvement would be to **warn in development when a dialog opens with no accessible name** (no `title` / `labelledBy` / `aria-label`) — the industry norm (react-aria, Radix) and consistent with the existing "A Dialog or Drawer needs an h1…" warning. It's left out here because an always-on warning fires for the many intentionally title-less modals in the test suite (one test asserts `warn` is called exactly once), so it needs its own change with test updates. Happy to follow up if the team wants it.

### DIFF
--- a/packages/dnb-eufemia/src/components/modal/ModalDocs.ts
+++ b/packages/dnb-eufemia/src/components/modal/ModalDocs.ts
@@ -112,7 +112,7 @@ export const ModalProperties: PropertiesTableProps = {
     status: 'deprecated',
   },
   dialogTitle: {
-    doc: 'The aria label of the dialog when no labelledBy and no title is given. Defaults to `Vindu`.',
+    doc: 'The `aria-label` used for the dialog when neither `title` nor `labelledBy` is given. Prefer a meaningful `title` or `labelledBy` for accessibility; this is only a generic fallback. Defaults to a localized string (`Separat vindu` for `nb-NO`).',
     type: 'string',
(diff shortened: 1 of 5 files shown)


tujoworker (approved):
The fallback and documentation now consistently use the intended casing.

PR #9068, opened 2026-08-21
fix(StepIndicator): use sentence case for en-GB overviewTitle

langz:
## Motivation

The en-GB `StepIndicator.overviewTitle` used title case **"Steps Overview"**, which is inconsistent with the sentence-case convention used throughout the locale (e.g. "Show options", "Upload documents", "Next page"). The Norwegian/Danish/Swedish values are single words, so this was the only multi-word one and stood out.

Aligned it to **"Steps overview"** and updated the two places that document the default (`StepIndicatorProps` JSDoc and `StepIndicatorDocs`). The value is used as the trigger's `aria-label`, so there is no visual impact.


tujoworker (approved):
The en-GB translation now uses the intended sentence case.

PR #9069, opened 2026-08-21
chore(analytics): add anonymous page-view `/collect` ingest endpoint

joakbjerk:
Adds a public `POST /collect` route so the docs portal can record anonymous page views without shipping the bearer token to the browser. The route relies on the Akamai edge lock (`X-Edge-Auth`) instead of the bearer token, while `/records` stays token-protected. Records are anonymous (`{type, path, env, timestamp}`) and stored as newline-delimited JSON under a unique per-batch key so events don't overwrite each other.

The optional `env` label lets a shared collector keep environments apart, and the Glue table gains matching columns so the page-view fields are queryable.

Part of adding page-view analytics to the Eufemia docs portal. The portal-side tracker is a separate PR (#9082). No Akamai change is needed — the existing `/analytics/*` edge route already reaches `/collect`.


tujoworker (changes requested):
The Lambda handles POST /collect, but the infrastructure does not add a corresponding API Gateway route. Akamai can forward the path only after the origin exposes it, so production requests will currently return 404. Please add the POST /collect route to tools/analytics/infra/main.tf and cover the deployed routing contract.

tujoworker (approved):

langz (approved):

PR #9070, opened 2026-08-21
fix(Modal): use sentence case for en-GB dialogTitle

langz:
## Motivation

The en-GB `Modal.dialogTitle` used title case **"Dialog Window"**, while `da-DK` ("Separat vindue") and `sv-SE` ("Separat fönster") use sentence case. Aligned en-GB to **"Dialog window"** for consistency.

This complements #9067, which applies the same sentence-case correction to `nb-NO` ("Separat Vindu" → "Separat vindu"). `dialogTitle` is the dialog's `aria-label` fallback (used only when neither `title` nor `labelledBy` is given), so there is no visual impact; two Dialog tests that assert the resolved string were updated.


tujoworker (approved):
The en-GB translation now uses the intended sentence case.

PR #9072, opened 2026-08-21
chore(Autocomplete): correct stale default for title prop

langz:
## Motivation

`Autocomplete`'s `title` prop was documented as `Defaults to \`Skriv og få alternativer\``, but the actual default is the localized value **`Skriv og velg`** (`Autocomplete.title`). The `title` prop resolves via `getTranslation`, and the input placeholder even falls back to it (`placeholder || title`).

The string "Skriv og få alternativer" was **orphaned** — it existed only in these two doc locations (JSDoc + `AutocompleteDocs`) and nowhere in the locales, defaults, or tests. This is the same doc-drift class as the recent Modal `dialogTitle` "Vindu" correction (#9067).

Doc-only change — no behavior, locale, or test impact.


tujoworker (approved):

PR #9073, opened 2026-08-21
chore(Slider): quote nb-NO default for add/subtract title props

langz:
## Motivation

`Slider`'s `addTitle` and `subtractTitle` docs quoted the en-GB values (`Increase (%s)` / `Decrease (%s)`), but the actual default is the localized value from nb-NO (Eufemia's default locale): **`Øk (%s)`** / **`Reduser (%s)`**.

Most component docs (Pagination, GlobalStatus, Autocomplete, …) quote the nb-NO default; this aligns Slider with that convention and makes the documented default accurate. Doc-only — no behavior, locale, or test impact.


tujoworker (approved):
The documented defaults now accurately represent the localized strings.

PR #9074, opened 2026-08-21
chore(StepIndicator): fix stepTitle documented default (missing colon)

langz:
## Motivation

The `StepIndicator.stepTitle` doc quoted `Defaults to \`Step %step of %count\``, but the actual default value ends with a colon: **`Step %step of %count:`** (matching the `stepTitle` locale value). Corrected the JSDoc and `StepIndicatorDocs` to include it. Doc-only — no behavior, locale, or test impact.


tujoworker (approved):

PR #9075, opened 2026-08-21
chore(markdown-tables-utils): prevent polynomial ReDoS in table parsing

langz:
Motivation:
- https://github.com/dnbexperience/eufemia/security/code-scanning/58
- https://github.com/dnbexperience/eufemia/security/code-scanning/59

Two CodeQL `js/polynomial-redos` alerts flagged the shared markdown table helpers. The table-separator matcher and the inline link matcher could run in O(n²) time on crafted input, which is a denial-of-service risk when parsing untrusted markdown.

The separator check now splits on pipes instead of relying on a single regex with ambiguous, adjacent whitespace quantifiers, and the inline link matcher excludes the opening `[`/`(` delimiters from its content classes so it fails fast on unbalanced brackets. Both keep matching linear while preserving the existing output.

### DIFF
--- a/tools/markdown-tables-utils/src/__tests__/markdownTablesUtils.test.ts
+++ b/tools/markdown-tables-utils/src/__tests__/markdownTablesUtils.test.ts
@@ -48,4 +48,37 @@ describe('markdownTablesUtils', () => {
       ],
     ])
   })
+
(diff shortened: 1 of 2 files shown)


tujoworker (approved):
Nice findings!

PR #9076, opened 2026-08-21
chore(Autocomplete): complete showOptionsSr documented default

langz:
## Motivation

`Autocomplete.showOptionsSr` documented its default as `Bla gjennom alternativer`, but the actual locale value is the full string **`Bla gjennom alternativer, lukk med esc knappen`**. Completed the JSDoc and `AutocompleteDocs` to match. Doc-only — no behavior, locale, or test impact (the Autocomplete test already asserts the full value).


tujoworker (approved):

Issue #9071, opened 2026-08-21
Support scrollable item list in inline `Field.MultiSelection`


RF0:
## 🚀 Feature Proposal

Support a scrollable item list for `Field.MultiSelection` with
`variant="inline"`, while keeping the search field and selected tags visible.

We are using `Field.MultiSelection` for selecting countries from a long list.
The desired behavior is that the search input and selected tags remain available
at the top, while only the selectable item list scrolls.

## Motivation

Long multi-selection lists can take up too much vertical space. In our case,
the list contains countries, so users need search and selected tags to stay
accessible while scrolling through the available options.

We are using `variant="inline"` because the default popover variant creates a
difficult scrolling experience for long country lists when the field is placed
far down on the page.

`Field.MultiSelection` renders the popover with:

```tsx
autoAlignViewportThreshold={0.75}
```

The Popover documentation describes this as keeping `placement="bottom"` until
the trigger reaches the bottom quarter of the viewport.

For a long list of countries, this means the list can remain rendered below the
input even when the field is already far down the page. The popover content is
then constrained by the remaining viewport height below the input. In practice,
this can leave only a small part of the country list visible, forcing users to
scroll through just a few countries at a time.

The inline variant avoids that dropdown positioning issue, but inline rendering
currently has no supported way to constrain only the item list height. Today,
this requires application code to depend on Eufemia’s internal DOM structure.
That makes the consuming application brittle if Eufemia changes its internal
markup or class names.

We would like a supported way to handle this use case without querying internal
Eufemia elements or manually mutating DOM styles.

## Example

Today, we use `Field.MultiSelection` with the inline variant, search field,
select-all, and selected tags enabled:

```tsx
<Field.MultiSelection
  variant="inline"
  showSearchField
  showSelectAll
  showSelectedTags
  selectedItemsCollapsibleThreshold={20}
  validateUnchanged
  minItems={1}
  label={label}
  data={filteredData}
  value={selectedCountries}
/>
```

To make only the item list scroll, while keeping the search field and selected
tags visible, we currently have to target an internal Eufemia element:

```ts
const selector =
  ".dnb-forms-field-multi-selection__items";

const items =
  containerRef.current?.querySelector(selector) as HTMLElement | null;

if (items) {
  items.style.maxHeight = "28rem";
  items.style.overflowY = "auto";
}
```

The screenshot below shows the desired behavior in our current implementation: the search field and selected country tags stay visible at the top, while only the list of countries below is scrollable.

[...]

PR #9077, opened 2026-08-21
fix(Field.MultiSelection): place selected items before search

tujoworker:
Selected items should be visible before users refine the available choices. This reorders the selected-items area ahead of search and removes redundant dividers while aligning the surrounding spacing with the intended UX.

Preview: https://codex-feat-multi-selection-s.eufemia-e25.pages.dev/uilib/extensions/forms/base-fields/MultiSelection#scrollable-inline-item-list


tujoworker:
This PR complements #9071 by establishing the intended layout order: selected items, search, then options. It does not add a supported way to constrain and scroll only the inline item list; that remains the scope of #9071.

langz (approved):

PR #9078, opened 2026-08-21
fix(renderWithFormatting): prevent polynomial ReDoS in link parsing

langz:
`renderWithFormatting` renders markdown-like text (links, bold, code) and is a publicly exported shared helper used by translations and Forms field labels. Its inline link matcher and the marker fast-path used content classes that did not exclude the opening `[` delimiter, so long runs of unbalanced brackets triggered O(n²) backtracking — a denial-of-service risk when formatting untrusted or interpolated strings.

Excluding `[` from the link/URL content classes makes matching fail fast on unbalanced brackets and stay linear, while preserving the existing output (including URLs that contain parentheses).


tujoworker (approved):

PR #9079, opened 2026-08-21
fix(Theme): prevent polynomial ReDoS in query parsing

langz:
`getTheme` reads the optional `?eufemia-theme=<name>` override from `window.location.search`, which is attacker-influenceable. It used a `/.*eufemia-theme=([^&]*).*/` regex whose leading and trailing `.*` cause O(n²) backtracking on a long query string that lacks the marker — a denial-of-service risk at app startup.

Replace it with `URLSearchParams`, which is linear and also decodes the value correctly. Behavior is otherwise unchanged (the existing query-override test still passes).


tujoworker (approved):

PR #9080, opened 2026-08-21
fix(input-masked): prevent polynomial ReDoS in numeric value cleanup

langz:
When a currency field uses a dynamic suffix (e.g. `krone` vs `kroner`), `cleanNumericValue` strips the mismatched trailing text before re-parsing. It did so with `/[^\d.,·-]+$/`, which backtracks in O(n²) time on a long non-numeric value ending in a digit — a denial-of-service risk when a long string value is fed to a masked numeric field.

Replace the regex with a linear trailing-character scan. Behavior is unchanged (the existing `Field.Currency` dynamic-suffix tests still pass). The helper is now exported so the behavior can be unit-tested directly.


tujoworker (approved):

PR #9081, opened 2026-08-21
feat(Field.MultiSelection): support scrollable inline item list

### LINKED ISSUE #9071: Support scrollable item list in inline `Field.MultiSelection`

tujoworker:
Long inline MultiSelection lists can consume excessive page height and force consumers to target internal markup. Add a supported `maxHeight` prop so only the options list scrolls while selected tags and search remain visible. Selected tags retain height animation when adding or removing tags changes their wrapped rows.

Closes #9071
Preview https://codex-feat-multi-selection-s.eufemia-e25.pages.dev/uilib/extensions/forms/base-fields/MultiSelection/demos#inline-variant


langz:
branch prefixing`codex` 😄

packages/dnb-eufemia/src/extensions/forms/Field/MultiSelection/MultiSelectionItemList.tsx
```diff
@@ -160,7 +162,19 @@ export function MultiSelectionItemList({
   }
 
   return (
-    <ScrollView className={clsx('dnb-forms-field-multi-selection__items')}>
+    <ScrollView
+      className={clsx('dnb-forms-field-multi-selection__items')}
```
langz, line 166: The new scrollable list should be keyboard-scrollable. `interactive="auto"` makes the container focusable only when it actually overflows, and activates the focus outline the stylesheet already defines for `&__items`:

```suggestion
      className={clsx('dnb-forms-field-multi-selection__items')}
      interactive={maxHeight === undefined ? undefined : 'auto'}
```

langz (commented):
**Keyboard access to the new scrollable region** (inline suggestion below). As shipped, the options list scrolls but its container isn't keyboard-focusable, so a keyboard-only user has no way to scroll it themselves. `ScrollView` already supports exactly this via `interactive="auto"` (focusable only when a scrollbar is actually present). Gating it on `maxHeight` keeps the popover variant untouched, and it also activates the `&__items:focus-visible` outline that's already defined in the stylesheet. Verified in a browser: once the container is focusable, PageDown / Arrow / End scroll the list.

langz:
I tested a bit on my laptop, and on my iPhone, and it looks and works very well.

My only concern is if users will clearly see and understand that the following list is scrollable? It could be difficult to see.

Is it normal that the scroll bar is always visible? 🤔 

This could just be the default/native experience, so I'm not flagging this as an issue, just a concern/though.

image

Here's a short video displaying that there's no scroll bar visual until actually trying to scroll in that area: https://share.icloud.com/photos/0a8G62wau6SrkEFhHcnk0zFig

langz (approved):

langz:
> I tested a bit on my laptop, and on my iPhone, and it looks and works very well.
> 
> My only concern is if users will clearly see and understand that the following list is scrollable? It could be difficult to see.
> 
> Is it normal that the scroll bar is always visible? 🤔
> 
> This could just be the default/native experience, so I'm not flagging this as an issue, just a concern/though.
> 
> image
> 
> Here's a short video displaying that there's no scroll bar visual until actually trying to scroll in that area: https://share.icloud.com/photos/0a8G62wau6SrkEFhHcnk0zFig

On a second thought, I think this is just native and the expected behavior of the browsers i'm using.
So I don't think this is anything to look into.

tujoworker:
> I tested a bit on my laptop, and on my iPhone, and it looks and works very well.
> 
> My only concern is if users will clearly see and understand that the following list is scrollable? It could be difficult to see.
> 
> Is it normal that the scroll bar is always visible? 🤔
> 
> This could just be the default/native experience, so I'm not flagging this as an issue, just a concern/though.
> 
> Here's a short video displaying that there's no scroll bar visual until actually trying to scroll in that area: https://share.icloud.com/photos/0a8G62wau6SrkEFhHcnk0zFig

I agree. But this is a general issue with the way macOS handles scrollbars. But there should be scrollbar visible initially actually. Not that this helps too much. But that should at least be an indicator for most users.

packages/dnb-eufemia/src/extensions/forms/Field/MultiSelection/MultiSelectionItemList.tsx
tujoworker: This is implemented now.

PR #9082, opened 2026-08-24
chore(Portal): add anonymous page-view tracking

joakbjerk:
Adds anonymous page-view tracking to the docs portal, sending only the pathname (no query, hash, identifiers, cookies or device storage), plus an environment label, to the analytics `/collect` endpoint. Events are buffered in memory and flushed with `sendBeacon` on `pagehide`/`visibilitychange`, or eagerly at the collector's batch limit, so navigation is never blocked and nothing is retried across reloads. Consecutive views of the same path (a re-mount) are recorded once.

Ships dark: tracking is off unless a build sets `VITE_ANALYTICS_ENDPOINT`, so this can merge without sending anything. Enable per environment by pointing it at the collector and setting `VITE_ANALYTICS_ENV` (e.g. `prod`); locally, a `.env.local` points at the dev logger or the real collector on demand. The beacon body is `text/plain` so the cross-origin request stays a CORS simple request and is delivered without a preflight (which `sendBeacon` cannot satisfy). The existing `/analytics/*` edge route already reaches the Lambda's `/collect`, so no new Akamai route is needed.

The backend `/collect` endpoint is #9069.


langz (approved):
Image

tujoworker (approved):

PR #9083, opened 2026-08-24
fix(Field.MultiSelection): add accessible name to the search field

langz:
This was found while reviewing #9081 (real-browser axe run against the preview build): the violation is present on the existing inline demo too, so it is pre-existing and independent of that PR's `maxHeight` feature.

## Why the placeholder does not already cover this

A reasonable question is "the search field already has a placeholder — isn't that read by screen readers?" It is not, and the distinction matters:

- A **native HTML `placeholder`** attribute *is* used as a last-resort fallback in the accessible-name computation (HTML-AAM / accname). If Eufemia emitted one, the field would technically have a name.
- But Eufemia's `Input` does **not** emit a native `placeholder`. It sets **`aria-placeholder`** on the input and renders the visible placeholder text as a separate decorative `<span>` (not linked via `aria-labelledby`/`aria-describedby`). `aria-placeholder` is a *hint* and is explicitly **excluded** from the accessible-name computation.

So the search input had **no accessible name at all** (no `<label>`, no `aria-label`, no `aria-labelledby`, only `aria-placeholder`), which is a WCAG 4.1.2 (Name, Role, Value) failure and the critical axe `label` violation.

### Verified against the base commit

Rendering the pre-fix component and inspecting the emitted input:

```
<input ... id="id-ra" aria-placeholder="Søk..." type="search" ...>
```

- native `placeholder`: none · `aria-label`: none · `aria-labelledby`: none · `<label for>`: none
- querying the searchbox *by* the placeholder text as its name → not found (placeholder is not the name)
- axe → **1 critical `label` violation** ("Form elements must have labels")

With this PR a visually-hidden `<label for>` is added, the field's accessible name becomes `Søk`, and axe reports **0** violations. No visual change.

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/MultiSelection/MultiSelection.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/MultiSelection/MultiSelection.tsx
@@ -723,6 +723,7 @@ function MultiSelection(props: FieldMultiSelectionProps) {
   const searchContent = (
     <MultiSelectionSearch
       show={showSearchField}
+      label={translation.searchLabel}
       placeholder={translation.searchPlaceholder}
       value={searchValue}
       disabled={disabled}
--- a/packages/dnb-eufemia/src/extensions/forms/Field/MultiSelection/MultiSelectionSearch.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/MultiSelection/MultiSelectionSearch.tsx
@@ -2,6 +2,7 @@ import { Input } from '../../../../components'
 
 export type MultiSelectionSearchProps = {
   show: boolean
+  label: string
   placeholder: string
   value: string
   disabled?: boolean
@@ -10,6 +11,7 @@ export type MultiSelectionSearchProps = {
 
 export function MultiSelectionSearch({
   show,
+  label,
   placeholder,
   value,
   disabled,
@@ -22,7 +24,8 @@ export function MultiSelectionSearch({
   return (
     <Input
       type="search"
-      label={false}
+      label={label}
+      labelSrOnly
       icon="loupe"
       iconPosition="left"
       placeholder={placeholder}
--- a/packages/dnb-eufemia/src/extensions/forms/Field/MultiSelection/__tests__/MultiSelection.test.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/MultiSelection/__tests__/MultiSelection.test.tsx
@@ -942,6 +942,28 @@ describe('MultiSelection', () => {
       )
     })
 
+    it('gives the search input an accessible name', async () => {
+      const data = [
+        { value: 'option1', title: 'Option 1' },
+        { value: 'option2', title: 'Option 2' },
+      ]
+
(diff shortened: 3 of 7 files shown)


langz:
## Why this is needed: `aria-placeholder` is not an accessible name (with a screen-reader demo)

To make the need reviewable — and testable with a real screen reader — here is a standalone demo plus a two-engine verification.

**▶️ Open the demo (one click, then use a screen reader):**
https://htmlpreview.github.io/?https://gist.githubusercontent.com/langz/c16784cf86d9a34b5a5d45527e9f596e/raw/multiselection-search-a11y.html

**Source / download:** https://gist.github.com/langz/c16784cf86d9a34b5a5d45527e9f596e

### How to test
1. Turn on a screen reader — **VoiceOver** (macOS): `⌘F5`. **NVDA** (Windows): `Ctrl+Alt+N`.
2. `Tab` into each of the three fields and listen to what is announced.
3. Or inspect each input in DevTools → **Accessibility** pane and read the computed **Name**.

The three fields look identical and all show the same “Søk…” hint — the only difference is in the accessibility tree:

| # | Field | What a screen reader announces |
|---|-------|-------------------------------|
| **A** | `aria-placeholder` only — **what Eufemia renders today** | *“search text field / edit, blank”* — **no name** |
| **B** | visually-hidden `<label for>` — **this PR** | *“Søk, search text field”* — **named**, no visual change |
| **C** | native `placeholder` attribute — *reference, not what Eufemia does* | browser-dependent |

### Verified programmatically (two independent engines)

Reproducing Eufemia's actual DOM (the visible “Søk…” is a decorative `aria-hidden` span; the input carries `aria-placeholder`, **not** a native `placeholder`):

| Case | `dom-accessibility-api` (W3C accname, used by Testing Library) | `axe-core` 4.12.1 `label` rule |
|------|:--:|:--:|
| **A** — `aria-placeholder` only (today) | name = `""` | **VIOLATION (critical)** — “Form elements must have labels” |
| **B** — sr-only `<label>` (this PR) | name = `"Søk"` | **PASS** |
| **C** — native `placeholder` (reference) | name = `""` | PASS |

### Takeaways
- `aria-placeholder` is a **hint**, explicitly **excluded** from the accessible-name computation — so the field had **no name at all** (WCAG 4.1.2 Name, Role, Value → the critical axe `label` violation).
- The confusion (“isn’t the placeholder read?”) comes from case **C**: a *native* `placeholder` is treated by some browsers/tools as a last-resort fallback name. But Eufemia never emits a native placeholder (it uses `aria-placeholder` + a decorative span), and any placeholder-as-name would vanish the moment the user types anyway.
- The fix adds a visually-hidden `<label>` (`labelSrOnly`) → accessible name becomes `"Søk"`, axe passes, **no visual change**.

This is also reproducible directly in the repo by rendering `Field.MultiSelection` with `showSearchField` at the base commit (`querySelector('input[type="search"]')` has no `<label for>`, no `aria-label`, only `aria-placeholder`; axe reports one critical `label` violation), which is what the added unit test locks in.

langz:
I tried testing this in chrome, and did not see the need for this PR

PR #9084, opened 2026-08-24
fix(Field.MultiSelection): scroll the focused option into view

langz:
## Problem

In `Field.MultiSelection`, tabbing through the options moved keyboard focus to checkboxes that stayed **scrolled out of view** — the focus indicator was off-screen (WCAG 2.4.7 Focus Visible).

**Root cause:** each checkbox's focusable `<input>` is an oversized (~2000×2000px), absolutely-positioned hit-area. The browser's native "scroll focused element into view" sees that huge box as already on-screen, so it never scrolls the list to reveal the focused row.

This affected **both variants**: the popover list (scrolls via CSS `max-height`) and the inline list (scrolls when a `maxHeight` is set). Verified in a real browser against the preview build.

## Fix

Scroll the focused row into view on `focus` in the shared list component (`MultiSelectionItemList`). Because it keys off the `focus` event, it covers **Tab, arrow keys and programmatic focus** uniformly, in both variants. `block: 'nearest'` only scrolls when the row isn't already visible, and the search field (which lives outside the list) never triggers it.

The popover's arrow-key handler previously did this manually for arrow keys only; that call is now redundant and removed, leaving a single code path.

## Tests

- New test: focusing an option directly (as Tab does) scrolls its row into view — in the **inline** variant, which had no scroll-on-navigation before.
- Existing arrow-key navigation test updated (focus now drives the scroll); still asserts the row scrolls with `{ behavior: 'smooth', block: 'nearest' }`.
- The existing "does not scroll when focus moves to the search input" test still passes.
- Full `MultiSelection` suite green (79); ESLint and type-check clean.

_Note:_ the component has visual-regression (screenshot) tests; this changes only scroll-on-focus behaviour, no static layout, so no snapshot changes are expected.

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/MultiSelection/MultiSelection.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/MultiSelection/MultiSelection.tsx
@@ -636,12 +636,6 @@ function MultiSelection(props: FieldMultiSelectionProps) {
             : navigable[(index - 1 + navigable.length) % navigable.length]
 
       next?.focus()
-
-      if (next !== searchInput) {
-        next
-          ?.closest('.dnb-forms-field-multi-selection__item')
-          ?.scrollIntoView({ behavior: 'smooth', block: 'nearest' })
-      }
     },
     [disabled, getCheckboxes, getSearchInput]
   )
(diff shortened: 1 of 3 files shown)


langz:
Closing this, false positive, was not able to reproduce this again.

PR #9104, opened 2026-08-25
chore(CI): select visual tests via resolved TypeScript dependency graph

langz:
## Why

A one-line `Upload` fix (#9097) triggered the entire visual-regression suite. Selective visual testing picks tests from the reverse dependency graph, but re-export barrels collapse into a single node — so changing one component fanned out to nearly every consumer of the barrel.

## What

Build the reverse-dependency graph from the TypeScript program instead: each import resolves to the module that actually defines the symbol, so re-export barrels are transparent and type-only imports are ignored. Selection now follows real usage — an `Upload` change runs the `Upload` tests, and an `Autocomplete` change runs the fields that render it.

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/contribute/getting-started/make-and-run-tests.mdx
+++ b/packages/dnb-design-system-portal/src/docs/contribute/getting-started/make-and-run-tests.mdx
@@ -131,6 +131,8 @@ Selection includes:
 
 Global impact still runs all screenshot tests when shared visual config/style paths are changed (for example `packages/dnb-eufemia/package.json`, `src/style/*` or it runs on the `main` branch).
 
+Reverse-dependency selection stops at library and extension barrel aggregates (for example `src/components/index.ts` or `src/extensions/forms/index.ts`). These re-export the whole surface, so walking through them would fan out a single component change to nearly every test. Per-feature barrels are still followed, so genuine shared-infra changes keep propagating.
+
 You can run the same logic locally:
 
 ```bash
--- a/packages/dnb-design-system-portal/src/docs/contribute/getting-started/make-and-run-tests.mdx
+++ b/packages/dnb-design-system-portal/src/docs/contribute/getting-started/make-and-run-tests.mdx
@@ -131,7 +131,7 @@ Selection includes:
 
 Global impact still runs all screenshot tests when shared visual config/style paths are changed (for example `packages/dnb-eufemia/package.json`, `src/style/*` or it runs on the `main` branch).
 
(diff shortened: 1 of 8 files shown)


langz:
Added a stacked smoke-test PR to demonstrate this fix live in CI: #9107.

It applies a single visual-neutral change to `Upload` (`UploadFileListLink.tsx` — the same file as #9097) on top of this branch. The **Run visual-regression tests** job there should report `Running 4/108 impacted screenshot test files` (only `Upload`, `Field.Upload`, `Value.Upload`, and the `Form/Appearance` demo that composes `Upload`) instead of the full 108-test suite the equivalent change triggered before this fix.

Draft/demo only — I'll close it once CI has shown the scoped run.

langz:
## Self-review update: converting to draft — found a correctness regression

Reviewing this against the codebase, the barrel-exclusion introduces **false negatives (under-selection)**, not just the intended reduction in over-selection.

Forms fields import base components through the components barrel, e.g. `SelectCountry` does `import { Autocomplete } from '../../../../components'`. Excluding `src/components/index.ts` from the reverse-dependency walk severs that genuine edge:

- Change to `Autocomplete`: **before** this PR → 48 impacted (≥0.4 → runs all, safe); **after** → **2** impacted, and `SelectCountry` is no longer tested.

Root cause: `depcruise` collapses a barrel into a single node carrying both genuine (`SelectCountry → Autocomplete`) and spurious (`Autocomplete.test → List`) edges. That same `components/index.ts` node is spurious-only for an `Upload` change but genuine for an `Autocomplete` change, so a static exclusion can't distinguish them — and it's required to fix the Upload-class explosion, so no clean subset exists.

The correct approach is **per-export / re-export resolution** (resolve `import { X } from '<barrel>'` to X's defining module) so the reverse-dependency graph is precise and no barrel heuristic is needed. Prototyping that now; will update or supersede this PR.

langz:
Reworked to the symbol-resolved approach discussed above — the barrel-exclusion heuristic is removed.

The reverse-dependency graph is now built from the TypeScript program: imports resolve to the module that defines the symbol (re-export barrels are transparent) and type-only imports are ignored. Validated on the real repo:

- `Upload` change → **4** tests (Upload, Field.Upload, Value.Upload, + the composing demo)
- `Autocomplete` change → **9** tests, now correctly including `SelectCountry` / `SelectCurrency` / `PhoneNumber` (the earlier heuristic dropped these)

New unit tests cover barrel resolution, type-only exclusion, and barrel transparency via a fixture TS program. Left as draft while CI runs (the graph build adds ~45s); happy to mark ready once it's green.

langz:
Validated end-to-end in CI via the two stacked smoke tests (both visual-regression jobs green):

- #9107 — isolated change to `Upload` → **Running 4/108** (scoped to the Upload tests)
- #9111 — shared `ProgressIndicator` (used via TS imports + demo composition) → **Running 18/108**, spanning the dependent components (Autocomplete, Dialog, Input, Pagination, Upload, several `Field.*`, …)

Marking ready for review.

tujoworker (approved):

PR #9109, opened 2026-08-25
chore(visual-regression): remove retired portal branch from base refs

langz:
Follow-up to #9096, which retired the legacy `portal` branch as a deployment source and removed its references from the CI workflows.

While reviewing #9096, one remaining reference to the `portal` branch was found outside the workflow files: the conditional visual-regression test-selection tool still listed it as a fallback base ref.

## Change

Remove `'origin/portal'` and `'portal'` from `DEFAULT_BRANCH_BASE_REFS` in `runScreenshotsConditionally/config.ts`.

## Why it's safe

These entries are candidate merge-base refs used to compute the changed-file set when no base ref is provided. In `git.ts` the refs are de-duplicated and each is tried with `git merge-base <ref> HEAD`; any ref that does not resolve is silently skipped, and nothing depends on the list's length or order. Removing the two retired-branch entries simply stops treating `portal` as a comparison base. No test asserts the contents of this list.

This completes the removal of `portal` branch references started in #9096.


tujoworker (approved):

tujoworker:
Nice you found this!

PR #9110, opened 2026-08-25
build: prepare RAIWork plugin releases

tujoworker:
Adds a manually dispatched release-preparation workflow for the Eufemia Web RAIWork plugin. It produces a source-attributed, checksummed artifact after validating the requested version and the live hosted MCP contract, giving maintainers a reproducible handoff for the private-first marketplace release flow.

Publishing remains explicit while RAIWork ownership is tied to a personal Entra identity and unattended workload identity is unavailable.

### DIFF
--- a/.github/workflows/raiwork-plugin-release.yml
+++ b/.github/workflows/raiwork-plugin-release.yml
@@ -1,0 +1,131 @@
+name: Prepare RAIWork plugin release
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: Version already set in plugin.config.json, without a leading v
+        required: true
+        type: string
+
+permissions:
+  contents: read
+
+concurrency:
+  group: raiwork-plugin-release
+  cancel-in-progress: false
+
+jobs:
+  prepare:
+    name: Validate and package v${{ inputs.version }}
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


langz (approved):

.github/workflows/raiwork-plugin-release.yml
```diff
@@ -0,0 +1,131 @@
...
+            )
+          }
+          NODE
+
+      - name: Run plugin checks
+        run: |
+          yarn workspace eufemia-raiwork-plugin lint
+          yarn workspace eufemia-raiwork-plugin test:types
+          yarn workspace eufemia-raiwork-plugin test
+          yarn workspace eufemia-raiwork-plugin release:check
```
langz, line 78: This seems to verify-remote call to https://server.eufemia.dnb.no/mcp/web. If that call fails, the step fails and the artifact never gets staged — even though the bundle's contents don't need the server to be up at build time.

Not worth looking into, but it just means that the release is dependent on the mcp server being available. All good

tujoworker:
Superseded by #9116, which generates and validates both the RAIWork and portable Agent Plugin distributions through one manually dispatched artifact workflow. The hosted MCP availability check discussed in review is retained intentionally as a release-only gate. Thanks for the review and approval.

PR #9116, opened 2026-08-26
chore(AI): publish portable agent plugin

tujoworker:
Adds an Agent Plugins 1.0 package with Claude compatibility alongside the existing RAIWork bundle, allowing Eufemia to be installed from the public `dnbexperience/eufemia-agent-plugins` marketplace and consumed as a Frontend Frida dependency.

Both distributions are generated from Eufemia’s canonical Agent Skills and share the existing `0.1.0` plugin version. A manual workflow produces validated release artifacts; RAIWork publication remains a reviewed UI promotion until its supported CLI exposes a documented upload contract.


langz (approved):
Great great 🎸 

One optional thought: `validateAgentPlugin` sums `totalBytes` and reports `fileCount`, but doesn't enforce the `MAX_PLUGIN_BYTES` / `MAX_PLUGIN_FILES` ceilings that `validateRaiworkBundle` does (bundle.ts:310–319). Can't trip today at ~21 KiB, but since this format ships to a public repo it might be worth exporting those two constants and reusing the same guard here for consistency. Not a blocker — up to you.

PR #9117, opened 2026-08-26
fix(HeightAnimation): hide open-on-find content in unsupported browsers

tujoworker:
## Motivation

The openOnFind feature relies on native hidden=until-found support. Without it, collapsed content can remain visible because the browser does not apply content-visibility: hidden.

Feature-detect the beforematch API and fall back to the existing keepInDOM hiding behavior when unavailable. This makes it safe for components that already retain content in the DOM to enable find-in-page behavior by default.


langz (approved):
Looks good to me :) 

Seems quite similar to https://github.com/dnbexperience/eufemia/pull/8976 which we decided to not include.

PR #9132, opened 2026-08-26
fix(MultiSelection): match any search word (OR) and drop unrelated what-input effect

langz:
Follow-up to #8053, addressing the two findings I flagged in review. Targets the feature branch so it can be folded into #8053.

## 1. Multi-word search: match ANY word (OR), like Autocomplete

MultiSelection required **all** search words to match (`matchedWords.length === searchWordsData.length`), while Autocomplete keeps items matching **any** word and reorders by relevance. So `alpha beta` filtered differently in the two components even though they now share the same `search` prop.

Changed MultiSelection to OR semantics (`matchedWords.length > 0`) + existing relevance reorder, mirroring Autocomplete. Multiple numeric terms still require all to match (unchanged — enforced by `passesNumericTermsCheck`).

Verified before the change (query `alpha beta`, one item matches only `alpha`, one only `beta`, one both):
- Autocomplete kept all three; MultiSelection kept only the both-match item.

After: both keep all three, best match reordered first. New regression test: `keeps items matching any search word (OR), like Autocomplete`.

## 2. whatInput: restore the previous value on close (no global leak)

The popover effect reset the global `whatInput.specificKeys` to `['Tab']` on close, but the app default (set once in `component-helper`) is `[]`, and nothing else touches it at runtime. So after the first open/close, every non-Tab key stopped registering as keyboard input **app-wide**, breaking focus-visible styling everywhere.

Verified: before, key `a` → `data-whatinput="keyboard"`; after one open+unmount, `a` → `"mouse"` and only `Tab` still counted.

Now the effect captures the previous value and restores it on cleanup. Added `whatInput.getSpecificKeys()` to support this, and made the effect a no-op when closed (no spurious global writes). New regression test: `restores global key-input detection after closing (no leak)`, plus a unit test for the getter. Also replaced the misleading "existing menu pattern" comment (no such pattern exists in the codebase).

## Tests
Local (Node 24.16, ICU-matched): whatInput (9), searchUtils (27), MultiSelection (86), Autocomplete (165) all pass. Lint + typecheck clean.

## Not included (minor, from the same review — happy to add if you want)
- Nested `children` aren't reordered by relevance (top-level `result.sort()` only). Confirmed reproducible but low impact.
- `SearchOptions` isn't re-exported from the public entry, so consumers can't `import { SearchOptions }` (object literals still work via structural typing).

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Field/MultiSelection/MultiSelection.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Field/MultiSelection/MultiSelection.tsx
@@ -26,8 +26,6 @@ import {
   calculateTotalScore,
   passesNumericTermsCheck,
 } from '../../../../shared/search'
-import whatInput from '../../../../shared/helpers/whatInput'
-import useIsomorphicLayoutEffect from '../../../../shared/helpers/useIsomorphicLayoutEffect'
 import withComponentMarkers from '../../../../shared/helpers/withComponentMarkers'
 import { createSharedState } from '../../../../shared/helpers/useSharedState'
 import { MultiSelectionTrigger } from './MultiSelectionTrigger'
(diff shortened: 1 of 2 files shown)


tujoworker (approved):
🤟

PR #9133, opened 2026-08-26
docs(Portal): improve heading anchor handling and feedback

tujoworker:
Heading anchor navigation currently combines a strong text-colour animation with a page jump, which makes linked sections feel disruptive and makes the permalink difficult to use without a pointer.

Use a subtle background-only tint, keep the viewport stable when setting the hash, and expose the trailing permalink to keyboard and screen-reader users.


langz (approved):
Good to me.
Vertical alignment of the # has changed a bit though, not sure what I like the most:

Before:


Image

Now:

Image

Personally, I think I like "Before" most, but does not really matter too much I would say. Functionality is the important part 😄 👍

joakbjerk (approved):

PR #9134, opened 2026-08-26
chore(analytics): add usage dashboard scaffold with snapshot workflow

joakbjerk:
Scaffolds an internal usage dashboard for the analytics service in `tools/analytics`.

## Why

We want a simple way to view the collected, anonymous usage statistics. This adds the dashboard UI and a data-refresh path now, with authentication and hosting deliberately left out of scope so the UI is reviewable early. Access control (Entra SSO) is tracked separately as a follow-up.

## What's here

- A static, dependency-free dashboard (key figures, per-day and top-pages bars, environment filter) that renders a bundled snapshot. All values use `textContent`, so record labels can't inject markup.
- A scheduled snapshot workflow that queries the analytics API using a token secret and bakes `data/records.json`. The token never reaches the browser; the only caller is the workflow runner.
- Sample data so the page renders before the first snapshot, plus a short README with preview and hosting notes.

## Not included (follow-up)

- Hosting and access control (restrict to authorised DnB users via Entra SSO).
- The workflow's publish-to-host step is stubbed until the hosting target is chosen.


tools/analytics/dashboard/styles.css
langz, line 1: As a demo/POC to be able to see/test data integration, etc, I think it's fine, but I would guess think that we would use Eufemia's own code/styling in the future 😄

langz (approved):
Image

tools/analytics/dashboard/app.js
```diff
@@ -0,0 +1,160 @@
...
+// DATA_URL; the page shows an empty state until a data source is wired up.
+
+const DATA_URL = './data/records.json'
+
+/** Normalise the stored shape to a common view model. */
+function normalise(record) {
+  const label = record.name ?? record.path ?? record.type ?? record.id ?? '—'
+  const when = record.createdAt ?? record.timestamp ?? ''
+  const day = typeof when === 'string' ? when.slice(0, 10) : ''
+  const env = record.env ?? ''
```
langz, line 11: I'e not validated if this data model matches what the real service provides.
I guess it does 🤞

tools/analytics/dashboard/styles.css
```diff
@@ -0,0 +1,152 @@
+:root {
+  color-scheme: light dark;
+  --bg: #f7f7f8;
+  --panel: #ffffff;
+  --ink: #14171a;
+  --muted: #5b6570;
+  --accent: #007272;
```
langz: Doesn't seem like this one is used

tools/analytics/dashboard/app.js
joakbjerk: It should match what we have now at least 😄

tools/analytics/dashboard/styles.css
joakbjerk: Yeah we will 😄

PR #9136, opened 2026-08-27
fix(Img): preserve internal error handling

tujoworker:
A consumer-provided `onError` currently replaces Img’s internal handler, preventing the error styling from being applied. Compose both handlers so the component retains its error state while still notifying the consumer.

Depends on #9135.

### DIFF
--- a/packages/dnb-eufemia/src/elements/img/Img.tsx
+++ b/packages/dnb-eufemia/src/elements/img/Img.tsx
@@ -4,7 +4,7 @@
  */
 
 import { useState } from 'react'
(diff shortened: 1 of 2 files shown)


langz (approved):
Is it worth calling out in the release notes that this is an observable behavior change? 

Previously a consumer `onError` replaced the internal handler, so `dnb-img--error` never applied. Now both run — so anyone (unknowingly) suppressing the error styling via their own `onError` will see it reappear. I see it's intendedd, but should we add a changelog line?

PR #9143, opened 2026-08-27
docs(Forms): document openOnFind for field help

tujoworker:
## Motivation

Inline field help supports openOnFind through the help configuration, but the shared Forms property documentation does not mention it.

Document the option alongside the existing inline help settings so consumers can discover the behavior.


langz (approved):
Nice

PR #9148, opened 2026-08-27
fix(Typography): align iOS Dynamic Type sizing

tujoworker:
Stacked on #9147.

Keep support for the user's iOS system text-size setting while aligning DNB's default basis text with the 18px desktop size. The correction stays on the typography scope so root-relative layout values continue to follow iOS Dynamic Type.

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/typography/font-size.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/typography/font-size.mdx
@@ -31,12 +31,13 @@ On iPhone and iPad, Eufemia follows the user's system text-size setting. This
 means that text may not have exactly the same pixel size as it does on macOS or
 other desktop platforms.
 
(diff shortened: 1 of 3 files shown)


langz (approved):
Looks good, 

Have we verified this on a real device? My understanding is that -webkit-text-size-adjust only drives WebKit's text-autosizing algorithm, which stays dormant on pages with a normal width=device-width, initial-scale=1 viewport — and I couldn't find anything indicating -apple-system-body activates it or that a sub-100% value scales text down below its specified size. So I'm worried the 94.117647% is a no-op and basis text is still 19.125px, while the docs now claim 18px.

tujoworker:
Closing after real-device testing: `text-size-adjust` scales text and line height, but not the `rem`-based control geometry. This distorts inputs and round buttons, so it is not a safe design-system default.

PR #9159, opened 2026-08-27
chore(analytics): add Entra sign-in to the dashboard

joakbjerk:
Draft, stacked on the dashboard scaffold PR (base `chore/analytics-dashboard`).

The dashboard shows internal usage statistics that should be visible only to authorised DnB users, including non-developers. This adds Entra (Azure AD) sign-in (OpenID Connect, no secret in the browser) so only assigned users can open it.

Scope caveat: on a static site this gates the UI, not a publicly served data file. Restricting the data itself needs host/edge auth or a token-protected API — a deliberate follow-up once hosting is decided.


tools/analytics/dashboard/app.js
```diff
@@ -1,11 +1,14 @@
...
 
+import { ensureSignedIn, signOut } from './auth.js'
+
 const DATA_URL = './data/records.json'
 
 /** Normalise the stored shape to a common view model. */
 function normalise(record) {
-  const label = record.name ?? record.path ?? record.type ?? record.id ?? '—'
+  const label =
+    record.name ?? record.path ?? record.type ?? record.id ?? '—'
```
langz, line 11: Not sure if this tool/app is configured with prettier?
Or perhaps it is, and we just forgot to run it in the bottom stacked branch https://github.com/dnbexperience/eufemia/pull/9134

Doesn't really matter, as all of this stack will go into a feature branch where I guess all commits will be squashed again when merged to main

langz (approved):

tools/analytics/dashboard/app.js
joakbjerk: Made sure the linter runs prettier in that file as well now 🙏

PR #9160, opened 2026-08-27
feat(Form.SubmitIndicator): add CSS border glow

tujoworker:
Introduces a CSS-only border glow for pending states in fields and submit buttons. This draft is an initial visual direction that we may iterate on based on feedback.

Preview: https://feat-submit-indicator-animat.eufemia-e25.pages.dev/uilib/extensions/forms/Form/SubmitIndicator/demos/#border-glow-on-all-fields


langz:
image

PR #9163, opened 2026-08-27
chore(build): parallelize cjs/es/esm library compilation

langz:
## Why

The library `postbuild` runs the cjs, es and esm Babel passes sequentially
(~80–100s of the e2e and build-check jobs). Each only reads `src` and writes to
its own `build/<target>` directory, so they don't need to run one at a time.

## What

Run the three Babel passes concurrently, waiting on each PID so a failure still
fails the build (`set -e` does not catch background jobs). Verified locally: the
output is byte-for-byte identical to the sequential build (27044 files, same
sizes), and a deliberately failing pass still exits non-zero.

### DIFF
--- a/packages/dnb-eufemia/scripts/postbuild/postbuild.sh
+++ b/packages/dnb-eufemia/scripts/postbuild/postbuild.sh
@@ -6,13 +6,23 @@ echo 'Postbuild started ...'
 
 yarn build:types:definitions
 yarn prettier:other:write
-yarn build:cjs
+
+# The cjs/es/esm Babel passes only read ./src and each writes to its own
+# ./build/<target> dir, so run them concurrently to cut build time. set -e does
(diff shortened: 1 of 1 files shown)


langz:
## Measured time savings

The three Babel passes dominate `postbuild`. Timed individually (locally):
`cjs` 26s + `es` 31s + `esm` 39s = **96s** run sequentially; in parallel the step
is bounded by the longest (~39s).

On CI the `Postbuild Library` step drops:
- build-check: **78s → 62s** (~20%)
- e2e: **103s → 67s** (~35%)

So ~15–35s off `postbuild` in each of build-check, e2e (and release / visual on
main). Output is byte-for-byte identical to the sequential build (27044 files,
same sizes), and a failing pass still fails the build (`set -e` + wait-on-PID →
non-zero exit).

joakbjerk (approved):

PR #9167, opened 2026-08-27
chore(analytics): add Entra-authenticated dashboard data API

joakbjerk:
The dashboard needs a data source that only authorised DnB users can read. This adds a browser-facing API that returns the analytics data, gated by an Entra (Azure AD) JWT authorizer so only assigned users can read it, kept separate from the edge-locked ingest API.


tujoworker:
Fixed both review findings in 5bfc5ec: the dashboard now queries page-view fields, and its snapshot stays under the existing records/* permission. Added regression coverage for both.

tujoworker (approved):
Looks good to me!

PR #9168, opened 2026-08-28
chore(CI): skip e2e library build on feature branches

langz:
Feature branches now run e2e against the Eufemia **source** instead of building the library first — the same thing visual-regression already does with `RUN_POST_BUILD`. `main` and `v*` still build and test the compiled dist.

**Why:** the library build (prebuild + postbuild) runs on every e2e push today, but the portal resolves the source directly when no `build/` exists. The compiled dist is already built and checked on every PR by `build-check` (build + `test:postbuild` + bundlesize), so gating it here only skips a redundant build. Build-output regressions are still caught by `build-check`, and e2e covers the dist end-to-end on `main`/`v*`.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -26,8 +26,11 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
-# No job-wide `env:` on purpose: only the public Algolia search config is
-# needed, and it is scoped to the portal build step (least privilege).
+env:
+  # Non-secret build flag only; the public Algolia search config stays scoped
+  # to the portal build step (least privilege). Feature branches skip the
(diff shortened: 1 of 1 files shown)


langz:
**Measured CI savings** — e2e job on this branch (feature → tests source) vs recent `main` runs (build the library):

| Phase | `main` (build) | this branch (source) |
|---|---|---|
| Prebuild Library | ~32s | 0s (skipped) |
| Postbuild Library | ~88s | 0s (skipped) |
| Build portal | ~39s | 28s |
| **Build phase** | **~159s** | **~28s** |

Attributable saving ≈ **2.2 min per feature-PR e2e run** (build phase ~159s → ~28s; the source portal build is even a touch faster than the dist one). Observed job total ~350s → 187s — the extra delta is Playwright retry variance, not from this change. The run passed on source, and `main`/`v*` still build + test the compiled dist.

tujoworker (approved):

PR #9169, opened 2026-08-28
fix(Typography): align iOS Dynamic Type tokens

tujoworker:
Stacked on #9147.

Prototype an iOS-only `16 / 17` correction for Eufemia's font-size and line-height tokens while leaving component geometry, spacing, radii, and icon sizes unchanged. This preserves Dynamic Type and keeps round controls proportional.

Custom raw `rem` text and a few bespoke component font values remain outside the token correction and need review before this is ready.


tujoworker:
Closing after local testing: adjusting the typography tokens also changes font sizes inside inputs. That introduces too many negative side effects for text alignment and placement, so this is not a safe design-system default.

PR #9170, opened 2026-08-28
chore(analytics): fetch dashboard data from the protected API

joakbjerk:
Wires the dashboard to its data source: after Entra sign-in the page requests an access token for the API scope and fetches the protected `/data` endpoint with it, showing an empty state when there is no data or no configuration. A rejected token triggers a fresh sign-in.

Depends on the dashboard data API (the `/data` endpoint) being deployed.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -79,11 +79,13 @@ jobs:
 
       - name: Audit dependencies
         # Audit the shipped (production) dependency tree of everything we ship:
-        # the @dnb/eufemia npm library and the @dnb/eufemia-mcp Lambda. Both must
-        # pass; dev/build tooling advisories are tracked via Dependabot instead.
+        # the @dnb/eufemia npm library, the @dnb/eufemia-mcp Lambda and the
+        # eufemia-analytics Lambda. All must pass; dev/build tooling advisories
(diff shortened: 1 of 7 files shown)


langz:
I opened #9174 against this branch with three small follow-up improvements from reviewing this PR:

1. **Gate the analytics workspace in `verify.yml`** — its lint/type-check/tests only ran via `analytics-lambda.yml` (push to `main`), so the dashboard tests added here never ran as a PR check. #9174 adds `eufemia-analytics` to the `lint`, `typecheck` and `test` jobs, matching the other tool workspaces.
2. **Format-check the dashboard test file** — the tool's `lint` only Prettier-checked `dashboard/**`, leaving `__tests__/dashboard.test.js` unchecked; widened the glob to cover `src` and `__tests__`.
3. **Harden `readSession()`** — a session persisted by an older build has no `accessToken` and would call the API with `Bearer undefined`; now it requires `accessToken` and forces a clean re-auth (with tests).

It targets this branch so it merges in before this PR does. All analytics checks pass locally (80 tests). No change to the normal signed-in path.

langz (approved):

tujoworker (approved):

langz (approved):

PR #9171, opened 2026-08-28
feat: add opt-in iOS Dynamic Type scaling

tujoworker:
iOS Dynamic Type changes the computed Apple body size, while Eufemia layouts rely on the root `rem` size. This opt-in bridge measures that preference before first paint and applies it coherently to typography and `rem`-based geometry.

It preserves the existing CSS behavior when JavaScript or measurement is unavailable.

### How it works

At the default iOS text setting, Apple’s `Body` size measures 17px. The blocking head script converts this to a 16px root while preserving the user’s Dynamic Type scale, then adds `data-eufemia-text-scale="apple"`. Because this happens before stylesheets and first paint, it does not cause a layout shift. If the script cannot run, the existing 17px `-apple-system-body` CSS remains as the fallback.

**17px basis:**
Simulator Screenshot - iPhone 17 Pro - 2026-08-31 at 10 47 51

**16px basis (new):**
Simulator Screenshot - iPhone 17 Pro - 2026-08-31 at 10 47 58

### DIFF
--- a/packages/dnb-eufemia/src/shared/TextScaleScript.tsx
+++ b/packages/dnb-eufemia/src/shared/TextScaleScript.tsx
@@ -1,0 +1,31 @@
+/**
+ * Opt-in text-scale helpers.
+ *
+ * `TextScaleHeadScript` runs synchronously in the document head before paint.
+ * `TextScaleClient` is a client-rendering fallback that runs in a layout effect.
+ */
+
+import useIsomorphicLayoutEffect from './helpers/useIsomorphicLayoutEffect'
+import { applyTextScale, getTextScaleScript } from './TextScaleScriptUtils'
+
+export { applyTextScale, getTextScaleScript } from './TextScaleScriptUtils'
+
+export function TextScaleHeadScript({ nonce }: { nonce?: string } = {}) {
+  return (
+    <script
+      nonce={nonce}
+      dangerouslySetInnerHTML={{
+        __html: getTextScaleScript(),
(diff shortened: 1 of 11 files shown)


tujoworker:
I tested this on a real iPhone. The solution is opt-in, so applications only enable it by adding the text-scale script. Without it, Eufemia falls back to the previous `-apple-system-body` behavior.

joakbjerk (approved):

packages/dnb-eufemia/src/shared/TextScaleScript.tsx
```diff
@@ -0,0 +1,31 @@
...
+      }}
+    />
+  )
+}
+
+/**
+ * Client-rendering fallback. This runs before React paints, but cannot prevent
+ * a flash when hydrating server-rendered HTML. Prefer `TextScaleHeadScript`.
+ */
+export function TextScaleClient() {
```
langz, line 28: Should this rather be a hook? `useTextScale` or such?
Or if we need the component, it could perhaps just be a one-line <TextScaleClient /> wrapper that just calls the hook

langz (approved):

packages/dnb-eufemia/src/shared/TextScaleScript.tsx
tujoworker: That's not possible 🤷‍♂️ because we want to block rendering until we now more.

PR #9172, opened 2026-08-28
chore(analytics): host the dashboard on public S3 + CloudFront

joakbjerk:
Hosts the analytics dashboard UI. Its data sits behind the Entra-authenticated `/data` API, so the shell itself carries no data or secrets and can be served as a plain public S3 + CloudFront site: the bucket stays private (read via an Origin Access Control) and access control is entirely the Entra sign-in plus the `/data` gate — no Lambda@Edge, edge auth or extra IAM role.

Go-live after the first deploy: add the CloudFront URL as an app-registration redirect URI and to `DASHBOARD_ORIGINS`, then drop `localhost`.

**Will merge 1st of September when in office just to be able to react quickly in case something needs to be fixed immediately after deplo**y 🙏

### DIFF
--- a/.github/workflows/analytics-lambda.yml
+++ b/.github/workflows/analytics-lambda.yml
@@ -158,6 +158,11 @@ jobs:
           cp "$GITHUB_WORKSPACE/tools/analytics/dist/lambda.zip" dist/
           cp -R "$GITHUB_WORKSPACE/tools/analytics/infra" .
 
+          # Include the dashboard static files so the deploy job can publish
+          # them to the CloudFront-fronted bucket. config.json is generated at
+          # deploy time from the GHE variables, so it is not copied here.
+          cp -R "$GITHUB_WORKSPACE/tools/analytics/dashboard" .
+
           # Include the deploy workflow so it lives on (and triggers from) the deploy branch
           mkdir -p .github/workflows
           cp "$GITHUB_WORKSPACE/tools/analytics/ghe-deploy-workflow.yml" .github/workflows/deploy.yml
--- a/tools/analytics/README.md
+++ b/tools/analytics/README.md
@@ -98,7 +98,11 @@ Because the OIDC deploy role's permissions boundary forbids `iam:CreateRole` (AD
 
 The dashboard is hosted separately as a static site:
 
-- **Dashboard bucket** (private, SSE-S3, public access blocked) holding the static UI, read only by CloudFront via an Origin Access Control.
-- **CloudFront distribution** serving the dashboard on its default `*.cloudfront.net` domain. The shell holds no data or secrets — access is gated entirely by the Entra sign-in and the token-protected `/data` API — so no Lambda@Edge, edge auth or extra IAM role is needed. The deploy job generates `dashboard/config.json` from the `ENTRA_CLIENT_ID`/`ENTRA_TENANT_ID` variables (redirect URI set to the CloudFront URL), syncs the files to the bucket and invalidates the cache. Add the CloudFront URL as a redirect URI on the app registration after the first deploy.
+- **Dashboard bucket** (private, versioned, SSE-S3, public access blocked) holding the static UI, read only by CloudFront via an Origin Access Control.
+- **CloudFront distribution** serving the dashboard on its default `*.cloudfront.net` domain. The shell holds no data or secrets — access is gated entirely by the Entra sign-in and the token-protected `/data` API — so no Lambda@Edge, edge auth or extra IAM role is needed. The deploy job generates `dashboard/config.json` (non-secret public identifiers: `clientId`/`tenantId` from `ENTRA_CLIENT_ID`/`ENTRA_TENANT_ID`, `redirectUri` set to the CloudFront URL, `apiBaseUrl` from the dashboard API endpoint, and `apiScope` = `api://<clientId>/Dashboard.Read`), syncs the files to the bucket and invalidates the cache.
+
+The CloudFront origin is added to the dashboard API's CORS automatically (the distribution domain is concatenated onto `DASHBOARD_ORIGINS`), so no second deploy is needed for cross-origin `/data` calls; `DASHBOARD_ORIGINS` only needs any extra origins such as a local-dev URL. After the first deploy, add the CloudFront URL as a redirect URI on the app registration.
+
+Prerequisites for sign-in and data to work end to end: the app registration must expose a `Dashboard.Read` scope under App ID URI `api://<clientId>` and issue v2 access tokens (`requestedAccessTokenVersion = 2`). The deploy role's CloudFront and S3 permissions are provisioned in the OIDC federation repo, alongside the Lambda execution role.
 
 Terraform state reuses the shared `eufemia-mcp-terraform-state` bucket under the `analytics/` key.
--- a/tools/analytics/README.md
+++ b/tools/analytics/README.md
@@ -68,9 +68,10 @@ Deployment mirrors the MCP Lambda pattern: public GitHub builds and tests, then 
 ```
 public GitHub (.github/workflows/analytics-lambda.yml)
   → test + build lambda.zip
-  → force-push dist/ + infra/ + deploy workflow to GHE repo `deploy` branch
+  → force-push dist/ + infra/ + dashboard/ + deploy workflow to GHE repo `deploy` branch
       → GHE (ghe-deploy-workflow.yml as .github/workflows/deploy.yml)
           → OIDC assume role → terraform apply
+          → generate dashboard/config.json → aws s3 sync → CloudFront invalidation
 ```
 
 - **Triggers** (`analytics-lambda.yml`): a push to `main` touching `tools/analytics/**` (or the workflow), or a manual `workflow_dispatch` from any branch. Analytics deploys on its own code changes, not on every Eufemia release (it has no Eufemia docs dependency, unlike the MCP server).
@@ -94,5 +95,10 @@ Because the OIDC deploy role's permissions boundary forbids `iam:CreateRole` (AD
 - **Athena workgroup** for the retrieve queries.
 - **Lambda function** (`nodejs22.x`) — its execution role is pre-created out-of-band, because the OIDC deploy role's permissions boundary forbids `iam:CreateRole` (ADR 0004); it is only referenced here.
 - **API Gateway HTTP API** with the two `/records` routes and throttling.
+
+The dashboard is hosted separately as a static site:
+
+- **Dashboard bucket** (private, SSE-S3, public access blocked) holding the static UI, read only by CloudFront via an Origin Access Control.
+- **CloudFront distribution** serving the dashboard on its default `*.cloudfront.net` domain. The shell holds no data or secrets — access is gated entirely by the Entra sign-in and the token-protected `/data` API — so no Lambda@Edge, edge auth or extra IAM role is needed. The deploy job generates `dashboard/config.json` from the `ENTRA_CLIENT_ID`/`ENTRA_TENANT_ID` variables (redirect URI set to the CloudFront URL), syncs the files to the bucket and invalidates the cache. Add the CloudFront URL as a redirect URI on the app registration after the first deploy.
 
 Terraform state reuses the shared `eufemia-mcp-terraform-state` bucket under the `analytics/` key.
--- a/tools/analytics/ghe-deploy-workflow.yml
+++ b/tools/analytics/ghe-deploy-workflow.yml
@@ -72,25 +72,27 @@ jobs:
           BUCKET=$(terraform output -raw dashboard_bucket)
           DIST=$(terraform output -raw dashboard_distribution_id)
           URL=$(terraform output -raw dashboard_url)
+          API_BASE_URL=$(terraform output -raw dashboard_api_endpoint)
 
           # Generate the per-host runtime config from the deploy variables. It
           # holds only non-secret public identifiers, so it is safe to serve.
-          # The redirect URI must be the CloudFront URL registered on the app.
+          # The redirect URI must be the CloudFront URL registered on the app;
+          # apiScope is the app registration's exposed "Dashboard.Read" scope.
           jq -n \
             --arg clientId "$ENTRA_CLIENT_ID" \
             --arg tenantId "$ENTRA_TENANT_ID" \
             --arg redirectUri "$URL" \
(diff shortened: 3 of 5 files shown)


langz:
> 🤖 **AI-generated review.** This was produced by RAIWork (an AI agent) at @langz's request. Please treat it as a set of **leads to check, not a verdict** — some points may be wrong, out of date, or already handled elsewhere. Have a look through and judge for yourselves which (if any) are real issues worth fixing. Nothing here is intended to block.

**Context / attribution:** since this branch was last rebased, the `/data` API, the Entra JWT authorizer, the CORS configuration and the dashboard's API client all landed in `main` separately. So this PR is now essentially *the hosting* — bucket, OAC, CloudFront distribution, publish step — plus the values it writes into `config.json`. A few findings below sit in `main` rather than in this diff; those are marked, and they're included only because this PR is what first puts them on the live path.

Overall the hosting pattern looks right: private bucket, full public-access block, CloudFront reading via an Origin Access Control, bucket policy scoped by both the service principal *and* the distribution ARN. The earlier concern about the shell reading data from its own public bucket is fully resolved now that `/data` exists.

---

## Most likely to bite on the first deploy

**1. Access-token version — could 401 every data call**

The authorizer requires `iss = https://login.microsoftonline.com/<tenant>/v2.0` and `aud = <client-id>`. Per [Microsoft's docs](https://learn.microsoft.com/en-us/entra/identity-platform/access-tokens), a `requestedAccessTokenVersion` of `null` or `1` issues **v1** tokens whose issuer is `https://sts.windows.net/<tenant>/` — which this authorizer rejects. Only `2` produces the `/v2.0` issuer together with the GUID audience the config expects, and portal-created single-tenant registrations default to `null`.

If that's the case here, the symptom is misleading: sign-in succeeds, then every data call fails and the UI shows its generic "try again later". Worth confirming on the app registration.

**2. App-registration prerequisites for the scope this PR writes**

The publish step hard-codes `api://<ENTRA_CLIENT_ID>/Dashboard.Read`. That needs the Application ID URI to be exactly `api://<client-id>` *and* an exposed scope named `Dashboard.Read`. If the URI was customised, Entra rejects the authorize request outright. Not currently mentioned in the bootstrap section.

**3. CORS can't include the dashboard until a second deploy**

`allow_origins` comes from `DASHBOARD_ORIGINS`, which can't contain the CloudFront domain before the distribution exists — so deploy #1 signs in and then fails every data call on CORS until someone edits the variable and re-runs. Terraform already knows the value, so this could be closed permanently:

```hcl
allow_origins = concat(
  var.dashboard_origins,
  ["https://${aws_cloudfront_distribution.dashboard.domain_name}"],
)
```

No dependency cycle — the distribution doesn't reference the API. *(The `cors_configuration` itself is in `main`.)*

Related: `dashboard_origins` is a `list(string)` supplied via `TF_VAR_dashboard_origins`, so the GHE variable's value has to be valid HCL list syntax — `["https://…"]`, not a comma-separated string. Easy to get wrong, not written down anywhere.

**4. Deploy-role permissions**

This looks like the first CloudFront resource anywhere in the repo, so the OIDC role likely needs CloudFront create/get/update/tag plus `CreateInvalidation`, and S3 `CreateBucket` / `PutBucketPolicy` / encryption / public-access-block, plus put/delete/list for the sync. The README documents the Lambda execution-role bootstrap in careful detail but says nothing about this — if the role's policy is narrow, the first apply fails on permissions.

**5. `Dashboard.Read` isn't actually enforced** *(route is in `main`)*

`GET /data` sets `authorization_type = "JWT"` with no `authorization_scopes`. Per Microsoft's [ID token claims reference](https://learn.microsoft.com/en-us/entra/identity-platform/id-token-claims-reference), an **ID token** for this same registration carries `aud` = the client-ID GUID and `iss` = `.../v2.0` — indistinguishable from the v2 access token to a validator checking only those two claims, and ID tokens carry no `scp` claim. So the scope this PR starts requesting is currently decorative.

Impact is limited (you still have to be an assigned user), but `authorization_scopes = ["Dashboard.Read"]` on the route closes it in one line.

---

## Worth changing in this diff

**6. The 403/404 → `index.html` rewrite could probably just go**

The comment says it's so "a refresh after the sign-in redirect lands on the app instead of an S3 error" — but the redirect URI is the distribution root and the app has no client-side router, so every URL it produces is `/`. Meanwhile the rewrite turns *every* missing object into HTTP 200 with the HTML shell, which has two consequences:

- A missing or unsynced `config.json` returns 200-with-HTML → `response.ok` is true → `response.json()` throws → `loadConfig()`'s catch returns `{}` → `ensureSignedIn()` returns `null` → the page renders **signed-out** rather than erroring. Low impact now that it's an empty shell, but it fails open rather than closed.
- A failed `app.js` or `styles.css` publish looks like a blank page instead of a 404.

Also worth noting: with OAC and no `s3:ListBucket` in the bucket policy, S3 returns 403 for missing keys — so the 404 rule is close to dead code either way.

**7. `Cache-Control` on the synced objects**

The invalidation step's comment says the new assets "are served immediately", but `index.html` / `app.js` / `auth.js` / `styles.css` aren't fingerprinted and S3 sends no cache headers, so browsers cache them heuristically and a CloudFront invalidation never reaches returning users. `config.json` is fine (fetched with `cache: 'no-store'`). Two extra flags on the `s3 sync` would make the comment true.

---

## Probably worth an explicit decision

**8. Viewer TLS is pinned to 1.0 by the default certificate**

Documented AWS behaviour: with `CloudFrontDefaultCertificate` set, ["CloudFront automatically sets the security policy to `TLSv1` regardless of the value that you set here"](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cloudfront-distribution-viewercertificate.html) — you can't raise it without an alternate domain name plus an ACM certificate. For a page handling Entra tokens that's probably worth recording as a conscious accept rather than discovering in a later scan. If the answer is "not acceptable", it becomes a custom domain — which would also dissolve #3 and the redirect-URI fragility below. (Note it'd need a second certificate in `us-east-1`; the existing one is regional in `eu-north-1` for API Gateway.)

---

<details>
<summary><b>Smaller things (click to expand)</b></summary>

- **Redirect URI + CORS origin are both derived from the generated CloudFront domain**, so replacing the distribution silently breaks sign-in *and* data until two things are updated by hand. Entra matches redirect URIs exactly, and `terraform output` emits no trailing slash — so register it without one.
- **No CSP.** The managed SecurityHeadersPolicy covers HSTS, `X-Frame-Options`, `Referrer-Policy` and `X-Content-Type-Options` but deliberately contains no `Content-Security-Policy`. Since the app has zero third-party dependencies and now holds an access token in `sessionStorage`, a tight custom policy (`default-src 'self'`, `connect-src 'self' login.microsoftonline.com <api-host>`) is cheap and is the main thing between an XSS and that token.
- **No `logging_config`** on the distribution, and no access logging or versioning on the bucket — the sibling `data` bucket in the same file *is* versioned, and versioning plus a `--delete` sync gives you a free rollback for a bad publish.
- **The publish sync is a deny-list** (`--exclude` per file); an allow-list (`--exclude '*'` plus explicit `--include`s) fails closed as the directory grows. Much lower stakes now that data comes from the API.
- `is_ipv6_enabled` defaults to `false`; enabling it is free.
- The section header reads "public S3 + CloudFront" two lines above "The bucket stays private" — slightly confusing for a future reader or auditor. "Public site, private bucket" might read better.
- `Post deploy summary` has no `if: always()`, so a publish failure also loses the endpoint/bucket output.
- No `offline_access`/refresh token, so a long-lived tab hard-navigates to Entra when the access token expires (~1h). Usually seamless via SSO — just noting it.
- `clearAuthRetry()` only runs on a successful response, so a 5xx leaves the retry marker set and a later genuine 401 skips its one retry. Trivial.
- **Top-level `README.md`** still describes `config.json` as generated from just `ENTRA_CLIENT_ID`/`ENTRA_TENANT_ID` — no `apiBaseUrl`/`apiScope`, no `Dashboard.Read` exposure, no token-version note, no `DASHBOARD_ORIGINS` step, no deploy-role permissions. `dashboard/README.md` *was* updated, so it's just the deploy-facing doc that's behind. Since that's the page whoever runs the first deploy will follow, it's probably the highest-value thing to fix here.

</details>

<details>
<summary><b>Things that looked suspicious but check out fine — listed so nobody re-chases them</b></summary>

- **`--exclude 'config.example.json'` / `'README.md'` work as intended.** Per the [AWS CLI filter docs](https://docs.aws.amazon.com/cli/latest/reference/s3/), each pattern has the source directory prepended, so they match those top-level files. (They wouldn't match same-named files in subdirectories, but there are none.)
- **`allow_headers = ["authorization"]` is sufficient**, even though `fetch(..., { cache: 'no-store' })` makes the browser append `Cache-Control` and `Pragma` request headers. The preflight only advertises author-set headers, and the actual request's headers aren't re-validated against the allow-list.
- **The token request omitting `scope` is not a bug.** Microsoft [documents it as optional](https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-auth-code-flow) on the authorization-code exchange. Passing `scopes()` explicitly would only make the returned token's target resource unambiguous.
- **The `$default` stage's trailing slash on `apiBaseUrl` is handled** by the `replace(/\/$/, '')` in `loadDashboardData`.
- **The empty-Entra-variable case already fails closed**: `variables.tf` validates `entra_client_id`/`entra_tenant_id` as non-empty and the Deploy step runs before Publish, so an unset variable aborts the apply instead of publishing a sign-in-less dashboard. Good.

</details>

---

## What looks good

OAC over a private bucket with a full public-access block; the bucket policy scoped by both the service principal and `AWS:SourceArn`; `depends_on` on the public-access-block avoiding the classic ordering race; the managed SecurityHeadersPolicy (with the nice note that a managed ID needs no extra deploy-role IAM); a separate API so the browser/JWT trust model stays isolated from the edge-locked ingest path; `allow_credentials` correctly left off for a bearer-token API; pinned action SHAs and `persist-credentials: false`; `noindex, nofollow`; the non-empty Terraform variable validations; and genuinely good unit coverage on the new API client (trailing-slash stripping, retry-once, malformed body, missing access token).

---

**One process note:** findings 1, 2, 3 and 5 only manifest at runtime and nothing in CI can catch them. A real end-to-end check after the first deploy — sign in, confirm the `/data` call returns 200 — would be worth more than any amount of further diff review.

langz (approved):
Approving. The hosting pattern looks sound — private bucket + OAC + full public-access block, policy scoped by both principal and `SourceArn`.

I've left a separate [AI-generated review comment](https://github.com/dnbexperience/eufemia/pull/9172#issuecomment-5478909646) with a list of things worth checking rather than blocking on — mainly a few runtime prerequisites (Entra access-token version, the `Dashboard.Read` scope exposure, the CORS origin chicken-and-egg on first deploy, and the deploy role's CloudFront permissions) that no CI check can catch. Worth a skim to see if any are real before the first deploy.

joakbjerk:
Thanks for the review — addressed the actionable items in 624a56f:

- **CORS chicken-and-egg (finding 3):** the `/data` API now `concat`s the live CloudFront origin into `allow_origins`, so cross-origin calls work on the first deploy — no manual `DASHBOARD_ORIGINS` edit + second deploy. No dependency cycle (the API doesn't reference the distribution).
- **403/404 → index.html rewrite (finding 6):** removed. No client-side router, the root is served by `default_root_object`, and the rewrite made a missing `config.json` fail open and masked failed asset publishes.
- **Cache-Control (finding 7):** `--cache-control 'no-cache'` on the sync so returning browsers revalidate (the invalidation alone doesn't reach already-cached clients).
- Plus: `is_ipv6_enabled`, dashboard bucket versioning (free rollback with `--delete`), `if: always()` on the summary step, reworded the header comment, and refreshed the deploy-facing README (config.json keys + v2-token / `Dashboard.Read` prerequisites).

Deferred as separate follow-ups (not this PR's diff):
- **`authorization_scopes = ["Dashboard.Read"]` on the `/data` route (finding 5)** — real hardening; lives in `main`, will do as its own small PR.
- **CSP** — needs a custom response-headers policy + a small deploy-role IAM addition; planned as a scoped follow-up.
- **TLS 1.0 floor (finding 8)** — inherent to the default `*.cloudfront.net` cert; recorded as a conscious accept for dev, revisit with a custom domain.

Findings 1/2 (app-registration v2 token + `Dashboard.Read` exposure) and 4 (deploy-role CloudFront/S3 permissions — already granted) are config-side and confirmed. Only remaining manual go-live step is registering the CloudFront URL as an app-registration redirect URI (no trailing slash); CORS is now automatic.

langz:
> Thanks for the review — addressed the actionable items in 624a56f:
> 
> - **CORS chicken-and-egg (finding 3):** the `/data` API now `concat`s the live CloudFront origin into `allow_origins`, so cross-origin calls work on the first deploy — no manual `DASHBOARD_ORIGINS` edit + second deploy. No dependency cycle (the API doesn't reference the distribution).
> - **403/404 → index.html rewrite (finding 6):** removed. No client-side router, the root is served by `default_root_object`, and the rewrite made a missing `config.json` fail open and masked failed asset publishes.
> - **Cache-Control (finding 7):** `--cache-control 'no-cache'` on the sync so returning browsers revalidate (the invalidation alone doesn't reach already-cached clients).
> - Plus: `is_ipv6_enabled`, dashboard bucket versioning (free rollback with `--delete`), `if: always()` on the summary step, reworded the header comment, and refreshed the deploy-facing README (config.json keys + v2-token / `Dashboard.Read` prerequisites).
> 
> Deferred as separate follow-ups (not this PR's diff):
> - **`authorization_scopes = ["Dashboard.Read"]` on the `/data` route (finding 5)** — real hardening; lives in `main`, will do as its own small PR.
> - **CSP** — needs a custom response-headers policy + a small deploy-role IAM addition; planned as a scoped follow-up.
> - **TLS 1.0 floor (finding 8)** — inherent to the default `*.cloudfront.net` cert; recorded as a conscious accept for dev, revisit with a custom domain.
> 
> Findings 1/2 (app-registration v2 token + `Dashboard.Read` exposure) and 4 (deploy-role CloudFront/S3 permissions — already granted) are config-side and confirmed. Only remaining manual go-live step is registering the CloudFront URL as an app-registration redirect URI (no trailing slash); CORS is now automatic.
> 

Thanks, have a nice evening :)

joakbjerk:
Right back at ya 🤝

PR #9175, opened 2026-08-31
fix(Dialog): prevent scroll jump to top when opening on long pages

langz:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1788172271827289

Can be demoed https://eufemia.dnb.no/uilib/components/dialog/demos or https://eufemia.dnb.no/uilib/components/help-button/#help-button-inside-a-suffix

Opening a `Dialog` (for example via `HelpButton`) on a page taller than the viewport scrolled the page to the top, which looked like a page refresh.

The scroll lock set the body to `height: 100%`. Once the locked `html` height is definite and the app authors `body { height: 100% }` (as the portal and most DNB apps do), a long page collapses to the viewport height, so the browser clamps the scroll position to the top when the Dialog opens.

The locked body now uses `height: auto` with `min-height: 100%`, so short documents still fill the viewport without collapsing long ones.

Introduced by #8985 (v11.10.0).

### DIFF
--- a/packages/dnb-eufemia/src/components/modal/__tests__/Modal.test.tsx
+++ b/packages/dnb-eufemia/src/components/modal/__tests__/Modal.test.tsx
@@ -1176,7 +1176,7 @@ describe('Modal component', () => {
     expect(outsideButton.getAttribute('aria-hidden')).toEqual('true')
   })
 
-  it('runs expected side effects on desktop', () => {
+  it('locks body scroll without fixing its height on desktop', () => {
     render(
       <Modal {...props}>
         <DialogContent />
@@ -1190,7 +1190,8 @@ describe('Modal component', () => {
     fireEvent.click(elem)
 
     expect(document.body.style.overflow).toBe('hidden')
-    expect(document.body.style.height).toBe('100%')
+    expect(document.body.style.height).toBe('auto')
+    expect(document.body.style.minHeight).toBe('100%')
     expect(document.body.style.boxSizing).toBe('border-box')
     expect(document.body.style.marginRight).toBe('0px')
     expect(document.documentElement.style.height).toBe('100%')
@@ -1291,7 +1292,8 @@ describe('Modal component', () => {
     expect(document.body.style.top).toBe('0px')
     expect(document.body.style.left).toBe('0px')
     expect(document.body.style.right).toBe('0px')
-    expect(document.body.style.height).toBe('100%')
+    expect(document.body.style.height).toBe('auto')
+    expect(document.body.style.minHeight).toBe('100%')
     expect(document.documentElement.style.height).toBe('100%')
     expect(
       document.documentElement.getAttribute('data-dnb-modal-active')
--- a/packages/dnb-eufemia/src/components/modal/bodyScrollLock.ts
+++ b/packages/dnb-eufemia/src/components/modal/bodyScrollLock.ts
@@ -40,7 +40,11 @@ const setOverflowHidden = () => {
     )
 
     bodyElement.style.overflow = 'hidden'
-    bodyElement.style.height = '100%'
+    // `height: auto` (not `100%`): with the html height above now definite, a
+    // `body` height of `100%` collapses a long page to the viewport and resets
+    // scroll to the top on open. `min-height` still fills the viewport if short.
+    bodyElement.style.height = 'auto'
+    bodyElement.style.minHeight = '100%'
     bodyElement.style.boxSizing = 'border-box'
     bodyElement.style.marginRight = `${scrollBarWidth}px`
 
@@ -49,7 +53,13 @@ const setOverflowHidden = () => {
         ;['overflow', 'height'].forEach((x) => {
           htmlElement.style[x] = htmlStyle[x] || ''
         })
-        ;['overflow', 'height', 'boxSizing', 'margin'].forEach((x) => {
+        ;[
+          'overflow',
+          'height',
+          'minHeight',
+          'boxSizing',
+          'margin',
+        ].forEach((x) => {
           bodyElement.style[x] = bodyStyle[x] || ''
         })
         htmlElement.style.removeProperty('--scrollbar-width')
--- a/packages/dnb-eufemia/src/fragments/drawer-list/__tests__/DrawerList.test.tsx
+++ b/packages/dnb-eufemia/src/fragments/drawer-list/__tests__/DrawerList.test.tsx
@@ -834,7 +834,7 @@ describe('DrawerList component', () => {
     rerender(<MockComponent open />)
 
     expect(document.body.getAttribute('style')).toBe(
-      'overflow: hidden; height: 100%; box-sizing: border-box; margin-right: 0px;'
+      'overflow: hidden; height: auto; min-height: 100%; box-sizing: border-box; margin-right: 0px;'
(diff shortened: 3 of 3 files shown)


joakbjerk (approved):
🦤

tujoworker:
Can you double check the PR that probably caused the issue in first place? https://github.com/dnbexperience/eufemia/pull/8985

langz:
> Can you double check the PR that probably caused the issue in first place? #8985

I can confirm that the PR that I state introduced it, actually introduced it 🤝

langz:
@langz — I checked the "does this reintroduce the Safari geometry issue that #8985 fixed?" question by testing in the real WebKit engine on the live previews **and** in real Safari (v26.6.2) locally. Summary below.

### ✅ The scroll-jump fix works
Opening the HelpButton dialog after scrolling down 1500px:
- **Production** (carries #8985's `body { height: 100% }`): scroll jumps `1500 → 0` on open — the reported bug.
- **This branch's preview**: scroll stays at `1500`. Fixed.

### ✅ It does not regress the Dialog's geometry
The overlay/content measurements are **identical** on production and this branch (`top: 0, height = viewport`). Reason: `.dnb-modal__content` and `.dnb-modal__overlay` are `position: fixed` with **no** containing-block-forming ancestor (no `transform`/`filter`/`contain`/`will-change` in the chain), so they resolve against the viewport and are independent of the body's height. Swapping the locked body height therefore can't distort them — confirmed live on both sites.

### ⚠️ One caveat worth flagging
`min-height: 100%` is **not** strictly equivalent to `height: 100%`. In real Safari (and it reproduces in every engine, since it's spec-correct) an element set to `height: 100%` whose containing block is the body:

| locked body style | body height | child `height:100%` |
|---|---|---|
| `height: auto` (pre-#8985) | 0 | **0 — collapsed** |
| `height: 100%` (#8985) | 200 | 200 — fills |
| `height: auto; min-height: 100%` (this PR) | 200 | **0 — collapsed** |

A `min-height`-derived height does not count as a "definite" height for resolving a descendant's percentage `height`. The Dialog doesn't rely on this (it's viewport-fixed, as above), so this PR is fine for Eufemia's own components. But #8985 was originally fixing a "legacy document" in Safari — **if** that original case relied on the locked body handing a concrete height down a `height: 100%` chain, this change would quietly bring it back.

### Recommendation
Attribution and fix both look correct. The one remaining check is to run this branch's preview against the exact document that motivated #8985 (the Slack thread linked there) to confirm that case still renders full-height. @tujoworker authored #8985, so he'd know the original repro.

langz:
@langz — I went back and checked specifically whether #8985 stays solved by this change.

**Short answer: no. As written, this PR reopens #8985** — for exactly the class of pages #8985 was about (documents that author a `height: 100%` layout chain down from `body`). The scroll-jump fix itself is correct and I could not fault it; the problem is that `min-height: 100%` does not do the job `height: 100%` was doing.

### Why

`min-height` does not make the body's height *definite*. A descendant `height: 100%` only resolves when its parent's height is definite — with `height: auto` it falls back to auto no matter what `min-height` floor is present. So every `height: 100%` in the app's own layout chain collapses to content height while the dialog is open. That is precisely what #8985's first commit — *"Preserve body height when opening modal"*, which removed the `height: auto` override entirely — set out to stop.

### Measured

Real modal DOM (captured from this PR's own preview deployment) plus the real compiled Eufemia stylesheets from that same deployment, so the geometry is the component's actual CSS and not a mock. Chromium and WebKit, 1280×800, identical results in both. "Legacy document" here is the structure Eufemia's own `resetLegacyStyles()` mixin targets — `html[xmlns]`, `body#dnbLayoutDefault`, `#wrapper > #column_content > #root` — with `height: 100%` down that chain.

| case | pre-#8985 `auto` | #8985 `100%` | **this PR** `auto` + `min-height` | proposal below |
|---|---|---|---|---|
| Long page — scroll position kept | ✅ 1200 | ❌ → 0 | ✅ 1200 | ✅ 1200 |
| Legacy short page — app's own full-height chain | ❌ 150px | ✅ 800px | ❌ **150px** | ✅ 800px |
| Legacy + portal root inside a `transform`ed ancestor, short page — overlay/content height | ❌ 150px | ✅ 800px | ❌ **150px** | ✅ 800px |
| Legacy in quirks mode, long page — scroll kept | ❌ → 0 | ❌ → 0 | ❌ → 0 | ✅ 1200 |

Row 3 is #8985's symptom verbatim — *"cause the modal layout to cover only part of the viewport"* — and it returns with this PR: the overlay and `.dnb-modal__content` measure 150px tall in an 800px viewport. That row needs the modal's containing block to come from the app's chain (a `transform`/`filter` on an ancestor, which is common).

Row 2 is the broader one and needs no transform at all: any legacy page with a full-height layout has that layout collapse to content height behind the overlay, which is visible through it.

### Proposal — satisfy both requirements

The two constraints are *don't collapse the page* (so the scroll survives) and *keep the height definite* (so the legacy chain resolves). A definite pixel height does both:

```ts
// capture before mutating, next to scrollBarWidth
const documentHeight = htmlElement.scrollHeight
…
bodyElement.style.height = `${documentHeight}px`
bodyElement.style.minHeight = '100%'
```

Everything else in this PR stays as-is — including adding `minHeight` to the restore list, which is needed and correct (I verified the style snapshot round-trips `height`, `min-height` and `overflow` back to their original inline values exactly).

This also fixes a case that none of the three variants handled: quirks-mode legacy pages currently lose the scroll position with `auto`, with `100%` **and** with `auto + min-height`, but keep it with a definite pixel height.

Trade-off worth stating: a pixel height is a snapshot, so it will not re-adapt if the page behind reflows while the dialog is open. The body is `overflow: hidden` and covered by the overlay, so it is not visible, and the `min-height: 100%` floor covers shrinkage — but it is a real difference from `auto`.

Side note: the Android path calls the same function, and there #8985 collapses the body to 800px while both `auto` and the proposal keep it at its full 4000px.

### Limits of this check

- The legacy structure and the transformed-ancestor variant are reconstructed from this repo, not from the original case in the Slack thread. Row 2 does not depend on that reconstruction; row 3 does.
- WebKit via Playwright, not real Safari on a real device.

So having #8985's original reporter re-test on this preview is still the thing that would settle it definitively.

### One small correction to the description

> once the locked `html` height is definite **and the app authors `body { height: 100% }`** (as the portal and most DNB apps do)

The app's own `body { height: 100% }` is not a precondition — I reproduced the jump with no author-side body height at all, because the lock's own inline `height: 100%` is sufficient. The half that *is* required is the lock's `html { height: 100% }`; with that left untouched there is no jump. As written it makes the bug read as app-specific when it affects every consumer.

langz:
I guess we could also manual test. 
I’m not 100% sure how you executed the manual test the last time you did this @tujoworker ?

tujoworker:
I did log in to the netbank and did change code in dev console.

langz:
> I did log in to the netbank and did change code in dev console.

Thanks, I'll give it a go tomorrow then 🙏

langz:
> > I did log in to the netbank and did change code in dev console.
> 
> Thanks, I'll give it a go tomorrow then 🙏

@tujoworker, I manually verified it, and #8985 still works as expected in this PR :)

PR #9177, opened 2026-08-31
feat(docs): version metadata (since/deprecatedIn/removedIn) and migration index

langz:
## Summary

Tags every documented prop/event with the Eufemia version it was introduced (`since`), deprecated (`deprecatedIn`), and/or removed (`removedIn`), and publishes a machine-readable **migration index**. This lets an AI agent (via the docs MCP server) answer *"is this available in the version I use?"* and surface upgrade notes from a single (latest) doc set.

## What's included

- **Schema** — `PropertiesTableProps` (`shared/types.tsx`) gains optional `since` / `deprecatedIn` / `removedIn`. Forward-compatible; the existing properties-table renderer ignores unknown fields.
- **Generator capture** — the LLM metadata generator carries the three fields through `DocEntry`, so author-set values flow into the emitted JSON automatically (and appear inline in `component_props`).
- **Inferred backfill** — `scripts/generateVersionMetadata.ts` (run via `yarn workspace eufemia-llm-metadata build:version-metadata`) walks each `*Docs.ts` file's git history and maps every introduction to the earliest release that contains it, writing the committed `version-metadata.json`. **Author annotations always win** at build time; inferred values are marked `sinceInferred`, and `sinceFloor` when only "at or before" can be established.
- **Migration index** — the docs build emits `build/docs/migrations.json` aggregating added / deprecated / removed per release, folding back prop removals that no longer exist in the current docs (e.g. the v10→v11 `snake_case` → `camelCase` rename that dropped ~340 property names in one release).
- **MCP tool `migration_index`** — filters by `component`, `fromVersion`/`toVersion`, and `changeType`. A call with no narrowing filter returns a compact per-version summary (the full index is ~576 KB); a filtered call returns full entries.
- **Worked examples + docs** — annotated Breadcrumb `backgroundColor` (`since` + `deprecatedIn`, both verified against history) and Button `iconPosition` (`since`), plus an authoring guide in the tool's README.

## Correction to the original assumption

The work started from the belief that structured `*Docs.ts` files only existed from **v11.0.0**, so `since` would be floored there. In fact they exist since **v10.21.0** (the "DocTable" convention, #3322). So `since` is derived accurately across the full history rather than floored at v11; values present in a file's first tracked commit are marked `sinceFloor` ("at or before", since the prop may predate structured docs). Verified independently — e.g. Input's Docs file first appears at v10.21.0 by both this method and `git describe`.

## Design notes

- **Author annotations are the source of truth** and surface inline in `component_props`. **Inferred** history is exposed only through the `migration_index` tool (from `migrations.json`), clearly marked — it is deliberately not injected inline into `component_props` (which stays authoritative) and does not appear in `llm.md`.
- `build/docs/migrations.json` is a build artefact (git-ignored); `version-metadata.json` is the committed, cheap-to-merge cache. The expensive git walk only runs when regenerating.

## Testing

- Tool suite: **71 passing**; docs MCP suite: **68 passing** (incl. new tests for the `migration_index` tool, the git-history derivation, migrations aggregation, and the schema/capture flow).
- Type-checking clean for both the tool and `@dnb/eufemia`; lint clean on new/changed files (including the strict `*Docs.ts` rules).
- Ran the full docs build and verified against the real output: author annotations appear in `component_props`, removed props appear in the migration index, and the tool returns correct filtered results.


langz:
This seems like a bit much work,, but could be nice to have

PR #9181, opened 2026-09-01
fix(Heading): restore level and lifecycle handling

tujoworker:
The hooks rewrite left counters and provider state tied to mount-time values. This caused stale heading levels, counter reuse between instances, hydration inconsistencies, lost refs, and incorrect rendering for zero text and custom elements.

This restores instance-aware counter cleanup and same-commit level updates, keeps provider props current, and defines `Heading.Level reset={1}` as the boundary for independent server-rendered trees.

Preview: https://fix-heading-hook-regressions.eufemia-e25.pages.dev/uilib/elements/heading


langz (commented):
Nice, I guess this fix some of the points @snorrekim reported?
If so, perhaps @snorrekim csn take a quick glance at this 🤞

langz:
@tujoworker I think there's one thing we need to fix before this goes in 🙏

`level` also accepts strings (`'1'`–`'6'`), and the new render-phase check compares the parsed number with the raw prop:

```js
state.prevLevel !== props.level // 3 !== '3' is always true
```

With a string that's always true, so the correction re-runs on every render. And when the correction wants a different level than the prop asks for, it never settles:

```jsx
<Heading level="1">a</Heading>
<Heading level="1">b</Heading>
```

`main` corrects those to h1 + h2 — here it throws _Too many re-renders_ and nothing renders at all. Same thing inside `Heading.Level reset={1}`. Numbers are unaffected.

This one matters a bit extra since our own portal passes strings — `AutoLinkHeader` defaults to `level = '1'`, and all MDX headings go through `<Header level="2" />` etc.

The smallest fix I found that converges *without* changing any of today's output — store the raw prop back, in both `Heading.tsx` and `HeadingProvider.tsx`:

```diff
-    state = { ...state, level: newLevel, prevLevel: level }
+    state = { ...state, level: newLevel, prevLevel: props.level }
```

together with `prevLevel?: InternalHeadingLevel | HeadingProps['level']`.

I compared that against `main` on every string and number case I could think of and the output is identical, tests green and types clean 🤝

(Worth noting the other obvious fix — comparing against the parsed `level` — also stops the crash, but it changes leveling wherever string levels are used, e.g. a page starting at `level="2"` then renders `h1`. So probably a separate discussion if we want strings and numbers to behave the same.)

Could you also add a test with `level="2"`? Looks like we don't cover string levels today.

langz (approved):

tujoworker:
Addressed in 8a2b2283786. The render-phase state now stores the raw `level` prop after correction, so string levels converge without changing their existing output. Added coverage for `level="2"`, repeated `level="1"` headings, and string levels on `Heading.Level`.

packages/dnb-eufemia/src/components/heading/Heading.tsx
```diff
@@ -334,10 +342,9 @@ Heading.Up = (props: HeadingStaticProps) => (
 Heading.Down = (props: HeadingStaticProps) => (
   <HeadingProvider decrease {...props} />
 )
-Heading.Reset = (props: HeadingStaticProps) => {
-  globalHeadingCounter.current?.reset()
-  return <HeadingProvider {...props} />
-}
```
langz, line 340: I guess this change is fine 🤝 
I could not find any tests for these components, I might have not looked too thorough 😅

langz (approved):

langz:
@tujoworker One thing I missed earlier: after 8a2b2283786, a level written as a **string** no longer matches the **same level written as a number** on `Heading.Level`.

```jsx
<Heading.Level level={3}>
  <Heading>Test</Heading>
</Heading.Level>

<Heading.Level level="3">
  <Heading>Test</Heading>
</Heading.Level>
```

|             | `main`  | this PR     |
| ----------- | ------- | ----------- |
| `level={3}` | h1      | h1          |
| `level="3"` | h1      | **h3**      |

And with a heading in front of it: `level={3}` gives h1 + h2 in both, `level="3"` gives h1 + h2 on `main` but h1 + **h3** here.

So the string variant now escapes the auto-correction — a page can start at h3, or jump h1 → h3. `level="2"` is unaffected (legal single step), and plain `<Heading level="…">` is unchanged, so our portal is fine. It's specifically `Heading.Level` + a string.

Cause: the initializer stores `prevLevel` as a parsed number, but the new check compares the raw prop, so a string always gets one extra correction pass that a number never gets.

Could we make both spellings behave the same? And `should preserve string Heading.Level correction behavior` fails on `main` (expects H3, gets H2), so today it locks the change in rather than guarding it — maybe rename it too 🙏

langz:
> The hooks rewrite left counters and provider state tied to mount-time values. This caused stale heading levels, counter reuse between instances, hydration inconsistencies, lost refs, and incorrect rendering for zero text and custom elements.
> 
> This restores instance-aware counter cleanup and same-commit level updates, keeps provider props current, and defines `Heading.Level reset={1}` as the boundary for independent server-rendered trees.
> 
> Preview: https://fix-heading-hook-regressions.eufemia-e25.pages.dev/uilib/elements/heading

I tried reproducing this, and was not able to, I guess this was just a false positive. Will look into it a bit 🤔

tujoworker:
Fixed in 14c19da5d7f. `HeadingProvider` now compares normalized numeric levels, so `level={3}` and `level="3"` follow the same correction path. Added regression coverage for both forms at the document root and after a preceding heading. Plain `Heading` string levels keep their existing legacy behavior.

packages/dnb-eufemia/src/components/heading/Heading.tsx
tujoworker: Added direct coverage for `Heading.Reset` in 14c19da5d7f. The test advances the parent context to `h3` and verifies that the static reset provider returns its child to `h2`. `Heading.Increase` and `Heading.Decrease` were already covered by the existing tests.

langz (approved):
⤴️

langz:
`text=""` or `text={false} `now render nothing where they previously fell back to rendering children.
I guess this is just a strange edge case, so probably nothing to fix I guess 👍

tujoworker:
Fixed in fe493da. Empty-string and false `text` values now fall back to children as before, while `text={0}` still renders `0`. Added regression coverage for all three cases.

langz:
Rebased onto the latest `main` — conflict-free.

PR #9182, opened 2026-09-01
chore(analytics): serve the dashboard from a read-only Lambda with a scheduled snapshot generator

joakbjerk:
The browser-facing dashboard `GET /data` path ran on the shared analytics Lambda and its execution role, which has S3 write and Athena access the read path does not need. This isolates it.

`/data` now points at a dedicated Lambda whose execution role can only read the snapshot object, and snapshot generation moves off the request path to an hourly scheduled job that runs under the existing analytics role. The route, JWT authorizer, CORS, and response shape are unchanged.

Depends on the deploy role gaining `iam:PassRole` on the new read-only role plus permission to manage the schedule (separate OIDC config PR), and on that role being created out-of-band. Hold merge until both are in place, otherwise the deploy apply fails.


joakbjerk:
Known non-blocking nits, tracked as a follow-up rather than expanding this PR:

1. **Alarms have no `alarm_actions`.** Intentional for now — state is console-visible but nothing notifies. Wiring an SNS target needs extra deploy-role IAM, so it's deferred.
2. **`snapshot_not_running` breaches on a fresh deploy.** Harmless while no actions are wired; it only becomes noise once (1) lands, so the two go together.
3. **~1h cold-start empty window** (one-time; the snapshot persists in S3 across later deploys). Seeded by invoking the generator once right after the first deploy.

Follow-up (one PR): wire SNS notifications + a deploy-time seed invoke of the generator (`aws_lambda_invocation`), which closes nits 2 and 3 together. Both need a small deploy-role permission bump, so they're batched.

Merge-order gate is satisfied: the out-of-band read-only role exists and the deploy-role `PassRole`/schedule/alarm permissions are merged in the OIDC config repo.

langz (approved):

langz:
Minor test-gap nit: now that `GET /data` is excluded from the ingest handler's routing contract, nothing verifies it's actually wired to the dashboard-read Lambda. Worth a small assertion that the `/data` integration `integration_uri` points at `aws_lambda_function.dashboard_read` to close the loop. Non-blocking.

joakbjerk:
> Minor test-gap nit: now that `GET /data` is excluded from the ingest handler's routing contract, nothing verifies it's actually wired to the dashboard-read Lambda. Worth a small assertion that the `/data` integration `integration_uri` points at `aws_lambda_function.dashboard_read` to close the loop. Non-blocking.

Nice, fikser det i morgen 😄
Nå må du legge deg! 😴

joakbjerk:
Added the assertion in `e59d3c09e83` — routes.test.ts now checks the dashboard `/data` integration's `integration_uri` targets `aws_lambda_function.dashboard_read`, closing the routing-contract loop.

PR #9183, opened 2026-09-01
chore(analytics): hide dashboard until sign-in resolves

joakbjerk:
The analytics dashboard rendered its full chrome (KPIs, charts) on load, then `app.js` ran the async sign-in check and redirected to Entra. Unauthenticated users saw the dashboard flash for a couple of seconds before being sent to the login page.

The dashboard content is now hidden by default behind a neutral "Checking sign-in…" placeholder, and only revealed once a session is resolved (or when sign-in is not configured, e.g. local preview). Since the login redirect never resolves its promise, the page stays on the placeholder and navigates away without ever exposing the dashboard.


langz (approved):
Image

PR #9184, opened 2026-09-01
fix(List): inset item borders

tujoworker:
Aligns List item borders with the Figma specification, where the outline is drawn inside the bounding box.

The pseudo-element border is now inset for all List items. This also removes the need for `overflow-clip-margin`, which Safari does not support.


langz (approved):
Thanks, looks good to me 🟢

PR #9189, opened 2026-09-01
feat(DatePicker): make adjacent month dates selectable

tujoworker:
Dates from the previous and next month can now be selected directly from the current calendar view.

Outside-month dates remain visually distinct without being treated as disabled. In range mode, repeated dates expose one control to assistive technology, and selecting an adjacent date moves the calendar and focus to the selected month.

Preview: https://agents-datepicker-clickable.eufemia-e25.pages.dev/uilib/components/date-picker/demos
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1752567655383199


langz (approved):
LGTM 👍 Nice fix — the isLastMonth/isNextMonth swap makes the direction load-bearing and correct. A couple of small notes below.

langz:
`isInactive` is now always equal to `isDisabled` (and `isSelectable` === `!isDisabled`), so the `isDisabled || isInactive` check is redundant. Fine to keep the flags for the public dayObject API, just noting it.

tujoworker:
I’m keeping both checks. `makeDayObject` initializes `isInactive` and `isDisabled` to the same value, but the public `onDaysRender` callback can set `isInactive` independently; our weekend example and tests rely on that to disable custom dates. `isDisabled` also needs to remain an independent check so `minDate` and `maxDate` cannot be overridden accidentally.

kimroen:
> the isLastMonth/isNextMonth swap makes the direction load-bearing and correct.

Love a load-bearing direction

Awesome feature, I've seen this feedback a few times in our apps too!

PR #9191, opened 2026-09-01
feat(Theme): deprecate `size` in favor of `density`

tujoworker:
The Theme `size` property describes component density rather than a specific size. This adds `density` as the canonical property across Theme and persisted theme state, while keeping `size`, `ThemeSizes`, and the legacy DOM hooks backwards compatible and deprecated until v13.

The v13 guide covers replacing the deprecated API.


langz (approved):
Great great, I like `density`, too me it it's way more meaningful than `size` :)

Should we share the type and values of the `density` property between `Space` and `Theme`? Or should they be kept separate?

Personally, if it makes sense, I like that the props have the same values for the same concept across our components.

Perhaps we'll make this connection at a later point :)

PR #9187, opened 2026-09-01
chore(Theme): migrate internal consumers off the deprecated `name`

langz:
Follow-up to #9186 (now merged).

#9186 makes `brand` the canonical Theme property and deprecates `name` (mirrored until v13). Several internal consumers still referenced the deprecated `name` — **8 places across 5 files**. Nothing breaks today because `name` is mirrored, but these all break at v13. This removes every internal reference to `name` (resolving via `useTheme`, which covers the deprecated `name`).

PR #9193, opened 2026-09-02
chore(Theme): make the portal theme handler brand-aware

langz:
Follow-up to #9186 and #9187 (both merged). Covers the review findings neither addressed: the **portal**, the **docs**, and **missing tests**. Each fix was reproduced as a failing test first.

## 1. Portal theme runtime disagreed with Eufemia

The portal's theme runtime shares the `eufemia-theme` localStorage key with Eufemia's `getTheme`/`setTheme`, but nothing types the two together — so every mismatch below was invisible to the compiler.

- **Startup activated the wrong brand** (caught in review — thanks @tujoworker). Four pre-hydration paths resolved `stored.name` only, so `{"brand":"carnegie"}` activated `ui` first → theme flash: build startup (`__loadEufemiaTheme`), dev startup (`getInitialTheme`), the pre-paint head script in `prerender-utils.ts`, and its copy in `prerender.mjs`. All four now resolve `brand` before `name` (`||`, matching `shared/Theme.tsx`).
- **Unknown brand disabled every stylesheet.** The pre-paint script never checked the page ships the brand; an unknown one matches no `link[data-eufemia-theme]`, so the disable loop left the page unstyled — reachable via `?eufemia-theme=`, but the real case is a brand a later release renames/removes (returning visitor stuck until they clear storage). Now validated against the theme links present, like `getTheme()`'s `isValidTheme()`. Build startup already no-ops on an unknown name.
- **Handler ignored `brand`.** A `brand`-only payload resolved to `'ui'` (selected brand silently lost); switching brand while a `brand` key was present wrote a diverged `{ brand: 'eiendom', name: 'sbanken' }`, after which Eufemia reported the old brand. Now resolves `brand ?? name` and re-derives both keys, never diverging.
- **Parsed `?eufemia-theme=` differently.** The regex `/.*eufemia-theme=([^&]*).*/` matched any param whose name merely *ends* with the key (`?x-eufemia-theme=sbanken`) and took the *last* repeated value, where Eufemia's `URLSearchParams` takes the *first*. Now uses `URLSearchParams` too.
- **Unknown brand cost you dark mode.** The fallback to the default discarded the rest of the stored state, including the `colorScheme` that `ColorSchemeScript` (rendered for FOUC prevention) keeps reading. Now keeps the state and replaces only the brand.
- **One source of truth.** `ToggleDarkMode`, `ChangeStyleTheme` and `PortalToolsMenu` wrote through the shim but read Eufemia's `getTheme`; since the shim applies the CSS and feeds `<Theme>` in `portal-app`, they now read it too, so the brand is validated on read. `portal-app` no longer spreads the deprecated `name` into its own `<Theme>`, and `ToggleDarkMode` drops an `as Record<string, unknown>` cast that defeated the `setTheme` signature.

The shim is deliberately **not** refactored to delegate to Eufemia's `getTheme`/`setTheme`: it stays free of library runtime imports, so the contract is documented at its top and locked by tests instead. Happy to do the deeper refactor separately.

## 2. Docs taught the deprecated property

Migrated to `brand`: `typography/Examples.tsx`, `typography/_helpers.tsx`, `components/logo/Examples.tsx` (x2), `components/avatar/Examples.tsx`, and `SidebarMenu`. Logo/Avatar were held back until **#9187** landed — their theme argument comes from `Logo`'s `svg` factory, which now carries `brand` (verified for both `brand` and the deprecated `name`); migrating early defaulted every themed logo and failed 11 visual-regression tests. No deprecated theme-name reads remain in the portal.

## 3. v13 guide: the shape that fails silently

`VisibilityByTheme` matches on raw object keys, so `visible={[{ name: 'sbanken' }]}` works today only because `name` is mirrored into the context. At v13 it stops matching and **the content just disappears**; TypeScript then reports `TS2353` for an inline object, but plain JavaScript still gets nothing. Added to the guide.

## 4. Tests for the paths that had none

- **`theme-handler-storage-contract.test.ts`** (new) — shim and Eufemia agree on the stored payload and query-parameter resolution, both directions, for brand-only / legacy name-only / mirrored payloads. Pins the one deliberate difference (only the portal validates the brand, because only it must have the CSS) and the `density`/`size` pair the portal carries through but does not manage (added after #9191).
- **Generated startup code** — dev and build modules evaluated with a brand-only payload, asserting the body brand class and the theme handed to the lazy loader.
- **Pre-paint head script** — run against real `link[data-eufemia-theme]` elements, asserting which stylesheet stays enabled and the published `globalThis.__eufemiaTheme`.
- **`prerender.mjs` drift guard** — that file's copy of the scripts isn't reachable from a unit test, so a guard pins it to the `prerender-utils.ts` copy; reverting either alone now fails. That is how this defect survived.
- **`shared/Theme.test.tsx`** — `Theme` resolving a brand from a context that provides only the deprecated `name`; that block had no coverage at all (removing it left all 46 existing tests green).

Legacy `{"name":...}` payloads are covered alongside every new case, so the v12 storage format keeps working in each startup path.

## Verification

- Portal touched suites 150/150; library `shared` suites green; `tsc --noEmit` 0 errors on both; lint + Prettier clean; rebased on current `main` (includes #9187, #9191).
- **Every fix is mutation-tested** — reverted in turn to confirm a test actually fails: brand resolution, query parsing, retained state, carried-through state, the four startup paths, the pre-paint validation, the `prerender.mjs` drift guard, and the `Theme.tsx` block.

## Deliberately left out

- **`VisibilityByTheme.Name`** (used on 3 docs pages) still returns the brand display name — the same "name means brand" confusion this rename targets. Adding `.Brand` and deprecating `.Name` is new public API, so it wants a maintainer decision.
- A defensive `brand ?? name` fallback in `ThemeWrapper`'s data attributes — unreachable today, so dead code.
- Stripping the deprecated `size` alongside `name` in `portal-app` — the shim never writes `density`/`size`, so that branch would be dead.

### DIFF
--- a/packages/dnb-design-system-portal/src/core/ChangeStyleTheme.tsx
+++ b/packages/dnb-design-system-portal/src/core/ChangeStyleTheme.tsx
@@ -7,7 +7,7 @@ import { Field } from '@dnb/eufemia/src/extensions/forms'
 
 export default function ChangeStyleTheme(props) {
   const themes = getThemes()
-  const { name } = getTheme()
+  const { brand } = getTheme()
   const { update } = useContext(Context)
 
   const data = Object.entries(themes).reduce((acc, [key, value]) => {
@@ -20,12 +20,12 @@ export default function ChangeStyleTheme(props) {
   return (
     <Field.Selection
       id="change-theme"
-      value={name}
+      value={brand}
       data={data}
       label="Change Brand"
       onChange={(value) => {
         update({ skeleton: true })
-        setTheme({ name: value as ThemeNames }, () => {
+        setTheme({ brand: value as ThemeNames }, () => {
           update({ skeleton: false })
         })
       }}
--- a/packages/dnb-design-system-portal/src/core/ChangeStyleTheme.tsx
+++ b/packages/dnb-design-system-portal/src/core/ChangeStyleTheme.tsx
@@ -1,8 +1,11 @@
 import { useContext } from 'react'
 import { Context } from '@dnb/eufemia/src/shared'
-import { getTheme } from '@dnb/eufemia/src/shared/Theme'
 import type { ThemeNames } from '@dnb/eufemia/src/shared/Theme'
-import { getThemes, setTheme } from '../../vite/client/shims/theme-handler'
+import {
+  getTheme,
+  getThemes,
+  setTheme,
+} from '../../vite/client/shims/theme-handler'
 import { Field } from '@dnb/eufemia/src/extensions/forms'
 
 export default function ChangeStyleTheme(props) {
--- a/packages/dnb-design-system-portal/src/core/ToggleDarkMode.tsx
+++ b/packages/dnb-design-system-portal/src/core/ToggleDarkMode.tsx
@@ -1,6 +1,5 @@
 import { useState } from 'react'
-import { getTheme } from '@dnb/eufemia/src/shared/Theme'
-import { setTheme } from '../../vite/client/shims/theme-handler'
+import { getTheme, setTheme } from '../../vite/client/shims/theme-handler'
 import type { ThemeColorScheme } from '@dnb/eufemia/src/shared/Theme'
 import { Field } from '@dnb/eufemia/src/extensions/forms'
 
@@ -22,7 +21,7 @@ export default function ToggleDarkMode(props) {
         { value: 'light', title: 'Light' },
       ]}
       onChange={(value: string) => {
-        setTheme({ colorScheme: value } as Record<string, unknown>)
+        setTheme({ colorScheme: value as ThemeColorScheme })
         updateColorScheme(value as ThemeColorScheme)
       }}
       {...rest}
--- a/packages/dnb-design-system-portal/src/docs/uilib/about-the-lib/releases/eufemia/v13-info.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/about-the-lib/releases/eufemia/v13-info.mdx
@@ -65,6 +65,15 @@ The deprecated `name` property and `data-name` attribute have been removed. Use 
 
 Replace `useTheme().name` with `useTheme().brand`. For persisted theme state, replace `getTheme().name` and `setTheme({ name })` with `getTheme().brand` and `setTheme({ brand })`.
 
+`VisibilityByTheme` also matches on the theme object, so replace `name` there too:
+
+```diff
+-<VisibilityByTheme visible={[{ name: 'sbanken' }]}>
++<VisibilityByTheme visible={[{ brand: 'sbanken' }]}>
+```
+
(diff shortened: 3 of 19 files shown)


tujoworker (changes requested):
Two issues need addressing before approval:

1. **Blocking:** The portal startup paths still resolve persisted theme state from `stored.name` only. `vite/client/plugins/eufemia-theme.ts` does this in both generated development and build code, and `vite/prod/prerender-utils.ts` / `vite/prod/prerender.mjs` do the same in the pre-paint script. With a brand-only payload such as `{ "brand": "carnegie" }`, these paths initially activate `ui`; the updated React handler corrects it only later, causing the wrong CSS/body class before hydration and a visible theme flash. This also means the new storage-contract test covers the shim but not the full portal runtime. Please resolve `stored.brand ?? stored.name` in every startup path and add focused generated-code/prerender coverage.

2. `src/shared/menu/SidebarMenu.tsx:433` is a normal `useTheme()` consumer but still reads the deprecated `name`. Unlike the deliberately deferred Logo/Avatar factory callbacks, it already receives canonical `brand` and should be migrated as part of this portal consumer cleanup.

langz:
Thanks — both were real, and point 1 was a genuine hole in my audit: I checked the portal components and the runtime handler but never the plugin's generated startup code or the prerender scripts. Same duplicated-storage-contract pattern I was chasing, and I walked past two more instances of it.

Fixed in 8a5c9da (+ 07635b1). All four startup paths now resolve `brand` before the deprecated `name`, using `||` to match the chain in `shared/Theme.tsx` exactly, with legacy `{"name":…}` payloads asserted alongside every new case.

`prerender.mjs` is the one I can't reach from a unit test — it carries its own copy of the script, which is exactly how the defect survived. There's now a guard pinning the two copies together; reverting either alone fails it.

Point 2: agreed, migrated. No deprecated theme-name reads left anywhere in the portal.

#9187 merged meanwhile, unblocking the three Logo/Avatar examples I'd deferred — visual regression is green.

The shim/Eufemia duplication is still deliberate (the shim avoids library runtime imports). Happy to unify separately if you'd rather have one implementation.

langz:
Follow-up while re-checking the CodeQL finding (already fixed in `07635b1`): the pre-paint head script never validated the resolved brand. An unknown brand matches no `link[data-eufemia-theme]`, so the disable loop switched off *every* theme stylesheet — completely unstyled page.

Reachable with `?eufemia-theme=nope`, but the case I care about is a stored brand that a later release renames or removes: returning visitors are stuck until they clear storage. `getTheme()` already guards this with `isValidTheme()`, so this just brings the pre-paint path in line.

Fixed in both copies of the script. The build path needed no change — `__loadEufemiaTheme()` already no-ops on an unknown name. Three tests added; reverting the fix fails all three plus the drift guard.

tujoworker (approved):

PR #9194, opened 2026-09-02
fix(Forms): preserve optional fields when editing array items

tujoworker:
Optional fields from a JSON Schema could become required after moving from `Iterate.PushContainer` to `Iterate.EditContainer`.

Required-state lookup now follows object properties and array item schemas instead of falling back to the root schema, so editing preserves each field's declared required state.

Reprod: https://stackblitz.com/edit/eufemia-starter-3godohm2?file=src%2FBeneficialOwners.tsx
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1788263705206119
Preview: https://fix-forms-json-schema-requir.eufemia-e25.pages.dev/uilib/extensions/forms/Iterate/EditContainer


langz (approved):
🚀

PR #9199, opened 2026-09-02
fix(Heading): recalculate dynamic levels

tujoworker:
Stacked on #9181 and intended to be reviewed after it.

Headings inside `Heading.Level` kept stale levels when an earlier heading or nested level provider was removed, and mounted children did not react when their provider level changed. This recalculates the affected scope while preserving mounted child state, and makes string and numeric level values behave consistently.

### DIFF
--- a/packages/dnb-eufemia/src/components/heading/__tests__/Heading.test.tsx
+++ b/packages/dnb-eufemia/src/components/heading/__tests__/Heading.test.tsx
@@ -148,11 +148,14 @@ describe('Heading component', () => {
     expect(elem[i + 1].textContent).toBe('[h2] Heading #12')
   })
 
-  it('should preserve string level correction behavior', () => {
-    render(<Heading level="2">Heading</Heading>)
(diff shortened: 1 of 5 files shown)


packages/dnb-eufemia/src/components/heading/__tests__/Heading.test.tsx
```diff
@@ -137,11 +137,14 @@ describe('Heading component', () => {
     expect(elem[i + 1].textContent).toBe('[h2] Heading #12')
   })
 
-  it('should preserve string level correction behavior', () => {
-    render(<Heading level="2">Heading</Heading>)
+  it.each([2, '2'] as const)(
```
langz, line 151: Not too important, but this generated the same it statement for the two tests.

langz (approved):

langz:
Rebased onto the latest `main` (on top of the freshly-rebased #9181 base) — conflict-free.

tujoworker:
Thank you @langz ❤️

PR #9197, opened 2026-09-02
feat(Form.Section): support async EditContainer completion

tujoworker:
Allow EditContainer's onDone callback to return a Promise so API-backed saves stay in edit mode until they succeed. Rejected saves keep the edited values available for retry and error feedback.

Motivation: [Slack discussion](https://dnb-it.slack.com/archives/CMXABCHEY/p1788184722094539)

Preview: [Async onDone example](https://feat-form-section-async-on-d.eufemia-e25.pages.dev/uilib/extensions/forms/Form/Section/EditContainer/demos/#async-ondone)

PR #9204, opened 2026-09-02
fix(Form.Section): keep onDone compatible with value-returning callbacks

langz:
Follow-up to #9197, which made `Form.Section.EditContainer` await a Promise returned from `onDone`.

That change narrowed a published prop type and breaks consumer builds on a minor release. Every claim below was measured by running the same input against the commit before #9197 (`d6d2401`), against current `main`, and against this branch.

## The break

`onDone` was typed `() => void`. TypeScript's void-return rule makes that the *most permissive* callback shape there is — a prop typed `() => void` accepts a callback returning anything. Narrowing it to `() => void | Promise<unknown>` removes that, because a union containing `void` does not get the rule.

So callbacks that exist in consumer code today stop compiling:

```tsx
onDone={() => list.push(value)}      // () => number
onDone={() => trackEvent('done')}    // () => boolean
onDone={() => saveToStore()}         // () => { id: string }
```

Verified with one probe file, one set of compiler flags, and the real exported component (not a replica):

| Callback shape | Before #9197 | On `main` | This PR |
| --- | --- | --- | --- |
| `() => boolean` | compiles | **TS2322** | compiles |
| `() => { id: string }` | compiles | **TS2322** | compiles |
| `() => number` | compiles | **TS2322** | compiles |
| `() => void` (control) | compiles | compiles | compiles |
| `async () => …` (control) | compiles | compiles | compiles |

The controls compiling in all three columns is what makes the middle column meaningful.

`onDone` is therefore typed `() => unknown`, which is as permissive as the original `() => void` while still awaiting a returned Promise at runtime. The Promise behaviour is documented on the prop and in both properties tables, rather than encoded in a type that rejects previously valid code.

## Why the runtime check is in the same PR

Relaxing the type on its own would be **worse than leaving it alone**, so the two cannot be separated.

The pending state was entered for any truthy return value, so the very callbacks that compilation now permits again reach `TypeError: result.then is not a function` — the Done button stops working and the section never leaves edit mode. I built a type-change-only variant to confirm this rather than assume it:

| `onDone={() => trackEvent('done')}` | On `main` | Type change only | This PR |
| --- | --- | --- | --- |
| Compiles | **no (TS2322)** | yes | yes |
| Switches to view mode | no | **no** | yes |
| Console error | – | **`result.then is not a function`** | none |

The middle column is the trap: a loud build error becomes a silent broken button. Fixing the type without the runtime would ship exactly that.

The returned value is now checked with `instanceof Promise`, which is how `useFieldValidation` already decides whether a callback result should be awaited.

Worth a line in the release notes either way: `async` callbacks already compiled before #9197, so anyone passing one today gets a behaviour change from fire-and-forget to awaited.

## Test

One test added — `should switch to view mode when onDone returns a non-Promise value`. It fails on `main` (reproduced twice) and passes here.

Verified: 33/33 in `EditAndViewContainer.test.tsx`; 8/8 test files in the `Form/Section` tree; lint clean; `test:types` produces a byte-identical error set to `main`. Nothing in the rendered output changed, so the `Form.Section` visual tests are unaffected.

## Not included

Findings from the same review that are behaviour or policy decisions rather than a compatibility break, kept out of this PR so it stays reviewable:

- **Keyboard focus is dropped and never restored.** Measured in Chromium on the #9197 preview: focus Done, press <kbd>Enter</kbd>, and `document.activeElement` becomes `<body>`, because a focused element that becomes `disabled` is blurred. It stays there after a failed save, leaving a keyboard or screen-reader user with no focus position next to the error. jsdom does not reproduce this. I have a verified fix for it (restore focus only when it was actually lost, so it is never stolen from the user) and can raise it separately.
- **No timeout.** Done and Cancel are both disabled while pending and there is no timer, so a request that never settles leaves the section with no way forward or back. `Form.Handler` solves this for submits with `asyncSubmitTimeout` (default 30s).
- **No minimum indicator time.** A fast save flashes the dots and the disabled state. `minimumAsyncBehaviorTime` (default 1s) exists for this.
- **Fields stay editable while saving,** unlike the submit path which disables all form elements. A user can keep typing after pressing Done, and on success the newly typed value is committed to view mode even though the save sent the previous one.
- **The pending indicator is rendered conditionally,** so the `role="status"` live region and its content appear in the same instant; `Form.SubmitButton` renders it unconditionally. Changing this alters the non-pending DOM, so it should go through the visual tests deliberately.

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/Form/Section/EditContainer/DoneButton.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/Form/Section/EditContainer/DoneButton.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useContext } from 'react'
+import { useCallback, useContext, useRef } from 'react'
 import SectionContainerContext from '../containers/SectionContainerContext'
 import ToolbarContext from '../Toolbar/ToolbarContext'
 import { useTranslation } from '../../../hooks'
@@ -6,6 +6,7 @@ import { Button } from '../../../../../components'
 import { check } from '../../../../../icons'
 import FieldBoundaryContext from '../../../DataContext/FieldBoundary/FieldBoundaryContext'
 import SubmitIndicator from '../../SubmitIndicator'
+import { useIsomorphicLayoutEffect as useLayoutEffect } from '../../../../../shared/helpers/useIsomorphicLayoutEffect'
 
 export default function DoneEditButton() {
   const { onDone, setShowError, isPending, setIsPending } =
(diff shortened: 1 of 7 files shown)


langz:
Continued in #9206.

This PR was closed, and GitHub does not allow reopening once the branch has been force-pushed — which had already happened here while narrowing the scope. #9206 carries the same branch and the current, narrowed diff.

Reopening this one would in any case show the wrong thing: its head is frozen at the earlier three-fix commit (+164/-9), not the narrowed change (+62/-8).

PR #9236, opened 2026-09-03
feat(MCP): expose Portal content workflow

tujoworker:
Makes the Portal contribution workflow available directly through the Eufemia MCP server. Agents using only the hosted or local MCP endpoint can retrieve the same focused editing, validation, pull request, preview, and portal-only deployment guidance without a separately installed skill.


langz (approved):
Wow, I should try that one time 🛕

PR #9237, opened 2026-09-03
fix(InputMasked): prevent update loops from equivalent masks

tujoworker:
Equivalent mask arrays and number-mask objects could be recreated during normal parent renders. This caused Maskito to repeatedly tear down and initialize, potentially reaching React’s maximum update depth.

Keep semantically equivalent masks stable while still rebuilding Maskito when the mask content changes.


langz (approved):
Nice, could this pattern be seen in other components that we have as well?

PR #9238, opened 2026-09-03
fix(Anchor): shorten external link tooltip hide delay

tujoworker:
External-link tooltips use the default 500 ms hide delay after the pointer leaves the anchor. This can make the tooltip overlap underlying content while the user has already moved on.

Use a shorter 100 ms hide delay for these tooltips. This preserves a brief delay while substantially reducing how long they cover nearby content.

### DIFF
--- a/packages/dnb-eufemia/src/components/anchor/Anchor.tsx
+++ b/packages/dnb-eufemia/src/components/anchor/Anchor.tsx
@@ -256,6 +256,7 @@ export function AnchorComponent(localProps: AnchorAllProps) {
           id={internalId + '-tooltip'}
           targetElement={tooltipRef}
           tooltip={tooltip}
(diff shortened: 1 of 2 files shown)


langz (approved):

langz:
Not sure if I actually like this or no, I'm not actually able to hover onto the tooltip and select the text.
https://www.w3.org/WAI/WCAG22/Understanding/content-on-hover-or-focus.html could be related?

tujoworker:
Good concern. The first version used 0 ms, which made the tooltip difficult to reach with the pointer. I've changed it to 100 ms instead. That keeps a short window for moving onto the tooltip, and the tooltip stays open while hovered, so its text can still be selected. It should preserve the hoverable behavior from WCAG 1.4.13 while reducing the lingering overlap compared with the default 500 ms.

PR #9241, opened 2026-09-03
fix(Form.Section): recover from a stuck async onDone via asyncSubmitTimeout

langz:
## Problem

When `Form.Section.EditContainer`'s `onDone` returns a Promise, the section disables its fields and both toolbar buttons until the Promise settles. A Promise that never settles (e.g. a hung request with no timeout) left the section disabled with **no way out**.

## Fix

`Form.Handler`'s async `onSubmit` already guards against this with `asyncSubmitTimeout` (30 seconds by default), after which the submit indicator reverts to normal. The section's `onDone` had no equivalent safety net.

`onDone` now re-enables the section after `asyncSubmitTimeout` when the Promise has not settled, staying in edit mode so the user can try again. This matches the rest of the form's async behavior and reuses the existing `asyncSubmitTimeout` prop — **no new public API**, and the deliberate "Cancel disabled during a pending save" behavior is unchanged.

## Why this approach

An earlier version of this PR re-enabled Cancel during the pending save and passed an `AbortSignal` to `onDone`. It was reworked to this smaller, consistent approach after review: an `AbortSignal` on `onDone` would have been a novel pattern used nowhere else in the forms system, whereas `asyncSubmitTimeout` is exactly how `onSubmit`, async `onChange`, and Wizard steps already recover from a hung async operation.

## Tests

- Added: the section re-enables its fields and buttons after `asyncSubmitTimeout` when `onDone` never settles, and stays in edit mode (verified it fails without the fix).
- Kept the existing "both buttons stay disabled while pending" test unchanged.
- Full `Form/Section` suite green (152 tests); typecheck, lint, and Prettier clean.

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Section/EditContainer/demos.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/Form/Section/EditContainer/demos.mdx
@@ -23,7 +23,7 @@ With `preventUncommittedChanges`, the user must select "Done" or "Cancel" before
 
 When `onDone` returns a Promise, the section stays in edit mode while it is pending. It switches to view mode when the Promise resolves and stays in edit mode when it rejects.
 
-While the Promise is pending, the fields inside the section and its "Done" and "Cancel" buttons are disabled. This keeps the submitted values from changing while the save is in flight. Make sure the returned Promise always settles, because one that never resolves or rejects leaves the section disabled.
+While the Promise is pending, the fields inside the section and its "Done" and "Cancel" buttons are disabled. This keeps the submitted values from changing while the save is in flight. If the Promise does not settle within `asyncSubmitTimeout` (30 seconds by default), the section re-enables itself and stays in edit mode, mirroring `Form.Handler`'s submit behavior, so a save that never settles cannot leave the section stuck.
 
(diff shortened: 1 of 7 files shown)


tujoworker (changes requested):
I found two blocking timeout-lifecycle issues:

1. The timeout only clears `isPending`; the original Promise remains connected to the section. If it resolves after recovery, it still switches the section to view mode after the user has resumed editing. If the user retries first, a late rejection from the older operation can also clear the newer operation’s pending state. I reproduced the late-resolution path. Give each save an operation token and ignore settlement from timed-out or superseded operations; invalidate and clear it on unmount as well.

2. The timeout path does not set `restoreFocusRef`, so when disabling the Done button moved focus to the document body, recovery leaves keyboard focus on the body. I reproduced this with the existing browser-focus simulation. Treat timeout recovery like rejection for focus restoration.

Please add regressions for late settlement/retry isolation and timeout focus restoration. This should also be rebased or stacked after #9252 so a configured `asyncSubmitTimeout` is inherited through `Form.Isolation`.

tujoworker:
Rebased on `main` and fixed in `e6163d9bddf`. Each async save now has its own operation token; timeout and unmount invalidate it, stale settlement cannot affect a retry, and timeout recovery restores focus only when focus fell back to the page body. Regression coverage now includes late resolution, stale rejection during retry, unmount, and timeout focus recovery.

tujoworker (approved):
The lifecycle and focus regressions are fixed now.

PR #9251, opened 2026-09-03
chore(Release): hide test commits from release notes

tujoworker:
Test-only commits are internal maintenance and add noise to user-facing release notes. Hide the test commit type from generated notes while keeping it recognized by the release configuration.


langz (approved):
Thanks!

PR #9252, opened 2026-09-03
fix(Forms): recover a stuck field state via asyncSubmitTimeout

langz:
## Problem

A field that waits for an async `onChange`, `onChangeValidator` or `onBlurValidator` sets a **per-field** pending state, and that state is what gates the form submit (`hasFieldState('pending')`). It was only cleared once the Promise settled.

So a Promise that never settles (a consumer bug, but a plausible one — a request that never resolves, a forgotten `resolve`) left the form permanently broken:

- a never-settling `onBlurValidator` left the input **disabled forever**
- a never-settling async `onChange` left the form **unsubmittable forever**, while the submit button re-enabled itself after `asyncSubmitTimeout` — so the primary action of the form silently did nothing

`asyncSubmitTimeout` only cleared the **form-level** state, never the per-field one.

## Fix

Give the per-field pending state the same `asyncSubmitTimeout` deadline that `Form.Handler` already applies to its own async submit.

All three consumers write that state through a single funnel (`setFieldState` in `useFieldError`), so one deadline covers async `onChange`, `onChangeValidator` and `onBlurValidator` — no scattered `setTimeout`s. The timer is replaced whenever a new state arrives and cleared on unmount.

## Note on behaviour

`asyncSubmitTimeout` now also limits how long a **field** waits, so it needs to be longer than the slowest async field work. One existing test set it to `1` ms purely to cut the submit indicator short; it now waits for the form to re-enable its fields instead of racing that timeout (it passes both before and after this change).

A validator that never settles still cannot let the form submit — a re-validation on submit starts the same never-settling work again. That is intentional: the fix removes the stale-state deadlock, it does not fabricate a validation verdict.

## Tests

`hooks/__tests__/useFieldError.test.tsx` — verified that the three new cases **fail without the fix** and pass with it:

- never-settling `onBlurValidator` → input re-enabled after the deadline
- never-settling async `onChange` → the form submits again
- never-settling `onChangeValidator` → the pending state is cleared
- the timer is cleared on unmount

Positive controls (pass before and after, so the tests are not vacuous):

- an async validator that settles before the deadline still reports its error and still blocks submit
- a validator still in flight keeps the field pending until it resolves

`useFieldError` backs every field, so the full `extensions/forms` suite was run: **159 files, 4765 tests passing**. Typecheck, ESLint and Prettier are clean.

## Related

- #9241 (`Form.Section` `onDone`) uses the same deadline for the section's own pending state.
- A never-settling `Field.Upload` `fileHandler` bypasses this funnel and is handled separately.
- The Bring connectors have no request timeout, but a hung request now recovers through this field deadline, so a second timeout there would be redundant.


tujoworker:
Rebased onto the latest `main` and pushed a follow-up fix. A field timeout now cancels any submit waiting on that field instead of resuming while async work may still be running, and the configured timeout now also reaches fields inside `Form.Isolation`. Regression tests cover the timeout race, normal completion, unmount cleanup, and the default timeout.

tujoworker (approved):
Reviewed the rebased revision and timeout follow-up. Timed-out fields now cancel deferred submission safely, `Form.Isolation` inherits the timeout, and the regression tests cover the race and timer lifecycle. Looks good.

PR #9254, opened 2026-09-03
fix(Field.Upload): recover a stuck file state via asyncSubmitTimeout

langz:
## Problem

`Field.Upload` sets two states before awaiting `fileHandler`: a per-file `isLoading` flag and a pending field state on the data context. Both were only cleared once the Promise settled (in a `finally`).

So a `fileHandler` that never settles left:

- the file **stuck in its loading state**, with its **delete button disabled forever**
- the field pending, which is what gates the form submit, so the form was **never submittable again** (verified: `onSubmit` called 0 times, submit button showing no sign of the problem)

## Fix

Give both the same `asyncSubmitTimeout` deadline that `Form.Handler` already applies to its own async submit: on expiry, clear the pending field state and turn off `isLoading` for the files this upload set loading.

`Field.Upload` writes the field state directly on the data context instead of going through the `useFieldError` funnel, so it needs its own deadline rather than being covered by the funnel fix.

Each upload keeps **its own** timer (in a `Set` cleared on unmount), so a later upload that succeeds does not cancel the recovery of an earlier one that never settles. Only the files that this upload set loading are touched, so a concurrent async delete keeps its own loading state.

## Tests

`Field/Upload/__tests__/Upload.test.tsx` — verified that the two new failure cases **fail without the fix** and pass with it:

- never-settling `fileHandler` → the delete button is re-enabled and the form submits again
- two uploads where the first never settles and the second succeeds → the first file still recovers (this one also **fails** if the implementation uses a single shared timer, which is why each upload keeps its own)

Positive control (passes before and after, so the tests are not vacuous):

- a `fileHandler` that settles before the deadline is not cut off, and the file stays loading until it resolves

Full `extensions/forms` + `components/upload` suites: **4957 tests passing**. One unrelated `Connectors/Bring` postal code test is flaky under parallel load — it fails on unmodified `main` in the same way (verified across repeated runs), so it is not caused by this change. Typecheck, ESLint and Prettier are clean.

## Out of scope

The base `Upload` component has the same shape for async `onFileDelete` and `onFileClick` in `components/upload/UploadFileList.tsx`, but it has no `Form.Handler` context to take the deadline from, so it needs a different mechanism and is left for a separate change.

## Related

- #9252 — the same deadline for the per-field pending state set through `useFieldError` (async `onChange` and validators).
- #9241 — the same deadline for `Form.Section`'s async `onDone`.


tujoworker (changes requested):
I found two blocking races in the timeout recovery:

1. Every `fileHandler` invocation uses the same field-level pending and `enableAsyncMode` flags. If upload A is still pending and upload B settles, B’s `finally` clears both flags. A submit then proceeds while A is unresolved. I reproduced this: `onSubmit` ran with A still carrying `isLoading: true`. Track active handler operations and clear the shared state only when none remain.

2. The timeout makes a file deletable but does not stop its handler. If the user deletes that file and the handler later resolves, the stale `newFilesLoading` result inserts the deleted file again. I reproduced the list changing from zero files back to one. Ignore late results for files that are no longer present, or otherwise invalidate the timed-out operation.

After #9252 is merged, the timeout path should also clear the pending state with `cancelPendingSubmit: true`; otherwise a submit attempted before the timeout resumes automatically while the handler is still running. Please add regression coverage for all three paths.

tujoworker (approved):
Reviewed the updated implementation at `5585002b6c3`. The concurrency, deferred-submit, and late-result races are covered and fixed, and the complete Forms suite passes. The code is production-ready; merge after #9252 and the new CI run.

tujoworker:
Fixed the three review findings and stacked this PR on #9252. Concurrent handlers now keep the field pending until every operation finishes or times out, timeouts cancel deferred submits, and expired handlers cannot apply late results. The complete Forms suite passes with 4,775 tests. Merge this after #9252.

PR #9256, opened 2026-09-03
fix(Form.Section): keep fields disabled during pending saves

langz:
An explicit `disabled={false}` could re-enable fields while an asynchronous section save was pending, allowing the displayed value to diverge from the value being saved. Make the pending lock authoritative through nested field providers, then restore normal provider precedence when the save settles.


tujoworker:
Rebased onto the latest `main` and pushed a follow-up fix. Pending UI locks are now separate from a field’s logical disabled state, so validation and errors stay intact. Regression tests cover nested providers, explicit disabled overrides, rejected section saves, and interactive submit confirmation.

tujoworker (approved):
Reviewed the rebased revision and follow-up fix. Pending UI locks no longer change logical field state or validation, and the regression tests cover nesting, overrides, rejected saves, and submit confirmation. Looks good.

PR #9258, opened 2026-09-03
fix(Forms): treat a validator that returns a Promise as async

langz:
## Problem

Whether a validator is asynchronous was decided from the **function** (`isAsync`), not from what it **returned**. A validator that returns a Promise without being declared `async` — for example `(value) => fetch(...).then(...)`, or any wrapper that returns a Promise — fell between the two paths, and the two validators failed in **opposite** directions.

Measured on `main`, with a validator returning a Promise that settles after 50 ms, submitting while it is in flight:

| validator | shape | `onSubmit` on `main` | `onSubmit` with this change |
|---|---|---|---|
| `onBlurValidator` | plain fn → Promise, **returns an error** | **1 — submitted an invalid value** | 0 |
| `onBlurValidator` | plain fn → Promise, no error | 1 | 1 |
| `onChangeValidator` | plain fn → Promise, returns an error | 0 | 0 |
| `onChangeValidator` | plain fn → Promise, **no error** | **0 — the form could never be submitted** | 1 |
| either | declared `async` | unchanged | unchanged |
| either | synchronous | unchanged | unchanged |

- `onBlurValidator` was never marked as validating, so nothing held the submit back and the form **submitted unvalidated**.
- `onChangeValidator` *was* marked as validating (it already detects the Promise at runtime), so the submit was correctly held back — but the data context never learned that the field validates asynchronously, so the queued submit was **dropped instead of continued** once the validation finished. Clicking submit again just repeated it, so the form could not be submitted at all.

There is no public escape hatch for this: the `hasAsyncBehavior` marker used by the Bring connectors is internal and not exported.

## Fix

1. In the onBlur path, detect the Promise on the **returned value** instead of on the function — the same runtime check the onChange path already does — so the field is marked as validating and the submit waits.
2. When a validator is first seen returning a Promise, tell the data context the field validates asynchronously (`enableAsyncMode`, the same field internal `Field.Upload` already sets). That is what arms the "continue the submit once the pending state resolves" path, so the queued submit resumes instead of being dropped.

This makes the documented contract — *"the function can be either asynchronous or synchronous"* — actually hold for both validators.

## Tests

`hooks/__tests__/useFieldValidation.test.tsx` — verified that the four new cases **fail on `main`** and pass with the change:

- `onBlurValidator` Promise with an error → the form does not submit
- `onBlurValidator` Promise with no error → the form does submit
- `onChangeValidator` Promise with no error → the form does submit
- a Promise-returning `onBlurValidator` shows the field as validating while it is pending

Positive controls (pass before and after, so the tests are not vacuous):

- an `onBlurValidator` declared `async` behaves identically
- a synchronous `onBlurValidator` still blocks and then releases the submit
- an `onChangeValidator` Promise with an error still blocks the submit

One existing test, `useFieldProps` → "should support mixed sync and async validators", typed into the field while an async validator was running. That worked only because the field was not marked as validating — the same gap this change closes. It now waits for the field to be enabled before changing the value, and it passes both **with and without** the source change.

Full `extensions/forms` suite: **159 files, 4767 tests passing** (two clean runs). A third run hit an unrelated `ValueBlock` help test that passes in isolation; this environment has a few load-sensitive tests. Typecheck, ESLint and Prettier are clean, with no new `exhaustive-deps` warnings.

## Out of scope

While measuring this I found a separate pre-existing issue: when a validator returns a list containing an async validator and validation re-enters while the first run is still in flight, the call-stack de-duplication can skip the async member, so it resolves as "no error". That is a different root cause and is left alone here.

## Related

- #9252 — recovers a stuck per-field pending state via `asyncSubmitTimeout`.
- #9254 — the same for a never-settling `Field.Upload` `fileHandler`.

### DIFF
--- a/packages/dnb-eufemia/src/extensions/forms/DataContext/Provider/Provider.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/DataContext/Provider/Provider.tsx
@@ -1485,12 +1485,14 @@ export default function Provider<Data extends JsonObject>(
         customStatus: undefined,
       })
 
-      const asyncBehaviorIsEnabled =
+      const shouldEnableAsyncBehavior = () =>
         (skipErrorCheck
           ? enableAsyncBehavior
           : // Don't enable async behavior if we have errors, but when we have a pending state
             !hasErrors() || hasFieldState('pending')) &&
         (enableAsyncBehavior || hasFieldWithAsyncValidator())
+
+      let asyncBehaviorIsEnabled = shouldEnableAsyncBehavior()
 
       if (asyncBehaviorIsEnabled) {
         setFormState('pending')
@@ -1512,6 +1514,12 @@ export default function Provider<Data extends JsonObject>(
             }
           }
         }
+      }
+
+      // A validator can reveal its async behavior only after it has been called.
+      if (!asyncBehaviorIsEnabled && shouldEnableAsyncBehavior()) {
+        asyncBehaviorIsEnabled = true
+        setFormState('pending')
       }
 
       let result: EventStateObject | undefined
--- a/packages/dnb-eufemia/src/extensions/forms/DataContext/Provider/Provider.tsx
+++ b/packages/dnb-eufemia/src/extensions/forms/DataContext/Provider/Provider.tsx
@@ -1485,14 +1485,16 @@ export default function Provider<Data extends JsonObject>(
         customStatus: undefined,
       })
 
-      const shouldEnableAsyncBehavior = () =>
+      const shouldEnableAsyncBehavior = (hasAsyncValidator: boolean) =>
         (skipErrorCheck
           ? enableAsyncBehavior
           : // Don't enable async behavior if we have errors, but when we have a pending state
             !hasErrors() || hasFieldState('pending')) &&
-        (enableAsyncBehavior || hasFieldWithAsyncValidator())
+        (enableAsyncBehavior || hasAsyncValidator)
 
-      let asyncBehaviorIsEnabled = shouldEnableAsyncBehavior()
+      const hadAsyncValidator = hasFieldWithAsyncValidator()
+      let asyncBehaviorIsEnabled =
+        shouldEnableAsyncBehavior(hadAsyncValidator)
 
       if (asyncBehaviorIsEnabled) {
         setFormState('pending')
@@ -1517,7 +1519,11 @@ export default function Provider<Data extends JsonObject>(
       }
 
       // A validator can reveal its async behavior only after it has been called.
-      if (!asyncBehaviorIsEnabled && shouldEnableAsyncBehavior()) {
+      if (
+        !asyncBehaviorIsEnabled &&
+        !hadAsyncValidator &&
+        shouldEnableAsyncBehavior(hasFieldWithAsyncValidator())
+      ) {
         asyncBehaviorIsEnabled = true
         setFormState('pending')
       }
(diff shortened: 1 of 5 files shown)


tujoworker (changes requested):
The first submit is still dropped when async behavior is discovered during that submit. `handleSubmitCall` calculates `asyncBehaviorIsEnabled` before it calls the field validators. For an untouched field or a programmatic submit, `startOnBlurValidatorProcess` detects the returned Promise and marks the field pending, but the current submit still has `asyncBehaviorIsEnabled === false`. It therefore neither awaits the callback nor registers `onSubmitContinueRef`. When a valid Promise resolves, `onSubmit` is not called until the user submits again.

I reproduced this on `384eb39e5fdebc918606e78fcc491cfe9d9daf1e` with an untouched `Field.String`, a non-`async` `onBlurValidator` returning a delayed `Promise<undefined>`, and `fireEvent.submit(form)`: expected one `onSubmit` call, got zero. The existing tests type first, so they discover async behavior before submitting and miss this path.

Please make the current submit continue when a validator first enters the pending state, and add this case as a regression test.

langz:
> The first submit is still dropped when async behavior is discovered during that submit. `handleSubmitCall` calculates `asyncBehaviorIsEnabled` before it calls the field validators. For an untouched field or a programmatic submit, `startOnBlurValidatorProcess` detects the returned Promise and marks the field pending, but the current submit still has `asyncBehaviorIsEnabled === false`. It therefore neither awaits the callback nor registers `onSubmitContinueRef`. When a valid Promise resolves, `onSubmit` is not called until the user submits again.
> 
> I reproduced this on `384eb39e5fdebc918606e78fcc491cfe9d9daf1e` with an untouched `Field.String`, a non-`async` `onBlurValidator` returning a delayed `Promise<undefined>`, and `fireEvent.submit(form)`: expected one `onSubmit` call, got zero. The existing tests type first, so they discover async behavior before submitting and miss this path.
> 
> Please make the current submit continue when a validator first enters the pending state, and add this case as a regression test.

Thanks, that sounds promising. 
Will take a stab at it tomorrow.

tujoworker:
Rebased onto `main` after #9261. The first-submit path now re-checks async behavior after field validation starts, so a Promise first discovered during submit resumes that same submit when validation completes. Added regression coverage for the untouched-field case.

tujoworker (changes requested):
The runtime Promise detection and first-submit continuation work, including an immediately resolved Promise. One blocking regression remains in the Provider re-check.

`shouldEnableAsyncBehavior()` is evaluated again after validation, but it does not distinguish newly discovered async behavior from an already-known async validator that was initially suppressed because the form had another error. Once that validator sets a pending field state, the `hasFieldState("pending")` exception flips the result to true and enables the deferred-submit path anyway. This changes existing Wizard validation timing/behavior and deterministically fails `should support navigating back and forth with async validators` at line 697 on this head; the same test passes on `main` and at the first PR commit. GitHub CI shows the same failure.

Please gate the post-validation activation on async capability actually transitioning from undiscovered to discovered during those callbacks, rather than re-running the full predicate. Keep the untouched-field regression, and ensure the existing Wizard test passes unchanged. The dependency-audit failure is a separate registry timeout.

tujoworker:
Fixed in `29954e2f52e`. Deferred submit continuation now starts after validation only when the validator was newly discovered as async. This keeps first-submit support for Promise-returning validators without changing Wizard behavior for validators that were already known async.

tujoworker (approved):
Approved at `29954e2f52e`. The regression is fixed and the code is production-ready; new CI is still running.

PR #9264, opened 2026-09-04
chore(ci): make dependency audit resilient to npm registry latency

langz:
## Problem

The shipped-dependency audit gate (`audit:ci`) has been failing in CI with:

```
➤ YN0001: RequestError: Timeout awaiting 'socket' for 60000ms
```

This is **not** a real advisory and **not** specific to this branch — every `verify` run across all branches has failed on the audit job since ~2026-09-04 04:44 UTC, while the identical commit passed at 2026-09-03 19:41 UTC.

## Root cause: npm advisory-endpoint latency (npm/cli#9950)

`audit:ci` runs `yarn npm audit --recursive --environment production --severity high`, a **POST** to `registry.npmjs.org` at `/-/npm/v1/security/advisories/bulk`. Per **npm/cli#9950**, that endpoint is under elevated latency: the POST can take minutes but still succeed (HTTP 200).

Yarn's `httpTimeout` defaults to **1m** and is the only timeout it sets on the request (`timeout: { socket }`), so it kills the slow-but-successful request at 60s. got does not retry it internally because **POST is absent** from got's retryable-methods list (`GET/PUT/HEAD/DELETE/OPTIONS/TRACE`) — raising `httpRetry` cannot change that. (Verified against the committed `yarn-4.18.0.cjs`; Yarn's own error text even says *"can be increased via httpTimeout"*.)

This is **not** #9255 (Yarn 4.16 → 4.18 / Node 24.16 → 24.20): the same commit passed while the endpoint was healthy.

## Fix

A shared composite action `.github/actions/audit` (mirroring `.github/actions/setup`):

1. **Wait out the latency.** Set `YARN_HTTP_TIMEOUT=3m` so a slow-but-successful audit completes instead of dying at 60s. Also set at the `verify` audit **job** level so the non-gating SBOM smoke step (which also calls `yarn npm audit`) is covered.
2. **Retry genuine transport blips** — as a backstop — on exactly the failures got itself treats as transient: its default `retry.errorCodes` (network errors) and `retry.statusCodes` (HTTP 408/413/429 + 5xx), with linear back-off, up to 2 attempts (each attempt can now wait out the 3-minute timeout, so the PR gate's worst case stays inside its 25-minute job limit).
3. **Fail fast, never mask.** A real `>=high` advisory (the security gate), an auth/config HTTP error (400/401/403/404), or a lockfile error fails immediately with its original exit code.
4. **Raise the capped job timeouts** (verify audit 10→25m, the two Lambda test jobs 10→20m) so a slow-but-passing audit fits.

Wired into `verify.yml`, `release.yml`, `mcp-lambda.yml`, `analytics-lambda.yml`.

## Verification

- All five YAML files parse; the four workflows resolve to the action with the correct workspace lists; every job checks out and sets up Node/Yarn before the audit step.
- The action's exact script, run under `bash -eo pipefail` against a stubbed audit, passes **18 scenarios**: success (single/multi), socket-timeout recovery, real-advisory fail-fast, persistent-outage exhaustion, HTTP 503/429/500 retried vs 404/401/403 failed fast, and the input guards.
- Root cause and the "POST isn't retried" / "raise httpTimeout" facts verified directly against the Yarn bundle.

## Caveat

This mitigates an **external** npm issue (#9950). If npm resolves the endpoint latency, audits return to sub-second. During a severe npm outage no client-side change can *guarantee* a pass — at that point failing the security gate is the correct outcome.

### DIFF
--- a/.github/actions/audit/action.yml
+++ b/.github/actions/audit/action.yml
@@ -1,0 +1,118 @@
+name: Audit dependencies
+description: >-
+  Run "yarn workspace <ws> audit:ci" for one or more workspaces, retrying only
+  on transient npm-registry transport failures. A real advisory (a >=high
+  finding — the actual security gate) or a lockfile/config error fails
+  immediately, so genuine findings are never masked or delayed.
+
+inputs:
+  workspaces:
+    description: >-
+      Newline-separated list of workspace names to audit, e.g.
+      "@dnb/eufemia". Each is passed to "yarn workspace <ws> audit:ci".
+    required: true
+  attempts:
+    description: >-
+      Maximum attempts per workspace before a transient (network/transport)
+      failure is treated as a registry outage and fails the job.
+    required: false
+    default: '3'
+
+runs:
+  using: composite
+  steps:
+    - name: Audit dependencies (retry transient registry failures)
+      shell: bash
+      env:
+        AUDIT_WORKSPACES: ${{ inputs.workspaces }}
+        AUDIT_ATTEMPTS: ${{ inputs.attempts }}
+      run: |
+        # GitHub invokes this with `bash -eo pipefail`. We deliberately turn off
+        # `-e`: `yarn ... audit:ci` exits non-zero both for a real advisory AND
+        # for a transient network failure, and we must CAPTURE that exit to tell
+        # the two apart rather than aborting on it.
+        set +e -u -o pipefail
+
+        # Why re-issue the whole command instead of raising retries/timeout:
+        # `yarn npm audit` sends a POST to the npm registry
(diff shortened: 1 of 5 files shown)


tujoworker:
Closing in favor of #9265. The npm incident evidence shows that the request can stay healthy for about five minutes before npm responds. Retries can help short transport failures, and this PR recovered two calls that way, but it still failed when the upstream stayed slow and adds a custom error classifier across four workflows. #9265 keeps the audit fail-closed and handles this case with a step-local timeout instead. A manual job rerun after npm recovers remains available for broader registry outages. Thanks for investigating this.

PR #9265, opened 2026-09-04
chore(CI): allow slow dependency audits to finish

tujoworker:
The npm audit endpoint can keep a healthy connection open for about five minutes before responding, as [documented in npm/cli#9950](https://github.com/npm/cli/issues/9950#issuecomment-5536505917).

Allow every CI audit request to wait six minutes instead of Yarn's default 60 seconds. Extend the affected job limits so sequential audits still have time to finish. Audits remain fail-closed when npm does not respond within six minutes or reports an advisory.


langz (approved):
🤞

PR #9271, opened 2026-09-04
fix(Tooltip): prevent flicker during rapid pointer re-entry

tujoworker:
Follow-up to #9248.\n\nA quick pointer leave and re-entry could still start the hide transition, especially with delays of 100 ms or less. Tooltip now keeps its open state until the hide delay expires and transfers hover ownership between the trigger and overlay. Returning to the trigger cancels both close timers and releases overlay ownership, including when the pointer crosses the physical gap.\n\nIf re-entry happens after hiding has started, the existing tooltip returns to its settled position without replaying the 200 ms entrance motion.

### DIFF
--- a/packages/dnb-eufemia/src/components/tooltip/TooltipWithEvents.tsx
+++ b/packages/dnb-eufemia/src/components/tooltip/TooltipWithEvents.tsx
@@ -13,7 +13,12 @@ import {
   useRef,
   useState,
 } from 'react'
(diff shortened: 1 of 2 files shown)


langz (approved):
Good good 🥇 

Starts to be quite some code related to this, but I guess it's fine and needed 🤝

langz (approved):
Works well on my phone

PR #9273, opened 2026-09-04
chore(CI): run the dependency audit only when the dependency tree changes

langz:
Moves the `audit` job out of `verify.yml` into a new `dependency-audit.yml`
that runs when the dependency tree or audit configuration changes, plus once
nightly and on manual dispatch.

## Why

The audit's result depends on the dependency tree and the audit configuration —
not on source changes. Measured on `main` over the last 60 days (327 commits):

| | commits | share |
| --- | --- | --- |
| Touch none of the trigger paths | 258 | **78.9%** |
| Touch the lockfile, a manifest, `.yarnrc.yml` or `release.yml` | 69 | 21.1% |

So four in five pushes currently re-audit a byte-identical dependency tree.

The nightly run covers the inverse case, which per-push triggering handles
badly: advisories are disclosed against dependencies that never change. Today
that surfaces on whichever branch happens to push next, at a random time.
Nightly on `main` makes detection deterministic and puts the failure where it
belongs. `17 4 * * *` UTC lands ~06:17 Oslo, so a red audit is waiting when
people arrive rather than interrupting the day.

## What is not being claimed

The `audit` job was the longest job in `verify.yml` when I measured it (5.4 min
median, 14 min worst) — but that was **during an npm incident**, and
`YARN_HTTP_TIMEOUT` is only a ceiling. **This PR is not a steady-state CI
speedup.** The job's own run on this branch took **46 seconds** now that npm has
recovered. What it buys is incident resilience — a degraded registry no longer
stalls every push — and better new-advisory coverage. In steady state it removes
about 45 seconds from four in five pushes.

## Why a separate workflow rather than a condition

The alternative was keeping the job in `verify.yml` behind a job-level `if:`.
That needs an extra job or a third-party action to compute changed files, where
`paths:` is native, evaluated server-side before any runner is claimed. It also
would not give us the nightly run: `verify.yml` has no `schedule:`, and adding
one there would run lint, tests, type-checks and builds nightly too.

## The job moves unchanged

Same steps, same `timeout-minutes: 30`, same secret-blanking `env:`, and the
same `Audit dependencies` job name so the check name is stable. Only `needs:
setup` is dropped, since the new workflow is standalone.

Verified mechanically, not by eye: both files are parsed and the two `audit` job
objects compared field-by-field — identical apart from `needs:`. Every other job
in `verify.yml`, and its trigger block, is asserted untouched against
`origin/main`.

Moving the whole job is deliberate rather than cherry-picking the audit step:
all three steps (`yarn dedupe --check`, the audit, the SBOM release smoke test)
are dependency and release-tooling concerns with the same input set. A side
benefit of running the SBOM smoke test nightly is that it also catches the
pinned `@cyclonedx/yarn-plugin-cyclonedx` version becoming unavailable, which
would otherwise surface mid-release.

## Trigger paths

Two additions beyond the obvious lockfile and manifests:

- **The root manifest is listed explicitly, as well as via the recursive glob.**
  GitHub documents `**` as "zero or more of any character" but does not state
  whether a leading `**/` also matches a root-level file. The root manifest
  carries `resolutions` and `packageManager`, so missing it would be the worst
  possible gap. Listing both spellings is correct under either reading and costs
  one line.
- **`.yarnrc.yml`** — `npmAuditIgnoreAdvisories` lives there. It is where
  `docs/vulnerability-management.md` tells you to add an advisory exception, so
  it directly changes what the audit reports. Without it, adding an exception
  would produce no re-run and no confirmation that it worked.

I checked GitHub's server-side path matcher empirically rather than trusting the
docs, using the existing `analytics-lambda.yml` filter as a control: over 194
`main` commits, 16 touched its filtered paths and 17 runs fired, with **zero
missed path changes and zero spurious runs** (the 17th is a window-boundary
artefact — its commit lands 4 s before the first run's timestamp).

## Costs and limits, stated up front

- **One duplicate install on lockfile changes.** This workflow runs the shared
  setup action itself, concurrently with `verify.yml`'s `setup` job. When
  `yarn.lock` changes, both miss the cache and both install; one save wins and
  the other is dropped. Same content, so it is harmless, but it is roughly one
  extra minute of a `ubuntu-latest` runner on the 21% of pushes that trigger
  this workflow. On the other trigger paths the cache hits and nothing installs
  — confirmed in this branch's run: `Cache hit occurred on the primary key […],
  not saving cache`.
- **Fork pull requests are unaffected, meaning still not covered.** Fork pushes
  do not fire the base repository's `push` workflows, so neither `verify.yml`
  nor this workflow audits them. No regression, but worth stating in a
  security-adjacent change rather than leaving implied.
- **The nightly run's failure notification goes to one person.** GitHub sends
  scheduled-run notifications to whoever created the workflow, and transfers
  them to whoever last edits the cron
  ([docs](https://docs.github.com/en/actions/concepts/workflows-and-actions/notifications-for-workflow-runs)).
  That is why this PR also adds a **README badge** for the workflow, so the
  nightly has a shared signal and not just a private email. Noted next to the
  cron so it is not tidied away later.

## Safety

- **`Audit dependencies` is not a required status check.** Read from branch
  protection on `main`: the required contexts are `Run build tests`,
  `Run analysis (javascript)`, `Run e2e tests`, `Run visual-regression tests`,
  `CodeQL`, `bundlewatch`, `Build and deploy (branch preview)`,
  `Validate PR title`, `Run lint`, `Run type checks`, `Run tests` and
  `security/snyk`. So no branch-protection impact.
- **`release.yml` keeps its own audit** of `@dnb/eufemia` before publish, so the
  published library stays gated regardless of this workflow.

One consequence worth stating plainly: **this change and "make the audit a
required check" are mutually exclusive.** A required check that does not run on
every push leaves the PR permanently pending. If you would rather have the audit
blocking, this PR is the wrong direction and should be closed.

## Documentation

`docs/vulnerability-management.md` pointed at `verify.yml`, so it had to be
updated. While in those sentences I also fixed three things I did not want to
re-assert or leave misleading:

1. It called `Audit dependencies` a **required** check. Branch protection does
   not list it — wrong before this PR.
2. It said the gate covers **two** workspaces. It covers three: `@dnb/eufemia`,
   `eufemia-mcp-lambda` and `eufemia-analytics`, all with an identical
   `audit:ci`.
3. "the audit and lockfile-dedupe checks **always run**" — true before, not
   after. Reworded to the actual intent: no flag or label bypasses the gate for
   a dependency change.

The cadence is now recorded in the control-inventory table the same way the
CodeQL row records its own, since that table is described in the doc as the
authoritative list of what runs today.

Happy to split the doc corrections into a separate PR if you would rather keep
this one purely mechanical.

`release.yml`'s "keep it in sync with `verify.yml`" comment now points at
`dependency-audit.yml`.

## Conflict with #9272

Closed PR #9272 edits this same audit step, so it conflicts. This PR is built
from `main` with the job moved unchanged, so it presumes nothing about whether
that gate is wanted. If #9272 is reopened it needs a small rebase to land in the
new file instead.

## Verification

- Both workflows and `release.yml` parse; the trigger block is asserted against
  the parsed object rather than eyeballed.
- All 16 tracked manifests match the path globs (same minimatch semantics),
  including the root one; negative controls confirm ordinary source edits,
  `docs/**` and `verify.yml` do **not** match.
- `prettier --check` passes on all five touched files. Note `docs/**` and
  `.github/**` are covered by no CI prettier check, so this was done locally.
- The nightly cron does not collide with the only other cron in the repo
  (`44 3 * * 5`, CodeQL).
- **Positive trigger test:** a commit touching
  `.github/workflows/dependency-audit.yml` — a trigger path — ran
  `Dependency audit`, which passed in 46 s, while `Verify` reported 7 jobs with no
  `Audit dependencies` among them. The check name still appears on the PR,
  reported by the new workflow.
- **Negative trigger test.** The last commit here touches only
  `docs/vulnerability-management.md`. Eight workflows fired on it — Verify,
  CodeQL, E2E, Visual Regression, Dev Preview, Deploy Preview, PR Checks,
  Consumer smoke — and `Dependency audit` did **not**. That is the behaviour this
  PR exists for, and it is now observed on this branch rather than inferred from
  the `analytics-lambda.yml` control.
- **Non-push context.** I dispatched the workflow manually on this branch to
  exercise the path the nightly `schedule` run will take, which has no push
  payload. All six steps passed in 41 s, including the SBOM smoke test. Worth
  knowing: `workflow_dispatch` worked even though the file is not on `main` yet.
- My first verification harness compared against `HEAD`, which after committing
  is the changed file — so it was comparing the change to itself. Rebased every
  assertion onto `origin/main` and re-ran; that is what the results above are.

### DIFF
--- a/.github/workflows/dependency-audit.yml
+++ b/.github/workflows/dependency-audit.yml
@@ -1,0 +1,159 @@
+name: Dependency audit
+
+# Audits the shipped dependency tree, checks the lockfile is deduplicated, and
+# smoke-tests the SBOM + audit generators that release.yml depends on.
+#
+# Split out of verify.yml because its result depends on the dependency tree and
+# audit configuration, not on source changes: roughly four in five commits touch
+# none of the paths below, so running it per push re-audits a byte-identical
+# tree. The daily schedule covers the other half of the problem — advisories are
+# disclosed against dependencies that never change, and a nightly run reports
+# those on main rather than reddening whichever branch happens to push next.
+#
+# release.yml keeps its own audit of the published library, so publishing stays
+# gated regardless of this workflow.
+#
+# No workflow-level `env:` on purpose: these steps only install, dedupe-check,
+# audit and generate an SBOM — none need repository secrets (least privilege).
+
+on:
+  push:
+    branches:
+      - '**'
+      - '!**--skip-ci'
+      - '!**--visual-reports'
+      - '!wip/**'
+      - '!experiments/**'
+    paths:
+      - 'yarn.lock'
+      # Both spellings on purpose. GitHub documents `**` as "zero or more of any
+      # character", but does not state whether a leading `**/` also matches a
+      # root-level file — and the root manifest is the one carrying
+      # `resolutions` and `packageManager`, so it must not be missed.
+      - 'package.json'
+      - '**/package.json'
+      # `npmAuditIgnoreAdvisories` (advisory exceptions) and `npmMinimalAgeGate`
+      # live here, so this file changes what the audit reports.
+      - '.yarnrc.yml'
+      - '.github/workflows/dependency-audit.yml'
+      # The SBOM smoke test below exists to mirror release.yml and pins a
+      # generator version that must match it, so changes there must re-run this.
+      - '.github/workflows/release.yml'
+  schedule:
+    # Daily, to catch advisories disclosed against an unchanged tree.
+    - cron: '17 4 * * *'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  audit:
+    name: Audit dependencies
+
+    runs-on: ubuntu-latest
+
+    timeout-minutes: 30
+
+    # This job runs untrusted third-party code (the CycloneDX generator via
+    # `yarn dlx` in the SBOM smoke test below). This workflow has no
+    # workflow-level `env:` (see the note at the top of this file), so no secrets
+    # reach this job today; these blanks are defense-in-depth in case a
+    # workflow-level `env:` is ever added, so the untrusted code still cannot
+    # read them.
+    env:
+      GH_EMAIL: ''
+      GH_NAME: ''
+      GH_TOKEN: ''
+      FIGMA_TOKEN: ''
+      FIGMA_ICONS_FILE: ''
+
+    steps:
+      - name: Git checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+        with:
+          persist-credentials: false
+
+      - name: Setup dependencies
+        uses: ./.github/actions/setup
+        with:
+          cache-version: ${{ secrets.CACHE_VERSION }}
+
+      - name: Check lockfile is deduplicated
+        # Guards against a plain `yarn install` reintroducing duplicate (and
+        # potentially vulnerable) transitive versions. Run `yarn dedupe` locally
+        # and commit yarn.lock if this fails.
+        run: yarn dedupe --check
+
+      - name: Audit dependencies
+        # Audit the shipped (production) dependency tree of everything we ship:
+        # the @dnb/eufemia npm library, the Eufemia MCP Lambda and the
+        # eufemia-analytics Lambda. All must pass; dev/build tooling advisories
+        # are tracked via Dependabot instead.
+        run: |
+          yarn workspace @dnb/eufemia audit:ci
+          yarn workspace eufemia-mcp-lambda audit:ci
+          yarn workspace eufemia-analytics audit:ci
+        env:
+          YARN_HTTP_TIMEOUT: '360000'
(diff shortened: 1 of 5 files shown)


langz:
Empirical confirmation of the caveat in the description, from this branch's own run: the relocated `Audit dependencies` job completed in **46 seconds** (12:12:56Z → 12:13:42Z).

That is the point I wanted to be honest about — the 5.4 min median I measured earlier was npm being degraded, not the audit's normal cost. npm has recovered and the job is back to well under a minute. So please read this PR as incident resilience plus nightly advisory coverage, **not** as a CI speedup. The steady-state time it removes from a push is about 45 seconds.

Also confirmed on this run: `Verify` now reports 7 jobs with no `Audit dependencies` among them, and the check name `Audit dependencies` still appears on the PR — reported by the new workflow instead.

tujoworker (approved):
Looks good. Running the audit nightly on main will surface newly disclosed advisories against an unchanged dependency tree, while the path-only push trigger keeps dependency and audit configuration changes covered. Including .yarn/** closes the remaining Yarn configuration and plugin gap.

PR #9274, opened 2026-09-04
fix(Upload): bound async onFileDelete and onFileClick

langz:
Upload's async file operations could leave a file stuck showing a spinner, with its delete button disabled and — inside a form — the field's validity pinned to it. This PR bounds those operations, and fixes four separate ways a file could end up stuck or reappear after deletion.

Everything here is `Upload`, `Field.Upload` and `Value.Upload`. 14 source files, 3 test files, 13 new tests.

## 1. An unbounded wait

A file waiting for an async `onFileDelete` or `onFileClick` shows a loading state and has its delete button disabled. Nothing bounded that wait, so a Promise that never settles left the file **permanently** stuck — no spinner ever stops, no retry is possible, and inside a form the file also carries the field's validity, so the form goes with it.

This is the same failure mode #9254 fixed for `fileHandler` and #9256 fixed for `Form.Section`'s Done button — just on the two remaining unbounded consumer Promises in Upload.

**Fixed with a deadline and an operation guard.** Exactly one of resolve, reject or timeout wins; a Promise settling after the deadline is ignored rather than acting on a file the user has since moved on from.

**A timed out deletion keeps the file** and shows an error message. It deliberately does *not* "resolve early": the resolve path is what removes the file, so resolving on timeout would delete a file the consumer never confirmed deleting. A timeout is the same event class as a rejection — the deletion did not complete — and the existing reject path already marks the file with an error, so the two now behave alike. A timed out click only stops the loading state; opening a file changes nothing, so there is nothing to report.

`Value.Upload` had the same gap in its local `loading` state, fixed the same way.

## 2. A Promise dropped unless the handler was declared `async`

Both callbacks are documented as `void | Promise<void>` (`types.ts`), but the implementation *predicted* which one it got from the function declaration, via `isAsync()`:

```ts
if (isAsync(onFileDelete)) {
  handleDeleteAsync(uploadFile)          // awaited
} else {
  onFileDelete({ fileItem: uploadFile }) // Promise dropped on the floor
  updateFiles(removeFile(uploadFile))    // file removed immediately
}
```

A handler that returns a Promise without being declared `async` — `onFileDelete={() => api.delete(id)}` — took the second branch. Its Promise was silently discarded and the file was removed from the list **before the consumer had done anything**. The public contract promised Promise support; the implementation only honoured it for the `async` keyword.

**Fixed by calling, then inspecting** — the result is checked instead of the declaration, matching `Form/Section/EditContainer/DoneButton.tsx`:

```ts
const result: unknown = onFileDelete({ fileItem: uploadFile })
if (!(result instanceof Promise)) { /* synchronous, unchanged */ }
```

## 3. Concurrent operations undoing each other

Both `Upload` and `Field.Upload` track the current file list in a `filesRef` that is only refreshed on render. Two operations settling in the same tick therefore both compute the next list from the same stale snapshot, and the second overwrites the first. That produces two distinct bugs, one in each file:

| | what the user sees |
|---|---|
| two deletions settle together | one deleted file **comes back** |
| two uploads settle together | one file **spins for good**, though its upload succeeded |

The second is the worse of the two. Because both uploads *succeeded*, both `asyncSubmitTimeout` deadlines from #9254 had already been cleared, so nothing remains to clear the stale flag: the spinner and the disabled delete button stay indefinitely, not for 30 seconds.

Both are fixed the same way — `filesRef` moves in step with `setFiles` instead of waiting for a render — and both were verified to fail on clean `origin/main`, so neither is a regression from this branch.

They are fixed here rather than deferred for two reasons. The deletion case is reached by a *second* class of handler once fix 2 lands (one returning a Promise without `async`), and shipping a change that widens the reach of a silent data-loss bug is not a good trade. The upload case is the identical bug one file over, so leaving it while fixing its sibling would be arbitrary.

## 4. An operation abandoned when the file list unmounts

The deadline work above initially cleared pending deadlines and emptied the operation set when the file list unmounted. That is the right instinct for component-local state, but these files live in a store that outlives the component, so it left a file marked as loading with nothing running to clear it. A remount inherits a spinner and a disabled delete button.

Measured against `origin/main`, reading the store through a subscriber that stays mounted:

| after the list unmounts | `main` | with the cleanup | now |
|---|---|---|---|
| the Promise resolves | file removed | stuck loading | file removed |
| the Promise never settles | stuck loading | stuck loading | recovered |

So the first row was a regression introduced within this PR, and the second is the very stuck file this PR exists to prevent — still stuck in the one case where the cleanup had cancelled the deadline. Removing the cleanup fixes both, and is a deletion rather than an addition. Writing to the store after unmount is what `main` already did on this path, so it is not a new risk, and the guard against two outcomes acting on one file is untouched.

`Value.Upload` deliberately keeps its cleanup: its loading state is local `useState`, so it disappears with the component and cannot be inherited.

## The deadline itself

`Field.Upload` and `Value.Upload` use the `asyncSubmitTimeout` of their `Form.Handler`, defaulting to the shared `DEFAULT_ASYNC_SUBMIT_TIMEOUT` from #9270. The base `Upload` keeps a constant of its own, and its comment records that the two are kept in step by hand.

Worth being precise about why, because an earlier revision of this description had it wrong: `components/` *can* import that constant. `yarn test:deps` only forbids circular dependencies and `defaults.ts` imports nothing, so wiring it up passes the check — I tried it. The reason to keep them separate is a layering convention rather than an enforced rule: exactly one non-test file under `components/` imports from `extensions/`, and that one needed its own carve-out in `.dependency-cruiser.js`.

Per review feedback, the base component's deadline is **not** a public prop. It is `_asyncFileOperationTimeout`, marked `@internal` and absent from the properties table, following `_omitInputShellClass` in `Input` and `DatePicker`. `Field.Upload` passes the form's value down through it. Taken literally the suggestion was to read #9270's constant directly, but that constant sits on the forms side of the layering described above, so the value stays where it was and only stops being public. Verified that the underscored prop does not reach the DOM, against a control proving that check can detect one.

#9270's constant enumerates what it governs, so it gains the two operations this PR puts behind it, in the wording #9356 introduced.

## New public surface

Just one thing: `errorDeleteTimeout`, the message shown on a file whose deletion timed out. A translation in all four source locales (`en-US` inherits from `en-GB`) plus an overridable text prop, exactly like its `errorLargeFile` and `errorUnsupportedFile` siblings.

## Documentation corrections

The `onFileDelete` and `onFileClick` docs are shared by three properties pages, and named the deadline by prop. That sentence was wrong on every one of them:

- On the `Field.Upload` and `Value.Upload` pages it named `asyncFileOperationTimeout`, which neither component accepts.
- On the base `Upload` page it claimed a surrounding `Form.Handler`'s `asyncSubmitTimeout` takes over. It does not — only `Field.Upload` and `Value.Upload` bridge it.

That last claim was measured rather than assumed: a base `Upload` inside `Form.Handler asyncSubmitTimeout={50}` was still loading long past 50 ms, while a control using the component's own deadline recovered. Both entries now refer to the deadline without naming a prop, and a test pins the boundary. Confirmed on a rebuilt portal that the corrected wording renders on all three pages.

`asyncSubmitTimeout`'s own doc is also extended, the way #9254 extended it when it widened what that deadline governs.

## One existing test changed

`Field.Upload > should support async fileHandler and onFileDelete in Iterate.Array` declared its `onFileDelete` mocks as plain functions returning a Promise, so `isAsync()` reported them as synchronous, the file was removed on click, and `resolveOnFileDeleteHandler2()` was a no-op. The test was passing while exercising the *opposite* path from the one its name claims. Now that the deletion is genuinely awaited, it has to wait for the file to leave the list before submitting the form.

## Verification

Every behaviour claim was checked by mutation rather than by trusting green. Each mutation below was applied and the listed tests confirmed to fail:

| Mutation | Tests that fail |
|---|---|
| remove the deadline | 3 base + 1 `Field.Upload` |
| `isAsync()` prediction instead of call-then-inspect | all 5 base |
| guard always claims (late settle acted on) | the late-settle test |
| timeout removes the file instead of keeping it | 2 base + 1 `Field.Upload` |
| drop the `Field.Upload` pass-through / the `Value.Upload` deadline | 1 `Field.Upload` + 1 `Value.Upload` |
| restore the unmount cleanup | both unmount tests |
| drop the synchronous `filesRef` update in `Field.Upload` | the concurrent-upload test |
| point the base deadline at the form's `asyncSubmitTimeout` | the `Form.Handler` boundary test |

323 tests pass across the touched areas, and 1029 across those plus `hooks` and `Iterate`. `test:types`, `test:deps`, `lint` and `prettier` clean. The portal build and `validate:links` pass across 1075 pages.

CI is green across every check, including the four with no local equivalent — visual regression, e2e, consumer smoke tests and the preview deploy.

Timing tests were treated with suspicion rather than accepted: the new ones were each run several times in isolation, and two failures seen along the way were confirmed load-induced by re-running the identical command unchanged.

## Known, and deliberately not fixed here

There is a third instance of the stale-snapshot family, and it needs its own change:

**Delete one file while another is uploading, and let both settle in the same tick — the deleted file is left spinning for good.**

Verified byte-identical on `origin/main`, so it is pre-existing and untouched by this branch. Unlike the two above it is not the same one-line fix: it is a race *between* two components that each hold their own snapshot of one shared store, so correcting it needs a synchronous read of that store. That means either adding new public surface to the `Upload.useUpload` hook — which is exactly what the review feedback above asked to avoid — or reaching into shared-state internals from the forms layer. Both are design decisions with a much larger blast radius than anything in this PR, so they belong in a change of their own. Reproduction is available.


langz:
I think this is a nice feature/fallback to have. But instead of providing a prop, I think we should just use the value from https://github.com/dnbexperience/eufemia/pull/9270

no point in exposing a prop for this, until we see the need.

PR #9277, opened 2026-09-04
fix(Filter): bound the async fetcher in useFilterAsync

langz:
## The bug

`useFilterAsync` sets `resultLoading: true` on the shared filter state as soon as filters or search change, and the fetcher's Promise is the only thing that clears it. Nothing bounded that wait, so a fetcher whose Promise never settles left `Filter.Content` showing a skeleton forever — no error, no result count, no way out.

Same failure mode as #9254 (`Field.Upload`'s `fileHandler`) and #9256 (`Form.Section`'s Done button), on a different unbounded consumer Promise.

## The fix

A 30 second deadline on the fetcher, the same value `asyncSubmitTimeout` defaults to. On expiry the loading state is cleared and `error` is set, so the consumer can show a message instead of an eternal skeleton.

There is no option for it, per review — the deadline is not configurable until something asks for it. `DEFAULT_ASYNC_SUBMIT_TIMEOUT` from #9270 therefore moves to `shared/defaults`, alongside the other values shared between components and the forms extension, and `extensions/forms/defaults` re-exports it so every forms consumer stays exactly as #9270 left it. A component reaching into the forms extension for a default would be the wrong direction, and there is now one definition rather than two.

The deadline is measured from the filter change that triggered the fetch, not from when the fetcher is actually invoked. That is deliberate: `resultLoading` is set at change time, so the stuck state the deadline protects against also begins there — which means it covers `debounce` as well as the fetch.

The `requestId` / `cancelled` guards already dropped stale results, but they would happily let a fetch that had *passed its deadline* settle and overwrite state the filter had moved on from — the deadline alone would have fixed the skeleton and left that hole open. Both outcomes now go through a single `claimRequest()`, so exactly one of resolve, reject or timeout takes effect, and the timer is cleared on settle and on unmount.

The timeout error carries `name: 'TimeoutError'`, following the platform convention set by `AbortSignal.timeout()`, so `error.name === 'TimeoutError'` is a reliable discriminator for a consumer that needs to render its own localized message. Without it the only way to tell a deadline apart from a fetcher rejection would be matching on an English message string.

**The fetch is ignored, not aborted.** The fetcher signature takes no `AbortSignal`, so real cancellation would be a separate, larger change.

## Docs

The portal's *Async data fetching* section documented `initialData` and `debounce` and nothing else, so with no option row in the properties table there was no place a consumer would learn that their fetcher has 30 seconds, or that `error` can now carry `TimeoutError`. That section now covers it.

It also states a limitation that was worth measuring rather than assuming: the hook fetches on filter *change*, so re-applying a value it already holds does not start a new fetch. Verified — a repeated `setFilter('/a', 'x')` after a deadline leaves the fetcher call count unchanged and the error in place, while changing the value clears it and refetches. There is no retry call, and the docs no longer imply one.

## Verification

4 new tests: never-settling fetcher recovers with an error; a late settle is ignored; a fetcher that settles inside the deadline is not cut off, and survives the deadline passing afterwards; and the deadline is spent on the `debounce` as well as the fetch.

That last one closes a gap worth naming: *measured from the change, so it covers `debounce`* was asserted in the hook comment, the option docs and the portal, and covered by nothing. Giving each fetch a full deadline of its own — the obvious refactor — passed every existing test. It now fails that one, and only that one.

They drive the real 30 second default with fake timers rather than a short value passed through an option, and pin the constant at 30000 so the prose that states it cannot drift silently.

Checked by mutation rather than by trusting green — each mutation was applied and the listed tests confirmed to fail:

| Mutation | Tests that fail |
|---|---|
| remove the deadline | never-settling + late-settle |
| drop the `settled` guard (late settle acted on) | late-settle |
| remove `clearTimeout` on the resolve path | *none* — the claim guard already makes the stale timer a no-op, so the third test pins the outcome, not that mechanism |
| measure the deadline from the fetch instead of the change | debounce-interaction |

`lint:ci`, `test:types`, `test:deps` clean. `src/components/filter`: 284 passed. Every consumer of the moved constant also run: `useFieldError` 9, `Field/Upload` 86, `Form/Section` 157, all passing.

One thing measured and deliberately left alone: after a deadline, `Filter.ResultCount` still shows the previous query's count. That is not introduced here — the same stale count is displayed throughout the new fetch, before any deadline, and equally after a fetcher rejection. The deadline only flips `resultLoading`; it does not change what is rendered. Worth a separate look, not this PR.

One caveat worth flagging, since it is not mine: `FilterActiveFilters > preserves search when clear all is clicked` fails when the filter directory is run as a whole. Measured against a clean `origin/main` worktree, sequentially, one suite at a time: **the same single test fails on both** — 1 failed / 281 passed on main, 1 failed / 284 passed here. It passes in isolation, so it looks like cross-file shared-state leakage in the filter tests. Untouched here.


langz:
I think this is a nice feature/fallback to have. But instead of providing a prop, I think we should just use the value from https://github.com/dnbexperience/eufemia/pull/9270

no point in exposing a prop for this, until we see the need.

PR #9283, opened 2026-09-07
feat(Tabs): animate active underline

tujoworker:
The selected underline currently jumps directly between tabs. This adds a shared indicator that animates between active items while preserving the initial static render and existing Badge geometry.

Preview: https://feat-tabs-animate-active-und.eufemia-e25.pages.dev/uilib/components/tabs/demos


joakbjerk (approved):

packages/dnb-eufemia/src/components/tabs/style/dnb-tabs.scss
```diff
@@ -365,6 +367,58 @@
     transition: padding 1s var(--easing-default);
   }
 
+  &__selection {
```
snorrekim, line 371: We now have two "selected" type classes that end up doing the same thing, but are not the same actual thing.

`.dnb-tabs__selection` and `.dnb-tabs__button selected`

And we can remove one and still have the exact same visual selection marker. So we right now have code that "hides" the non-animated marker, but why do we have it at all, should we not trust the animated marker?

This might risks regressions that are not caught.

But I'm not sure what the best way to do this is.

Animation code is by nature very custom, and we can't really predict exactly what is required to implement it in one component vs the next. So we can't really create a grand unified code structure, but that makes it more important to figure out a unified naming structure, and some internal best practices.

Animation elements names, and whether or not they should always supplement or replace non-animated elements. And how they are turned off or on.

My general feel is that at the least:
- We should have a unified way to turn off animations
- even if we turn off all animations, the component should work
- if the animated element supersedes (overlaps) a non-animated element, should it not actually be removed?

In this case. I think the two overlapping elements work together in a way. One marks the current immediately, the other slides over to it. But perhaps their naming should be more closely connected?

Again, I do't know the best practice here, but I can foresee that it might cause issues because animation code can be both very complex, and at the same time completely superfluous (can be disabled).

packages/dnb-eufemia/src/components/tabs/style/dnb-tabs.scss
snorrekim, line 296:
```suggestion
        &::after {
          background-color: var(--token-color-stroke-action-hover);
          opacity: 0.25;
        }
```

Makes the animation look better by not pre-marking the selected one. So the animated marker actually seem to move over, instead of simply paintning a path to a marker that already exists.

packages/dnb-eufemia/src/components/tabs/style/dnb-tabs.scss
tujoworker: Good point. The existing button underline is kept intentionally as the SSR/no-JS fallback, so Tabs still has a selected marker before measurement and when JavaScript is unavailable. Once the measured underline is ready, it replaces the fallback visually. Reduced motion still works because the global reduced-motion styles shorten the transition. I added a short comment by the selector to make this relationship explicit.

packages/dnb-eufemia/src/components/tabs/style/dnb-tabs.scss
tujoworker: Tried this locally and agree, it makes the movement read more clearly. Applied the 0.25 opacity.

PR #9292, opened 2026-09-08
chore(Portal): test for duplicate ids in all mdx files and on some rendered pages

snorrekim:
Travels down the imports of each mdx file to check if there are duplicated ids.

Also tests some rendered pages to catch duplicate ids caused by render logic outside of the doc files.


snorrekim:
Right now I am running it without any fixes to the mdx files to test if it works.

snorrekim:
There is an edge case where we have rendering gates around content:

```
<VisibilityByTheme hidden="sbanken">

## In general

</VisibilityByTheme>

<VisibilityByTheme visible="sbanken">

## In general

</VisibilityByTheme>

```

Might just have to ignore, or produce a warning

PR #9294, opened 2026-09-08
chore(SidebarMenu): add responsive navigation, resizing and notifications

tujoworker:
Extends SidebarMenu with reusable responsive inline/Drawer composition, resize handling, notification indicators, refined hierarchy states, and the corresponding documentation and regression coverage.

This is separated from the Portal integration so the component API and behavior can be reviewed independently.

Preview: https://feat-sidebar-menu-responsive.eufemia-e25.pages.dev/uilib/extensions/sidebar-menu


joakbjerk (approved):

langz:
An accordion with `collapsible={false}` loses its own badge as soon as a descendant carries a notification badge. `<SidebarMenu.Accordion id="products" text="Products" badge={3} collapsible={false}>` with one notification item inside renders just `Products` — the `3` is gone and no dot takes its place. The same markup on `feat/sidebar-menu-extension` renders `Products3`, so it's new here.

`showNotificationIndicator` drops the badge from `content`, but `accordionIndicator` is gated on `collapsible`, so on the non-collapsible branch nothing is rendered in its place. A non-collapsible accordion never hides its children, so there's nothing to summarise there — gating the suppression on `collapsible` as well should be enough.

---

**Verified fixed** on `545b564` — suppression is gated on `collapsible` now, so `collapsible={false}` keeps its `3`, and the collapsible case still gets the dot. Thanks! 🙂

langz:
`onOpenChange` fires with a stale closure. The `useMemo` in `SidebarMenuResponsiveProvider` lists `[isHydrated, isMobile, resolvedOpen]`, so `setOpen` keeps whichever `onOpenChange` it captured the last time the memo ran.

I put the provider inside a component holding an unrelated `route` state, opened the drawer while `route === 'a'`, changed it to `'b'`, then closed — the handler reported `a:false`. Same on the controlled path: a value bumped from `0` to `9` before opening still came through as `0`. Any consumer whose handler reads other state gets the wrong value.

Holding the callback in a ref and calling `ref.current` avoids it without adding renders. Worth noting `open`, `defaultOpen` and `onOpenChange` on the provider have no tests at all, which is presumably why this got through.

---

**Verified fixed** on `545b564` — `onOpenChangeRef` keeps it current, and there's a test for it now. Thanks! 🙂

langz:
The resize gesture never handles `pointercancel`, so an aborted drag leaves the page in a stuck state. After `pointerdown` → `pointermove` → `pointercancel`, `dnb-sidebar-menu-resize-handle--resizing` is still on `<html>`, and a later stray `pointermove` keeps resizing — I watched it go to `500px` with no button held.

That class applies `cursor: ew-resize !important` and `user-select: none` to `html *`, so the whole document keeps the resize cursor and can't select text until the next complete pointerdown/pointerup. `touch-action: none` on the handle does prevent the most common cause, so it won't happen often — but `setPointerCapture` plus a `pointercancel` listener in `startResize` would close the lifecycle properly, and would also cover the pointer being released outside the window.

---

**Verified fixed** on `545b564` — `pointercancel` and `setPointerCapture` are both wired up now. Thanks! 🙂

langz:
`separates the selected section badge from its checkmark` isn't capturing the state it's named for. Its baseline is byte-identical to `matches the declarative menu` — same md5 for both `ui` and `sbanken` — and shows the closed dropdown in its initial state, so the `column-gap` rule it was added to guard never renders. It passed in 7.7s on the previous run while asserting nothing.

`[role="option"]:nth-of-type(2)` does resolve to the Business option in a DOM test, so it isn't the selector; I couldn't work out why the click has no effect under Playwright.

Also worth knowing while you look at the red visual check: 43 of the 45 failures are identical on `feat/sidebar-menu-extension`, and the 26 `List` + 1 `Timeline` ones come from the icon path refactor in #9328, which regenerated only the 8 `Icon-*` baselines. These two are the only screenshots failing here that aren't already failing on the base, so they need re-recording regardless — probably worth fixing the simulation first so the new baseline actually shows the open list.

---

**Resolved** on `545b564` — the baseline is no longer a byte-copy (different md5 and dimensions), and per my follow-up the `column-gap` rule is in the captured pixels. Thanks! 🙂

langz:
Arrow-key resizing desyncs once the sidebar is capped in CSS. `getTargetWidth()` re-reads the rendered width on every keypress, so with the target capped at 340px five `ArrowRight` presses give 336 → 352 → 356 → 356 → 356 — it sticks at cap + step, and `aria-valuenow` then reports 356 while the element is actually 340. `aria-valuemax` still advertises the viewport width, so AT is told there's room left when there isn't.

The `minWidth` doc says to "use layout CSS for any additional responsive constraints", so this is the configuration the docs point people at. Holding the written width in state as the source of truth, rather than re-reading the DOM on each keypress, would keep `aria-valuenow` honest and stop the keyboard feeling stuck.

---

**Verified fixed** on `545b564` — the written width is the source of truth now instead of re-reading the DOM each keypress, and `getMaximumWidth()` honours a px cap so `aria-valuemax` follows. Thanks! 🙂

langz (approved):
Approving. The responsive composition and the resize handle are both nicely separated from the Portal work, and the coverage on the declarative and data paths is good.

I've left five separate comments. The `collapsible={false}` badge one is the only thing I'd want fixed before this lands, since it's a behaviour change against the base rather than a new-feature rough edge — the rest are fine as follow-ups.

langz:
On a mobile-width first paint both the inline menu and the hamburger are visible at the same time.

`ResponsiveInline` is JS-only — `useSidebarMenuResponsive().isMobile ? null : children` — and `useMediaQuery` returns `false` on the server and on the first client render, so the full inline menu ends up in the prerendered HTML. `ResponsiveTrigger` is CSS-driven (`display: none` plus `allBelow(medium)`), so it shows straight away. Loading `/uilib/extensions/sidebar-menu/demos/` at 420px with JS blocked: inline nav `display: block` and visible, hamburger visible. After hydration it's correct — nav gone, `data-sidebar-menu-responsive-visible="true"`.

`info.mdx` says "`ResponsiveInline` renders its children only above the breakpoint" and that the CSS-driven visibility avoids layout shifts during hydration, so this looks unintended.

Related: the `medium` in `.dnb-sidebar-menu-responsive-trigger` is hardcoded, so the `breakpoint` prop has no effect pre-hydration — `breakpoint={800}` still flips at `medium`.

The Portal only avoids this because it adds its own `.desktopNavStyle { allBelow(medium) { display: none } }`. A consumer following the docs wouldn't.

---

**Update on `a147dbc`** — the first-paint overlap is fixed (visibility is CSS-driven before hydration now), so at mobile width only the hamburger shows. The `breakpoint` half still stands though, and it's measurable: the CSS flips at the hardcoded `medium` (60em / 960px), so with `breakpoint={800}` the SSR markup at **850px** renders the mobile layout — hamburger shown, inline nav hidden — when 850 > 800 should be desktop. 700px and 1100px both agree; it's only the window between a custom `breakpoint` and `medium` that's wrong until hydration 🙂

---

**Verified fixed** on `8a6a741` — the scoped first-paint CSS does exactly the job. With `breakpoint="50em"` the pre-hydration markup now renders desktop at 850px (it rendered mobile before), and 700px/1100px still agree. Named breakpoints resolve properly too — `small`→40em, `large`→72em, `medium`→the bundled CSS — so no silent fallback. `styleNonce` for CSP is a nice touch. Thanks! 🙂

langz:
`ResponsiveTrigger`'s `aria-controls` never resolves to an element.

The trigger sets `aria-controls={controls}` and the Drawer gets `id={id}`, but `Modal` never renders that id — it becomes `contentId || `dnb-modal-${id}``, so the real ids are `dnb-modal-responsive-menu-drawer` and `dnb-modal-responsive-menu-drawer-content`.

Checked on the demos page at 420px: `aria-controls="responsive-menu-drawer"` with no matching element, both closed and after opening. This is an axe `aria-valid-attr-value` failure, and it applies to the defaults too — `controls` and `id` both default to `sidebar-menu-responsive-drawer`.

Passing `contentId={id}` from `ResponsiveDrawer` would line them up. `SidebarMenuResponsive.test.tsx` only asserts the attribute value, never that the target exists, so nothing catches it today.

---

**Verified fixed** on `545b564` — `contentId={id}` lines them up, and `aria-controls` now resolves to the modal content element. Thanks! 🙂

langz:
Separate from the `pointercancel` and arrow-key points above — the defaults themselves look wrong for a published component.

`minWidth = 1` and no `maxWidth`, so `getMaximumWidth()` falls back to `window.innerWidth`. Out of the box a drag can take the sidebar to 1px or to the full viewport; in the Portal that leaves the header 16px wide with the search field unusable. So the Portal not passing limits is the symptom rather than the cause.

There's also no `resize` listener, so a written width is never re-clamped when the viewport shrinks — the sidebar can end up wider than the window with only double-click to recover, and `resolvedMaxWidth` is computed once so `aria-valuemax` goes stale on window resize too.

Sane defaults (something like 240/560) plus a `resize` handler that re-clamps would cover both.

---

**Verified fixed** on `545b564` — 240/560 defaults plus a `resize` listener that re-clamps and refreshes `aria-valuemax`. Thanks! 🙂

langz:
The new user-facing strings are hardcoded English in library code, and no locale file has a `SidebarMenu` entry:

- `'Close menu'` / `'Open menu'` — `ResponsiveTrigger` `title`
- `dialogTitle = 'Menu'` — this one is the Drawer's accessible name
- `'Resize sidebar'` — `ResizeHandle` `aria-label`
- `aria-label="Contains notifications"`

`LOCALE` defaults to `nb-NO`, so a Norwegian app announces these in English. Grepping `aria-label="[A-Z]` across `src/components` and `src/extensions` returns only sidebar-menu and `VippsWalletButton` (a brand name) — everything else goes through `getTranslation`.

#8986 already has `'Contains current page'` and `sectionLabel = 'Menu section'`, so one `SidebarMenu` locale block would cover both PRs.

---

**Verified fixed** on `545b564` — all five strings are in `da-DK`, `en-GB`, `nb-NO` and `sv-SE` (and `en-US` inherits). Thanks! 🙂

langz (approved):
Approving to keep the stack moving — the responsive split and the resize handle are a good addition.

The five points above are worth sorting before this reaches `main` though, in particular the pre-hydration state on mobile and the resize defaults, since both affect consumers of the published component rather than just the Portal.

The red visual regression looks unrelated: the 45 failures are identical to the ones on #9296 and none are unique to this branch — most likely the icon path refactor in #9328 needing regenerated baselines on `main`.

langz:
`containsNotifications` is missing from the locale files. `SidebarMenuAccordion.tsx:177` reads `translation.containsNotifications`, but the four locale files only got `sectionLabel` and `containsCurrentPage`.

Three things break from that one key:

- **Type check**: `Property 'containsNotifications' does not exist on type '{ sectionLabel?: string; containsCurrentPage?: string; }'`. I ran `tsc --noEmit` on this head and on the base — the base has no sidebar-menu errors, so this is the only new one.
- **Unit test**: `prioritizes nested notifications over information badges` fails, since it asserts `aria-label="Inneholder varsler"`.
- **Runtime**: the indicator renders as `<span class="dnb-sidebar-menu__accordion__notification-indicator" role="img"></span>` with no accessible name at all — so it ends up worse than the hardcoded string it replaced, since a `role="img"` without a name is an axe violation.

Worth knowing it cascades: #9296 and #9295 inherit it, and it accounts for most of the red on those two right now (type checks, build tests, publish).

Should just need the key added to `en-GB`, `nb-NO`, `da-DK` and `sv-SE` 🙂

---

**Verified fixed** on `545b564` — `containsNotifications` is in all four locale files now. Thanks! 🙂

langz:
Thanks for picking this up — the dropdown genuinely opens now, so it's no longer a byte-copy of the closed menu. But I don't think it's capturing the state it's named for yet.

The screenshot targets `[role="option"][aria-selected="true"]`, and the demo sets `defaultActiveSection="personal"`. Personal has no badge — the `badge={9}` is on the **Business** section. Since the clicks that selected Business were dropped, both new baselines are 374×56 and show just a person icon, `Personal`, and the checkmark.

The rule being guarded needs a badge present to match at all:

```scss
.dnb-drawer-list__option--selected:has(.dnb-sidebar-menu__badge)
  .dnb-drawer-list__option__inner {
  column-gap: 0.5rem;
}
```

`:has(.dnb-sidebar-menu__badge)` can't match Personal, so `column-gap` still never renders in the captured pixels.

Selecting Business before the capture — or pointing the test at a section that actually has a badge — should do it.

---

**Resolved** — superseded by my follow-up below; the rule is in the captured pixels after all. Sorry for the round-trip! 🙂

langz:
Small follow-up on the arrow-key fix: `resolvedMaxWidth` can now only ever go *down*.

When CSS caps the width, `setWidth` lowers it to the rendered width (`SidebarMenuResizeHandle.tsx:102`), but nothing ever raises it back — `getMaximumWidth` is a `useCallback` on `[maxWidth, minWidth]` that reads `window.innerWidth` without depending on it, and there's no resize listener in the component. So after hitting the cap once, `aria-valuemax` stays pinned at the capped width for the life of the component, even if the viewport grows and the cap relaxes.

Dragging still works past it, since `setWidth` clamps against a live `getMaximumWidth()` call — so it's only the advertised value that goes stale. Recomputing the maximum on resize would sort this and the original "valuemax advertises the whole viewport" point in one go 🙂

---

**Verified fixed** on `545b564` — there's a `resize` listener now that recomputes the maximum, so it can rise again: `aria-valuemax` went 400 → 560 when I grew the viewport. Thanks! 🙂

langz:
Three things about the notification dot don't match what the code does now.

`types.ts` (3 places) and `SidebarMenuDocs.ts` (2 places) say a nested notification is "indicated on collapsed ancestor accordions", while `info.mdx` says "The dot remains visible when the accordion expands". The code matches `info.mdx`, so the properties table — the one people actually read — is the wrong one.

Now that `showNotificationIndicator` is gated on `collapsible`, a `collapsible={false}` accordion never shows the dot either. That looks like the right call, since its children are always visible, but neither doc mentions it.

And `hasNotificationBadge` walks `element.props.children`, so it can't see past a component boundary. Rendering `<SidebarMenu.Accordion><MyItems /></SidebarMenu.Accordion>` gives no dot, while an array and a `SidebarMenu.Group` in the same position both work. Worth a caveat in `info.mdx`, since pulling a sub-menu out into its own component is a normal thing to do.

---

**Verified fixed** on `545b564` — all five places say "collapsible" now, and `info.mdx` gained the caveat that detection only follows direct `SidebarMenu` children. Thanks! 🙂

langz:
Separate from the defaults point above — the reset has no keyboard equivalent.

`handleKeyDown` handles only `ArrowLeft` and `ArrowRight` (`SidebarMenuResizeHandle.tsx:198`, `:205`), and `resetWidth` is wired only to `onDoubleClick` (`:230`). So a keyboard user can nudge the width but has no way back to the CSS default.

The APG window splitter pattern lists `Enter` with no "(Optional)" marker — collapse the pane, or restore the splitter to its previous position — while `Home` and `End` are both marked optional. `Enter` looks like the natural home for this, and it would also give the 1px case a way out without reaching for the mouse.


---

**Verified fixed** on `545b564` — `Enter` resets the width now. ArrowLeft wrote `240px`, Enter cleared the property. Thanks! 🙂

langz:
An inactive section holding a notification gives no signal at all.

Accordions summarise nested notifications automatically, but sections don't. With a `Section` containing an item that has `badgeProps={{ variant: 'notification' }}`, the dropdown renders `["Personal", "Business"]` with no badge and no dot on either option.

Nothing in the docs promises otherwise, so this isn't a bug — but a section hides an entire tree, which is exactly where an unseen notification matters most, and the new `badge` / `triggerBadge` props are effectively the manual workaround for its absence.

Deriving the section badge the way `hasNotificationBadge` already does for accordions would make the two consistent. Either way it's worth stating in `info.mdx`, because right now the asymmetry is silent.


---

**Addressed** on `545b564` — `info.mdx` now states that sections don't summarize nested notifications and points at `badge` / `triggerBadge`, plus notes the accordion detection only follows direct children. Exactly what I was after, thanks! 🙂

langz:
Correction to [my earlier comment](https://github.com/dnbexperience/eufemia/pull/9294#issuecomment-5632640358) — I was looking at the previous head, sorry.

`Examples.tsx:46` now puts `badge={2}` on the `personal` section, which is the one `defaultActiveSection` selects, so `:has(.dnb-sidebar-menu__badge)` does match. Both regenerated baselines (374×56) show the person icon, `Personal`, the red `2` and the checkmark with the gap between them — so the `column-gap` rule is in the captured pixels, and this one is already done.


---

(My own correction — resolving to keep the thread tidy.)

langz:
A percentage `max-width` on the resize target turns the handle shrink-only.

`getMaximumWidth()` does `Number.parseFloat(getComputedStyle(target).maxWidth)` and treats the result as pixels, but computed `max-width` keeps percentages verbatim — I checked in Chromium, Firefox and WebKit and all three return `"50%"`. So `50%` becomes `50`, which is below the new `minWidth` of 240, and `getMaximumWidth()` falls back to `minWidth`.

One `ArrowRight` press from 320px, one case per test so the renders don't share a DOM:

| target CSS | `aria-valuemax` | 320px → |
| --- | --- | --- |
| none | 560 | 336px |
| `600px` | 560 | 336px |
| `100%` | 240 | **240px** |
| `50%` | 240 | **240px** |

Dragging is capped the same way, since `setWidth` clamps against the same function. `rem`, `min()` and `calc()` are all fine — browsers resolve the first two to px, and `calc(100% - 2rem)` gives `NaN` which falls through to `Infinity`.

`max-width: 100%` on a sidebar is an ordinary thing to write, and the `minWidth` doc points straight at it ("use layout CSS for any additional responsive constraints"). Only trusting the computed value when it ends in `px` should do it 🙂

---

**Verified fixed** on `450fb31` — `getMaximumWidth()` now only trusts the computed value when it ends in `px`. `100%` and `50%` behave identically to the `none` and `600px` controls: `aria-valuemax` 560, and one ArrowRight takes 320px to 336px. Thanks!

langz:
The compound default import now pulls Drawer and Modal into every consumer.

`info.mdx` says "only the subcomponents you use are included", and this PR adds "Stay modular and tree-shakeable through compound exports" to the design goals — but `SidebarMenu.ts` re-exports a namespace object (`import * as _default from './SidebarMenuExport'`), which a bundler can't see through. Bundled both branches with esbuild, React external, minified:

| import | base | this PR | Drawer/Modal in output |
| --- | --- | --- | --- |
| `import { Container }` | 234 250 B | 235 656 B | no |
| `import SidebarMenu` | 234 628 B | **266 642 B** | **yes** |

That's +32 KB minified and +10 KB gzipped (72.8 → 82.9 KB), around +14% — and `import SidebarMenu from '@dnb/eufemia/extensions/sidebar-menu'` is the import in the docs, in `Examples.tsx` and in the Portal. `dnb-modal` goes from 0 to 28 occurrences.

The named import still shakes cleanly (+1.4 KB), so `SidebarMenuContainer`'s new `useOptionalSidebarMenuResponsive` import isn't the cause — it's the namespace re-export. Either the tree-shaking claim wants softening, or the responsive parts want their own entry point 🙂

---

**Verified fixed** on `450fb31` — dropping the default export and documenting `import * as SidebarMenu` does it. Using only `.Root` through the namespace now bundles to 37 956 B with no Modal, Drawer or Dropdown, byte-identical to `import { Root }` and down from 266 616 B. One caveat for anyone reading later: the table above predates the `Container` to `Root`/`Data` split, so it's the ratio that carries over, not the names. Thanks!

langz:
Two small things on the properties page.

`SidebarMenuResizeHandleProperties` still carries the pre-fix defaults: `minWidth` has `defaultValue: '1'` and `maxWidth` says "Defaults to the browser viewport width". The component now defaults to 240 and 560, and `info.mdx` says so too — so the properties table is the odd one out, and it's the page people actually read.

And the responsive parts have no table at all. `properties.mdx` gained only `SidebarMenu.ResizeHandle`, so `ResponsiveProvider`, `ResponsiveTrigger`, `ResponsiveDrawer` and `useResponsive` are prose-only, and `open` / `defaultOpen` / `onOpenChange` aren't mentioned outside the type. Related while you're in there: the return type of `useResponsive()` isn't exported, so consumers can't type it, and it currently hands out `drawerScrollElement` and `setDrawerScrollElement`, which read like internals 🙂

---

**Verified fixed** on `450fb31` — `minWidth` and `maxWidth` now document 240 and 560, and the page gained `ResponsiveProvider`, `ResponsiveTrigger`, `ResponsiveDrawer` and `useResponsive` tables. `useResponsive()` is exported and typed now too (`SidebarMenuResponsiveValue`), narrowed to `close` / `isMobile` / `open` / `setOpen` / `toggle`, so the internals are no longer handed out. Thanks!

langz:
`onClick` passed to `ResponsiveTrigger` is silently dropped.

`<Button {...props} … onClick={toggle} />` spreads first and overrides after, so a consumer handler never runs — I get 0 calls while the Drawer still toggles fine. Same story for `aria-haspopup` and `aria-expanded`.

`ButtonProps` advertises all three, so there's no type error to hint at it. Either compose it (`onClick={(event) => { props.onClick?.(event); toggle() }}`) or `Omit` them from `SidebarMenuResponsiveTriggerProps` 🙂

---

**Verified fixed** on `450fb31` — a consumer `onClick` now fires once and the Drawer still toggles, and `aria-expanded` / `aria-haspopup` are omitted from the props type. Thanks!

langz:
Crossing the breakpoint while the Drawer is open fires `onOpenChange` twice and focuses a hidden button.

Opening at `30em` and then moving to `70em` gives `[true, false, false]` — both the provider's `!isMobile && resolvedOpen` effect and the Drawer's `onClose` run for the same transition. `onClose` also calls `triggerRef.current?.focus()`, and at that width the trigger has just been given `data-sidebar-menu-responsive-visible="false"`, i.e. `display: none`, so the focus restore has nothing to land on.

Nothing covers the transition today, which is presumably why it slipped through. Skipping the focus restore when `!isMobile`, and letting one of the two paths own the close, would tidy both.

For what it's worth I checked the scarier version of this first — Modal does clean up its scroll lock and `aria-hidden` when it unmounts mid-open, so the page isn't left stuck 🙂

---

**Verified fixed** on `450fb31` — crossing the breakpoint now gives a single `[true, false]` and focus is no longer moved to the hidden trigger. I checked the normal path too: closing with Escape at mobile width still returns focus to the trigger, so the guard is scoped correctly. Thanks!

langz:
A few naming notes. I checked each one against the rest of the library first and ended up dropping most of my list, so these are the ones that genuinely deviate.

**`isMobile`** is the only screen-size flag in Eufemia named after a device rather than a size. `useMedia()` returns `isSmall` / `isMedium` / `isLarge`, destructured in `Breadcrumb.tsx:191`, `FilterDate.tsx:76` and `FilterSortButton.tsx:50`; and where a `useMediaQuery` result gets its own local name it still uses size language (`isSmallScreen`, `AccordionContent.tsx:246`). `isMobile` occurs 23 times (plus 6 for the new `isMobileRef`), all in `SidebarMenuResponsive.tsx` and `SidebarMenuRoot.tsx`, and nowhere else in `src` — and unlike those locals it's public, since `useResponsive()` hands it out. It also claims a little more than it knows: a narrow desktop window isn't a phone. `isSmall` or `isSmallScreen` would match the house vocabulary and stay truthful about what was actually measured.

**`rootSelector`** on `ResizeHandle` sits one import away from `SidebarMenu.Root` with an unrelated meaning ("the closest shared layout ancestor"). `<SidebarMenu.ResizeHandle rootSelector=".layout">` next to `<SidebarMenu.Root>` reads like it might be selecting the Root. Where Eufemia does use `root*` it means the component's *own* root element — `rootRef` on `DrawerListPortal` — whereas here it means an ancestor the component doesn't own. (The one other `rootSelector` in `src` is a local in `InteractionInvalidation.ts`, meaning something else again.) `<what>Selector` itself is well established and public — `targetSelector` on Tooltip and Popover — so it's only "root" doing double duty here. `scopeSelector` or `ownerSelector` would sidestep it.


For the record, things I looked at and decided were fine, in case they come up: `Root` (matches `Menu.Root`, `Filter.Root`, `Stat.Root`, same `XRoot.tsx` → `Root` file pattern); `triggerBadge` (`trigger*` is well established — `triggerProps` 15, `triggerAttributes` 9, `triggerOffset` 8 — and it maps onto Dropdown's own `content` / `selectedValue` split); bare `open` / `defaultOpen` (69 and 11 uses, against a single namespaced counterpart in Filter's `defaultPanelOpen`, and bare matches Radix); the four `Responsive*` parts (60 of 133 namespace members are multi-word, and `Submit*` / `Push*` already repeat a prefix); `--sidebar-menu-*` custom properties (698 component-scoped in the repo against 9 `--dnb-*`); and the nested `__a__b` classes (8 other blocks do the same) 🙂

---

**Verified fixed** on `545b564` — renamed to `isSmallScreen` (and `isSmallScreenRef`) throughout. Thanks! 🙂

langz:
Small follow-up on the `onClick` fix — the `defaultPrevented` half of it can't fire.

`Button` dispatches through `dispatchCustomElementEvent`, which hands the handler a shallow spread copy of the synthetic event (`{ ...eventObjectOrig.event, ...eventObjectOrig }`, `component-helper.ts:167`). So the `defaultPrevented` that `ResponsiveTrigger` reads is a snapshot taken before the consumer's handler runs, and `preventDefault` isn't copied across at all. Logging from inside a consumer `onClick`:

```
before = false          typeof args.preventDefault = undefined
args.event.preventDefault()
afterOuter = false      afterNested = true      aria-expanded = true
```

So `!event.defaultPrevented` is always true, the branch is unreachable, and a consumer has no way to suppress the toggle: `args.preventDefault()` throws `is not a function`, and `args.event.preventDefault()` only marks the nested event.

`event.event?.defaultPrevented` does flip to `true`, so reading the nested event would make the guard behave as intended — otherwise it's probably cleaner to drop it. `calls a supplied trigger onClick before toggling` passes either way, so nothing pins it today 🙂

---

**Update on `545b564`** — fixed, and I overstated this one: the guard does fire, and `onClick={({ event }) => event.preventDefault()}` suppresses the toggle correctly. The direct `(e) => e.preventDefault()` throws, but a plain `Button` behaves the same and `tsc` accepts it library-wide, so that's pre-existing and nothing to do with this PR. Sorry for the noise! 🙂

langz:
Also from the narrowing change: `useResponsive()` now returns a fresh object on every render.

It used to hand back the memoized context value; now it builds `{ close, isMobile, open, setOpen, toggle }` on each call, so the identity changes even when nothing inside it has. With a consumer re-rendering for unrelated reasons I get five renders, five distinct identities, and a `useEffect(…, [responsive])` firing on all five.

Not a correctness bug by itself, but it quietly defeats `useMemo` / `useEffect` / `React.memo` on the hook result, which is a fair thing to expect to work on a public hook. Wrapping the return in a `useMemo` over those five values would restore it.

The narrowing itself is a nice change, by the way — `drawerScrollElement` and friends are properly off the public surface now 🙂

---

**Verified fixed** on `545b564` — memoized now. Re-ran it with a consumer re-rendering for unrelated reasons and the identity holds, changing only when a value actually changes. Thanks! 🙂

langz:
Nice feature 🙂 Small thing on the dismissible desktop example — the collapse doesn't actually hide the menu.

`.sidebar` goes to `width: 0`, but nothing clips it: `ResponsiveInline` is `display: contents` so it can't, and the example CSS never sets `overflow`. With the real markup and the compiled stylesheet the labels stay painted — "Personal customers" ends up as a vertical column of single letters down the left edge, full viewport height. Same in Chromium, Firefox and WebKit. `inert`/`aria-hidden` work fine, so it's gone for screen readers and for clicks, just still visible on screen (which is also why a11y checks won't flag it).

Where it ends up depends on how the sidebar is positioned, which the example doesn't specify: with the grid `.layout` from the resize example above it sits on top of `.content`, and with the dismissible block on its own it pushes `.content` below the fold instead. Either way nothing is hidden.

`overflow: hidden` on `.sidebar` in the example fixes it: collapsed is clean, expanded is unchanged, and both the resize handle (`position: fixed`) and the portaled section dropdown are unaffected.

---

**Verified fixed** on `8a6a741` — and `overflow: hidden auto` is better than what I suggested, since it keeps vertical scrolling. Re-measured with the new documented CSS in all three engines (menu text and page content colour-separated so they can't be confused): menu ink after collapse goes from ~500px to **0** in Chromium, Firefox and WebKit. The added note that the sidebar must own the clipping because `ResponsiveInline` is `display: contents` is a good call. Thanks! 🙂

langz:
One on the new pre-hydration script — as published it doesn't run, and it fails silently 😅

`getPreHydrationScript()` serialises `applyPreHydrationState` with `Function.prototype.toString()`, but the `polyfill-corejs3` plugin (`usage-pure`, applied to every `BABEL_ENV` except `es`) rewrites three built-ins *inside* it into module-scope imports: `rules.push(…)` and `closedItems.includes(…)` become `…InstanceProperty(x).call(x, …)`, and `JSON.parse` becomes `_JSON$parse`. Those identifiers don't exist inside the injected inline `<script>`, so it throws on first use — and every path is wrapped in `try {} catch {}`, so nothing surfaces.

It's the shipped output, not just a local artifact: in the published `11.13.0` tarball 134 root-level files and 137 under `cjs/` import `core-js-pure`, while `es/` has 0. The root build is what `main`/`module` point at, and `build:copy` copies `build/esm/` there before `rm -rf build/esm`.

Injecting the script the way the component does, in Chromium with the state seeded:

| entry | style element | rules |
| --- | --- | --- |
| `@dnb/eufemia/es` | yes | 6 |
| root (`main`/`module`) | **no** | **0** |
| `cjs/` | **no** | **0** |

No page errors in any of them. The unit tests can't see it because vitest runs the TS source, where those built-ins are still native.

Worth being clear on severity: nothing consumes it yet, and when it fails you get today's behaviour (a flash of the wrong open state) rather than a broken page — so this isn't urgent. It just won't work when #9296 starts using it, and it won't say so.

Avoiding the three rewritten built-ins inside the serialised function fixes it: `list[list.length] = item`, `indexOf(id) === -1`, and `globalThis['JSON']['parse'](value)` (computed access, so the plugin doesn't match it). With those three, the root build produces the same 6 rules as `es`.

Minor: `PreHydrationScript` and `getPreHydrationScript` aren't in `info.mdx` or `properties.mdx` yet.

---

**✅ Fixed in `ca5a19cb` — verified.** All three published entry points now run the inline script successfully (6 CSS rules created, 0 page errors in Chromium):

| entry | before | after |
|---|---|---|
| `@dnb/eufemia/es` | 6 rules | 6 rules |
| `@dnb/eufemia` (root/esm) | **0 rules** | 6 rules |
| `@dnb/eufemia` (cjs) | **0 rules** | 6 rules |

Compiled output for all three Babel envs now contains zero `core-js-pure` imports. The new `postbuild.test.ts` guard is a good one — it imports the *built* module per stage and executes the serialised string via `Function(...)` with no module scope, which is exactly the failure mode; I mutation-tested it against the pre-fix code and it correctly fails there. It also ran on a real build in CI (45 tests, `BUILD_MINI` unset, so all 3 stages).

langz:
Thanks for working through all of these 🙂

I've gone back over the whole list against `8a6a741` and everything else I raised is now fixed and verified — the collapse clipping, the pre-hydration breakpoint, the keyboard reset and `aria-valuemax` at a CSS cap, `pointercancel`, `aria-controls` resolving, the locale keys, the notification-dot docs, and the non-collapsible accordion badge. That's 25 resolved. The touch/hover guard in `SidebarMenuSection` is a good addition too, and `overflow: hidden auto` was the better call over the plain `overflow: hidden` I suggested.

The one thing still outstanding is the pre-hydration script comment above — the Babel core-js rewrite inside the serialised function, which makes the bootstrap a silent no-op in the published root and `cjs` builds. It isn't urgent, since nothing consumes it yet and it degrades to today's behaviour, but it will fail quietly once #9296 starts using it.

Happy to approve as soon as that one's sorted 👍

langz:
The `postbuild.test.ts` guard is a good addition — it executes the built module per stage with no module scope, which is exactly the failure mode, and I confirmed it fails against the pre-fix version.

One gap while you're in there: it only exercises the open-items half. The mock returns `[]` for anything other than `[data-open-items-storage-key]`, so the scroll-restore block never runs, and the assertions only look at the appended `<style>`. The two blocks have separate `try/catch`, so a helper reference in the scroll half is swallowed independently — I put one at the top of that block and the guard still passed.

Correcting myself on two things from my first version. `preserves an exact stored scroll position before hydration` does cover that block, so it isn't untested — but it imports the source, so it can't see a build-time rewrite: the same mutation passes there too. So the scroll half has no coverage against *this* class of bug specifically, which is narrower than I first put it. And extending the mock is about six lines rather than the two I said — a `view` with the three `style` methods, `closest`, `querySelector`, the extra `querySelectorAll` branch and a storage key. With that plus an assertion on the restored `scrollTop`, the clean build passes and the broken one fails; I checked both ways.

---

**✅ Fixed in `1bf7be4e` — verified.** The mock now supplies a `view` and a `scrollMenu`, and asserts the restored `scrollTop`. I ran your exact mock against a clean build and against one where I broke the scroll block: clean passes, broken fails on `scrollTop` (0 instead of 120) while the two style assertions still pass — so the new assertion is doing the work. Gap closed.

langz (approved):
That's the last one from me. The pre-hydration script now runs in all three published entry points (6 rules, no errors in root, `cjs` and `es`), and the new build-stage guard would have caught the old version. Everything else on the list stays verified at `f5a2c3c`.

Approving 👍

packages/dnb-design-system-portal/src/docs/uilib/extensions/sidebar-menu/demos.mdx
```diff
@@ -19,3 +19,12 @@ Use `sections` or `data` to render the same recursive structure from JavaScript
 This example is placed inside a `ScrollView` with a viewport-aware maximum height and persists its scroll position in session storage.
 
 <Examples.SidebarMenuData />
+
+### Responsive drawer
+
+The same data renders as an inline menu on larger screens and behind a hamburger button on smaller screens. `ResponsiveProvider` owns the breakpoint, Drawer state, and optional collapsed-inline state, while `ResponsiveDrawer` handles the dialog, focus trapping, Escape key, backdrop, left-side close button, and focus return.
+
+Drag the demo viewport to the medium breakpoint or narrower to reveal the hamburger button and try the drawer.
+For a resizable desktop sidebar that can be dragged past its minimum width and restored from the same trigger, see [Dismissible desktop navigation](/uilib/extensions/sidebar-menu/info#dismissible-desktop-navigation).
```
langz:
```suggestion
For a resizable desktop sidebar that can be dragged past its minimum width and restored from the same trigger, see [Dismissible desktop navigation](/uilib/extensions/sidebar-menu/#dismissible-desktop-navigation).
```

langz:
Not sure if we need to look into this.

But my thoughts regarding the following video is:
- Why does the dropdown open above/at the top, as it's not visible then, and the user will have to scroll to find it.
- When scrolling I see the dropdown having a higher z-index than the search/menu bar, is that correct? Not sure if it's worth looking into.

https://github.com/user-attachments/assets/96da39ca-6a91-4868-9f2e-4e7eb5c21952

langz (approved):
Both of my last two points are now closed, and I checked each by breaking the thing the test is meant to protect:

- `1bf7be4e` — the pre-hydration guard now covers the scroll half. Against a build where I broke that block it fails on `scrollTop` while the style assertions still pass, so the new assertion is carrying its weight.
- `725184d8` — the dimming test now fails with `!sectionSelectorOpenRef.current` removed, so it pins the fix rather than passing either way.

I also confirmed the rebase was content-free for this PR — its own patch is unchanged apart from blob hashes and line offsets, so nothing I verified earlier needs redoing. Re-approving on `1bf7be4e`; `Run tests` and the visual-regression suite were still finishing when I looked.

langz:
`c9ce3b3` looks right — I checked the three-level case too, and the innermost accordion still animates after the outer reopens, so nothing is over-gated.

The test that came with it doesn't guard it, though. `animates only the outer accordion when open descendants return` passes unchanged against the pre-change source, and also with each of the three parts removed individually — the `animate` gating, the `ancestorsOpen` value, and the two animation handlers. The reason is the observable it picked: `--animating` on the inner element comes out identical either way in jsdom, at every step of that sequence.

`--parallax` on the inner element does discriminate. Without the change the inner is already primed to animate when the outer reopens; with it, it isn't. Swapping that one class in the existing assertion fails on the old source and passes on the new one — I checked both directions.



---

**✅ Fixed in `e482b0db` — verified.** Swapping the assertion to `--parallax` does the job. Against the pre-change source it now fails on exactly that line (`Received: … dnb-height-animation--parallax`), and it fails just as well with `ancestorsOpen` neutralised to `true`, while the clean head passes. So the two parts that carry the fix are pinned now. Gap closed — thanks!

langz (approved):
Approving 👍

Thank you for working through all of this so thoroughly — it was a long list and you've closed every item on it. A few were genuinely awkward to chase: the Babel core-js rewrite that made the pre-hydration script a silent no-op in the published root and `cjs` builds, the dimming that only stuck when the mouse happened to land on one particular option, and the option nodes being swapped out mid-gesture. You also went further than asked more than once — adding the build-stage guard rather than only fixing the three built-ins, and then extending it to cover the scroll half once I pointed out the gap.

All 17 checks are green on `c9ce3b3`. I re-verified the two fixes from earlier today by breaking what they protect, and both now fail without the fix.

Two things left, neither blocking this PR: the note above about the new nested-accordion test not yet guarding its change, and on #8986 the double-tap report is worth a retest on your phone now that the `DrawerList` cause has landed. #8986's visual-regression run also failed on `c99047e9`, but the two failing screenshots are Autocomplete and RadioButton, the shared files are byte-identical to the previous green run, and no baselines were touched — so that one just needs a re-run.

langz:
On the z-index half I can give you a definite answer: yes, the open list does sit above the header, and it isn't this PR. The drawer-list portal is `z-index: 3300` and the portal header is `3201`, so the list wins by design. I measured the same stack on `/uilib/components/dropdown/demos/` as on the SidebarMenu demos — byte-identical, same portal z, same header z — so it's generic `DrawerList` portal behaviour rather than anything the sidebar introduces.

The "opens at the top so you have to scroll" half I could not reproduce reliably, so I don't want to guess at it. My desktop harness never got the list to paint with a non-zero height on either page, including the plain `Dropdown` control, so every number I got there was an unpainted element and worth nothing. It's likely the same family as the first-open positioning already looked at above, which also turned out to be pre-existing `DrawerList` behaviour — but if you can say which page and viewport the video is from, I'll measure it properly rather than assume.

langz (approved):
Approving 👍

`e482b0db` closes the last item. I re-ran the mutation check in both directions: the strengthened assertion fails against the pre-change source on exactly that line (`Received: … dnb-height-animation--parallax`) and fails just as well with `ancestorsOpen` neutralised to `true`, while the clean head passes — so the two parts that carry the fix are pinned now rather than the test passing either way. Full sidebar-menu suite is 127/127 green and every check on `e482b0db` is passing, including the visual-regression job that was red earlier in the stack.

That's the whole list closed. I've hidden the resolved threads to make the remaining conversation readable.

Two things left over that are **not** blocking and don't need anything from you before this lands:

- The "dropdown opens at the top" half of my video question is still open. The z-index half I could answer definitively — it's generic `DrawerList` portal behaviour, identical on `/uilib/components/dropdown/demos/` — but I couldn't reproduce the positioning reliably enough to say anything useful. Happy to chase it separately if you can point me at the page and viewport.
- Tiny coverage note for whenever you're next in that file: removing `onAnimationStart`/`onAnimationEnd` (leaving `descendantsReady` at its initial value) keeps all 127 tests green, and I couldn't get it to produce an observable difference in jsdom either — including a three-level nest where the innermost still animated identically. That may simply be jsdom having no real transitions, so I'm not claiming the handlers are redundant; just flagging that nothing currently pins them.

---

Thank you for seeing this one all the way through — it was a long list and you closed every item on it, several of which were genuinely unpleasant to chase. The Babel `core-js` rewrite that made the pre-hydration script a silent no-op in the published root and `cjs` builds was the standout: an easy one to dismiss as a local artifact, and you not only fixed it but added a build-stage guard that executes the serialised string with no module scope, which is exactly the failure mode. The percentage `max-width` parsing, the double `onOpenChange` on breakpoint crossings, and the namespace re-export dragging Drawer and Modal into every consumer were all in the same category — small-looking, awkward to verify.

I also appreciate that you pushed back where you were right to, and that a few of your fixes were better than what I suggested: `overflow: hidden auto` over my plain `overflow: hidden`, and narrowing `useResponsive()` rather than just documenting the internals it was handing out. The component is in noticeably better shape than when it arrived. Nice work 🙂

langz:
`458bc9d1` doubles the gap when an accordion contains a `SidebarMenu.Group`.

The new padding is attached with a descendant selector, so it lands on **every** `.dnb-sidebar-menu__list` inside the accordion — including the one `SidebarMenu.Group` renders:

```css
.dnb-sidebar-menu__accordion .dnb-sidebar-menu__list {
  padding-block-start: var(--sidebar-menu-accordion-gap);
}
```

`.dnb-sidebar-menu__group__list { padding: 0 }` can't hold it back — one class against two, so the accordion rule wins on specificity regardless of order.

Measured in Chromium against the real rendered markup, with the compiled stylesheet, distance from the accordion trigger to the first item:

| composition | `458bc9d1~1` | `458bc9d1` |
| --- | --- | --- |
| accordion → items | 8px | 8px |
| accordion → accordion → items | 60px | 60px |
| **accordion → group → items** | **8px** | **16px** |

The `padding-top` on the group's own list is the giveaway: `0px` before, `8px` after, on top of the 8px the content list already contributes. The other two compositions are untouched, so it's specifically the group case.

`info.mdx` points people straight at this composition — "Use `SidebarMenu.Group` for a titled category whose items should always remain visible" — and it's reachable from the data API too, since `type: 'group'` renders the same `SidebarMenuGroup`. Nothing catches it today: the 127 sidebar-menu tests pass either way, and the visual-regression suite is green, so no baseline covers a group inside an accordion.

Worth noting the fix looks like it was already half-written — `458bc9d1` adds `className="dnb-sidebar-menu__accordion__content"` to the `HeightAnimation`, but that class isn't referenced anywhere in the repo. Scoping the rule to it puts the padding on exactly one element:

```scss
.dnb-sidebar-menu__list {
  gap: 0.5rem;
  padding: 0;
}

&__content > .dnb-sidebar-menu__list {
  padding-block-start: var(--sidebar-menu-accordion-gap);
}
```

I compiled that and re-measured all three: group goes back to 8px, plain stays 8px, nested stays 60px, and the group list's `padding-top` returns to `0px` while the accordion content list keeps its `8px`.

Everything else on `458bc9d1` looks good to me — dropping `compensateForGap` for a padding that lives inside the animated box is a much cleaner way to get the collapsed state right than the `margin-block-start: -0.5rem` hack it replaces, and removing the two `!important` margin rules from the pre-hydration script is a nice simplification on top. All checks are green on this head. It's just the selector scope 🙂

PR #9295, opened 2026-09-08
chore(Portal): move the home page to MDX

tujoworker:
Moves the Portal home page to MDX, replaces the legacy MainMenu home implementation, and adds the corresponding LLM metadata renderer.

This is separated from the SidebarMenu Portal integration so the homepage and metadata migration can be reviewed independently.


joakbjerk (approved):

packages/dnb-design-system-portal/src/shared/home/Home.tsx
```diff
@@ -0,0 +1,151 @@
...
+          across web and native platforms.
+        </P>
+      </section>
+
+      <nav aria-label="Get started">
+        <Card.List className={actionGridStyle}>
+          <Card.ListItem>
+            <Card.Action
+              href="/quickguide-designer"
+              rel="external"
```
joakbjerk: are these external links?

langz:
I feel like when I press these buttons that it's more of a reload happening compared to when clicking other links:

image

langz:
Should the animation of the cards only happen on hover, and not on tabbing?


https://github.com/user-attachments/assets/59f9fac2-9682-40c8-9a01-f106534700dc

packages/dnb-design-system-portal/src/shared/home/Home.tsx
```diff
@@ -0,0 +1,151 @@
...
+  heroStyle,
+  introStyle,
+  resourceGridStyle,
+  resourcesStyle,
+  rootStyle,
+  titleStyle,
+} from './Home.module.scss'
+
+const resources = [
+  { title: 'Images', url: '/uilib/usage/best-practices' },
```
langz, line 24: I was not able to find any info regarding Images at this URL.

tools/eufemia-llm-metadata/src/extensions/mdx/home.ts
```diff
@@ -0,0 +1,32 @@
...
+        '# Welcome to Eufemia',
+        '',
+        "Eufemia is DNB's design system, providing resources for designers and developers to create consistent and efficient experiences across web and native platforms.",
+        '',
+        '- [Design](/quickguide-designer) – Figma UI kits and more',
+        '- [Develop](/uilib/getting-started) – Get started with installation guides',
+        '',
+        '## Resources',
+        '',
+        '- [Images](/uilib/usage/best-practices)',
```
langz, line 21: I was not able to find any info regarding Images at this URL.

langz (approved):

langz:
The old home page linked to six sections: `/brand/`, `/design-system/`, `/icons/`, `/contribute/`, `/uilib/` and `/quickguide-designer/`. The new one links Design and Develop plus the resources — so Brand and the change log are no longer linked from the home page.

Is that intentional? Neither declares `sidebarMenu:` frontmatter, so they don't show up in the new sidebar either. `/brand` ends up with no inbound links anywhere — I could only reach it through the search field. The change log is milder, still two hops away via `/uilib/getting-started` → `/design-system/contact`.

langz:
Found the cause of the reload you noticed in [this comment](https://github.com/dnbexperience/eufemia/pull/9295#issuecomment-5630366365) — the two hero cards never reach the router.

`Card.Action` only routes through a router component when you give it an `element`. `CardAction.tsx:105` does `const AnchorElement = element || 'a'`, and with no `element` it falls through to a bare `<a href={href || to}>` at `:123`. Both hero cards pass only `href` (`Home.tsx:57` and `:90`), so their clicks are plain browser navigations.

The Resources links just below go through the portal's `Anchor`, which substitutes `PortalLink` → react-router's `Link` (`Anchor.tsx:56` and `:25`). That's the difference you were feeling.

I measured it by setting a value on `window` before each click and checking whether it survived:

| link | result |
| --- | --- |
| Design → `/quickguide-designer` | full document load |
| Develop → `/uilib/getting-started` | full document load |
| Icons → `/icons` | client-side |
| Grid → `/uilib/layout/grid` | client-side |

The two cards also end up on `/quickguide-designer/` with a trailing slash while the client-side ones stay on `/icons` — the server round-trip showing.

`Card.Action`'s own doc comment points at the fix — "Route path for use with a router Link component (e.g. react-router). Pass a custom `element` that accepts `to`" — so `element={Link}` from `../tags/Anchor` on both cards should make them behave like every other internal link.

Unrelated but in the same two lines: both cards carry `rel="external"` (`Home.tsx:59` and `:92`) for internal portal pages. `Card.Action` just forwards it to the anchor so nothing breaks today, but it does advertise `/quickguide-designer` and `/uilib/getting-started` as off-site.

packages/dnb-design-system-portal/src/docs/index.mdx
```diff
@@ -0,0 +1,8 @@
+---
+title: 'DNB Design System'
```
langz, line 2: Going through `PortalLayout` means the home page now inherits the "Suggest an edit" button, and it is live on the preview:

```
curl -sL https://chore-portal-home-mdx.eufemia-e25.pages.dev/ | grep -o 'suggest-edit[^"]*index.mdx'
-> suggest-edit/?page=%2F&source=...%2Fsrc%2Fdocs%2Findex.mdx
```

It renders as a lonely secondary button below the 100vh hero. Inviting everyone to edit the front page is very much in the open source spirit, but maybe not the first impression we are after. `hideEditLink: true` in the frontmatter does it, same as `contribute/suggest-edit.mdx`.

packages/dnb-design-system-portal/src/e2e/pageNavigation.spec.ts
```diff
@@ -110,6 +110,33 @@ test.describe('Page Navigation', () => {
       await waitForApp(page)
     })
 
+    test('uses client-side navigation for the home action cards', async ({
+      page,
+    }) => {
+      for (const { name, path } of [
+        { name: /Design/, path: '/quickguide-designer' },
+        { name: /Develop/, path: '/uilib/getting-started/' },
```
langz, line 118: The `toHaveURL` support you added to the page filter is what keeps `/quickguide-designer` alive in the e2e build, which is neat. But `collectForOfStringValues` only understands `for...of` over arrays of string literals, and this one iterates objects, so `path` is never resolved.

I ran the real collector on this branch:

```
collectE2eTestPages() -> 41 pages
/quickguide-designer   => IN allowlist
/uilib/getting-started => MISSING
```

It is not in `E2E_TEST_FALLBACK_PAGES` either (that one has `/contribute/getting-started`). Since `build:e2e` sets `IS_E2E=1`, the Develop card navigates to a route that got filtered out of `virtual:portal-pages` — and both assertions here (URL + `portalNavigationMarker`) are perfectly happy with that. So the test is green while visiting a page that does not exist, which is a bit too zen for my taste. A heading assertion like the Design test has would catch it.

tools/eufemia-llm-metadata/src/extensions/mdx/home.ts
```diff
@@ -0,0 +1,34 @@
...
+        return content
+      }
+
+      const markdown = [
+        '# Welcome to Eufemia',
+        '',
+        "Eufemia is DNB's design system, providing resources for designers and developers to create consistent and efficient experiences across web and native platforms.",
+        '',
+        '- [Design](/quickguide-designer) – Figma UI kits and more',
+        '- [Develop](/uilib/getting-started) – Get started with installation guides',
```
langz, line 17: This is a carbon copy of `Home.tsx` — same intro sentence, same eight resource links. They are in sync today (I checked, all ten targets resolve to real docs), but nothing keeps them that way, and the copy has already started drifting: `/uilib/getting-started` here vs `/uilib/getting-started/` in the component. The first person to add a link to the home page will not think to look in `tools/`. Could the list live in one place and be imported?

packages/dnb-design-system-portal/src/shared/menu/Card.tsx
```diff
@@ -1,12 +1,11 @@
 /**
  * Card
- *
  */
 
 import type { HTMLAttributes, JSX, ReactNode } from 'react'
 import { Link } from '../tags/Anchor'
 import { Button, Card, Lead, P } from '@dnb/eufemia/src'
-import { cardItemStyle } from './MainMenu.module.scss'
+import { cardItemStyle } from '../parts/ListComponentsOverview.module.scss'
```
langz, line 8: `MenuCard` is only used by `contribute.mdx`, and it now reaches into a stylesheet named after a completely different component. `cardItemStyle` was at least neutrally homeless in `MainMenu.module.scss`; now `ListComponentsOverview.module.scss` owns nine lines that two unrelated consumers share. A `Card.module.scss` next to this file would spare the next person the treasure hunt.

packages/dnb-design-system-portal/vite.config.ts
```diff
@@ -211,6 +211,7 @@ export default defineConfig({
       '@mdx-js/react',
       'react/jsx-runtime',
       'react-markdown',
+      'remark-gfm-react-markdown',
```
langz, line 214: Curious what pulled this one in — `remark-gfm-react-markdown` is only imported by `PropertiesTable`, which this PR does not touch. Did moving the home page to MDX shuffle the dep discovery and give you a dev server re-optimize? No objection, I would just like a sentence in the description so it does not look like it wandered in on its own.

langz (commented):
Had another look now that the routing is sorted. One thing I think should be fixed before merge, plus a few smaller notes.

langz:
Picking up my [earlier note on the removed home links](https://github.com/dnbexperience/eufemia/pull/9295#issuecomment-5630410672): Brand and a change log are both back under Resources, so that part is sorted. `/design-system` did not make it back, and it now has no way in.

The old home page's first card pointed at it. On this branch nothing does:

```
grep -rn -e '](/design-system)' -e 'href="/design-system"' src vite
-> no matches

curl -sL <preview>/uilib/getting-started/ | grep -o 'href="/design-system[^"]*"'
-> /design-system/about/living-system, /design-system/contact
```

Only the sub-pages are linked. The section page itself (`design-system.mdx`, `title: 'About Eufemia'`, `menuTitle: 'Change log'`, the one rendering `EUFEMIA_CHANGELOG.mdx`) has no inbound link and no `sidebarMenu:` frontmatter, so it is not in the sidebar either. The new `What's new` resource goes to `/uilib/changelog`, which is the GitHub releases list — a different page.

Same question as last time then: intentional? If the old change log is on its way out, fine by me. If it is meant to stay, it needs an entry point somewhere.

langz:
Took a deeper look at the front page as a whole this time, and there is one thing I would like sorted before this lands.

### `/` no longer hydrates

React throws #418 (`the server rendered text didn't match the client`) on every load of the new home page. Reproduced 3/3 against the previews:

```
#9295  /                          React #418    1983 nodes    277 kB raw / 30 kB br
#9296  /                          clean          174 nodes     47 kB raw / 10 kB br
#9296  /uilib/components/button/  React #418    4226 nodes
eufemia.dnb.no /                  clean
```

The hydrated DOM contains no `_R_` ids at all, only `_r_` ones — and per `mountId()` in react-dom 19.2.5 the `_R_` form is emitted while `isHydrating` and `_r_` only on a plain client render. So React is discarding the prerendered tree and rebuilding it. You can see it in the SidebarMenu responsive `<style>`, served as `sidebar-menu-_R_9_` and ending up as `sidebar-menu-_r_0_`.

Correction to my first pass: that `<style>` difference is the bail-out, not necessarily its trigger. It is the only text difference between the served HTML and the client render that is not a post-mount update — the `⌘ K` search hint is the other one, and #9296 `/` patches that too while staying clean — so it is the best lead I have, but I have not proven it is the first mismatch.

Either way it is not something this PR introduced: `/uilib/components/button/` behaves the same on #9296. But #9296 kept the front page out of it, since the old home page prerendered without header and without sidebar, and this PR opts it in. Net effect is 230 kB of extra markup (20 kB over the wire) on the most visited page, thrown away on arrival.

Chasing the actual mismatch belongs in #9296. I would just rather `/` not be the page that puts it in front of everyone without that being a deliberate call.

### Minor, while we are deleting things

`changelogVersion` loses its last consumer together with `MainMenu.tsx`:

```
grep -rn "changelogVersion" src | grep -v buildInfo.ts
-> no matches
```

`build-info.ts` still reads and parses `EUFEMIA_CHANGELOG.mdx` on every build to produce it, and `buildInfo.ts`, `vite-env.d.ts`, the test mock and five assertions in `build-info.test.ts` still carry it around. `releaseVersion` and `buildVersion` are still used by the footer — this one now goes nowhere.

Two smaller ones in passing: the `test-page-filter.ts` header still says it extracts URLs from `page.goto()` calls, which is no longer the whole truth after the `toHaveURL` addition, and `createHomeExtension` is the one import breaking the alphabetical order in `registry.ts`.

tools/eufemia-llm-metadata/src/extensions/mdx/home.ts
langz: One more argument for a single source: `validate:links` only walks `index.html` files (`vite/prod/validate-internal-links.mjs`, the `path.basename(filePath) !== 'index.html'` skip), so the generated `.md` copies are never link-checked. If this list drifts into a dead URL, nothing in CI notices — only the LLM readers, and they will not file an issue about it.

packages/dnb-design-system-portal/src/docs/index.mdx
tujoworker: Added `hideEditLink: true` to the home page frontmatter. The built homepage test now also asserts that the edit link is absent.

packages/dnb-design-system-portal/src/e2e/pageNavigation.spec.ts
tujoworker: Split the cards into literal-route tests so both destinations enter the partial E2E allowlist. The Develop case now asserts the real `Getting Started` heading as well as client-side navigation.

tools/eufemia-llm-metadata/src/extensions/mdx/home.ts
tujoworker: Moved the title, intro, actions, and resources into one Portal-owned JSON source consumed by both `Home` and the Markdown renderer. That also keeps the trailing slash and all resource URLs identical.

packages/dnb-design-system-portal/src/shared/menu/Card.tsx
tujoworker: Moved `cardItemStyle` into `Card.module.scss` beside `Card.tsx`. `ListComponentsOverview` keeps its own local style instead of sharing ownership.

packages/dnb-design-system-portal/vite.config.ts
tujoworker: Removed `remark-gfm-react-markdown` from `optimizeDeps`; it was unrelated to this change and did not need to be part of the PR.

tujoworker:
Worked through the home-page feedback and pushed the final stacked revision in `ba29126e339`.

Highlights:
- homepage content now has one Portal-owned source used by both React and generated Markdown
- Design and Develop both use client-side routing; Develop also asserts the real destination heading in the partial E2E build
- homepage hides the edit link
- MenuCard owns `Card.module.scss`
- removed the unrelated `remark-gfm-react-markdown` prebundle entry
- restored Brand, What's new, and the separate Eufemia change log entry point
- removed the now-unused `changelogVersion` build/package pipeline
- kept registry imports ordered
- added homepage and general Portal hydration regression coverage

Validation: 67 focused unit tests, both type checks, JS/SCSS lint, production E2E build, and five focused homepage Chromium tests. The inline review threads were replied to and resolved.

packages/dnb-design-system-portal/src/e2e/pageNavigation.spec.ts
```diff
@@ -2,6 +2,38 @@ import { test, expect } from '@playwright/test'
...
+  page.on('console', (message) => {
+    if (message.type() === 'error') {
+      errors.push(message.text())
+    }
+  })
+
+  await page.goto('/')
+  await waitForApp(page)
+
+  expect(errors.filter((message) => message.includes('#418'))).toEqual([])
```
langz: This assertion can never fail, so `/` still ships the hydration bail-out — now with a green test in front of it. Same goes for the `/uilib/components/button/` variant further down.

`playwright.config.e2e.ts` runs Firefox, and there `message.text()` for this error is just `"Error"`; the code only lives in `message.args()[0].message`. (`page.on('pageerror')` stays empty, so that is not the way round it either, and a dev build never emits `#418` at all since it is a production-only minified code.)

Against the preview, in Firefox:

```
--- firefox / ---
  errors.filter(text includes #418) -> []               <- the assertion
  message.text()                    -> ["Error"]
  message.args()[0].message         -> ["Minified React error #418; ..."]
  React ids in the DOM              -> _R_: 0, _r_: 21  <- client render
```

`eufemia.dnb.no/` is clean for comparison.

Reading `message.args()[0].message` makes it catch the real thing, and then it fails, which I think is the honest outcome. Asserting that the prerendered `_R_` ids survive would be even more direct.

Chasing the mismatch itself still belongs in #9296 — I would just rather not merge coverage that reads as if it were already handled here.

langz (commented):
Thanks for working through all of this. Routing, the shared `HomeData.json`, the hidden edit link, `Card.module.scss`, the e2e allowlist and the `changelogVersion` cleanup all check out on my side, and `/design-system` has an entry point again — I have resolved those threads.

One thing I would still like fixed before merge: the new hydration coverage cannot fail, and the page it was added for still does not hydrate.

_Update 22 Sep: fixed in `edcc93f7964` and re-verified — see the approval below._

langz:
Thanks for the patience on this one, @tujoworker — nine commits, one `MainMenu` retired to a nice farm upstate, and a front page that no longer asks visitors to edit it the second they arrive. Good trade.

Apologies for turning `/` into a small science project along the way. In my defence, it is the one page where literally everyone shows up, so it gets the pedantic reviewer.

One assertion away from a yes from me.

packages/dnb-design-system-portal/src/e2e/pageNavigation.spec.ts
tujoworker: Good catch. Both hydration checks now inspect the console argument value, including Error.message, instead of relying on message.text(). The underlying #9296 hydration fix has also been rebased onto current main, and the corrected Firefox hydration checks are green on the latest #9296 run. #9295 Firefox E2E is green too.

tujoworker:
The hydration feedback is fixed in `edcc93f7964`: both homepage and general Portal checks now inspect the actual console argument and `Error.message`, so Firefox React #418 errors cannot pass unnoticed. The parent hydration mismatch is fixed in #9296. This PR is cleanly stacked, mergeable, approved, and all 20 current-head checks are green.

langz (approved):
The last item is sorted. `captureConsoleErrors` now reads `message.args()` and unwraps `Error.message`, so Firefox's bare `"Error"` text can no longer hide a React #418 — and `/` genuinely hydrates now.

Re-checked the preview at `edcc93f7964`:

```
/                          clean   _R_ ids survive, no _r_
/uilib/getting-started/    clean
/quickguide-designer       clean
/uilib/components/card/    clean   _R_: 3, _r_: 0
/design-system             clean
```

Both hero cards keep the `window` marker across the click and land on the right heading, the resource links all resolve, `index.md` matches the rendered page one-for-one, and the head still carries the title, description and `og:` tags. Nothing left from the earlier rounds — all threads stay resolved.

Nice work getting `/` back to a single prerendered page. Approving.

tools/eufemia-llm-metadata/src/extensions/mdx/home.ts
```diff
@@ -0,0 +1,47 @@
...
+        ),
+        '',
+        '## Resources',
+        '',
+        ...homeData.resources.map(
+          ({ title, url }) => `- [${title}](${url})`
+        ),
+      ].join('\n')
+
+      return content.replace(/<Home\b[^>]*\/>/g, markdown)
```
langz, line 44: String replacement, so `$&`, `` $` ``, `$'` and `$$` inside `markdown` are substitution patterns rather than text. `HomeData.json` has no `$` today, and nothing would tell us if one arrived — `validate:links` does not read `.md`. Every other extension in this folder uses the function form, including `listComponentsOverview.ts` and `relatedComponents.ts`, which inject markdown exactly like this.

```suggestion
      return content.replace(/<Home\b[^>]*\/>/g, () => markdown)
```

packages/dnb-design-system-portal/src/e2e/pageNavigation.spec.ts
```diff
@@ -33,6 +33,37 @@ async function readConsoleMessage(message: ConsoleMessage) {
...
+  expect(
+    await page.evaluate(
+      () =>
+        (window as Window & { portalNavigationMarker?: boolean })
+          .portalNavigationMarker
+    )
+  ).toBe(true)
+}
+
+test('home page hydrates without recoverable errors', async ({ page }) => {
```
langz, line 54: `#418` is the minified code, so this and `hydrates a Portal page without recoverable errors` only assert anything against a production build. On the dev server React logs the unminified `Hydration failed because...` instead and both pass regardless — which is why the other prerender tests here open with `if (await isDev(page)) return`.

That guard, or widen the filter to `#418|Hydration failed`.

langz (commented):
Two small ones from the final pass. Neither blocks the merge.

langz:
Approved — see [the review above](https://github.com/dnbexperience/eufemia/pull/9295#pullrequestreview-5275580095). `edcc93f7964` closes out everything from the earlier rounds, and I have re-checked the preview myself rather than taking the CI run on trust.

The two notes I left afterwards are non-blocking. Pick them up or wave them through, neither should hold the merge.

PR #9215, opened 2026-09-02
chore(Portal): remove theme readiness markers

tujoworker:
The portal labels many components as ready for Sbanken or Carnegie based on manually maintained page metadata. Remove the sidebar badges, the readiness metadata from 99 MDX pages, and the authoring guide for adding these labels so the portal no longer presents them as a readiness guarantee.

PR #9296, opened 2026-09-08
docs(Portal): integrate new SidebarMenu navigation

tujoworker:
Replaces the Portal navigation with SidebarMenu, defines the navigation structure in page metadata, and adds the Portal-specific layout and regression coverage.

The reusable compact, responsive, resizable navigation behavior is reviewed separately in #9393. The Portal home MDX migration is reviewed independently in #9295.

Deploy preview: https://docs-portal-sidebar-menu.eufemia-e25.pages.dev/


langz:
`/uilib/helpers` and its three sub-pages are no longer in the sidebar. On `main` the menu links `/uilib/helpers`, `/uilib/helpers/classes`, `/uilib/helpers/functions` and `/uilib/helpers/sass` — on the branch preview none of them appear. The pages still build and respond, and `/uilib/helpers/functions#debounce` is linked from five other docs pages, so the content is live, just unreachable from the menu. Visiting `/uilib/helpers/` also leaves the sidebar without an `aria-current` item at all, so there's no indication of where you are.

`/uilib/extensions` is dropped as well, but all four of its children (`forms`, `payment-card`, `sidebar-menu`, `vipps-wallet-button`) are placed elsewhere, so that one looks intentional.

It comes from `createConfiguredRoot` in `SidebarMenuData.ts` — `orderedChildren` is built and then thrown away when `includePageAs`, `pageParent` and a synthetic `path` are combined, which is exactly what `uilib.mdx` does:

```ts
...(config.includePageAs && !config.pageParent ? [ … ] : []),  // skipped, pageParent is set
...(pageOwnsChildren ? [] : orderedChildren),                  // discarded
```

The `pageParent` docs added in this PR say it places the page link "while leaving the synthetic tree in place", so I don't think this is intended. Giving `uilib/helpers` its own `sidebarMenu.parent` would fix it, or `pageOwnsChildren` could require `!config.pageParent`.

Worth noting the accepted inline snapshot in `SidebarMenu.test.tsx` feeds `uilib/helpers` and `uilib/extensions` into the fixture and contains neither, so the loss is baked in and tests won't catch a regression here.

langz:
The header background is gone, so page content scrolls visibly through it. `.headerStyle` loses both `background-color` and `border-bottom` in this PR, and the only replacement added is `.leadingStyle::before`, which sits inside `allBelow(medium)` and only covers the hamburger and logo.

I compared the header band pixels at scroll 0 and scroll 1400 on `/uilib/components/button/` at 1440px: on `main` they are byte-identical (opaque), here they change. `elementsFromPoint` inside the band returns `code.dnb-code` and `span.dnb-span`, and visually the body text runs into the search field:

`For the secondary variant, the recommended sizes are default and larg|` — cut off mid-word where the search box starts.

The `border-bottom` removal is asserted in `responsiveness.spec.ts`, so that part is clearly intended — but nothing covers the background, so I think that one is collateral. Re-adding `background-color: var(--token-color-background-neutral)` on `.headerStyle` should be enough.

langz:
The header collapses to 488px on a 1440px viewport on the 404 page (any page passing `hideSidebar`).

`Layout.module.scss` resets `--aside-width: 0` for `.hideSidebarStyle`, and `StickyMenuBar.module.scss` does `width: calc(100% - var(--aside-width))`. A unitless `0` can't be subtracted from a percentage, so the declaration is invalid at computed-value time and `width` falls back to `auto` — the flex header then shrinks to its content.

Verified in the browser: `CSS.supports('width', 'calc(100% - 0)')` is `false` while `calc(100% - 0px)` is `true`, and setting the custom property to `0px` snaps the header back to 1440px.

So `--aside-width: 0px` fixes it. `left: var(--aside-width)` is unaffected since a bare `0` is a valid length there, and `.fullscreenStyle` never hits it because the header isn't rendered in fullscreen.

langz:
The resize handle has no limits — the rendered button exposes `aria-valuemin="1"` and `aria-valuemax="1440"`. Dragging it fully right at 1440px leaves the sidebar at 1424px and the header at 16px: the search field is an unreadable sliver and the GitHub/Figma/tools buttons are gone entirely. Dragging to the other extreme forces a horizontal scrollbar via `min-width: 20rem` on the scroll content.

`SidebarMenu.ResizeHandle` isn't given `minWidth`/`maxWidth` in `SidebarMenu.tsx`, so it falls back to the extension defaults of `1` and `window.innerWidth`.

It resets on reload since the width isn't persisted, so it's recoverable — but something like `minWidth={240} maxWidth={560}` would stop you getting there.

Semi-related: the width isn't persisted at all, while open-accordion state and sidebar scroll position both are. Was that deliberate?

langz:
`fragments.mdx` and `pagination.mdx` replace `accordion: true` with `theme: ['sbanken', 'carnegie']` rather than just dropping it:

```diff
-accordion: true
+theme: ['sbanken', 'carnegie']
```

#9215 (9f3609ea32f) removed exactly that line from exactly those two files — 100 removals across 103 files, 0 additions — and it's already an ancestor of this branch. So this looks like a rebase resolution that kept the wrong side. The full history of the line in `pagination.mdx` is now: added in #2569, removed in #9215, re-added here.

Under the Carnegie theme it also produces a live 404. The new `.portal-sidebar-menu__theme-badge--carnegie` rule in `SidebarMenu.module.scss` points at `/images/carnegie_badge.png`, which was deleted in that same commit, so the badge renders as a plain black circle and each render fires a failed request. Confirmed on the branch preview: `404 carnegie_badge.png`.

`uilib/extensions/sidebar-menu.mdx` carries the same `theme:` line from the base branch. Dropping it from all three removes both problems.

langz (approved):
Approving to unblock the stack — the direction here is right, and moving the navigation structure into page metadata is a good call.

The five points above should be sorted before this reaches `main` though, in particular the missing `/uilib/helpers` pages. Two of them are one-liners (`--aside-width: 0px`, and the header `background-color`).

Note the red visual regression looks unrelated to this PR: the 45 failures are identical to the ones on #9294 and there are none unique here — likely the icon path refactor in #9328 needing regenerated baselines on `main`.

langz:
Heads up, this one has a merge conflict now — `packages/dnb-design-system-portal/src/e2e/focusmode.spec.ts`, from #9335 landing on main.

The red checks here are mostly inherited from the missing locale key I commented on in #9294, so they should clear once that's fixed and this is rebased.

langz:
Removing the platform sections left one e2e test behind that now contradicts its new sibling.

`matches the section selector surface from Figma` still does `await menu.getByRole('combobox').click()`, while the new `omits the platform selector without other platforms` in the same `test.describe('Portal SidebarMenu')` block asserts `await expect(menu.getByRole('combobox')).toHaveCount(0)`. Both share the `beforeEach` that goes to `/uilib/components`, so they can't both hold.

`SidebarMenu.tsx` has no `sections`/`Dropdown` references left and I don't see a `.dnb-dropdown` in the sidebar on the branch preview, so the click should just time out. E2E is still pending on this commit, so it hasn't surfaced yet — separate from the locale key in #9294.

Guessing that test was meant to go with the sections; the extension's own screenshot tests still cover the dropdown surface. The rest of the removal looks clean — no leftover `sections`/`sectionLabel`/`portal-ios-overview` references and the icon imports are gone.

langz:
The new resize handle sits outside every landmark, which trips axe on each page that has a sidebar.

`SidebarMenu.tsx:211` renders `SidebarMenu.ResizeHandle` as a sibling of `ResponsiveInline`/`ResponsiveDrawer` rather than inside the `nav` that `menu()` returns, so it lands directly under the layout wrapper:

```
div._portalStyle > div._wrapperStyle > button[role="separator"].dnb-sidebar-menu-resize-handle
```

axe's `region` rule ("all page content should be contained by landmarks") then fires on `.dnb-sidebar-menu-resize-handle`. On this preview I get it on `/uilib/components/button`, `/uilib/getting-started` and `/icons`; the same run against production is clean and has no such element, so it arrives with this PR.

Worth being clear about the weight: `region` is tagged `best-practice` in axe (also RGAA 9.2.1) rather than WCAG 2 A/AA, so this is tidiness rather than a blocker. Rendering the handle inside `aside#portal-sidebar-menu` would clear it, and reads better anyway — a window splitter belongs with the pane it resizes, which is the element it already names in `aria-controls`.

Two things I checked while in there, both fine or already known: the `aria-controls="portal-sidebar-menu"` target does resolve (`aside#portal-sidebar-menu` is present), and `aria-valuemin="1"` / `aria-valuemax="1440"` are unchanged from [the earlier comment](https://github.com/dnbexperience/eufemia/pull/9296#issuecomment-5631187067).

langz:
On first or second render or such, the search bar shifts/stutters or changes width:

https://github.com/user-attachments/assets/d9a396f7-bd02-4052-9513-c41615257085

langz:
On first or second render or such, the search bar shifts/stutters or changes width, also the arrow/caret of the opened menu item sort of "blinks" or it's first hidden, then appears as open:

https://github.com/user-attachments/assets/f6d0c40c-fb9c-4700-a65d-731c6eb7dde6

langz:
I'm no designer, but my first thought is that it's quite a lot of space between the menu items in the sidebar menu:

image

langz:
I find it a bit strange that VippsWalletButton is listed under actions in the sidebar menu, but not under actions in this list:
https://docs-portal-sidebar-menu.eufemia-e25.pages.dev/uilib/components/overview#actions

image

I guess this could happen in other places as well, I've not cross-checked all these other categories that we group components into

langz:
The Portal's hamburger animation reaches into the design system and gives a Figma-owned icon a haircut.

`StickyMenuBar.module.scss:132-141` staggers `#toggle-sidebar-menu .dnb-icon path` via `:nth-child(2)`/`(3)`, which only works if the icon is three separate `<path>` elements — hence `d="M1 14h14M1 8h14M1 2h14"` being split across `hamburger.svg` and the auto-generated `hamburger.tsx` (`/** This file is auto generated by convertSvgToJsx.ts */`). Verified on the preview: three paths, delays `0.2s` / `0.27s` / `0.34s`.

Two snags with paying for it there:

- `hamburger.svg` is Figma-managed. It's in `src/icons/dnb/icons-svg.lock` with node id `21:0` and a `figma-alpha-api` URL, and `assets/icons` is exactly `extractIcons`' `destDir` — so the next icon sync re-downloads the single-path version and quietly takes the animation with it.
- `hamburger_medium.svg` is untouched and still one path, so the two sizes now disagree and the stagger silently no-ops at `iconSize="medium"`.

Rendering the three bars locally in the Portal would get the same effect without the design system footing the bill.

langz:
A couple of leftovers, I think from the platform sections removal.

`platform` is threaded through the GraphQL query (`SidebarMenu.tsx:87` and `:112`), three type definitions and four assignment sites in `SidebarMenuData.ts` (`:225`, `:314`, `:361`, `:402`), plus `platform: web` in 9 MDX files — and nothing ever reads it. `renderSidebarMenuItems` picks its props explicitly so it never reaches the DOM, `grep -r platform src/extensions/sidebar-menu` comes back empty, and no test touches it. It's a very well-plumbed pipe to nowhere.

Same in `SidebarGraphics.ts`: `cog`, `laptop`, `layout_grid` and `play` are imported and registered (`:37-44`, `:78-85`), but no `icon:`/`pageIcon:` frontmatter references them — not here, and not on #9295 either. `graphics[item.icon]` is the only lookup, so they're purely along for the ride.

`brush`, `edit`, `lightbulb`, `newspaper` and `home` are all genuinely in use, so it's just those four.

langz:
Two assertions that can't fail anymore.

`responsiveness.spec.ts:98` asserts `#toggle-main-menu` has count 0 — but this PR deletes `HomeButton`, so that id now appears in exactly one place in the whole repo: that line. Confirmed 0 on the preview. Green, but for the wrong reason.

And `pageNavigation.spec.ts:24` (`should not be able to open portal tools`) drops its `await page.click('#portal-tools', { force: true })`, so with `javaScriptEnabled: false` it now only checks that `#switch-grid` is absent — which holds on every page whether or not a menu exists. `#portal-tools` is still rendered (count 1 on the preview), so it wasn't removed because the element disappeared. Was the click failing for some other reason?

langz:
`should have same amount of elements` → `should list each element once` gives up the parity check, and I don't think it had to.

The old test compared the sidebar count against the page list. The new one only asserts the page list has no duplicates, which would still pass if the sidebar and the page disagreed completely. On `/uilib/elements/` on the preview I count 10 child links in the sidebar and 10 in `#tab-bar-content` — so the numbers still line up, the selector just needed the same `data-sidebar-menu-id` treatment that `components` and `extensions` got in this PR.

Small one in the same file: `expandAllSidebarItems` now ends on `await page.waitForTimeout(500)` (`:11`). It'll be fine right up until it meets a slow CI runner.

langz:
The new frontmatter is doing a lot of work for how little of it is written down.

`sidebarMenu` is now used in 32 MDX files across 13 distinct keys: `order` (27), `title` (26), `parent` (19), `static` (12), `root` (10), `platform` (9), `path` (9), `includePageAs` (7), `id` (7), `icon` (7), `pageOrder` (6), `pageParent` (2), `hideStatus` (1). `documentation.mdx` documents `sidebarMenuPlacement`, `sidebarMenuDividerBefore` and one `pageParent` example. So the two most-used structural keys, `parent` and `static`, aren't mentioned, and the next person adding a page gets to reverse-engineer `SidebarMenuData.ts` to find out where it lands.

`groups` has the opposite problem — typed, implemented in `addConfiguredGroups` and covered by tests, but not used by a single MDX file.

Unrelated drive-by while I was in there: `prerender-utils.ts` loses the `buildRedirectHtml` JSDoc block in this PR, which looks accidental.

langz:
`uilib/intro.mdx` quietly loses `fullscreen: true`.

Production `/uilib/intro/` renders with no header and no sidebar; on the preview it gets both, and `main` narrows from 1152px to 768px. Given it's now a "Quick intro" entry in the Guides menu that makes sense — a menu item pointing at a page that hides the menu would be a bit rude — just flagging it since it isn't in the description and nothing covers it.

langz:
Just a comment about that I find it interesting that the search bar and buttons in the top right corner has a fixed position without sort of a "sticky banner" or such. It just "floats" there ⛵ 

Usually I feel like there's sort of a full width banner that's fixed to the top, not only search bar and buttons with what appears as a "transparent background".

https://github.com/user-attachments/assets/7f50c30e-20ca-41c6-ad9f-60aace791bd2


 But I like this approach a lot on my phone:

image

packages/dnb-design-system-portal/src/docs/uilib/intro.mdx
```diff
@@ -1,5 +1,8 @@
 ---
-fullscreen: true
```
langz, line 2: I guess this makes sense, however when pressing the "next"-button, it gets into fullscreen mode:

Image

Image

langz:
I guess we still have to support menu items that works as accordion, and menu items that works as accordion and as a link. But from a user perspective, I find it a bit strange that 2 things that looks visually the same behave different when I click them. See in the following video where all my clicks result in an accordion opening, but the last one navigates me to a page and also closes the menu. All my first clicks on visually same items did not have that behavior:

https://github.com/user-attachments/assets/d93d0277-7fcd-465d-a115-e6377ec1ad20

langz:
Not sure if it's possible to also just dim the top section that's above the menu here, as we do with what's to the right of the menu when opened.
Instead of just making it black when the menu opens.

I guess it applies to both top of the menu, and bottom of the menu. 

https://github.com/user-attachments/assets/1ae6328d-d400-47bc-9ca7-669724ffa578

https://github.com/user-attachments/assets/f24c9347-29a5-4d72-bf0c-b36d689fae03

langz:
This is a really nice piece of work. 🎉

Really looking forward to having this on the Portal 👏

langz:
**CI, as of `da8392ca`** — `Run e2e tests` and `Run visual-regression tests` are both red, `main` is green on both, and the counts haven't moved since `329a6b6`: **6 e2e failures / 150 passed** and **24 screenshot failures**. The branch has also gone `CONFLICTING` against `main` (`declaration.d.ts` and `ListComponents.tsx` changed on both sides, `SidebarMenuContext.tsx` deleted here but modified upstream).

### The four drag specs — not the runner, and not where the pointer lands

`sidebarResize.spec.ts:60` / `:392` / `:435` and `sidebarMenuCompact.spec.ts:160`.

`:392` is a controlled experiment inside a single test. The real-mouse drag at `:412-416` passes (`238px`, then `240px` after the spring-back), and the very next drag, `dragSidebarToWidth(page, resizeHandle, 700)` at `:431`, gets `240px` instead of `560px`. Same handle, same test, same viewport — so pointer position can't be the difference.

What changed is the delivery. `dragSidebarToWidth` moved from `page.mouse.move/down/move/up` — which is what `main` still has, and what `:412-416` still does — to `resizeHandle.dispatchEvent('pointerdown', { pointerId: 1 })` plus `window.dispatchEvent(new PointerEvent(…))` at `:28-55`. `sidebarMenuCompact.spec.ts:177-190` does the same. `locator.dispatchEvent` calls `element.dispatchEvent()` directly, so hit-testing never enters into it.

The portal suite runs Firefox (`playwright.config.e2e.ts:20`), and Firefox rejects a `pointerId` that doesn't belong to a live pointer. Minimal repro against the same Playwright build, one button with a `pointerdown` listener:

```
firefox  synthetic:  ["handler-start","NotFoundError: Element.setPointerCapture: Invalid pointer id"]
firefox  real mouse: ["handler-start","after-capture"]
chromium synthetic:  ["handler-start","after-capture"]
chromium real mouse: ["handler-start","after-capture"]
```

`SidebarMenuResizeHandle.tsx:302` is `event.currentTarget.setPointerCapture?.(event.pointerId)` — `?.` guards a missing method, not a throwing one — so the `NotFoundError` leaves `handlePointerDown` before `startResize()` on `:303` ever runs. The drag silently does nothing, which is exactly the `384px` / `240px` / `384px` in the three `sidebarResize` failures and the `scrollWidth > clientWidth` poll at `sidebarMenuCompact.spec.ts:200`.

Smallest fix is to put `dragSidebarToWidth` back on `page.mouse` the way `main` has it, and give the compact spec the same treatment. Wrapping `setPointerCapture` in a `try`/`catch` would make the handle survive synthetic pointers too, but I'd rather not reshape product code to suit a test when the test can drive a real one.

### The other two e2e failures

`pageHeading.spec.ts:12` and `routeFocus.spec.ts:41` time out clicking a sidebar link with `element is not stable` and a neighbouring item intercepting the pointer. Neither file is touched by this PR and both pass on `main`, so the taller animated menu is what makes them miss. `expandAllSidebarItems` in `pageLists.spec.ts` waits for `.dnb-height-animation--animating` to settle; these two don't.

### The 24 screenshots

Unchanged, and all `SidebarMenu for ui` / `SidebarMenu for sbanken` — cause in the `ResponsiveProvider` comment.

langz (approved):

langz:
`preload-styles.runtime.ts:23` still hides `nav#portal-sidebar-menu`, but the sidebar is an `<aside>` now — the e2e specs in this PR were updated from `nav#portal-sidebar-menu` to `#portal-sidebar-menu` for exactly that reason. That inline `<style>` is what keeps `?fullscreen`, `?focusmode` and `?data-visual-test` from flashing the prerendered sidebar before hydration.

I've now measured it rather than just reading the selector. Loading `/uilib/components/button/?fullscreen` with the app bundle blocked, so only the inline script runs:

- the injected `<style id="portal-preload-style">` is present and contains `nav#portal-sidebar-menu{display:none!important}`
- `document.querySelector('nav#portal-sidebar-menu')` → `null`
- `header.sticky-menu` → `display: none` (that rule still works)
- `#portal-sidebar-menu` → `display: flex`, 384px wide, visible
- `.dnb-app-content` → starts at `left: 303px` instead of `0`

So in fullscreen the header disappears on time and the sidebar doesn't, and the content column sits 303px off until React takes over.

Dropping the `nav` is the quick part, but it isn't sufficient on its own: injecting the corrected `#portal-sidebar-menu{display:none!important}` into that same page leaves `.dnb-app-content` at `left: 303px`. The reserved space lives above the aside, so the rule needs to reach the wrapper too. `.dnb-app-content{margin-left:0!important}` on the next line no longer does anything either, since `.contentStyle` computes `margin-left: 0px` already.

Nothing will catch a regression here — `visual-test-detection.test.ts:55` and `:68` assert the CSS string verbatim, `nav` included.

langz:
Non-blocking, but the two new `prefers-reduced-motion` escape hatches can't win on specificity.

`Layout.module.scss:231` and `StickyMenuBar.module.scss:58` compile to a bare `.contentStyle` / `.headerStyle` (0,1,0), while the rule they're cancelling is `:global(.dnb-sidebar-menu-resize-handle--transition-ready) &` → `.dnb-sidebar-menu-resize-handle--transition-ready .headerStyle` (0,2,0). The media query doesn't change that, so the 300ms `left`/`width` transition still runs with reduced motion on. Nesting the media query inside the `--transition-ready` block (or reusing `utilities.reducedMotion()`, which the extension styles already do) covers both.

While in there: `Layout.module.scss:224` transitions `margin-left` on `.contentStyle`, which it no longer sets — the offset moved to the flow-based wrapper.

langz (approved):

langz:
Correction to my CI comment above — I read the Playwright output backwards on the geometry assertion, sorry.

In `expect(layout.headerLeft).toBe(layout.sidebarRight)`, `Received: 61` is `headerLeft` and `Expected: 384` is `sidebarRight`, not the other way round. Checking the preview at 1280×900: `--aside-width` resolves to `4rem`, the header and `.dnb-app-content` both start at 64px and the inline wrapper is 64px wide — the layout is correct. The `<aside>` stays 384px and is clipped by the wrapper, which is the box the assertion happens to measure. Same for `toHaveCSS('display', 'flex')`: `.dnb-sidebar-menu-responsive-aside` sets no `display`, so `block` is right.

So those two are assertions left over from before compact mode rather than a broken header — the job still needs to go green, but there's nothing to chase on the layout.

The 24 visual failures are a different matter though, and I think they're pointing at a real bug — separate comment below.

langz:
The Portal's `ResponsiveProvider` makes every `SidebarMenu` on a docs page collapse to an icon rail.

`Layout.tsx:161` wraps the whole page — header, sidebar *and* `Content` — in `ResponsiveProvider`, and `SidebarMenuRoot.tsx:579` now derives `dnb-sidebar-menu--compact` from `useOptionalSidebarMenuResponsive()?.isCompact`. Any `SidebarMenu` rendered inside page content therefore inherits the Portal's compact state, even though it has nothing to do with the Portal's own navigation.

Measured on the preview, `/uilib/extensions/sidebar-menu/demos/`, no query flags:

| viewport | `sidebar-menu-declarative` | `sidebar-menu-data` |
| --- | --- | --- |
| 1280px | `--compact`, action 44px, label 0px | `--compact`, action 44px, label 0px |
| 1400px | normal, action 880px | normal, action 848px |

So on any laptop inside the Portal's compact range the two static examples render as unlabelled icon rails. #9393 serves the same demo page from an extension tree that differs by four lines of SCSS and doesn't do this — it has no Portal-level provider — which is also exactly what the 24 failing `SidebarMenu for ui` / `SidebarMenu for sbanken` screenshots are catching, and #9393's visual job is green. Worth saying explicitly: those baselines are right, they shouldn't be regenerated.

Where to fix is a judgement call. The Portal can't easily narrow the provider, since the trigger lives in the header and the aside in the wrapper, so it may be cleaner for `SidebarMenuRoot` to take compact from the `ResponsiveInline` it is actually inside rather than from the provider — which would land in #9393.

Still reproducing on `da8392ca`, and the 24 screenshots are still red, so this one is outstanding.

langz:
Some of this is the intended consolidation, but the prerendered payload grows on every page, so probably worth a conscious yes rather than discovering it later.

`/uilib/components/button/`, production vs preview:

- page: 370,812 → 527,780 bytes (29.2 → 45.5 KB gzipped, +56%)
- `#portal-sidebar-menu` subtree: 82,109 → 230,594 bytes (7.8 → 23.3 KB gzipped)
- inside the sidebar: 162 → 305 anchors, 162 → 317 `<li>`, 31 accordions of which 27 ship collapsed

Two things stack here: `shouldIncludeSidebarPrefix` merges `uilib`, `icons`, `quickguide-designer` and `contribute` into one menu, and every collapsed accordion still ships its children so the pre-hydration script can open them before paint. Both follow from the design — it's ~16 KB gzipped and a 305-anchor DOM on each of the prerendered pages that I wanted to put a number on.

Cheap and related: the data pipeline in `SidebarMenu.tsx` isn't memoised, so it re-runs on every render of `SidebarLayout`. Against the real docs set in the portal's test environment I measure ~39ms for `prepareNav` + `groupNavItems` (pre-existing) plus ~27ms for `applySidebarMenuPlacement` + `createUilibSidebarStructure` + `groupComponentsByCategory`, so ~66ms a pass. `SidebarLayout` now also subscribes to the responsive context via `useResponsive()`, so that runs again on hydration and on every drawer/compact change, not only on route changes. A `useMemo` keyed on `location.pathname` and `showAll` would cover it.

langz:
Everything still open from my side, in one place. None of it blocks — the one blocking item is in the separate comment about the compact rail.

## Never matches or never runs

- `SidebarMenu.module.scss:43-45` still lists `uilib-design-language-web`, `uilib-usage-web` and `uilib-about-web`. The ids actually rendered are `uilib-components-category-*`, `uilib-usage-customisation-*`, `uilib-layout-*`, `uilib-elements`, `uilib-extensions-forms-Form-Section` and `uilib-about-the-lib-releases-*`, so only the component categories get the muted title colour.
- `Layout.module.scss:223` `&:has([data-portal-home])` — nothing in this branch emits `data-portal-home`; it arrives with #9295.
- `scroll-position.runtime.ts:65` calls `restoreScrollPosition({ smooth: true, restoreWindow: false })`, which is a no-op now that the element branch is gone — the body only does something when `restoreWindow` is true, and it's the only caller passing `false`. `smooth` is never `true` anywhere else either, yet `scroll-position.test.ts` asserts `behavior: smooth ? 'smooth' : 'auto'` is present, so the test pins code that can't run.
- `SidebarMenu.test.tsx:1116` (`does not configure portal accordions as initially open`) checks that no MDX sets `sidebarMenuOpen` or `sidebarMenu.open`. Neither exists in `SidebarMenuConfig` or in the GraphQL query, so it can't fail.
- The theme badge has nothing left to render — no page carries `theme:` frontmatter any more, which leaves the `theme` field in the query (`SidebarMenu.tsx:75`), `useTheme()` (`:128`), `addThemeBadges` (`:190`), `ThemeBadge` (`:203`), `themeTitles` (`SidebarMenuData.ts:158`) and the `.portal-sidebar-menu__theme-badge` block with nothing to do. `addThemeBadges` also walks every item on each render to attach `suffix: undefined`. If `theme:` is coming back it's just waiting for data — worth knowing for that day that the badge is a `<span title="…">` with no text content, so screen readers won't announce it.
- `sidebarMenuPlacement` has zero pages using it (the only hit in `src/docs` is the example in `documentation.mdx:15`) while being carried through the query (`SidebarMenu.tsx:76`), the `NavItem` type and four functions in `SidebarMenuData.ts`. `pageIcon` is documented at `documentation.mdx:45` and likewise has no users.

## Still worth a look

- **Omitting `sidebarMenu` fails silently.** A typo in `sidebarMenu.parent` throws `Sidebar menu parent not found` and fails the prerender, which is the right call. Leaving the config out doesn't: adding `uilib/new-section.mdx` plus `uilib/new-section/child.mdx` with no `sidebarMenu` drops both from the menu without a warning, since `createUilibSidebarStructure` returns only the configured roots. Today that hides only `uilib/shared/media-query/properties`, an include-only fragment, but the next top-level section will go the same way.
- **The outer landmark lost its name.** Production renders `nav#portal-sidebar-menu` with `aria-labelledby="toggle-sidebar-menu"`; here it's `aside#portal-sidebar-menu` with no name, wrapping a `nav` with `aria-label="Section Content Menu"`, so the landmark list gains an unnamed `complementary` next to the named `navigation`. axe is clean either way, it just adds a stop with nothing to say.
- **The search field still shifts after hydration.** `.portal-search` is steady at 280px, but the `<input>` inside it goes 260 → 236px about 200ms in. `SearchBar.tsx:45` reads the `⌘ K` / `Ctrl K` hint through `useSyncExternalStore`, which returns `null` on the server on purpose, so the `<kbd>` only mounts after hydration. Reserving the space during SSR would settle it.
- **The resize handle is only grabbable in its leftmost quarter.** At 1440px the handle's box is `x: 378 → 402` (24px) while the `<aside>` ends at `384`, and hit-testing each pixel with `elementFromPoint` gives 6 of 24; the overhang resolves to `main#dnb-app-content`. The handle is `position: absolute; z-index: 3` inside the aside while `.contentStyle` carries `z-index: 2` a level up — different stacking contexts, so the higher number doesn't help. Keyboard resize is unaffected, and the specs work around it now by pressing at `x + 5`.

## Questions

- **How is this meant to land relative to the neighbours?** `72c4821` carries the full extension diff — `SidebarMenuResponsiveAside.tsx`, the `SidebarMenuResponsive` rework, the `SidebarMenuRoot` compact scoping and +387 lines of SCSS. That is the same work as #9393, which is still a draft with no reviews on it, so if this merges first that half lands on `main` reviewed only through this PR. It may well be the plan — I just don't want to approve on the assumption someone else has the other half. Which way are you thinking?
- **What should the sections outside the unified menu show?** `shouldIncludeSidebarPrefix` merges `uilib`, `icons`, `quickguide-designer` and `contribute`, so a designer opening the Quick Guide gets the full UI library tree. The other side is `/brand/`, where the sidebar is a single item, with `/license/` and `/design-system/` similar — and those don't get the Home entry either, because `createUilibSidebarStructure` returns early before `includeHome` when nothing in the section is configured, so the only way back out is the logo.
- **`scopeId` is consumer-supplied and lands in a selector string.** `SidebarMenuResponsive.tsx:618` does `` `[data-sidebar-menu-responsive-scope="${scopeId}"]` `` and that string is rendered into the inline `<style>`. Nothing reachable from the Portal since it's a literal — just asking whether it's worth escaping at the boundary now that it's public API, plus a word in the docs about keeping it unique per provider.
- **Can the new hydration guard fail in dev?** `pageNavigation.spec.ts:148` filters on `#418`, the minified code, so a dev run logging React's full non-minified message would pass regardless. Its sibling in the same block opens with `if (await isDev(page)) return`.

## One measurement, for a conscious yes

The prerendered payload grows on every page. `/uilib/components/button/`, production vs preview: page 370,812 → 527,770 bytes (29.2 → 45.5 KB gzipped, +56%), the `#portal-sidebar-menu` subtree 82,109 → 230,594 bytes, and 162 → 305 anchors inside it with 31 accordions of which 27 ship collapsed. Runtime via CDP `Performance.getMetrics`, three loads each at 1440×900: `TaskDuration` 1.181s → 1.547s, `RecalcStyleDuration` 0.232s → 0.384s, DOM nodes 3,462 → 4,307. Both drivers follow from the design — merging four sections into one menu, and shipping collapsed children so the pre-hydration script can open them before paint.

One cheap lever: the data pipeline in `SidebarMenu.tsx` isn't memoised, so it re-runs on every render of `SidebarLayout` (~27ms for `applySidebarMenuPlacement` + `createUilibSidebarStructure` + `groupComponentsByCategory`, on top of ~39ms of pre-existing `prepareNav` + `groupNavItems`). `SidebarLayout` also subscribes to the responsive context via `useResponsive()`, so it re-runs on hydration and on every drawer/compact change, not only on route changes. A `useMemo` keyed on `location.pathname` and `showAll` would cover it.

langz:
Following up on the two search bar recordings from before — the wrapper is fixed now, but there's still a small jump left, so leaving those open for a moment.

Sampling `/uilib/components/button/` on the preview every 100ms from navigation commit, 1440px:

```
none, none, none, 280/260, 280/260, 280/236, 280/236, 280/236, … (stable)
```

`.portal-search` itself is rock-steady at 280px, which is the part that got fixed. The `<input>` inside it goes 260 → 236px roughly 200ms after first paint. It's the `⌘ K` / `Ctrl K` hint: `SearchBar.tsx:45` reads it through `useSyncExternalStore`, which returns `null` on the server on purpose (`does not render a platform-specific hint on the server`), so the `<kbd>` only mounts after hydration and takes ~24px off the field.

Reserving the space during SSR — a fixed-width placeholder, or `padding-inline-end` on the input regardless of whether the hint is there — would settle the last of it.

The caret part of the second recording does look sorted: the active accordion arrives expanded in the prerendered HTML and the pre-hydration script pins the icon state, and I see no state change across 24 samples after load.

langz:
One thing I want to check rather than assume: how is this meant to land relative to the two neighbouring PRs?

As it stands, `329a6b6` carries the full extension diff — 10 files, +1184/−47 under `packages/dnb-eufemia`, including the new `SidebarMenuResponsiveAside.tsx`, the `SidebarMenuResponsive` rework, the `SidebarMenuRoot` compact class and +387 lines of SCSS. That is the same work as #9393, give or take four SCSS lines that only exist here. #9393 is still a draft with `REVIEW_REQUIRED` and zero reviews on it, so if this merges first, the "reviewed separately in #9393" part of the description ends up on `main` reviewed only through this PR — and #9393 becomes a mostly-empty follow-up.

That's not automatically wrong: it may well be the plan to let this carry the extension work and close #9393, or to get #9393 out of draft and merge it first. I just don't want to approve something on the assumption that someone else is looking at the other half. Which way are you thinking?

Related, and much smaller: #9295 is based on this branch rather than `main`, and `Layout.module.scss:235` (`&:has([data-portal-home])`) only does something once #9295 lands, so that rule ships inert in between. Fine if the order is this → #9295, just noting it in case they end up separated.

langz:
Three things I bumped into that are more questions than findings — happy to be told they're intentional or out of scope.

**Does the theme badge still have a job?** No page carries `theme:` frontmatter on this branch any more — the last one went when `theme: ['sbanken', 'carnegie']` came off `sidebar-menu.mdx`, which was the right call given the Carnegie badge 404. That leaves the badge itself with nothing to render: the `theme` field in the query (`SidebarMenu.tsx:75`), `useTheme()` (`:128`), `addThemeBadges` (`:190`), `ThemeBadge` (`:203`), `themeTitles` (`SidebarMenuData.ts:158`), `NavItem.theme` / `themes` (`:30`, `:62`, `:209`) and the `.portal-sidebar-menu__theme-badge` block including the `--sbanken::before { content: 'Sbanken' }` rule. `addThemeBadges` also walks every item on each render to attach `suffix: undefined`. If the plan is to reintroduce `theme:` on the Sbanken-ready pages then it's just waiting for its data, and one thing to keep in mind for that day is that the badge is a `<span title="…">` with no text content, so screen readers won't announce the "ready for use with the X theme" part. If it isn't coming back, it looks like it could follow `platform` out the door.

**Was dropping the label on the outer landmark deliberate?** Production renders `nav#portal-sidebar-menu` with `aria-labelledby="toggle-sidebar-menu"`, so the landmark list reads `navigation: Hide section content menu`. Here it's `aside#portal-sidebar-menu` with no name wrapping `nav` with `aria-label="Section Content Menu"`, so the list gains an unnamed `complementary` alongside the named `navigation`. axe is clean either way and one unnamed complementary is legal, it just adds a stop with nothing to say when you page through landmarks. An `aria-label` on the aside, or letting `ResponsiveAside` inherit one, would close it.

**What should the sections outside the unified menu show?** `shouldIncludeSidebarPrefix` merges `uilib`, `icons`, `quickguide-designer` and `contribute`, so those four now share the same 305-item menu — a designer opening the Quick Guide gets the full UI library tree. The other side of that is `/brand/`, where the sidebar is literally one item (`Brand → /brand`), and `/license/` and `/design-system/` are similar. They also don't get the Home entry, because `createUilibSidebarStructure` returns early before `includeHome` when nothing in the section is configured, so the only way back out of those pages is the logo. Both halves may be exactly what you want — I mostly wanted to check the small-section case wasn't collateral.

For the record, two things I suspected and then couldn't substantiate: the drawer and the aside share `openItemsStorageKey` and `scrollPositionStorageKey`, but the stored offset gets re-derived from the active item on load, so I couldn't produce any cross-contamination. And the resize handle's SSR `aria-valuenow="240"` corrects itself to `384` as soon as the layout effect measures.

langz:
I went looking for why the drag-based specs fail and it reproduces on the preview: **only the first 6px of the 24px resize handle are clickable**, so pointer resizing barely works.

At 1440px on `/uilib/components/button/`:

- the handle's box is `x: 378 → 402` (24px wide)
- the `<aside>` ends at `x: 384`
- hit-testing each pixel across the handle with `elementFromPoint` gives `111111000000000000000000` — 6 hits out of 24

The overhanging 18px resolve to `main#dnb-app-content` instead. The handle is `position: absolute; z-index: 3` inside the aside, while `.contentStyle` carries `z-index: 2` a level up, so for the part that sticks out past the aside the content wins — different stacking contexts, so the higher number doesn't help.

Consequences I could measure:

- `page.mouse` press at the handle's centre (`x ≈ 390`): no `pointerdown` on the handle at all, `--sidebar-menu-runtime-expanded-width` never set, width stays `384px`. That is exactly `sidebarResize.spec.ts:154` — `Expected: "520px" / Received: "384px"` — and Playwright clicks element centres by default, which is why four of the six e2e failures are the drag-driven ones (`sidebarResize.spec.ts:60/392/435` and `sidebarMenuCompact.spec.ts:160`).
- Press at `x + 2` instead: works perfectly, `384px → 524px`.
- Keyboard resize was never affected: 10× <kbd>→</kbd> gives `544px` and a matching `aria-valuenow`.

So it isn't runner flakiness, and the specs are right to fail. For a user it means the strip looks grabbable across its full width but only responds in the leftmost quarter, right on the border.

Slightly awkward: this looks like a side effect of moving the handle inside `aside#portal-sidebar-menu`, which was the right fix for the landmark issue from the first round. Keeping it there and letting its hit area escape the aside's stacking context — or lowering `.contentStyle` below it — would give back the full 24px.

langz:
Nice catch on the hydration one — I can confirm the shipped preview logs `Minified React error #418` (`args[]=text`) on `/uilib/components/button/` while production is clean, and pinning `scopeId` is exactly the right lever since the mismatching text is the generated first-paint CSS inside the `<style>`.

I also checked the preload change against the real markup rather than waiting for the rebuild: loading `?fullscreen` with the app bundle blocked, the shipped rule leaves the sidebar at 384px and `.dnb-app-content` at `left: 303px`, and swapping in `.dnb-sidebar-menu-responsive-inline{display:none!important}` puts it back to `left: 0, width: 1440`. Targeting the wrapper rather than the aside is the right call — that's the box holding the width. Marked that one resolved.

Two small things that came with the new commits, both questions rather than asks:

**The new hydration guard may not be able to fail in dev.** `pageNavigation.spec.ts` filters on `#418`, which is the minified code, so a dev run logging React's full non-minified hydration message would pass regardless. Its sibling in the same block (`prerendered content should stay visible during JS hydration`) opens with `if (await isDev(page)) return`. Would it be worth the same guard, or matching on something present in both builds?

**`scopeId` is now consumer-supplied and lands in a selector string.** `SidebarMenuResponsive.tsx:618` does `` `[data-sidebar-menu-responsive-scope="${scopeId}"]` `` and that string is rendered into the inline `<style>`. While it was internally generated that was fine; as a public prop a value containing a quote would end up shaping the rule. Nothing reachable from the Portal since it's a literal, and the docs entry is clear, so mostly asking whether it's worth escaping or constraining at the boundary now that it's part of the API — along with a word about keeping it unique per provider, since two providers sharing a `scopeId` would share their first-paint CSS.

langz:
**Correction to my drag diagnosis above** — the hit area isn't what the four drag specs are failing on, and I'd rather say so before it sends you at the wrong thing.

`sidebarResize.spec.ts:392` turns out to be a controlled experiment inside a single test. The real-mouse drag at `:412-416` passes (`238px`, then `240px` after the spring-back), and the very next drag, `dragSidebarToWidth(page, resizeHandle, 700)` at `:431`, gets `240px` instead of `560px`. Same handle, same test, same viewport — so pointer *position* can't be the difference.

What actually changed is the delivery. `dragSidebarToWidth` moved from `page.mouse.move/down/move/up` — which is what `main` still has, and what `:412-416` still does — to `resizeHandle.dispatchEvent('pointerdown', { pointerId: 1 })` plus `window.dispatchEvent(new PointerEvent(…))` at `:28-55`. `sidebarMenuCompact.spec.ts:177-190` does the same. `locator.dispatchEvent` calls `element.dispatchEvent()` directly, so hit-testing never enters into it.

The portal suite runs Firefox (`playwright.config.e2e.ts:20`), and Firefox rejects a `pointerId` that doesn't belong to a live pointer. Minimal repro against the same Playwright build, one button with a `pointerdown` listener:

```
firefox  synthetic:  ["handler-start","error:NotFoundError: Element.setPointerCapture: Invalid pointer id"]
firefox  real mouse: ["handler-start","after-capture"]
chromium synthetic:  ["handler-start","after-capture"]
chromium real mouse: ["handler-start","after-capture"]
```

`SidebarMenuResizeHandle.tsx:302` is `event.currentTarget.setPointerCapture?.(event.pointerId)` — `?.` guards a missing method, not a throwing one — so the `NotFoundError` leaves `handlePointerDown` before `startResize()` on `:303` ever runs. The drag silently does nothing, which is exactly the `384px` / `240px` / `384px` we get in the three `sidebarResize` failures and the `scrollWidth > clientWidth` poll at `sidebarMenuCompact.spec.ts:200`. It also explains why this only started with these specs: nothing else in the suite drives a synthetic pointer.

Smallest fix is to put `dragSidebarToWidth` back on `page.mouse` the way `main` has it, and give the compact spec the same treatment. Wrapping the `setPointerCapture` call in a `try`/`catch` would make the handle survive synthetic pointers too, but I'd rather not reshape product code to suit a test when the test can just drive a real one.

The 6-of-24-px hit area I measured earlier still stands on its own merits — it just isn't what's red in CI. Sorry for the detour.

tujoworker:
Worked through the full review and pushed the fixes in `72c4821a5bb`.

Highlights:
- restored Helpers and removed stale theme metadata / missing Carnegie asset use
- bounded resizing to 240–560px, kept width intentionally session-local, and moved the handle inside the sidebar landmark
- fixed compact width propagation so the header/search do not stutter or reserve 24rem at compact widths
- added an opaque desktop header surface while retaining the mobile floating treatment
- removed dead platform/group/icon plumbing and restored meaningful E2E assertions
- restored element-list parity coverage without a fixed timeout
- documented the active `sidebarMenu` frontmatter contract and restored redirect JSDoc
- restored the original Figma-managed hamburger; the animated three-bar icon is local to ResponsiveTrigger
- kept the sidebar visible throughout the intro sequence
- aligned the component overview / related lists / LLM catalog with categorized extensions such as VippsWalletButton
- fixed fullscreen/focusmode preload CSS, reduced-motion specificity, and Portal hydration through a stable responsive scope
- scoped compact state to ResponsiveInline so embedded SidebarMenu demos keep their normal rendering
- replaced synthetic pointer events with real hit-target drags for Firefox

I kept the existing linked-accordion behavior: those rows are both real page destinations and expandable parents. Turning them into button-only folders would remove direct navigation, while splitting the row into separate link/toggle targets is a separate component API/design decision rather than a Portal integration fix.

Validation: 67 focused unit tests, type checks, JS/SCSS lint, production E2E/visual builds, 33 focused Chromium E2E tests, 9 compact/resize interaction tests, and clean local hydration. Live visual regression is green on this revision; E2E is currently still running.

langz:
> Worked through the full review and pushed the fixes in `72c4821a5bb`.
> 
> Highlights:
> - restored Helpers and removed stale theme metadata / missing Carnegie asset use
> - bounded resizing to 240–560px, kept width intentionally session-local, and moved the handle inside the sidebar landmark
> - fixed compact width propagation so the header/search do not stutter or reserve 24rem at compact widths
> - added an opaque desktop header surface while retaining the mobile floating treatment
> - removed dead platform/group/icon plumbing and restored meaningful E2E assertions
> - restored element-list parity coverage without a fixed timeout
> - documented the active `sidebarMenu` frontmatter contract and restored redirect JSDoc
> - restored the original Figma-managed hamburger; the animated three-bar icon is local to ResponsiveTrigger
> - kept the sidebar visible throughout the intro sequence
> - aligned the component overview / related lists / LLM catalog with categorized extensions such as VippsWalletButton
> - fixed fullscreen/focusmode preload CSS, reduced-motion specificity, and Portal hydration through a stable responsive scope
> - scoped compact state to ResponsiveInline so embedded SidebarMenu demos keep their normal rendering
> - replaced synthetic pointer events with real hit-target drags for Firefox
> 
> I kept the existing linked-accordion behavior: those rows are both real page destinations and expandable parents. Turning them into button-only folders would remove direct navigation, while splitting the row into separate link/toggle targets is a separate component API/design decision rather than a Portal integration fix.
> 
> Validation: 67 focused unit tests, type checks, JS/SCSS lint, production E2E/visual builds, 33 focused Chromium E2E tests, 9 compact/resize interaction tests, and clean local hydration. Live visual regression is green on this revision; E2E is currently still running.

Thanks, looks very good, I can do a final round of manual testing soon :)

langz:
Rechecked on `72c4821`. `Run visual-regression tests` is green now and the four drag specs pass — the compact scoping and the `page.mouse` switch both did the job, thanks. Two e2e failures are left, and they turn out to share one cause that I think matters beyond CI.

**Hovering the compact rail pushes every menu item down, so you click the neighbour of what you aimed at.**

Firefox at 1280×900 on `/uilib/components/`, which is inside the Portal's compact range (`compactAt="large"` plus `compactOffset="10em"`):

| | rail idle | rail hovered |
| --- | --- | --- |
| group titles rendered | 0 of 6 | 6 of 6 |
| `…/components/anchor` top | 548 | 588 |
| `…/components/button` top | 600 | 640 |
| `…/components/autocomplete` top | 820 | 900 |
| what's at `(32, 622)` | the Button link | the `li` gap between rows |

It snaps back on unhover, so it's not a one-off settle. The shift is cumulative — the further down the list, the bigger it gets.

The mechanism is in `dnb-sidebar-menu.scss`: the collapsed rail sets `.dnb-sidebar-menu__group__title` / `__header` / `__section-text` to `max-height: 0; padding-block: 0` at `:849`, and `.dnb-sidebar-menu--compact:is(:hover, :focus-within)` at `:880` puts both back (`:941`, `:950`). Every title that appears adds its own height to the flow, so everything below it moves.

For a user that means aiming at an icon, having the rail expand as the pointer arrives, and landing on the row below. For Playwright it's the same thing one step more visible: it measures the target, moves the pointer, the row moves out from under it, and the click goes to a neighbour — in the previous run it actually navigated, the tab panel ends up on `slider/demos` while the test is still on `/uilib/components/`. That's `pageHeading.spec.ts:12` and `routeFocus.spec.ts:41`, and it's why they burn the full 30s instead of failing fast.

Keeping the vertical rhythm identical between the two states would fix both at once — reserving the title height while collapsed (`opacity` / `visibility` rather than `max-height: 0` plus `padding-block: 0`), or simply not revealing the titles on hover. Adding a settle-wait to the two specs would turn CI green too, but it would leave the pointer behaviour as is.

Measured against the branch preview, which is still serving `da8392ca` — nothing in `72c4821` touches this path, and the two specs fail the same way on both.


One for the merge checklist, since the old CI comment is now stale: the branch is still `CONFLICTING` against `main` — `declaration.d.ts` and `ListComponents.tsx` changed on both sides, and `SidebarMenuContext.tsx` is deleted here but modified upstream.

langz:
Thanks for the sweep — you closed more in one commit than I managed to write up in a morning, which feels suspiciously like being out-reviewed by the author. 😄

Somewhere in here I have learned more about `setPointerCapture` and Firefox's strong opinions on made-up pointer ids than I ever intended to, and I regret nothing. The Portal navigation looks genuinely lovely now.

One icon rail left to persuade to stand still when you hover it, and I think we're there 👏

tujoworker:
Fixed the remaining compact-rail issue in `9e52ab2a521`: group/header space is now reserved while collapsed, all row Y-positions are asserted unchanged through hover expansion, and the expanded rail is above Portal content so nested links remain clickable. I also restored the original sidebar-click coverage, rebased onto current `main`, and kept the shared component-list filtering compatible with the new virtual-page source. The latest revision is mergeable and all checks are green, including Firefox E2E and visual regression.

langz:
When doing a refresh (command + r), the sidebar menu jumps a bit:


https://github.com/user-attachments/assets/576b6cd4-f3b9-43b7-97a5-8f8bfd4cd9e0

langz:
Not an issue, but I find it a bit strange that when scrolling all the way to the bottom, then trying to scroll further, that the background showing up then is black and not in the same color of the page:


https://github.com/user-attachments/assets/61bead66-4c2d-4790-acf4-0d5c38dcbc51

Same with top:

https://github.com/user-attachments/assets/71cf6b39-a865-4ec6-a1a6-015af46eb3d4

langz:
The search field itself moves at hydration now, not just the `<input>` inside it.

`/uilib/components/button/` at 1280px, sampled from navigation: at ~320ms `.dnb-input__shell` sits at `left: 844, width: 260`; at ~730ms, once the `⌘ K` hint mounts, it's `left: 836, width: 268`. That's the element carrying the inset border and the radius, so the visible box jumps 8px left and grows 8px while the magnifier slides 32px.

Production holds the shell at `left: 592, width: 512` straight through hydration — there only the `input` and the button move inside it. The difference is `SearchBar.module.scss`: `.autocompleteStyle` pins `.dnb-autocomplete__inner` and `.dnb-autocomplete__shell` to `width: 100%`, but `.dnb-autocomplete__row` is left content-sized, so it absorbs the `<kbd>` when it arrives. `.dnb-autocomplete__row { width: 100% }` in the same block pins the box; reserving the hint width during SSR would settle the inner shift too.

I think this is the remaining half of [the](https://github.com/dnbexperience/eufemia/pull/9296#issuecomment-5710481945) [two videos](https://github.com/dnbexperience/eufemia/pull/9296#issuecomment-5710497441) further up.

Rechecked the rest on `9e52ab2a`: the compact rail holds still (0 of 12 row tops move on hover), the embedded demos render at full width again, resize is bounded to 240–560, and the preload CSS targets the right element. I've hidden the comments this revision closes.

langz:
The LLM catalog and the site now disagree about which extensions count as components.

`componentCatalog.ts` keeps every `uilib/extensions` page three segments deep, and its `getCategoryId` falls back to `'other'` when `category` is missing. `forms.mdx` has a title, no `draft`/`hideInMenu` and no `category` — so "Forms for applications" lands in the generated markdown under Other, right next to the `[Eufemia Forms](/uilib/extensions/forms/)` link the overview already writes by hand. The site runs the same pages through `isCategorizedComponentSlug`, whose allow-list is `payment-card`, `sidebar-menu` and `vipps-wallet-button`.

`componentCatalog.test.ts` mirrors `componentCategoryOrder`, `excludedSlugs` and `MAX_VISIBLE_RELATED` from the portal source, but nothing mirrors the new allow-list — which is how this drifts quietly. Mirroring `categorizedExtensionSlugs` across, with one more sync test, would close both.

Same function, unrelated: the `split('/').length !== 3` guard runs after `readFrontmatter`, so the build reads all 517 `.mdx` files under `uilib/extensions` in order to keep 4. Moving the guard above the read skips 513 of them.

langz:
On main branch, the top section is the same background color as rest of the page:

image

In this branch, it's black:

image

langz:
Not a problem, just a thought.

UI Library, Quick Guide - Designers, Icons Library and Contribution Guide have the same menu/sidebar menu, but About Eufemia and Brand has its own sidebar menu with different entry points.

I find it a bit strange that each of these links/entry points do not have a separate menu/sidebar menu:

image

langz (approved):
Approving on `9e52ab2a`. Everything I had against this is gone — the compact rail stands still under the pointer, the embedded demos render at full width again, resizing is bounded, the preload CSS targets the right element, and CI is green the whole way through, Firefox E2E and visual regression included.

The two comments left open are polish rather than gates: 8px of jitter on the search field at hydration, and a Forms entry that only shows up in the generated markdown. Neither needs to hold this back, and both are easy to pick up separately.

Really nice piece of work — the Portal navigation is in a much better place now 👏

langz:
There's still quite a few open/existing comments/reviews that we could potentially look into:
- https://github.com/dnbexperience/eufemia/pull/9296#issuecomment-5773017442
- https://github.com/dnbexperience/eufemia/pull/9296#issuecomment-5773061322
- https://github.com/dnbexperience/eufemia/pull/9296#issuecomment-5773418038
- https://github.com/dnbexperience/eufemia/pull/9296#issuecomment-5773527613

PR #9298, opened 2026-09-08
chore(Portal): store URL hash and portal render flags in page-view path

joakbjerk:
Page-view tracking recorded only the pathname, so navigations that differ only by the URL hash (e.g. component doc anchors) or by a portal render mode looked identical in analytics.

The tracked path now also carries the hash and a small set of portal render params (`fullscreen`, `focusmode`, `eufemia-theme`). Minimisation happens at the collector: `normalizeTrackedPath` keeps the pathname, the allow-listed render params (flags key-only, theme only for a safe token, canonical order) and an anchor-shaped fragment, and drops everything else — notably free-text docs search terms — so nothing incidental is stored, regardless of what reaches the edge-locked ingest route.

Tracking still ships dark until `VITE_ANALYTICS_ENDPOINT` is set.

### DIFF
--- a/packages/dnb-design-system-portal/src/core/FocusModeCodeContext.tsx
+++ b/packages/dnb-design-system-portal/src/core/FocusModeCodeContext.tsx
@@ -7,6 +7,7 @@ import {
   useState,
 } from 'react'
 import type { ReactNode } from 'react'
+import { useLocation, useNavigate } from 'react-router'
 
 const FOCUS_MODE_CODE_PARAM = 'focusmode'
 
@@ -27,6 +28,8 @@ export function FocusModeCodeProvider({
 }: {
   children: ReactNode
 }) {
+  const location = useLocation()
+  const navigate = useNavigate()
   const [focusModeCodeId, setFocusModeCodeIdState] = useState<
     string | null
   >(null)
@@ -38,7 +41,7 @@ export function FocusModeCodeProvider({
       return // stop here
     }
 
-    const params = new URLSearchParams(window.location.search)
+    const params = new URLSearchParams(location.search)
     const value = params.get(FOCUS_MODE_CODE_PARAM)
 
     if (value) {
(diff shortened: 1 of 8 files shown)


joakbjerk:
Addressed the review by enforcing the minimisation at the ingest boundary instead of the client:

- **Server-side enforcement (was findings 1–3):** the render-param allow-list and hash handling moved into the collector (`normalizeTrackedPath` in `portal-view.ts`), applied in `buildPortalViewRecord`. The stored path is now safe by construction regardless of what reaches the edge-locked route. The portal just sends the full in-app path.
- **Hash bounded:** the fragment is kept only when it's anchor-shaped (`#slug`), otherwise dropped.
- **Canonical order:** allow-listed params are sorted, so `?fullscreen&eufemia-theme=ui` and the reverse produce one record.
- **Stale doc comment fixed.**
- **No cross-package duplication (finding 5):** the allow-list now lives only on the server; the client no longer carries it.

Tests cover `normalizeTrackedPath` directly, including `focusmode` reduced to key-only and the canonical-order and fragment-drop cases.

joakbjerk:
Staying with server-side minimisation and **accepting #1 as a documented defense-in-depth tradeoff** rather than restoring the client allow-list. Minimisation is centralised in the collector's `normalizeTrackedPath`; the residual risk is only that a docs search term transits in the beacon body (TLS, edge-locked route that doesn't log POST bodies, dropped at ingest, never stored) — which satisfies the original brief. The render params are portal-controlled, so any client/server drift would only mean missing analytics, never a privacy leak.

Added a comment at `buildTrackedPath` recording that the raw send is intentional (do not add client-side stripping), so the decision is visible and doesn't get re-litigated. Also noted that `SAFE_PARAM_VALUE` matching `ENV_PATTERN` is coincidental, kept separate on purpose.

Description fixed to describe collector-side minimisation. Anchor event volume is an accepted product choice (hash is intentionally part of the path); `MAX_BUFFER=50` flushes fine.

tujoworker (approved):

PR #9299, opened 2026-09-08
chore(Portal): record page-view status (ok, not_found, error)

joakbjerk:
Page-view analytics could not tell a normal view apart from a 404 or a render error — every event looked the same, so a bad link or a crashing page was invisible in the data.

Each page view now carries a `status`:

- `not_found` — the visited path fell through to the catch-all 404 route (detected from the matched splat route). The attempted pathname is already captured, so you can see *which* URLs 404.
- `error` — a render error caught by the error boundary, recorded against the path the user was on.
- `ok` — everything else (the default).

Dedup now keys on path *and* status, so a status change on the current path (e.g. a page that renders then errors) is still recorded.

The stored record and the `portal_views` Glue table gain a `status` column. It is schema-on-read, so older rows read NULL and no migration is needed.

Stacked on #9298 (base `feat/portal-analytics-tracked-path`) — merge that first. Tracking still ships dark until `VITE_ANALYTICS_ENDPOINT` is set.

### DIFF
--- a/packages/dnb-design-system-portal/src/core/ErrorBoundary.tsx
+++ b/packages/dnb-design-system-portal/src/core/ErrorBoundary.tsx
@@ -5,6 +5,7 @@ import { Code, CopyOnClick, P } from '@dnb/eufemia/src'
 
 type ErrorBoundaryProps = {
   children: ReactNode
+  onError?: (error: Error) => void
 }
 
 type ErrorBoundaryState = {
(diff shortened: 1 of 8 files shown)


tujoworker (approved):
Reviewed the current revision. No blocking issues found. Focused portal and analytics tests pass, and all PR checks are green.

PR #9319, opened 2026-09-10
chore(analytics): move the dashboard snapshot to a combined per-source shape

joakbjerk:
Prepares the dashboard snapshot to carry more than portal page views: it becomes a single combined object with one named section per data source, so upcoming sources (MCP usage, Nucleus component usage) can each add a section without reshaping the envelope or touching existing charts.

Deploy sequencing (dev): re-run `bootstrap-dashboard-iam.sh` to grant `snapshots/*`, then deploy, then invoke the snapshot generator once to seed the new snapshot key. Kept as a draft until that ordering is done so the read Lambda does not briefly 503.


langz:
Am I reading the deploy path right that merging triggers the apply directly here (push to `main` touching `tools/analytics/**` → `terraform apply -auto-approve`)? If so, does the IAM step need to land before merge rather than between merge and deploy?

And has `bootstrap-dashboard-iam.sh` already been re-run with the new path? Per the README the read role has only `s3:GetObject` and no `s3:ListBucket`, which I think means a missing key comes back as `AccessDenied` rather than a 404 — so would we sit on a permanent 503 rather than a brief one? Or is something seeding the key first that I've missed?

langz (approved):

joakbjerk:
Both spot-on, thanks — verified in the code and fixed.

**Deploy path:** confirmed — a merge to `main` under `tools/analytics/**` builds and force-pushes to the GHE deploy repo, which runs `terraform apply` via OIDC, so it's effectively auto-deploy on merge with no manual window. The dashboard read role is now in place ahead of that (scoped to `s3:GetObject` on `snapshots/*`).

**Cold start:** you're right that with `GetObject` and no `ListBucket`, a missing key is `AccessDenied` (503), not `NoSuchKey` — so the empty-state path wouldn't trigger before the first generator run. Rather than broaden the role with `ListBucket` (a prefix-conditioned grant doesn't reliably give the 404, since `s3:prefix` isn't set on a `GetObject`), I've added a post-apply step that invokes the generator once so the snapshot exists immediately — it also fails the deploy if the generator errors, turning a would-be silent stale/503 into a loud failure. The read role stays `GetObject`-only.

Separately, the frontend now shows deploy-aware copy for a 503 instead of a generic error (#9347).

PR #9320, opened 2026-09-10
chore(analytics): add anonymous MCP usage tracking to the dashboard

joakbjerk:
Stores anonymous MCP server usage and surfaces it on the dashboard — request counts, which tools are used, which components are looked up, and which doc paths are read. No query text or personal data.

Deciding the storage layout now (13-month raw rows vs a kept-forever daily rollup for year-over-year comparison) is cheap while there is no data to migrate; doing it after the producer ships would mean a backfill.

Deploy prerequisite: re-run `bootstrap-analytics-iam.sh` before deploy — the analytics role's Glue permission was pinned to a single (now stale) table; it now covers the database's tables so the generator can query the new ones.

Stacked on #9319; the MCP producer is a follow-up, so the section renders empty until it lands.


tujoworker (approved):

joakbjerk:
Nice catches on both 👍 One small thing on the noncurrent-version cleanup: it's on `mcp-usage/`, but those files use unique keys and are never overwritten — so there are no old versions there to clean up.

The folder that *does* get overwritten is `mcp-usage-daily/` (the generator rewrites `dt=…/agg.json` each run), and it has no lifecycle rule — so under bucket versioning its old versions pile up.

Should `noncurrent_version_expiration` point at `mcp-usage-daily/` (or both) instead? Might be misreading — let me know!

langz:
One thing I'd fix before merge: `buildMcpUsage` is awaited inside the `snapshot` object literal, so if either new query fails, `writeSnapshot` never runs — we lose the whole snapshot, including the portal views that were just fetched fine. I tried it with a Glue `AccessDenied` on `mcp_usage` and with a timeout on the rollup read; both write zero objects, and the dashboard just keeps serving the last good snapshot. `snapshot_empty` won't catch it either, since `SnapshotRecordCount` is emitted after the write and the alarm treats missing data as not breaching.

That's also exactly what we get if the `bootstrap-analytics-iam.sh` re-run is missed on deploy. A try/catch around `buildMcpUsage` falling back to `EMPTY_MCP_USAGE` (already exported, already used by `dashboard-read`) keeps the new section blank — which is how it renders until the producer lands anyway — and lets the rest of the dashboard keep updating.

langz (approved):

joakbjerk:
Dug into this — my original comment was right, and the fix is now pushed.

Checked every prefix against how it's written (versioning is enabled, so only overwritten keys leave noncurrent versions):

- `snapshots/dashboard.json` — overwritten hourly → noncurrent versions pile up, **no cleanup rule**
- `mcp-usage-daily/<dt>/agg.json` — recomputed each run → same, **no cleanup rule**
- `mcp-usage/…`, `portal-views/…` — unique keys, write-once → never create noncurrent versions (so the old rule here cleaned nothing)

Rather than add a per-prefix rule for each overwritten prefix (which is how the gap happened), I went **bucket-wide**: one `noncurrent_version_expiration` with an empty filter, `noncurrent_days = 7`. It covers `snapshots/` and `mcp-usage-daily/` plus any future overwritten prefix, and is a harmless no-op on the write-once ones. Removed the dead noncurrent clause from `mcp-usage/`. Current versions are untouched, so the live snapshot and the durable rollup are unaffected.

`terraform fmt`/`validate` pass. `noncurrent_days` is easy to dial if 7 feels off.

PR #9325, opened 2026-09-10
chore(mcp): capture anonymous MCP server usage for the dashboard

joakbjerk:
Fills the dashboard MCP usage section by capturing, server-side, which registered tool each request called and, where applicable, the component or doc path it acted on. No query text or personal data is queued or stored: component and path arguments are structurally validated first.

Usage records are sent to an encrypted SQS queue with a one-second producer deadline. A separate, concurrency-limited Lambda batches queue messages into partitioned S3 objects for Athena. Storage retries therefore stay off the MCP response path; exhausted messages are retained in a dead-letter queue, with alarms for both a stalled source queue and DLQ arrivals.

Depends at deploy time on the `mcp_usage` tables from #9320.

Deploy prerequisites: before applying Terraform, update the pre-created MCP role with `sqs:SendMessage` on the source queue, and create `eufemia-<env>-mcp-usage-consumer-role` with basic Lambda logging, SQS receive/delete/get-attributes access on the source queue, and `s3:PutObject` on the analytics bucket `mcp-usage/*` prefix. The deploy role also needs permission to manage the SQS queues, event-source mapping, consumer Lambda, log group, and CloudWatch alarms.

### DIFF
--- a/tools/mcp-lambda/infra/main.tf
+++ b/tools/mcp-lambda/infra/main.tf
@@ -7,6 +7,8 @@ locals {
   }
 }
 
+data "aws_caller_identity" "current" {}
+
 # Lambda function
 resource "aws_lambda_function" "mcp" {
   function_name = local.function_name
@@ -31,6 +33,12 @@ resource "aws_lambda_function" "mcp" {
 
       # Shared secret for the X-Edge-Auth origin check (injected by Akamai).
       EDGE_AUTH_SECRET = var.edge_auth_secret
+
+      # Anonymous MCP usage capture. The analytics data bucket (same account,
+      # owned by the analytics stack) and the environment label stored on each
+      # record. The bucket name is deterministic, so no cross-stack reference.
+      DATA_BUCKET = "eufemia-${var.environment}-analytics-${data.aws_caller_identity.current.account_id}"
+      USAGE_ENV   = var.environment
     }
   }
 
--- a/tools/mcp-lambda/package.json
+++ b/tools/mcp-lambda/package.json
@@ -15,6 +15,7 @@
   },
   "type": "module",
   "dependencies": {
+    "@aws-sdk/client-s3": "3.1075.0",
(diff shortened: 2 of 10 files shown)


tujoworker (approved):

langz:
The component field looks inverted for multi-word components: the docs server lowercases the name and the docs are hyphenated files, so `date-picker` resolves but stores an empty component, while `DatePicker` stores the name and returns ENOENT. That means the 22 hyphenated components can never be attributed to a successful call, and where a name does appear it's a failed lookup — could we allow hyphens and normalise the stored value the same way `normalizeName` does?

langz:
Heads up — the normalized-doc-name commit lowercases the component but `usage-capture.test.ts:67` still expects `Button`, so the mcp-lambda suite is red on `6d74029` (1 failed / 86 passed). Nothing in CI catches it, since `eufemia-mcp-lambda` isn't in the test/lint/typecheck jobs in `verify.yml`. I've fixed that line in #9344 along with the same normalisation gap in the `path` field, so it can come in with that if you'd rather not touch this branch.

joakbjerk:
🚧 **Hold merge — deploy prerequisite pending (code is approved and ready).**

Merging triggers `terraform apply`, which creates the SQS queue + DLQ and two CloudWatch alarms. The deploy role doesn't yet have SQS-management or CloudWatch-alarm permissions, so an apply would fail with `AccessDenied`. A separate OIDC-federation change to grant those (ARN-scoped) is being prepared.

Please don't merge until that grant lands and its deploy run is green — I'll comment here once it's unblocked.

joakbjerk:
✅ **Unblocked.** The OIDC grants are now live on the deploy role — the SQS, CloudWatch-alarm, and consumer-role permissions merged and the "Deploy OIDC Roles" run applied them (verified on the live policy). `terraform apply` will no longer 403 on the queue, alarms, or the consumer-role lookup.

Safe to merge whenever — code is approved and green.

PR #9287, opened 2026-09-08
chore(Portal): enable anonymous page-view analytics on the release build

joakbjerk:
Turns on the anonymous page-view tracking that has been shipping dark since it was added. The `release` portal build now sets `VITE_ANALYTICS_ENDPOINT` (the edge-locked `/collect-portal-views` route) and `VITE_ANALYTICS_ENV=prod`; preview and dev builds set neither, so they stay dark.

All page-view shape work has merged (#9298, #9299, #9326, #9343, #9351) and the enriched record round-trips end to end (edge → Lambda → `portal_views`, all dimensions). Merging this, then the next release, ships tracking live with the first real rows already on the final shape.

The collected data is anonymous (no identifiers, cookies or device storage) and out of DPIA scope — see EDS-960.

PR #9326, opened 2026-09-10
chore(Portal): record locale, theme, and color scheme as page-view dimensions

joakbjerk:
> [!NOTE]
> #9298 and #9299 (path enrichment + the `status` column) have merged, so this now sits directly on `main`. It's the last page-view **shape** change before the flip: #9326 → #9287.

Adds three dimensions to each page view: the active component language (`locale`), theme (brand), and the resolved color scheme (`light`/`dark`). Today we only know *which* page was viewed; this also captures the state it was viewed in — e.g. this page, in Swedish, with the Sbanken brand, in dark mode. These sit alongside `status` as attributes of a view; user actions (code copy, outbound clicks) will be separate event types later.

All three columns are additive and schema-on-read (old rows read NULL, no migration). Landing this before the tracking flip (#9287) keeps the first real rows on the final shape.

The values are validated server-side at the edge-locked ingest boundary as low-cardinality, non-identifying tokens; the client normalizes `locale` before sending so a stale stored value can't drop a beacon batch, and resolves an `auto` color scheme to the scheme the user actually saw. Tracking reads existing locale/theme/color-scheme preferences but creates no device storage of its own, so it stays inside the anonymous envelope.


tujoworker (approved):
Looks good. The dimensions are now restricted to the portal's supported values, the client normalizes stale locale data, and the stored record contract is documented.

langz (approved):
Approving — I went through the client normalisation, the ingest validation and the Glue schema change, and checked the parts that could bite: the portal's locale/theme lists and the collector's allow-lists agree today, the `colorScheme` key does resolve to the `colorscheme` column, an unset colour scheme resolves the same way as the portal's own boot script, and the deploy is safe whichever side lands first. No new test failures locally.

One non-blocking note: the allow-lists in `portal-view.ts` duplicate the portal's own lists, and an unrecognised value rejects the whole batch rather than just the field. Harmless while tracking is off, but worth a look before #9287 flips it on — coercing an unknown dimension to `unknown` would keep the low-cardinality guarantee without the coupling.

PR #9336, opened 2026-09-11
fix(Drawer): respect the dynamic viewport height

tujoworker:
Uses the dynamic viewport height so mobile browser controls do not cover the bottom of Drawer content, while retaining the existing viewport-height fallback.


langz (approved):

tujoworker:
We tested it on Android in this [PR](https://github.com/dnbexperience/eufemia/pull/9296) and it did the trick.

PR #9337, opened 2026-09-11
docs(Portal): fix local MCP docs server setup instructions

joakbjerk:
Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1789115978091599

The local MCP setup on the Tools page fails with `-32000: Connection closed` when followed as written, so people fall back to the hosted server.

Two independent mistakes in the instructions, each enough to break startup:

- It points `node` at `mcp/mcp-docs-server.js`, which is the runtime-agnostic core that only exports functions. The runnable stdio entry is `mcp/mcp-stdio.js`.
- It tells you to install `@modelcontextprotocol/sdk`, but the server imports `@modelcontextprotocol/server` (v2, which exposes the `./stdio` subpath it uses). The in-repo `src/mcp/README.md` already uses the correct package.

Reported on Slack against Eufemia 11.13.0.

### DIFF
--- a/packages/dnb-design-system-portal/src/docs/uilib/usage/first-steps/tools.mdx
+++ b/packages/dnb-design-system-portal/src/docs/uilib/usage/first-steps/tools.mdx
@@ -219,7 +219,7 @@ Run the server from your project (where `@dnb/eufemia` is installed):
     "eufemia": {
       "command": "node",
       "args": [
-        "${workspaceFolder}/node_modules/@dnb/eufemia/mcp/mcp-stdio.js"
+        "${workspaceFolder}/node_modules/@dnb/eufemia/mcp/mcp-server.js"
       ]
     }
   }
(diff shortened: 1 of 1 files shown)


joakbjerk:
Coordination note: this docs change should land **after** #9339 is released.

#9337 corrects the local-server command, but the stdio server in 11.13.0 (and any version before #9339) still fails at startup — `validateDocsRoot` runs against the unresolved `'<pending>'` docs root, so the corrected docs would point at a server that still can't start (just with a different error). #9339 is the code fix that makes it actually run.

The portal docs are version-agnostic (one live site), so the local-server section only truly works once readers are on the release that contains #9339. Suggested sequence: merge/release #9339 first, then merge this. Optionally add a "requires `@dnb/eufemia` ≥ <version>" note here once that release number is known.

tujoworker (approved):

PR #9340, opened 2026-09-11
chore(deps): resolve the joi and @vitest/mocker security advisories

langz:
Closes the three open Dependabot alerts. All three sit in build/test tooling — the shipped dependency trees were, and remain, clean.

`@vitest/browser` and `@vitest/browser-playwright` were left on 4.1.10 when #9305 bumped `vitest` to 4.1.11. Both pin `@vitest/mocker` exactly, so the vulnerable 4.1.10 copy stayed in the lockfile (GHSA-82fw-gwwq-j7x9) and their exact `vitest` peer requirement stopped matching the installed runner.

The root `joi: 17.13.4` resolution predates the `start-server-and-test` 2.0.1 → 3.0.11 bump in #8775, which replaced `wait-on` 7 (joi `^17`) with `wait-on` 9 (joi `^18.2.1`). The pin has since been forcing an out-of-range major downgrade onto the joi 17 line, which is where both remaining advisories live (GHSA-gg4h-3hg2-grpc, GHSA-6w3j-5fw6-r9vr). `wait-on` is the only consumer, so dropping the resolution lets joi resolve to a patched 18.x and the override is no longer needed.


joakbjerk (approved):
Bumpernickel ✅

PR #9342, opened 2026-09-14
chore(Portal): validate generated internal links

tujoworker:
Broken internal links and heading anchors can currently reach production unnoticed.

Validate the prerendered Portal output against a generated page and anchor map, and fail the build when an internal page, anchor, or redirect target cannot be resolved. Live component previews are excluded because they intentionally contain example URLs.


langz (approved):
Looks good :) When does this "check" run?

tujoworker:
@langz Could you take a look at this approach? It generates a map from the complete prerendered pages and anchors, then fails the build when an internal page, anchor, or redirect target cannot be resolved. It runs once per PR in the complete preview build and is skipped for the intentionally partial E2E and visual builds.

langz:
> @langz Could you take a look at this approach? It generates a map from the complete prerendered pages and anchors, then fails the build when an internal page, anchor, or redirect target cannot be resolved. It runs once per PR in the complete preview build and is skipped for the intentionally partial E2E and visual builds.

Will do :)

langz:
Nice approach — I like it. Validating the rendered output rather than the source is the right call: that is what caught the links our components generate, not just the ones we hand-write.

I pushed two commits:

- **af645df** — the "skip anything with a dot in the last path segment" rule was a little too blunt: it hid the four `v4.x-info` release pages, and also `/mailto:tobias.hoegh@dnb.no`, which is broken on two live pages today. Targets are now matched against the files the build actually emitted, so nothing is skipped by name. That needed a small fix in the portal `Anchor` too, which prefixed scheme-less hrefs with `/` and handed `mailto:` to the router.
- **4b43d49** — moved the validation to the end of the prerender step, so it also sees the markdown copies, `llms.txt` and fonts that are written after the pages.

Verified against a full local build and the preview build — all green.

Two small non-blockers if you want them: `internal-link-map.json` (1.3 MB) ships to the live site with no consumer, and since the check runs inside the release job, a broken doc link would currently stop an npm publish.

Thanks for taking this on, good to have the gate in place!

tujoworker:
> a broken doc link would currently stop an npm publish.

I think that should not happen, because the check runs both on PRs and from main -> release in our pipeline. So its fine as is.

PR #9343, opened 2026-09-14
chore(Portal): record the referrer category as a page-view dimension

joakbjerk:
Records how a page view was reached as a coarse `referrer` category — `search`, `internal`, `direct` or `external` — so we can see whether people arrive from a search engine versus a link elsewhere or a same-origin navigation. Only the category label is stored; the referrer URL and host are never sent, so the view stays anonymous.

It reads `document.referrer` at capture and maps it to a bucket client-side (a small allow-list of search-engine hosts → `search`; same origin → `internal`; empty → `direct`; anything else → `external`), mirroring how the existing locale/theme/color-scheme dimensions are captured. The category is validated at the ingest boundary and stored in a new nullable `referrer` Glue column.

Stacked on #9326; ships dark until the collector endpoint is set.


langz:
`document.referrer` belongs to the document, not to the navigation — `pushState` never touches it. All real navigation in the portal is client-side, so the arrival referrer gets stamped onto every view for the rest of the tab's life. Less a dimension, more a tattoo.

Verified in Chromium from a genuine cross-origin arrival: client-side navigation, reload and the back button all leave it unchanged; only a plain `<a>` full page load moves it. So one visitor arriving from Google and reading four pages yields `search, search, search, search` — which measures how much people read, not where they came from.

`internal` has the inverse problem. It needs a full document load, and the only ones left are the demo links on the Anchor/typography/elements pages (MDX 3 keeps literal `<a>` as plain HTML) plus a few `Button`/`Card.Action` demos with an internal `href` and no `element={Link}`. So it faithfully counts "clicked the link that exists to demonstrate underlines".

Five-line fix: read `document.referrer` once per document and label later views `internal`. All 22 tests pass before and after, which is the actual tell.

joakbjerk:
Good catch — and thanks for the Chromium verification, that's the crux. Fixed: read `document.referrer` once for the arrival view, then label every later in-app navigation `internal`, so the category counts where a visit started rather than how much was read. Added a test asserting `search, internal, internal` for a search arrival that then reads two more pages.

langz (approved):
Tattoo removed — laser clean.

Re-verified on the rebased head: put `analyticsReferrerCategory()` straight back into the buffer and it prints `search, search, search` again, so your new test is doing real work rather than nodding along. Mutated the host suffix match and the ingest allow-list too — both bite. Lookalikes behave: `evil-google.com` and `google.com.phishing.example` both land in `external`. 98 + 24 green, types and prettier clean.

Two nits for whenever, neither blocking: `mail`/`docs`/`drive`/`translate.google.com` all suffix-match into `search`, and `google.no` is dead weight — `.no`, `.se`, `.dk` and `.co.uk` all 301 to `google.com`, so the referrer says `www.google.com` regardless.

Approving. It measures where people came from now, instead of how stubborn they are.

tujoworker (approved):

joakbjerk:
Thanks for the mutation-testing pass 🙏 Applied both nits:

- Search-host matching is now exact (apex or `www.`) instead of a suffix match, so non-search Google subdomains (`mail`/`docs`/`translate.google.com`) bucket as `external`, not `search`. Added a regression test for that case.
- Dropped `google.no` (and skipped the other ccTLDs) since they 301 to `google.com` — the referrer host is always `www.google.com`, so those entries never matched. Also switched `yahoo.com` → `search.yahoo.com`, its actual web-search host, now that matching is exact.

Heads-up: the branch has since been rebased onto `main` and `referrer` converted to the coerce-to-`unknown` pattern to match `48be823`, so a stale value drops the one dimension rather than the batch.

PR #9344, opened 2026-09-14
chore(mcp): record doc paths under one canonical key

langz:
Follow-up to #9325, targeting `feat/mcp-usage-tracking`. Rebased onto `6d74029` — the component half of this branch is dropped, since that commit already fixes it. What remains is the same defect in the sibling field.

### Doc paths are still split across spellings

`normalizeDocsPath` accepts a path with or without a leading slash, converts back-slashes, and collapses empty and `.` segments before reading the file, so these are all successful reads of the same document:

| path sent | `docs_read` returns | stored on `6d74029` | stored here |
| --- | --- | --- | --- |
| `/uilib/components/button.md` | the file | `/uilib/components/button.md` | unchanged |
| `uilib/components/button.md` | the file | *(blank — dropped)* | `/uilib/components/button.md` |
| `/uilib//components/./button.md` | the file | verbatim, separate row | `/uilib/components/button.md` |
| `/uilib/components/button.md/` | the file | verbatim, separate row | `/uilib/components/button.md` |
| `\uilib\components\button.md` | the file | *(blank — dropped)* | `/uilib/components/button.md` |
| `/uilib\components/button.md` | the file | *(blank — dropped)* | `/uilib/components/button.md` |

`/`, `///` and `/./` were also stored as paths in their own right, and `docs_list` prefixes split between `/uilib/components` and `/uilib/components/` while a back-slashed prefix was dropped outright.

Same root cause as the component name: the validator was written against the canonical example in the tool description rather than against what the resolver actually accepts. Since the rollup groups on the exact stored string — `GROUP BY dt, tool, component, path` in Athena with no `lower()`, and blank values skipped outright in `sumBy` — a dropped path is absent from the totals rather than merely unlabelled, and a split one double-counts as two documents.

### Guarantees unchanged

Traversal and length are still checked on the raw value, before any collapsing, so canonicalisation cannot mask a `..` the check would otherwise catch. That check is slash-agnostic, so the back-slashed forms are covered by it too: every `..` spelling still drops, including mixed ones like `/uilib/.\..\x.md`. Query strings and fragments are still stripped first. The character set is untouched, and path casing is deliberately left alone: `normalizeDocsPath` does not lowercase and resolution hits a case-sensitive filesystem, so the stored path stays exactly what resolved.

`canonicalDocsPath` reimplements the server's normalisation rather than importing it, to keep this module dependency-free. A test pins the two together against the real `normalizeDocsPath`, so they cannot drift apart the way the validator and resolver already did twice. The first version of that pin was five forward-slash examples, which let the back-slashed divergence through — the same mistake one level down. It now enumerates each shape the normaliser treats as significant, back-slashes included, rather than a list of paths that happen to work.

Two things checked and left alone: `PATH_PATTERN`'s character set rejects none of the 1088 `.md`/`.mdx` paths in `main`, and none of them contain `..` or exceed the 512-character bound. The `docs_list` trailing-slash collapse is not over-broad either — the server does `withLeadingSlash(normalizeDocsPath(prefix)).replace(/\/?$/, '/')`, force-appending the slash, so every spelling returns an identical list.

### One unrelated fix carried along

`6d74029` lowercases the component but only updated `mcp-usage.test.ts`, so `usage-capture.test.ts:67` still expects `Button` and the suite is red on that commit. One-line assertion updated here so this branch is green. Nothing in CI would have told you at the time: `eufemia-mcp-lambda` was absent from the test, lint and typecheck jobs in `verify.yml`. That is fixed on `main` now (#9346), but this branch and its base predate it, so the checks here still do not run the suite — merging `main` into `feat/mcp-usage-tracking` would give it that check.

### Verification

Verified locally, since this branch predates the CI coverage: 106 tests pass, typecheck and lint clean. On `6d74029` the suite is 1 failed / 86 passed; with this branch it is 106 passed. The new path assertions fail against `6d74029` with the already-canonical path passing on both sides as a control, and the back-slash assertions fail against the first commit here for the same reason.


langz:
One gap left, and it is the same class as the one this fixes: `normalizeDocsPath` also does `.replaceAll('\\', '/')`, so the server resolves `uilib\components\button.md` to the same document — but `canonicalDocsPath` splits on `/` only, the backslash survives into `PATH_PATTERN`, and the record is dropped entirely rather than merely split.

Reproduced against the real resolver (temp docs root, `createNodeDocsSource`), with the forward-slash forms as controls:

| path sent | server resolves it | stored on `58068b7` |
| --- | --- | --- |
| `/uilib/components/button.md` (control) | yes | `/uilib/components/button.md` |
| `uilib/components/button.md` (control) | yes | `/uilib/components/button.md` |
| `\uilib\components\button.md` | yes | *(blank — dropped)* |
| `/uilib\components/button.md` | yes | *(blank — dropped)* |

Same on the sibling field: `\uilib\components` returns the identical file list through a real `docs_list` call, and stores blank.

What makes it worth fixing rather than just noting is the guard. The parity `it.each` pins five inputs and all five are forward-slash, so the test meant to stop `canonicalDocsPath` and `normalizeDocsPath` drifting does not cover the divergence that already exists — the same "written against the canonical examples" shape this PR is correcting.

`.replaceAll('\\', '/')` before the split closes it: 100 tests still pass, every backslash variant lands on `/uilib/components/button.md`, and the traversal guard is untouched because `raw.includes('..')` is slash-agnostic — every `..` form still drops, verified with and without the change. Pushing that plus backslash cases in the parity list and in both argument lists.

Two things I checked and discarded, for the record:

- `PATH_PATTERN`'s character set is not too narrow — it rejects none of the 1088 `.md`/`.mdx` paths on `main`, and none of them contain `..` or exceed the 512-character bound.
- The `docs_list` trailing-slash collapse is not over-broad. The server does `withLeadingSlash(normalizeDocsPath(prefix)).replace(/\/?$/, '/')`, force-appending the slash, so every spelling returns an identical list — verified through real `docs_list` calls with a `uilib/components.md` decoy that a looser prefix would have swallowed.

PR #9347, opened 2026-09-14
chore(analytics): friendlier dashboard message for a not-yet-generated snapshot

joakbjerk:
When the dashboard API has no snapshot yet — most likely right after a deploy, before the generator's first run — the read endpoint returns 503, and the page showed a generic "the data API returned an error (503)" message that reads like something is broken.

This maps a 503 to softer, deploy-aware copy asking the viewer to refresh shortly, while other error statuses keep the generic message with the code.


langz (approved):
🍳

PR #9348, opened 2026-09-14
chore(CI): decouple portal link validation from the release

langz:
Follow-up to #9342.

The link validation ran inside the portal build, so any pipeline that builds the portal could fail on a documentation link — including two that have nothing to do with docs:

- **The release.** The portal build sits in the `build` job, before the package is prepared and before the consumer smoke tests. `action` has `needs: build`, so it never publishes if that job fails. A mistyped heading anchor could hold back an `@dnb/eufemia` release.
- **The icons pipeline.** `icons-lib.yml` builds the portal only to take screenshots, then commits the updated icon snapshots. A broken link elsewhere in the docs would stop the Figma icon sync from ever committing.
- **E2E and visual builds** had to opt out with `IS_E2E` / `IS_VISUAL_TEST` env vars.

Whether a broken link should be fatal is a per-pipeline decision, so this moves it out of the build script and into the pipelines. The build only builds; a new `validate:links` script runs the check where it belongs.

| Pipeline | Before | After |
|---|---|---|
| PR preview | blocking, inside the build | blocking, explicit step — unchanged gate |
| Release | blocked the npm publish | reports in its own job, publish unaffected |
| Icons | could block the icon sync | does not run it |
| E2E / visual | opted out by env var | does not run it |

The release job unpacks the uploaded Pages artifact rather than rebuilding, so it validates the exact output that gets deployed. Nothing has `needs:` on it — a dead link produces a red check without holding back the publish or leaving the docs site stale, and every PR already blocks on the same check.

### The validator also stopped passing on nothing

While testing the artifact restore I mislaid the output directory and the check reported `✓ Validated internal links across 0 pages` and exited 0. Now that the script can be run on its own, an empty or misplaced output would have been a permanent false green, so it fails with a clear message instead.

### Verified

Full portal build: the build no longer validates, `yarn validate:links` reports 1,075 pages, and a deliberately broken link still exits 1. I reproduced the new release job locally — packed the output into an uncompressed `artifact.tar` the way `upload-pages-artifact` does, unpacked it into a clean `public/`, and got the same 1,075 pages, including with hidden files stripped as the action strips them. `actionlint` is clean on both workflows.

Note on #9345: it edits the same file, so I placed the new guard clear of its hunk. I merged the two branches locally to confirm they combine without conflict and that the combined tree still passes.


langz:
The guard I added here only covered a `public/` that exists and is empty. With the directory absent — which is where you are the first time you run `yarn validate:links` by hand — `findOutputFiles` ran before the guard and died in `readdirSync` with a raw `ENOENT ... scandir` instead of "Build the portal first." Still exit 1, so never a false green, just the wrong message in the one case a person is most likely to hit, and the case CI never sees because the release job creates the directory before unpacking.

f1c1d70 skips the scan when the directory is absent, so both states report through the same guard. Verified all four cases: missing and empty each give the guard message and exit 1, a full build still validates 1,075 pages, and a deliberately broken link and heading anchor still exit 1. Re-checked against #9345 as well — still merges clean, same 1,075 pages, its 7 tests pass.

PR #9351, opened 2026-09-15
chore(analytics): adopt snake_case for stored field names

joakbjerk:
`createdat` is not a deliberate style — it's `createdAt` flattened to lowercase by Athena/Glue, which lowercases all identifiers. snake_case is the standard warehouse column convention and survives lowercasing readably (`created_at`, not `createdat`). This renames the stored portal-view field/Glue column accordingly.

Local TS variables stay camelCase; only the serialized key and Glue column become `created_at`. No data migration: tracking still ships dark and Glue is schema-on-read, so the destroy/recreate of the column is fine with no real data yet.

Sibling to #9326 (`colorScheme`→`color_scheme`). Should land after #9326 and #9343 merge, since both also touch the portal-view record shape.


langz (approved):
Good good 🙂 `createdat` always read like a typo that slipped through customs — `created_at` is much kinder on the eyes.

Checked the Glue columns against what `buildPortalViewRecord` actually emits and they line up exactly (`created_at, env, path, status, timestamp`), and nothing SELECTs the column yet, so this is about as free as a rename gets 🤝

PR #9352, opened 2026-09-15
fix(NumberFormat): normalize minus signs across locales

tujoworker:
Intl.NumberFormat returns different minus characters by locale, which makes otherwise equivalent formatted values look inconsistent. Normalize rendered number, currency, percentage, Stat, semantic-part, and ARIA output to the proper Unicode minus sign (U+2212). Keep U+002D for editable number inputs and copied values so keyboard input, parser round-trips, and spreadsheet pasting remain compatible.

Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1789461772635629


langz (approved):

HenrikT (approved):

langz (commented):
Three things, one of them load-bearing.

### `cleanNumber` can no longer read what `NumberFormat` writes

`normalizeMinusSign` stamps U+2212 onto every locale, but `NUMBER_CHARS` is still `'\\-0-9,.'` and `cleanNumber`'s leading-char strip is still `/^[^0-9-]/`. Both only speak U+002D, so the sign isn't mis-parsed — it's deleted.

Verified on both branches, same input:

| | `main` | this PR |
| --- | --- | --- |
| `formatNumber(-1234.5, { locale: 'nb-NO' })` | `-1 234,5` | `−1 234,5` |
| `cleanNumber(…that…)` | `-1234.5` | `1234.5` |
| `<NumberFormat.Number clean>…that…</NumberFormat.Number>` | `-1 234,5` | `1 234,5` |

A debit that quietly turns into a credit is a bold feature for a bank.

Same root in the copy path: `copySelection` defaults to `true`, and the non-`cleanCopyValue` branch of `buildReturn` derives `cleanedValue` from `display` — so a plain copy now hands out `−1234,5` in every locale, where `main` gave ASCII everywhere. The `minusSign → '-'` case you added is the right idea, it just needs a twin on the `else`.

### 9 failing tests

In isolation (the three others that fail in a full forms run are pre-existing flakes — they pass alone on both branches):

```
FAIL src/extensions/forms/Value/Currency/__tests__/Currency.test.tsx  (5)
FAIL src/extensions/forms/Value/Number/__tests__/Number.test.tsx      (4)

Expected: "-12 345,68 kroner"
Received: "−12 345,68 kroner"
```

### `Stat.Trend` didn't get the memo

`Trend.tsx` still hard-codes `sign: '-'`:

```
nb-NO  Stat.Percent -4.98  →  − 4,98 %
nb-NO  Stat.Trend   -2.1%  →  -2.1%
```

`stat-root-and-label` puts a `Stat.Currency` and a `Stat.Trend` in the same box, so it's the original bug with a new postcode.

tujoworker:
Thanks, good catches! Fixed the sign round-trip and copy paths, aligned Stat.Trend, and updated the missed Value and Slider cases. I also added regressions for both copy modes and currency-before-sign locales, and the full unit and visual suites are green.

langz (commented):
All three from last round are gone — verified locally on `ccd0b468ffe`: `cleanNumber('−1 234,5')` → `-1234.5`, the default `cleanedValue` is ASCII again in both locales, and `Stat.Trend` now renders `−`. Full run across number-format, stat, slider, input-masked and forms: 5962 passing, and `Run tests` is green where it was red. Collapsing the old comment.

One red check left, `Run visual-regression tests` — 15 snapshots, and they're the expected ones rather than anything broken:

```
List (ui, sbanken)           ×4   inside card, ± scroll view
Stat (ui, sbanken)           ×4   root-and-label, currency-default
NumberFormat                 ×4   default, compact, locales, currency
NumberFormat with skeleton   ×2   default, currency
Filter (ui)                  ×1   date-and-selection
```

The tell that it's only the glyph: `NumberFormat > have to match percent numbers` is *not* in the list. Percent was already U+2212 on `main`, so that baseline never moved — every diff is a surface that used to be ASCII. Visual regression is green on `main` (`f3e1a0f`), so it's this change and nothing else.

Baselines need regenerating, then it's done.

tujoworker:
Added some mdx docs about it as well.

langz (commented):
Code side is done — verified on `60575ae3dd5`: the round-trip, both copy paths, `Stat.Trend` and `Run tests` are all green now. The docs section is a good addition too, and it quietly fixes the old "em dash" that was actually an en dash. Thanks for turning all of that around so fast, and for the extra regressions on the copy modes and currency-before-sign locales.

Baselines are 10 of 15, though. The visual job just came back red on the five that weren't regenerated:

```
✗ List for ui      > list inside card
✗ List for ui      > list inside card without scroll view
✗ List for sbanken > list inside card
✗ List for sbanken > list inside card without scroll view
✗ Filter for ui    > date and selection filter
```

Both of those demos do render negatives — `list-card` has `Value.Currency value={-500}` and `-99`, and `filter-date-selection` has `amount: -245` and `-3999`. Easy pages to miss, since the diff is one glyph wide and it is not a number component you would think to look at.

Regenerate those five and it is green.

langz (commented):
All green — `64184d1a29b` is `clean`, visual regression included. The last five were the ones that mattered, and the regenerated set now lines up exactly with the fifteen that moved.

Verified end to end: U+2212 everywhere on display, ASCII on both copy paths and in editable fields, `cleanNumber` reads Eufemia's own output back correctly, and `Stat.Trend` matches the rest.

Worth noting it fixed something nobody reported, too. `sv-SE` is U+2212 in CLDR and `cleanNumber` only spoke ASCII, so Swedish was silently turning negatives positive on round-trip before this — `−1 234,5` came back as `1234.5` on `main` and comes back as `-1234.5` now.

Thanks for the quick turnaround, and for writing the convention into the docs rather than leaving it to be rediscovered by whoever hits it next.

PR #9354, opened 2026-09-15
fix(NumberFormat): normalize formatted minus signs

HenrikT:
Use the mathematical minus sign consistently across formatted numbers, currencies, and percentages.


HenrikT:
Superseded by #9352, which centralizes the U+2212 normalization and preserves ASCII minus signs for copy values.

PR #9363, opened 2026-09-16
fix(Translation): insert formatMessage values literally

langz:
`formatMessage` passes the argument value straight into `String.prototype.replace` as a replacement string. That string has its own syntax — `$&`, `` $` ``, `$'` and `$$` are patterns, not literals — so any value containing them was rewritten instead of inserted:

| call | before | expected |
| --- | --- | --- |
| `formatMessage('Total: {amount}', { amount: 'a$&b' })` | `Total: a{amount}b` | `Total: a$&b` |
| `formatMessage('Ref {code}', { code: 'A$$B' })` | `Ref A$B` | `Ref A$$B` |

Values reach this path from consumer application data, so this is silent corruption of whatever the app passes in.

The placeholder name was also interpolated into the regex unescaped, letting a name like `user.name` match `{user_name}`.

`escapeRegexChars` already exists in `component-helper` and is used by number-format, `highlightText` and `searchUtils` — this brings `formatMessage` in line with it.


langz:
Self-review done — I consider this ready to approve.

I re-verified the substitution against the cases that worried me, all behaving correctly: repeated placeholders, `$&` / `` $` `` in values, numbers, `undefined`, objects, and keys containing `-`, a space, a comma, or an empty key. Coercion is unchanged from before (`String(value)` matches what `replace` did implicitly), so non-string args behave exactly as they did.

All 18 checks are green.

joakbjerk (approved):

PR #9364, opened 2026-09-16
chore(Element): keep the base class when className ends with it

langz:
`Element` skips adding its base `dnb-*` class when the incoming `className` already contains it. The check used `new RegExp(`${tagClass}(\\s|$)`)`, which has no boundary at the start, so a consumer class that merely *ends* with the tag class counted as a match and the base class was dropped:

```jsx
<Span className="x-dnb-span" />   // class="x-dnb-span"        — dnb-span missing
<Ul className="a b custom-dnb-ul" />  // class="a b custom-dnb-ul" — dnb-ul missing
```

The result is silently unstyled output. `Element` backs `Typography`, so this applies to every HTML primitive (`P`, `Span`, `H1`–`H6`, `Ul`, `Li`, `Td`, …).

Comparing whole class names fixes it and also removes a `RegExp` compiled on every render of every primitive.


langz:
Self-review found a regression in my own change — fixed in d89c8aeaff5.

`internalClass` can hold **several** class names (`Img` passes `clsx('dnb-img', hasError && 'dnb-img--error')`), and my whole-string `includes()` could never match a multi-token value, so those classes would be re-added when they were already present.

Comparing per name also removes a duplicate the original regex left behind: with `internalClass="dnb-img dnb-img--error"` and `className="dnb-img"`, both the old code and my first attempt rendered `dnb-img` twice. Only the missing names are added now.

I checked every committed snapshot for duplicate class tokens first — there are none, so no recorded output changes.

joakbjerk (approved):

PR #9365, opened 2026-09-16
chore(MCP): apply the transport security requirements to the docs HTTP server

langz:
The [MCP Streamable HTTP spec](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports) states three transport requirements. The docs HTTP server met none of them by default:

| Requirement | Before |
| --- | --- |
| **MUST** validate the `Origin` header (DNS rebinding) | not validated; only `Host` was checked, and only when `MCP_ALLOWED_HOSTS` was set |
| **SHOULD** bind to loopback when running locally | defaulted to `0.0.0.0` |
| **SHOULD** authenticate connections | no token configured by default |

So the out-of-the-box posture was an unauthenticated server reachable on every network interface.

This adds an `Origin` allowlist (`MCP_ALLOWED_ORIGINS`, defaulting to loopback origins), changes the default bind to `127.0.0.1`, and logs a warning when the server is bound to another interface without `MCP_AUTH_TOKEN`.

Hosting on a public URL still works — `run-mcp-http-server.sh` already sets `HOST` explicitly, and the README now documents the variables to set alongside it.

Note the deployed `tools/mcp-lambda` was already correct and fails closed; this is the self-hostable variant shipped in the package.


langz:
Self-review turned up one more thing — fixed in 17125190128.

The `Host` and `Origin` guards were registered *after* `express.json({ limit: '4mb' })`, so a request that was about to be rejected still had its body parsed first. Both guards only read headers, so they now run before the parser.

Added a test that pins the order: a disallowed `Origin` with a malformed body must return 403, not the 400 the parser would produce. I verified it fails against the previous order.

Moving the `Host` guard is a line I did not otherwise need to touch — say the word if you would rather keep that out of this PR.

joakbjerk (approved):
Reviewed the latest revision against the source. Cleanly maps the three MCP Streamable HTTP transport requirements onto the self-hostable server: loopback-default Origin allowlist, loopback bind, and a warning when exposed without a token. Both header guards now run before `express.json` (rejects before the 4mb parse, with a test pinning 403-before-400), and the Origin logic checks out.

Two non-blocking notes:
- Explicit `MCP_ALLOWED_ORIGINS` replaces the loopback default, so a browser client on `localhost` then gets 403 — reads as intended, worth a one-line README note.
- Origin matching is exact-string, so a config typo (trailing slash, `:443`) fails closed silently. Fine for canonical browser origins; a doc note could help.

LGTM.

joakbjerk:
One more caveat I left out above: a cross-origin no-cors GET omits `Origin`, so it passes the guard — but the response is opaque, and any actual read needs `cors` mode, which forces an `Origin` and gets blocked. So `MCP_AUTH_TOKEN` + `MCP_ALLOWED_HOSTS` remain the real defense for an exposed host; the Origin allowlist is the anti-DNS-rebinding layer, not a standalone one. Worth a one-line note in the README.

langz:
Thanks @joakbjerk — all three points addressed in 48af1aec1bc.

**Exact-string matching.** I went past the doc note and normalised instead, because I could reproduce the footgun: against a browser sending `https://eufemia.dnb.no`, a configured `https://eufemia.dnb.no/` and `https://eufemia.dnb.no:443` both failed to match. Entries are now compared as canonical origins via `new URL().origin`, so trailing slash, default port and casing all match. Anything that is not a valid origin is skipped and reported at startup rather than silently never matching. Covered by tests, including the startup warning.

**Explicit list replaces the loopback default.** Documented — the README now says to list the local origins too when a browser on `localhost` should keep working.

**Not a standalone control.** Documented as a DNS-rebinding guard that does not replace `MCP_AUTH_TOKEN` and `MCP_ALLOWED_HOSTS` on an exposed host. I deliberately wrote it as what the guard is *for* rather than spelling out the request shape that sidesteps it, since this repo is public and `AGENTS.md` asks us not to document bypass paths.

Let me know if you'd rather the normalisation were its own PR — happy to split it.

joakbjerk:
Verified `48af1aec` — normalization looks right: canonical `new URL().origin` on both sides, invalid entries skipped and surfaced at startup, malformed/absent incoming Origin fails closed. Good call fixing the footgun rather than just documenting it, and agreed on framing point 3 as what the guard is *for* rather than the bypass shape. No need to split — the normalization resolves a nitpick from this PR and belongs with the allowlist it hardens. Still LGTM.

PR #9368, opened 2026-09-16
chore(Eufemia): enforce eslint-plugin-security rules

langz:
The lint script runs with `--quiet`, which reports errors only, while `eslint-plugin-security` ships every rule as a warning. None of its findings could surface — not locally, and not in CI. This raises the recommended set to errors so the rules the repo already opted into actually gate the build.

Turning them on exposed four regexes in the postbuild test that backtrack exponentially. In `([\r\n][^}]*)+` the inner `[^}]*` also matches newlines, so it overlaps itself: 22 non-matching lines already take 120 ms and the cost grows roughly fivefold every two lines. Two of the four sit in `not.toMatch` assertions, where not matching is the expected outcome — the case that triggers the blowup. Removing the redundant outer group matches the same input in linear time.

The remaining findings are false positives from the `safe-regex` heuristic. Each was measured against adversarial input and stays linear, so they are suppressed with a short reason rather than rewritten.


langz:
Re-reviewed this branch against the source. Ready for approval from my side.

**What the review changed**

Three suppression comments claimed more than the code supports, so they were corrected rather than left as-is:

- the import matcher in `style-imports.test.ts` is bounded by the required `from`, not by the quote;
- the relative-time alternatives in `DateFormat.test.tsx` do overlap (`uker?` and `uke`), so the safety comes from the alternation not being repeated, not from disjointness;
- the spacing pattern in `SpacingUtils.ts` was described imprecisely.

The rule promotion now also preserves any options the plugin sets on a rule, so a future `eslint-plugin-security` upgrade cannot drop them silently while being raised to `error`.

**Verification**

- Every suppressed regex was benchmarked over a doubling input (2k → 32k adversarial chars). All stay flat at well under 1 ms, so each is linear and the suppression is safe.
- `eslint --print-config` confirms 11 security rules resolve to `error` and that the three deliberate opt-outs (`detect-object-injection`, `detect-non-literal-fs-filename`, `detect-non-literal-regexp`) still resolve to `off`.
- The four rewritten regexes could not be exercised locally, because `test:postbuild` needs a full build. CI covers it: `Run build tests` → `postbuild:ci` → `yarn test:postbuild`, and it is green here. That is the real validation of the rewrite against actual build output.
- Equivalence of the rewrite was additionally fuzzed over 200k random CSS-like inputs, with no divergence.

All 17 checks pass.

PR #9369, opened 2026-09-16
refactor!: enforce component-name prefix on exported component types

langz:
## Motivation

The `component-types/require-component-prefix` rule already existed but ran as a **warning** with 100 violations, so exported component type names had drifted (e.g. `AmountProps`, `TextProps`, `Svg`, `ItemBasicProps`). Generic, unprefixed type names are ambiguous for consumers and conceptually collide across components.

This enables the rule as an **error** and renames all 100 flagged exported types so each name carries its component prefix (e.g. `ItemBasicProps` → `ListItemBasicProps`, `AmountProps` → `StatAmountProps`, `Svg` → `LogoSvgInput`).

## ⚠️ Must go into our next breaking release (v13)

14 of the renamed types are part of the public API (re-exported from their component entry), so renaming them is a **breaking change** for consumers importing those types. **This has to go into our next breaking release.** The public renames are documented as a search-and-replace migration table in the v13 release info.

The renames are **type-only** — they were performed via the TypeScript language service so every reference is updated, and the runtime API (components, hooks and constants) is unchanged. The same-named Icon runtime constants (`ValidIconType`, `DefaultIconSizes`) keep their names; only the types were renamed.

## Notable decision

The spacing infrastructure keeps its canonical `Spacing`/`InnerSpace` names (e.g. `SpacingProps`, re-exported from `shared/types` and used across ~100 files) via the rule's `allowlist` rather than being renamed to a `Space*` prefix. Happy to rename these instead if you'd prefer.


langz:
Heads-up for reviewers/release: this renames public (exported) type names, so it is a **breaking change** and must be merged as part of our **next breaking release (v13)** — not into a minor/patch. Migration guidance is included in `v13-info.mdx`.

langz:
## Review — ready to approve

Re-reviewed and rebased onto latest `main`. The previously failing **`Run type checks`** is now **passing**; CI shows 0 failures.

### Finding, fixed

`Run type checks` runs **eight workspaces**, not just `@dnb/eufemia`. I had only verified the library, so I missed that the portal imports the renamed type:

```
ListSummaryFromEdges.tsx(8,8): error TS2614: Module '@dnb/eufemia/src/components/Heading'
  has no exported member 'InternalHeadingLevel'
```

Fixed the portal usage (`InternalHeadingLevel` → `HeadingInternalLevel`) and then swept every non-library workspace for all 91 renamed names — that was the only real code reference (the other hits are the portal's own unrelated local `ContentProps` and the migration tables in the release notes).

All eight workspaces now type-check clean.

### Corroboration for the naming

Worth noting for the review: **v11 already performed this exact renaming with the same convention**, so these names are consistent with established precedent rather than newly invented. From `v11-info.mdx`:

| Before | After |
| --- | --- |
| `InfinityScrollerProps` | `PaginationInfinityScrollerProps` |
| `CalendarNavigationEvent` | `DatePickerCalendarNavigationEvent` |
| `ContentWrapperProps` | `TabsContentWrapperProps` |

My `PaginationInfinityScrollerProps` is literally the name v11 already documented for that type.

### Still requires a breaking release

Unchanged: 14 of the renamed types are public, so this must go into **v13**. Migration table is in `v13-info.mdx`.

### Verified

- `require-component-prefix`: **0** violations, rule enforced as `error`
- All 8 workspace type checks: 0 errors; ESLint 0 errors; Prettier clean
- Icon's runtime constants (`ValidIconType`, `DefaultIconSizes`) deliberately keep their names — only the types were renamed

PR #9371, opened 2026-09-16
chore(Portal): enable the achievable strict TypeScript checks

langz:
Moves the portal toward `strict` by enabling every strict-family flag that can be enabled today, rather than leaving the package with a single opt-in (`strictFunctionTypes`).

### What is on now

`alwaysStrict`, `noImplicitThis`, `strictBindCallApply`, `useUnknownInCatchVariables`, plus `noFallthroughCasesInSwitch` and `noImplicitOverride`. This brings the portal closer to the posture `packages/dnb-eufemia` already has.

### Why not full `strict`

The portal imports `@dnb/eufemia/src/...` directly in 255 files, so the library source is part of the portal's type program. Measured per flag:

| flag | total errors | in portal | in library |
| --- | --- | --- | --- |
| `alwaysStrict` | 0 | 0 | 0 |
| `noImplicitThis` | 1 | 1 | 0 |
| `strictBindCallApply` | 5 | 5 | 0 |
| `useUnknownInCatchVariables` | 13 | 12 | 1 |
| `noImplicitOverride` | 5 | 5 | 0 |
| `noFallthroughCasesInSwitch` | 0 | 0 | 0 |
| **`noImplicitAny`** | 1393 | 548 | **845** |
| **`strictNullChecks`** | 3003 | 468 | **2535** |

The last two cannot be enabled from this package — they are blocked on making the library strict-clean first, which is its own piece of work.

I also measured `noImplicitReturns` (not part of `strict`): 55 errors, all in the portal, but every one is an intentional idiom — validators returning nothing when valid, `useEffect` with a conditional cleanup, and Vite plugin hooks returning nothing when they do not handle an id. It would push `return undefined` into documentation examples that users copy, so I left it off.

### How the errors were fixed

No `any`, no `unknown`, no casts:

- Caught values are narrowed with `instanceof Error` before reading `.message` or returning them. `Error` is exactly what `ValidatorReturnSync` and `EventReturnWithStateObject` accept.
- The `this` context of a debounced function is now typed. `DebounceHelpers` (previously the file-local `ReturnHelpers`) is exported from `shared/helpers`, so consumers can annotate `this` instead of guessing the shape. Documented under helpers.
- React class members got explicit `override`.
- The `CodeBlock` form mocks narrow with `element instanceof HTMLFormElement` instead of comparing a tag name string, which is what made `submit`/`action` resolve.

### Verified

`tsc --noEmit` is clean for the portal and for all three tool packages that extend its config (`analytics`, `eufemia-css-optimizer`, `eufemia-llm-metadata`), and for `dnb-eufemia` after the debounce rename.


langz:
Self-review: trimmed the docs snippet in 058b4201aa2 — it repeated the whole async example above it, when the only new information is the `this` annotation.

Two things worth a reviewer's attention:

1. **This PR touches the library.** Exporting `DebounceHelpers` from `shared/helpers` is not portal-local. It is needed to type `this` in the debounced validator example; the alternative was duplicating the helper shape inline in the docs. Happy to split it out if you would rather keep the packages separate.
2. **`noImplicitReturns` is deliberately off.** All 55 of its errors are in portal code, so it is technically achievable, but every one is an intentional idiom — validators returning nothing when valid, `useEffect` with a conditional cleanup, Vite hooks declining an id. It would push `return undefined` into examples users copy.

I re-confirmed `tsc --noEmit` is clean for the portal, `dnb-eufemia`, and the three tool packages that extend the portal config.

PR #9372, opened 2026-09-16
chore: move component types into sibling types.ts files

langz:
## Motivation

`component-types/no-inline-type-exports` had 24 violations — component files carrying large blocks of exported type definitions (up to 29 in `DrawerList.tsx`, 24 in `Pagination.tsx`, 22 in `NumberFormat.tsx`). The rule exists to keep component files focused on rendering logic and make types easier to find and reuse.

This brings the count to **0**.

## What changed

The exported type definitions moved out of the component `.tsx` files into a sibling `types.ts` (20 new files).

**No public API changes.** Each component re-exports its types with `export * from './types'`, so every existing import path keeps working — both deep paths like `@dnb/eufemia/components/button/Button` and the component entries that already do `export * from './Button'`. This also means no importing file needed to change.

Two files needed different treatment:

- **Icon** — the size constants (`DefaultIconSizes`, `ValidIconType`, `ListDefaultIconSizes`, `DefaultIconSize`) and the types derived from them via `typeof` can't be separated without a circular import, so both now live in a new `sizes.ts`. `Icon.tsx` re-exports it, so the constants remain publicly available under the same path.
- **NumberFormat** — it's a namespace barrel whose "type exports" are re-exports from surrounding modules. Those now route through `types.ts`.

## Validation

- `component-types/no-inline-type-exports`: 24 → **0**
- `tsc --noEmit`: 0 errors; ESLint: 0 errors; Prettier clean
- Tests for every affected component: **1,864 passed**
- Verified public types *and* the Icon value exports still resolve from their original entry paths

Note the diffstat is near-neutral (+3596/−3671) because types were moved rather than duplicated.


joakbjerk (approved):

langz:
## Review — ready to approve

Re-reviewed after rebasing onto latest `main`. All three previously failing checks — **`Run build tests`**, **`Build and smoke-test consumers`** and **`Publish`** — are now **passing**; CI shows 0 failures.

### Finding, fixed

All three failures shared a single root cause: `postbuild.test.ts` asserted that `InputProps` is declared **inside** the emitted `Input.d.ts`. This refactor moves the types to a sibling `types.ts`, so that assertion no longer matched.

Before changing the test I verified the guarantee it protects still holds:

- `types.d.ts` **is** emitted next to `Input.d.ts`, so the re-export is not dangling
- a consumer importing `InputProps` / `ButtonProps` from the published component path still resolves correctly through `export type * from './types'`

So the assertions moved to the emitted `types.d.ts`, and the test now *also* checks that both files are emitted and that the re-export is present — slightly stronger than before.

### On your two commits

Both are improvements, thanks:

- **`export type * from './types'`** is better than my plain `export *` — it states the intent explicitly and is friendlier under `isolatedModules`. I re-verified consumer resolution still works through the type-only form, and that it still doesn't count toward the rule (it's an `ExportAllDeclaration`, not a named export).
- **Enforcing the rule as `error`** — verified this is safe against current `main`: the new SidebarMenu extension exports 0 inline types, and the rule's scope covers `components`/`elements`/`fragments`, not `extensions`.

I also re-checked your updated `/export type \* from/` assertion against real emitted output — the d.ts contains `export type * from './types';`, so it matches.

### Verified

- `no-inline-type-exports`: **0** violations; ESLint 0 errors
- `tsc --noEmit` 0 errors; Prettier clean
- Consumer type resolution confirmed against freshly emitted declarations
- No public API change: every existing import path keeps working

packages/dnb-eufemia/src/components/autocomplete/types.ts
```diff
@@ -0,0 +1,361 @@
...
+  value: string
+  previousValue: string | number | null
+  event: SyntheticEvent | Event
+}
+
+export type AutocompleteMode = 'sync' | 'async'
+
+export type AutocompleteAlign = 'left' | 'right'
+
+export type FormLabelLabelDirection = 'horizontal' | 'vertical'
```
langz: Nineteen types went from private to public without being asked.

Moving a type into `types.ts` forces an `export` on it, and `export type * from './types'` then ships it. So declarations that were deliberately bare `type X = …` on `main` are now part of the API: all thirteen `Autocomplete*` aliases here, plus `FormLabelLabelDirection` (this line), `ReactRouterLink` (`button/types.ts:54`), `IconType` (`icon/types.ts`), `AccordionHeaderIconIcon`, and `SectionSpacingProps` / `SectionReturnParams`.

`import type { FormLabelLabelDirection } from '@dnb/eufemia/components/Autocomplete'` doesn't compile on `main` and does here — which is a funny thing for an Autocomplete entry to hand out. The two odd ones out are also the only new lint warnings in the PR: `require-component-prefix` goes 100 → 102, and it's exactly `FormLabelLabelDirection` and `ReactRouterLink`. The rule is telling us they don't belong on the public surface.

packages/dnb-eufemia/src/components/accordion/AccordionHeader.tsx
```diff
@@ -27,25 +26,26 @@ import {
...
+  AccordionHeaderDescription,
+  AccordionHeaderDescriptionProps,
+  AccordionHeaderIcon,
+  AccordionHeaderIconProps,
+  AccordionHeaderProps,
+  AccordionHeaderTitle,
+  AccordionHeaderTitleProps,
+} from './types'
+
+export type * from './types'
```
langz: One `types.ts` per directory plus `export type *` means the siblings start wearing each other's clothes.

`AccordionHeader` now exports `AccordionProps`, `AccordionGroupProps`, `AccordionInstance` and `accordionDefaultProps`; `Accordion` returns the favour with twelve `AccordionHeader*` types. Same trade going on between `Radio` ↔ `RadioGroup` and `ToggleButton` ↔ `ToggleButtonGroup`. 58 names in total now resolve from an entry they didn't on `main` — `import type { RadioProps } from '@dnb/eufemia/components/radio/RadioGroup'` being my favourite of the bunch.

Nothing breaks, it's all additive, and it's easy enough to live with — but it's not "no public API changes", and once it's out there we own it. Worth being a deliberate choice rather than a side effect of the star.

Note that plain named re-exports aren't the escape hatch here: `no-inline-type-exports` counts `export type { X, Y }` specifiers too (`rules/no-inline-type-exports.js`, case 3), so `Accordion` would be straight back over the threshold. Splitting the shared `types.ts` per component is the option that actually holds.

packages/dnb-eufemia/src/components/number-format/types.ts
```diff
@@ -0,0 +1,125 @@
+/**
+ * Types for NumberFormatBase
```
langz, line 2: This one says `NumberFormatBase`, but it's also what backs the `NumberFormat` entry.

Broader nit while I'm here: 18 of the new files open with a `Types for <Component>` banner, and not one of the `types.ts` files already on `main` (dialog, popover, upload, tooltip, slider, flex…) has one. A file called `types.ts` next to `Button.tsx` is fairly self-explanatory. `internal-types.ts` earns its header because it explains *why* it exists — these don't.

langz:
Reviewed at `dcb5e254d65`. The move itself is clean — I diffed the resolved export surface of every touched module and barrel against `main`, and nothing was lost: no JSDoc dropped, 0 ESLint errors across `src`, Prettier happy, no new `tsc` errors. `NumberFormatValue` and the two `NumberFormatInternal*` types were off/on the wrong side of the fence in the earlier revision, but the follow-up commit sorted both — `internal-types.ts` is a nice touch.

What's left is the "**No public API changes**" line in the description, which isn't quite true in the other direction: 19 previously private types are now exported, and 58 names now resolve from entries they didn't before. Details in the two inline comments. Additive, so nobody's build breaks — but it's still API we're on the hook for, and the description is what ends up in the release log.

One more for the pile: flipping `no-inline-type-exports` to `error` is unrelated to the move. Splitting it out would make this revertable without also giving up the gate.

langz:
Carrying over the one open item from my earlier comment, now that the surface question is settled at `cec9914d901` — public exports and the emitted `.d.ts` both match `main` exactly.

The `no-inline-type-exports` `warn` → `error` flip (`d707bfe8f54`) is still a separate concern from the move. If the refactor ever needs backing out, the gate goes out with it. Own PR?

langz:
## Review — no breaking consequences for consumers

Verified at `23af426`.

### Public API is identical to `main`

I compared the fully-resolved structure of every exported symbol across **128 modules** (all `components/*`, `elements/*`, `fragments/*`, `extensions/*` barrels, the root `index.ts`, and the 25 deep component paths consumers import directly):

| | Result |
| --- | ---: |
| Exports removed | **0** |
| Exports added | **0** |
| Value ↔ type kind changed | **0** |
| Types structurally changed | **0** |

The only diffs are 3 types (`IconIcon`, `ButtonIcon`, `AutocompleteProps`) where TypeScript now *prints* `import("…/icon/types").IconType` instead of the bare `IconType`, because the alias moved to a sibling module. Same underlying types.

### Runtime output is safe

- `export type *` is fully erased by Babel in all three build targets (`cjs`, `es`, `esm`) — no phantom `require('./types')` anywhere.
- All 9 new sibling type modules compile to `export {};`. Only `accordion/types.js` and `icon/sizes.js` carry runtime values, both leaf modules with zero imports, so no new import cycles.
- Value re-exports survive: `Icon.js` emits `export * from "./sizes.js"` and the CJS star-export loop correctly skips `default`, so `exports.default = Icon` stays intact.
- Scanned all 6,489 relative specifiers in the emitted build — every JS module resolves.
- Loaded the built output in Node: `Accordion.Group`, `Radio.Group`, `ToggleButton.Group`, all 7 `NumberFormat` variants, plus `DefaultIconSize`, `DefaultIconSizes`, `ValidIconType`, `ListDefaultIconSizes`, `accordionDefaultProps` and `COPY_TOOLTIP_TIMEOUT` all resolve.
- No tree-shaking or bundle-size regression: esbuild output for 8 entry points is within 6 bytes of `main` (slightly smaller).

### Declarations resolve for consumers

A simulated consumer importing 44 public types and 9 value exports from the emitted `.d.ts` tree — including the new sub-component paths `accordion/AccordionHeader`, `radio/RadioGroup`, `toggle-button/ToggleButtonGroup` and `number-format/NumberFormatBase` — typechecks under `moduleResolution` **bundler**, **node10** and **node16**, in `strict` mode.

No new TypeScript version floor: `export type *` needs TS ≥ 5.0, but `main` already ships it in `dialog/Dialog.d.ts`, `popover/Popover.d.ts`, `upload/Upload.d.ts` and `forms/Iterate/Array/Array.d.ts`.

### Tooling

Barrel codegen reads directory names under `src/components/`, not files inside them, so the new `types.ts` files cannot produce bogus entries — the generated barrels are unchanged. `verifyGeneratedEntriesCommitted`, `copyTypeScriptFiles`, `validatePackageContents` and `.npmignore` are all unaffected.

### Checks

`tsc --noEmit` clean in `dnb-eufemia` and in the portal. `--noUnusedLocals` finds no dead imports left behind. Prettier clean. ESLint 0 errors, and warnings for the package go **down** 422 → 398. 1,927 tests pass across the 20 affected components. Merges cleanly with current `main`.

### Conclusion

Safe to merge — approving. ✅

Two checks are still running (`Run tests`, `Run visual-regression tests`); worth a glance before hitting merge.

langz:
## Reviewed at `23af42602c3` — safe for consumers, good to merge 👍

I went looking for a way this could bite someone downstream and couldn't find one.

**Nothing leaves the public surface.** Compared the resolved exports of *every* module under `src` — 2,654 on `main` vs 2,681 here — and then did the same against the emitted `.d.ts`:

- shipped modules removed: **none**
- exports removed from any module: **none**
- only delta: `accordion/types.d.ts` gains `AccordionProps`, `AccordionVariant`, `AccordionHeading`, `AccordionIcon`, `AccordionIconPosition`, `AccordionChangeEvent` — purely additive, and the natural home for them now

No `exports` map on the package, so every deep path a consumer might already import (`components/button/Button`, `components/accordion/types`, …) still resolves to the same thing.

**The runtime is untouched.** Compiled all 54 changed files with the repo's Babel config on both branches and diffed the output: every emitted `.js` is byte-identical to `main` except `Icon.js`, and that one is just the constants moving to `sizes.js` behind an `export *` — same four values, same shape. The 27 new type modules compile to empty stubs with zero `require()`, so nothing new gets pulled in at runtime and they tree-shake away. bundlewatch agrees.

`export type *` was already the house pattern here (Dialog, Popover, Upload, `Iterate.Array` all shipped it before this PR), so no new TS floor for consumers.

**Checks:** 16/16 green on this SHA, including `Run build tests`, `Run type checks` and — the one I cared most about — `Build and smoke-test consumers`. Locally: `tsc --noEmit` 0, ESLint 0 errors, `no-inline-type-exports` 0 (down from 24), `require-component-prefix` back to 100 = same as `main`, Prettier clean, and the touched suites pass. `Run visual-regression tests` is still churning; given the runtime is byte-identical I'd be surprised by anything there, but worth a glance before merging.

Only thing I'd still consider is lifting the `warn` → `error` flip (`d707bfe8f54`) into its own PR, per the note above — not a blocker, just tidier to revert. Approving as is.

langz:
Approval carries over to `2f10841cd6e` — the branch got rebased onto `main` (picking up #9373) right as I posted.

Re-checked rather than assumed: the PR's own patch is byte-identical before and after the rebase (`diff` of the two diffs is empty), so it was a clean replay with no semantic tangle against the DrawerList change. Re-ran the export-surface comparison against the *new* `main` — still no removed modules, no removed exports, same single additive delta on `accordion/types.ts`. `tsc` 0, ESLint 0, DrawerList + Autocomplete suites green.

12 checks already passed on the new SHA; the rest are still running.

PR #9374, opened 2026-09-16
fix: resolve genuine exhaustive-deps stale-value bugs

langz:
## Motivation

`react-hooks/exhaustive-deps` has 38 warnings in production code. Most are **benign** — stable `*Ref` objects and state setters, intentional mount-only effects, and cases that intentionally depend on specific members (e.g. `datePickerProps.onOpen`) rather than a fresh-every-render object. Blindly "fixing" those would regress memoization or re-run behaviour.

This is a **targeted pass**: fix the genuine stale-value bugs and the clearly-dead dependencies, and leave the benign ones alone. It resolves **7 warnings (38 → 31)** with no new ones.

## Genuine stale-value bugs (added the missing dependency)

- **`ListFormat`** — the `result` `useMemo` (deps `[format, list, locale]`) also read `className`, `listType` and the rest props, so changing *only* `className`/`props` returned stale output. Since the rest props are a new object every render, adding them would defeat the memo — so the text formatting is memoized and the list element is rendered inline, where the props are always current. **Includes a regression test** (verified it fails on the previous code and passes now).
- **`Upload`** — the file-processing `useCallback` used `allowDuplicates` for dedup logic but omitted it, so toggling the prop had no effect.
- **`PopoverContainer`** — the alignment `useLayoutEffect` reads `autoAlignViewportThreshold` but omitted it, so changes didn't re-align.
- **`PhoneNumber`** — `callOnChange` used `toEvent` but omitted it.

## Dead-dependency removals (no behaviour change)

- **`Modal`** — `handleSideEffects` was listed on `toggleOpenClose` but unused there.
- **`useExternalValue`** — `getValueSnapshot` was listed but unused.

## Correction after review

I originally also removed `showAllErrorsRef.current` from the `Provider` shared-state effect, on the reasoning that a mutable ref is never a valid dependency. **That was wrong and CI caught it** — `Form.Outlet` validation broke.

A `ref.current` in a dependency array is still *compared on every render*, so it does gate whether the effect re-runs; it simply doesn't itself trigger a render. The effect needs it to re-sync when the error visibility flips. It has been restored with a comment explaining why it is intentional, which is why this PR resolves 7 warnings rather than 8.

## Left as-is (benign, for transparency)

Stable refs/setters (`useFieldError`, `useFieldValidation`, `Modal`, `Provider`), intentional mount-only effects (`Accordion` "Constructor", `ProgressIndicatorCircular` startup), partial-object deps that would regress if widened (`MediaQuery` `props`, `Date` `datePickerProps.*`), and the ref-in-cleanup notices in the Wizard router hooks.

## Validation

- `exhaustive-deps` (prod): 38 → **31**, 0 newly introduced
- `tsc --noEmit`: 0 errors; ESLint: 0 errors; Prettier clean
- Tests across every touched area (forms DataContext + Form, ListFormat, Modal, Popover, Upload): **1307 passed**


langz:
## Review — ready to approve

Re-reviewed after rebasing onto latest `main`. **CI is green** (18 passing, 0 failing), including `Run tests`, which is the check that caught the problem below.

### One finding, fixed

CI caught a genuine error in my own change. I had removed `showAllErrorsRef.current` from the `Provider` shared-state effect, arguing a mutable ref is never a valid dependency. **That was wrong** — `Form.Outlet` validation broke.

A `ref.current` in a dependency array *is* still compared on every render, so it does gate whether the effect re-runs; it simply doesn't itself trigger a render. The effect relies on that to re-sync when the error visibility flips. Restored with a comment, and the PR description now reflects the corrected count (**38 → 31, 7 resolved**, not 8).

I verified the failure reproduces locally without the dependency and passes with it.

### Verified

- `exhaustive-deps` (prod): 38 → **31**, 0 newly introduced
- `tsc --noEmit` 0 errors, ESLint 0 errors, Prettier clean
- **1307 tests** across every touched area (forms DataContext + Form, ListFormat, Modal, Popover, Upload)
- The `ListFormat` regression test was confirmed to fail on the old code and pass on the new

The remaining 31 are the benign cases (stable refs/setters, intentional mount-only effects, partial-object deps) deliberately left alone.

PR #9375, opened 2026-09-16
chore(Eufemia): enforce react-hooks/exhaustive-deps

langz:
The lint script runs with `--quiet`, which reports errors only, while `react-hooks/exhaustive-deps` ships as a warning. Its 59 findings could not surface — not locally, and not in CI. This raises it to an error so a stale closure is caught when it is introduced.

Two of the findings were real, each covered by a regression test that fails without the fix:

**ListFormat** memoized the rendered list element on `[format, list, locale]` while the body also read `variant`, `listType`, `className` and the rest props. Passing an inline `value` array hid this, because a new array on every render invalidated the memo anyway — with a stable `value`, changing only `className` kept the old class. Memoizing JSX bought little, so only the text variant, which builds an `Intl.ListFormat`, stays memoized.

**Upload** read `allowDuplicates` inside `onInputUpload` without listing it, so toggling the prop had no effect until some other dependency happened to change.

The remaining findings are deliberate and now carry a reason instead of being silent. No other dependency array is touched: in this hook graph an added dependency changes how often callbacks and effects re-run, which deserves its own change and its own test.

Documentation examples and test components are exempted in config rather than annotated inline — they seed state from mount-only effects, and the examples are read as copy-paste samples where a lint directive would teach the wrong thing.


langz:
Re-reviewed every suppression on this branch against the code it sits on. Seven were wrong, and the review changed the code rather than the wording in three of them.

**Three were not deliberate omissions at all**

`PopoverContainer` (`autoAlignViewportThreshold`), `ListAllProps` (`log`) and `PhoneNumber` (`toEvent`) each *read* the omitted value in the hook body, so each was a stale closure of the same kind this PR already fixes in `ListFormat` and `Upload`. They are primitives or memoized callbacks, so they are now listed instead of suppressed.

I had reverted the first two in an earlier revision while chasing forms-suite failures I assumed I had caused. That was wrong: sampling `main` without my changes reproduced the same failures with a shifting set of tests. The suite is timing-sensitive and the repo already accounts for it with `retry: isCI ? 2 : 0` in `vitest.config.ts`. Under CI retry semantics the full forms suite is green (161 files, 4846 tests).

**One was a rationale I had invented**

The Modal `handleSideEffects` entry is now simply removed. `toggleOpenClose` does not read it, and its identity is never observed — it appears in no dependency array, the two effects that call it have no dependency array, and every other caller goes through `toggleOpenCloseRef`. Removing it is inert, which is better than a comment justifying a dependency that does nothing.

**Three claimed more than the code supports, and were reworded**

- Re-running the `Accordion` effect would not create duplicate instances, since cleanup removes first. The real reason is that `thisInstance` is rebuilt on every render.
- The `useFieldValidation` values are a stable ref and callbacks with empty dependency lists, not values that change per render.
- `onValidationError` is not read by the callback that lists it; it is only used by the validator below.

**The remaining 24 suppressions were each checked against the source**

Every "stable identity" claim was traced to its `useRef` or `useReducer` declaration; `datePickerProps` and `thisInstance` were confirmed to be rebuilt per render; `getAjvInstance` was confirmed to cache into a ref; `showAllErrorsRef.current` was confirmed to be assigned during render; and the `setActiveState` case was confirmed to be a temporal-dead-zone constraint, so listing it would throw rather than fix anything.

**Worth a reviewer's judgement**

`ListFormat` now builds its list element during render instead of inside a `useMemo`. That is the actual fix — the memo was dropping prop updates — and React elements are cheap, but it is a behaviour change in a shipped component rather than a pure lint change.

CI is still running on this revision; it should be green before approving.

PR #9379, opened 2026-09-16
chore(Docs): unbreak lint on main for SidebarMenu docs types

langz:
## Motivation

`main` currently fails `yarn lint` with **10 `docs-types/warn-supported-types` errors**, all in `SidebarMenuDocs.ts`. The rule was promoted from `warn` to `error`, and SidebarMenu (#8986) landed values the rule rejects. This unblocks `main`.

## 1. Restore two corrupted values (data left behind by an old autofix)

An earlier version of the `avoidPipeUnion` autofix split on **every** `|`, including ones inside generic arguments. It turned:

```ts
type: 'Omit<BadgeProps, "content" | "children">'
```

into two malformed fragments that are still in the file today:

```ts
type: ['Omit<BadgeProps, "content"', '"children">']
```

That renders as two nonsense types in the generated property table. The splitting itself is **already fixed upstream** by `splitTopLevelUnion` (which tracks bracket depth), so this PR only restores the corrupted data — no rule change needed for it.

## 2. Recognise two valid TypeScript forms the rule could not express

The remaining eight errors are correct TypeScript:

- **Template literal types** — `` `${number}em` `` (the `breakpoint` prop)
- **Indexed access types** — `BadgeProps["content"]`

Each gets its own narrow pattern:

```js
const TEMPLATE_LITERAL_TYPE_PATTERN = /^`[^`]*`$/
const INDEXED_ACCESS_TYPE_PATTERN = /^[A-Z][A-Za-z0-9_.]*\[".+"\]$/
```

I deliberately did **not** just allow `"` in the general capitalised pattern — that would also un-flag genuinely malformed values like `'Omit<BadgeProps, "content"'`, which should keep failing. Verified that the corrupted form is still reported after this change, so the rule keeps its teeth.

## Validation

- `yarn eslint --quiet ./src`: 10 errors → **0** (main is green again)
- docs-types plugin tests: **98 passed**, with new fixtures covering template literal + indexed access types
- `test:types`: 0 errors; Prettier clean


langz:
Superseded by #9378, which landed the same fix on `main` — and did it slightly better:

- `INDEXED_ACCESS_PATTERN` there also allows a generic before the index (`Foo<Bar>["baz"]`), which mine did not.
- `TEMPLATE_LITERAL_TYPE_PATTERN` there requires an actual `${...}` interpolation, so it is stricter than my `/^\`[^\`]*\`$/`.

Both changes also restore the two corrupted `Omit<BadgeProps, "content" | "children">` values identically. Nothing here to add, so closing as redundant.

PR #9382, opened 2026-09-16
chore(analytics): scaffold Nucleus component-usage storage and dashboard section

joakbjerk:
## Motivation

We want to back Eufemia's value with data on which components consuming apps actually use. This lands the "our side" of that pipeline so the storage and dashboard read-path are ready ahead of the producer.

It adds the anonymous component-usage storage (a `component_usage` Glue table + `component-usage/` prefix), a shared record module, a best-effort snapshot section, and an empty-state dashboard panel — mirroring the existing MCP-usage pattern.

Intentionally out of scope: the Nucleus bundler plugin that emits the data, the cross-team OIDC grant that lets its CI write to the prefix, and the final record schema. The section renders an empty state until a producer lands, and the schema is cheap to evolve while there is no data.


tools/analytics/src/lambda/snapshot.ts
```diff
@@ -173,10 +295,23 @@ export async function handler(): Promise<{
     mcpUsage = EMPTY_MCP_USAGE
   }
 
+  // The component-usage section is additive too; a failure here must not discard
+  // the portal views or MCP section already built. Fall back to empty.
+  let componentUsage: ComponentUsageSection
+  try {
+    componentUsage = await buildComponentUsage(bucket)
```
langz: The infra test next door is literally called *"allows the snapshot generator to complete three Athena queries"* and guards `timeout >= 90`. That's 3 × the 25s poll window in `retrieve.ts` (`MAX_POLLS * POLL_INTERVAL_MS`), plus a little air.

This line makes it five, and they all queue up politely behind each other. I mocked Athena and drove the real handler to count:

```
QUERY COUNT:    5
MAX CONCURRENT: 1
  1. SELECT path, env, "timestamp" FROM "db"."portal_views" …
  2. SELECT dt, tool, component, path, count(*) … mcp_usage
  3. SELECT dt, tool, component, path, count   … mcp_usage_daily
  4. SELECT dt, app, component, version, count(*) … component_usage
  5. SELECT dt, app, component, version, count   … component_usage_daily
```

And one slow query burns exactly 25 000 ms of the budget before it gives up (fake timers, Athena parked on `RUNNING`):

```
FIRST QUERY TIMEOUT AFTER (virtual ms): 25000
RESULT: Error: Athena query timed out
```

5 × 25s = 125s against the 90s `timeout` in `infra/main.tf:752`. When that lands the Lambda is killed *before* `writeSnapshot`, so the whole dashboard goes stale — portal views and MCP included — and the careful per-section fallback right here never gets its turn.

`buildMcpUsage` and `buildComponentUsage` don't touch each other, so `Promise.all` buys the headroom back without touching the timeout, and halves the wall-clock while it's at it. Otherwise the timeout wants ~150s, and the test name needs to learn to count past three.

tools/analytics/src/records/component-usage.ts
```diff
@@ -0,0 +1,181 @@
...
+}
+
+/**
+ * Canonical component name for stable rollups. Mirrors the docs resolver's
+ * lowercase convention (see the MCP usage track, #9344) so different import
+ * spellings — `Button` vs the deep path `@dnb/eufemia/components/Button` — never
+ * fragment into separate rows when Athena groups by component.
+ */
+export function normalizeComponentName(name: string): string {
+  return name.trim().toLowerCase()
```
langz: The doc just above promises that `Button` and the deep path `@dnb/eufemia/components/Button` "never fragment into separate rows". `trim().toLowerCase()` is not quite that ambitious:

```
"Button"                         -> "button"
" Button "                       -> "button"
"@dnb/eufemia/components/Button" -> "@dnb/eufemia/components/button"
```

Two rows, still. The test a screen down is named "collapses import spellings to one canonical name" and then only tries `Button` and `' Button '`, so nothing catches it.

There's no producer yet, which means this comment *is* the spec the Nucleus plugin gets written against — and if it takes the promise at face value and emits deep paths, the rollup fragments quietly while the raw rows you'd need to reprove it expire after 395 days. Either grab the last path segment here, or let the comment admit it only does case and whitespace.

tools/analytics/src/lambda/retrieve.ts
```diff
@@ -205,3 +208,60 @@ export async function retrieveMcpUsageDaily(): Promise<McpUsageDaily[]> {
...
+
+/** Read the full durable daily component-usage rollup for the dashboard section. */
+export async function retrieveComponentUsageDaily(): Promise<
+  ComponentUsageDaily[]
+> {
+  const database = requireEnv('GLUE_DATABASE')
+  const table = requireEnv('GLUE_TABLE_COMPONENT_USAGE_DAILY')
+  const workgroup = requireEnv('ATHENA_WORKGROUP')
+
+  const query = `SELECT dt, app, component, version, count FROM "${database}"."${table}"`
```
langz: No `LIMIT`, and `readResults` pages until `NextToken` runs dry — so this hauls every row `component_usage_daily` has ever held into a 256 MB Lambda and then sums it in JS.

`retrieveMcpUsageDaily` does the same, granted, but the two aren't in the same weight class: `tool × component × path` is bounded by what the MCP server offers, while `app × component × version × day` mints fresh rows every time an app bumps Eufemia. 50 apps × 100 components × a couple of versions is ~15k rows/day, ~5M/year — at 1000 rows per `GetQueryResults` page that's thousands of sequential round-trips, inside the same 90s the other comment is already fretting about.

The section only ever wants three ranked breakdowns and a total. `GROUP BY component` / `app` / `version` in Athena hands back ~150 rows instead of five million, and `sumComponentUsageBy` gets to retire early. Cheap to do now, awkward once there's data.

langz (approved):
Nice mirror of the MCP track — the trailing-slash lifecycle trap is called out *and* pinned by a test, the `dt` interpolation is properly fenced, and making the tail recompute best-effort so the durable history survives a blip is a genuinely nicer failure mode than the MCP section has today.

Three things worth a look, all in comments below. The Athena query budget is the one I'd actually fix before this deploys; the other two are a doc that over-promises and a read that gets expensive once there's something to read.

Approving — none of it blocks a scaffold with no producer behind it yet.

PR #9384, opened 2026-09-16
chore(Portal): fix component lists

snorrekim:
Some lists of components did not properly filter the correct pages. Our "fake" graphql parser is as barebones as can be, and is vulnerable to regressions.

There was also a separate graphql parser being used when generating llm .md files. This one was more correct, but caused a mismatch between llm lists and actual page lists.

- Implement a different way to get page data
  - add `regularMdxEdges` export that contains the most commonly requested list of pages
  - add `globPath` helper to allow glob filtering of `regularMdxEdges`
- `<List*>` components now generate edges using `regularMdxEdges` and `globPath`
- generate llm `.md` lists using the same code as the the `<List*>` components
  - move the actual filtering of `regularMdxEdges` in to a separate `.ts` file so it can be run by both
  - add a test that ensures `<List*>` component adhere to the requirements that the llm generators rely on


packages/dnb-design-system-portal/vite/client/plugins/portal-pages.ts
```diff
@@ -356,8 +411,14 @@ ${routeDefs.join('\n')}
 ];
 
 export const allMdxNodes = [
-${nodeEntries.join('\n')}
+${allMdxNodes.map((node) => `  ${JSON.stringify(node)},`).join('\n')}
 ];
+
+export const regularMdxNodes = [
```
langz, line 417: `regularMdxNodes` goes out as a second full JSON copy of nodes that are already sitting in `allMdxNodes` six lines up — `tableOfContents` and all. I built the portal on both branches to see what that costs on the entry chunk (re-measured after the rebase, against `main` @ `0722d09`):

```
                                        raw        gzip
main                               1 506 675     243 926
this PR                            1 727 996     285 812    +40.9 kB
  ↳ regularMdxNodes by reference   1 531 405     254 213    +10.1 kB
  ↳ …and without picomatch         1 508 249     244 916     +1.0 kB
```

So the duplicated payload is 31.6 kB gzip and picomatch is the other 9.3 kB, and it all lands in `index-*.js` — everyone pays it on first load, to render lists that only ever read `slug`, `title` and `description`.

The nodes in `regularMdxNodes` are the very same object references `toRegularMdxNodes()` filtered out of `allMdxNodes`, so we can emit indices and let the browser do the lookup:

```js
export const regularMdxNodes = [${regularMdxNodes
  .map((node) => allMdxNodes.indexOf(node))
  .join(',')}].map((i) => allMdxNodes[i]);
```

I ran that through a build (the 254 213 row above) and checked the result round-trips identical to today's output — 329 nodes, no `-1`, same order. Same array, we just stop buying it twice.

packages/dnb-design-system-portal/src/shared/parts/ListSummaryFromEdges.tsx
```diff
@@ -13,9 +13,19 @@ import { basicComponents } from '../../shared/tags'
...
+   *
+   * Lists keep their page selection in `listEdges.ts` and pass the result
+   * here. Those are plain functions taking the pages as an argument, so the
+   * markdown generator for the LLM docs can import and call them instead of
+   * re-deriving the selection from the component source — one definition of
+   * what each list contains.
+   *
+   * Also accepts the legacy `{ node }` shape from `useStaticQuery`.
+   */
+  edges: Array<MdxNode | { node: MdxNode }>
```
langz, line 28: The legacy `{ node }` shape has no callers left. All 18 `List*` components in this PR hand over a plain array, and no `.mdx` renders `<ListSummaryFromEdges>` directly — so the union and the `'node' in edge` narrowing below exist to support a caller this same PR removed.

Making it `MdxNode[]` would keep the "one definition of what a list contains" story tidy, and save the next person wondering which shape is the real one.

Non-blocking.

packages/dnb-design-system-portal/src/declaration.d.ts
```diff
@@ -29,8 +29,27 @@ declare module 'virtual:portal-pages' {
 
   /**
    * Every MDX page, unfiltered and unsorted.
+   *
+   * Prefer importing from `src/core/portalPages`, which exposes this data
```
langz, line 33: `src/core/portalPages` does not exist — not on this branch, not on the base, not on `main`. The only `portalPages` in the tree is `portalPagesPlugin` over in `vite/client/plugins/`.

Slight plot twist too: all 18 `List*` components in this PR import `regularMdxNodes` straight from `virtual:portal-pages` — precisely what this line tells the next person not to do.

Either the wrapper got left behind in another branch, or the advice can go.

packages/dnb-design-system-portal/src/shared/parts/ListSummaryFromEdges.tsx
snorrekim: Fixed. Removed unused type.

packages/dnb-design-system-portal/src/declaration.d.ts
snorrekim: Fixed. Removed the comment, was left over form a previous commit.

packages/dnb-design-system-portal/vite/client/plugins/portal-pages.ts
snorrekim: Fixed. Now uses a list of indexes instead of the entire object.

langz (approved):

PR #9386, opened 2026-09-17
chore(analytics): rebuild the dashboard UI with Eufemia components

joakbjerk:
## Motivation

The analytics dashboard shipped as a deliberately dependency-free HTML/CSS/JS placeholder while the pipeline and data API were built out. Now that those work end to end, the UI should use Eufemia so it matches the rest of the portal, stays maintainable, and dogfoods our own design system.

This rebuilds `tools/analytics/dashboard/` as a React + Vite app on `@dnb/eufemia`: header, environment `Dropdown`, KPI cards, a per-day bar list, and ranked Eufemia tables for top pages and MCP usage. Sign-in, the `/data` fetch, environment filtering, the empty/error states, and the "last updated" meta are preserved; the sign-in and data logic moved to typed modules with their own tests. It still deploys as a static site to S3 + CloudFront under the existing strict CSP.

## Follow-up PR's

- Adding dark mode 
- Update layout
- More?


langz:
**The production build is a blank page.**

I built it, served it, and opened it:

```
TypeError: Cannot read properties of null (reading 'useContext')
```

`#root` stays empty. Nothing renders at all.

The cause is in `dashboard/package.json`: `react` and `react-dom` are pinned to `19.2.5`, while the rest of the monorepo is on `19.2.8`. Yarn dutifully hands the dashboard its own nested copy at `tools/analytics/node_modules/react`, Eufemia — imported as `@dnb/eufemia/src` — keeps resolving the root `19.2.8`, and Vite cheerfully bundles both. Grep the output chunk and you'll find `e.version="19.2.5"` *and* `e.version="19.2.8"` living together. Only one of them ever gets a dispatcher attached, and it isn't the one Eufemia's hooks are calling.

This isn't a local quirk — the committed `yarn.lock` resolves both `react@npm:19.2.5` and `react@npm:19.2.8`, so CI reproduces it every single time. It just never opens the page, which is why 20 green checks are telling you absolutely nothing here.

Bumping both to `19.2.8` (matching `eufemia-starter`) collapses it to one React, and the dashboard renders properly — header, meta line, empty states and all. Verified end to end.

Given the whole point of the PR is the UI, a tiny smoke check that mounts the built app and asserts `#root` isn't empty would be worth its weight. This failure mode is completely invisible to `build`, `lint`, `test:types` and the unit tests.

langz:
**`space="large"` is quietly doing the opposite of what it looks like.**

`App.tsx:171`. `Flex.Stack` already defaults to `gap="medium"`, so `space` isn't widening anything between the cards — it's a spacing prop, and it expands to all four sides:

```
dnb-space__top--large dnb-space__right--large dnb-space__bottom--large dnb-space__left--large
```

`.dnb-space__left--large { margin-left: var(--spacing-large) }` has the same specificity as `.dashboard { margin: 0 auto }`, and lands ~209 kB further down the stylesheet — `App.scss` gets emitted first, because `App.tsx` imports it before `main.tsx` pulls in Eufemia's style. So the spacing class wins, `auto` never gets a turn, and the dashboard sits shoved against the left edge.

Measured on the built page at 1280px: **32px on the left, 224px on the right.** Swapping to `gap="large"` gives a tidy 128/128 and actually spaces the sections apart, which I assume was the intent all along.

joakbjerk:
Thanks @langz — all three addressed:

- **Blank page (dual React):** bumped `react`/`react-dom` to `19.2.8` to match the rest of the monorepo, so the bundle no longer co-bundles two React copies. `yarn.lock` now resolves a single `react@npm:19.2.8` and the app mounts.
- **Layout shoved left:** switched the outer `Flex.Stack` from `space="large"` to `gap="large"`, so the section spacing no longer emits a left margin that beat `.dashboard { margin: 0 auto }` — it's centered now.
- **Smoke test:** added `App.test.tsx` that mounts the app and asserts it renders (heading + sections, not a blank `#root`) — the failure class that build/lint/type checks can't see.

Bonus from the `main` merge: #9382 had shipped a component-usage section the React rewrite lacked, so I ported it (Top components / by app / by Eufemia version, plus the freshness count and demo data) to avoid regressing it.

One honest caveat on the smoke test's scope: it mounts the source app through vitest (single React), so it guards mount-time crashes but doesn't reproduce the production two-React *bundling* itself — the real guard against that remains the version pin + single-React lockfile.

langz (approved):

PR #9387, opened 2026-09-17
chore(analytics): lock dashboard CloudFront origin to the Akamai edge

joakbjerk:
The dashboard's raw `*.cloudfront.net` origin was only CORS-blocked. This adds a viewer-request CloudFront function that returns 403 for any request missing the shared `X-Edge-Auth` secret Akamai injects, so the static shell is reachable only through the edge domain. The Entra JWT authorizer on `/data` remains the real data control; this hardens the static-shell delivery path.

Requires the analytics deploy role's new CloudFront Function permission to be deployed first.


langz:
**The lock looks right. I'd just like to know someone's holding the key.**

`X-Edge-Auth` is documented in this repo for the *ingest* path only — README L21, and the `origin_domain` output ("Custom origin hostname for Akamai to point at") is the API Gateway domain. The dashboard is a different origin entirely: no custom domain, Akamai pointing straight at `*.cloudfront.net` (`main.tf:951`). So this is a second Akamai behaviour, and nothing in here tells me it injects the header.

If it doesn't — or injects a secret other than `var.edge_auth_secret`, which this reuses — the first `terraform apply` 403s every dashboard user, and the way back is another apply, not a toggle. Worth confirming before merge rather than discovering it from the other side.

langz:
**`infra/main.tf:801` didn't get the memo.**

> Access control is entirely the Entra sign-in plus the /data API's JWT authorizer — there is deliberately no Lambda@Edge and no edge auth here.

That sits ninety lines above `aws_cloudfront_function.dashboard_edge_auth`. The next person to read that block will believe the comment over the resource, and the comment now argues against the PR it's sitting in.

langz:
**`tools/analytics/README.md` still describes the world as it was this morning.**

- **L110** — "so no Lambda@Edge, **edge auth** or extra IAM role is needed". The "extra IAM role" half still holds (CloudFront Functions don't take an execution role), the middle word doesn't.
- **L112** — "hard-blocking direct access is a **separate follow-up**". This *is* the follow-up. It has arrived.

While you're in there: `outputs.tf:39` still advertises `dashboard_url` as the "Public CloudFront URL of the dashboard" — after this lands, that's a URL that returns 403 to anyone who takes the description at face value.

langz (approved):
Approving — the mechanism is sound and the implementation is tidy.

Verified at `882ec22b004`: the function is correct CloudFront-JS 2.0 (duplicate `X-Edge-Auth` headers arrive comma-joined, so they fail closed rather than sneaking through), `jsonencode` keeps the secret a safe JS literal, and Terraform propagates the `sensitive` flag through `templatefile` so `code` won't show up in plan output. The deploy job `cp -R`s `infra/`, so `infra/functions/` comes along, and nothing in the pipeline makes an HTTP request to the dashboard — so the apply itself won't trip over its own lock. Both test files pass (6/6), Prettier clean.

Three comments above. Only the Akamai one is worth blocking on, and that's a question for you rather than a defect in the diff — the other two are docs that now say the opposite of the code.

joakbjerk:
Confirmed — the edge side is already in place, so the first apply won't 403 real users.

- The dashboard's Akamai property was updated to send the shared edge secret to this origin, using the same mechanism and the same shared value as the existing API origin lock. That change has been deployed and activated successfully.
- This CloudFront function validates that same shared value, so traffic arriving through the edge continues to serve normally while requests that don't come through the edge are refused.
- The rollout is sequenced so the edge change is live before this enforcement merges — verified the activation succeeded first.

Thanks for making sure someone's holding the key.

PR #9391, opened 2026-09-17
chore(Portal): escape page metadata interpolated into prerendered HTML

langz:
The prerender pipeline built the page title, meta description, Open Graph tags, and redirect pages by interpolating MDX frontmatter straight into HTML template strings. A stray `"` in a frontmatter `description` ends the `content` attribute; a `<` in a `title` ends the `<title>` element. Nothing caught that except a reviewer noticing the character in a YAML block.

Not a vulnerability: every value comes from in-repo MDX, so reaching it already requires write access to the repository, and fork previews are never published. It is still a poor place for correctness to depend on human proofreading, so this escapes the values instead.

The new `escapeHtml` lives in its own `.mts` module rather than in `prerender-utils.ts`, following the convention the file header already documents for `md-paths.mts`: `prerender.mjs` mirrors `injectHtml` inline and cannot import the `.ts` version, so a helper with no Vite-resolved imports belongs where both the build and the tests can use the same copy.

Two related things came out of the same pass:

- The interpolated values were passed as the replacement argument of `String.replace`, where `$&`, ``$` ``, `$'` and `$1` are substitution patterns. A title containing a dollar sign was rewritten rather than inserted. The replacements that carry authored content now use function replacers, which insert verbatim. This also covers the rendered page HTML going into `#root`, where a `$&` in body copy had the same effect. The manifest-derived replacements (hashed asset filenames) are left as they are.
- Escaping changes existing output for titles that legitimately contain `&`, which now render as `&amp;` in the markup and unchanged on screen.


langz:
Reviewed the diff and ran it against the branch: 82 tests pass, `tsc --noEmit` is clean, Prettier is clean.

The framing in the description is right, and the `$&` / `` $` `` / `$'` half is the more valuable one. Escaping guards against a character nobody has written yet; the replacement-pattern substitution was already silently rewriting any title containing a dollar sign. Good call leaving the manifest-derived replacements as plain strings rather than converting everything on principle.

I also checked the hydration side, since escaping changes the prerendered bytes. `usePortalHead` assigns through `textContent`, so `Cards &amp; Lists` parses back to the same string the client writes — no mismatch, no visible change. Worth noting that the `head-description` branch there sets `textContent` on a `<meta>`, which never reaches the `content` attribute, so the prerendered value is the only one that element ever has. Pre-existing and out of scope, but it means escaping `desc` matters for the lifetime of the page, not just the first paint.

Three things:

### 1. `emotionCss` is the one authored-content replacement still passing a string

`vite/prod/prerender-utils.ts:401` and `vite/prod/prerender.mjs:570`:

```js
html = html.replace('</head>', `${emotionCss}\n</head>`)
```

`emotionCss` is page-derived — `entry-server.tsx` pulls the `<style data-emotion>` tags straight out of the rendered tree — so it sits in the same bucket as the `#root` case that did get a function replacer, one line above it. A single `content` declaration is enough:

```
content:"$&"   →   content:"</head>"
```

which closes the head early and spills the rest of the stylesheet into the body. Same shape as the bug the PR fixes, so it reads as an oversight rather than a decision.

`reactPreloadLinks` at `prerender-utils.ts:423` has the same shape but only ever carries manifest hrefs, so leaving that one alone is consistent with the stated rule.

### 2. The `prerender.mjs` copy — the one that actually ships — is unguarded

`prerender-utils.test.ts:1049` already has a guard for exactly this hazard, pinning `headThemeScript` and `bodyThemeScript` across both files, with the comment *"Pin them together so a fix to one can't silently miss the other."*

Extracting `escapeHtml` into a shared module means the escaping *function* can't drift, but the call sites can, and the tested copy is not the copy that builds the site. Worth extending that guard over the metadata block, or stating in the `html-escape.mts` header why it isn't extended.

### 3. The ampersand test doesn't pin what its name claims

`vite/__tests__/prod/html-escape.test.ts:9`:

```ts
it('escapes the ampersand first so entities are not double-escaped', () => {
  expect(escapeHtml('&lt;')).toBe('&amp;lt;')
})
```

Two mismatches. The name says "not double-escaped"; the assertion shows a pre-existing entity being re-escaped — correct for raw text input, just not what the name describes. And "first" implies an ordering hazard that a single regex with a callback doesn't have, since `replace` never rescans its own output.

More concretely: `'&lt;'` contains no `<`, so a naive chained-`.replace` implementation passes this test too. The regression it names is not the regression it catches. `expect(escapeHtml('<')).toBe('&lt;')` is the assertion that fails on the chained version. The existing case is still worth keeping, under a name like *"escapes an ampersand in text that already looks like an entity"*.

### Smaller, take or leave

- `escapeHtml(meta.mdPath)` inline at `prerender-utils.ts:386` forces Prettier to break the call across three lines mid-template. The function already hoists `desc` and `fullUrl` above the tag list; `mdPath` would read better next to them.
- `HTML_ENTITIES[char]` typechecks only because the portal tsconfig leaves `noImplicitAny` off. A `Record<string, string>` annotation keeps it building if that is ever tightened.

langz:
## Pre-merge verification

All 20 checks green. No conflicts with `main`, and nothing on `main` has touched `vite/prod/` since the branch point.

Checked against the branch (not a stale checkout):

- 86 tests pass in `vite/__tests__/prod/`; `tsc --noEmit` and Prettier clean.
- `html-escape.mts` loads under plain Node's type stripping, so the `.mts` import from `prerender.mjs` resolves at build time — same path `md-paths.mts` already uses.
- The parity guard fails when `prerender.mjs` drifts. Verified by reverting the Emotion CSS replacer in `prerender.mjs` alone and watching `escapes and injects page metadata the same way` fail, then restoring.
- Escaping introduces no hydration mismatch: `usePortalHead` writes through `textContent`, so `Cards &amp; Lists` parses back to the string the client sets.

## One latent issue, not a blocker

`escapeHtml` calls `value.replace(...)`, so it throws on anything that isn't a string. `getPageMeta` produces its values with a cast, not a conversion:

```ts
let title = (node.frontmatter.title as string) || ''
```

`as string` is erased at runtime. Frontmatter is YAML, so `title: 10.5` or `title: 2024` arrives as a number and the cast waves it through:

```
BEFORE (template literal):  "10.5 | Eufemia"
AFTER  (escapeHtml):        TypeError: value.replace is not a function
```

Interpolation used to stringify anything. Escaping turns the same input into a crash that takes down the whole prerender, reported from `html-escape.mts` with no mention of the page or the field responsible.

Not reachable today — no non-string `title` or `description` in any of the 332 frontmatter entries under `src/docs/`. It only bites the first time someone writes an unquoted version or year as a page title, which is a plausible thing to write.

If it's worth closing, the honest place is the cast in `getPageMeta` rather than a coercion inside `escapeHtml`, since the type lie is what lets a non-string reach a `value: string` parameter in the first place. Equally reasonable to leave it and let the build tell us.

packages/dnb-design-system-portal/vite/__tests__/prod/prerender-utils.test.ts
```diff
@@ -924,4 +1089,52 @@ describe('prerender-utils', () => {
       }
     })
   })
+
+  // prerender.mjs carries its own copy of injectHtml and buildRedirectHtml,
+  // and that copy is the one that builds the site. Pin the regions that must
+  // stay identical so a fix to one can't silently miss the other. Only
+  // annotation-free regions can be compared this way — the surrounding code
+  // differs by its types.
+  describe('parity with the prerender.mjs copy', () => {
```
langz, line 1098: The guard covers two of the three replacements this PR converts, but not the `<div id="root">` one — and that is the one with the widest exposure, since `appHtml` carries every page's rendered body.

Verified on this branch: revert only the `#root` function replacer in `prerender.mjs`, leave `prerender-utils.ts` correct, and all 86 tests still pass. The copy that builds the site would go back to substituting `$&` / `` $` `` / `$'` out of the app HTML, and nothing here would notice, because these tests exercise the `.ts` copy.

A third region closes it. The markers are unique in both files, and I confirmed it passes as-is and fails on the reverted copy:

```ts
it('inserts the app HTML the same way', () => {
  const from = 'let html = template.replace('
  const to = 'scrollRestoreScript}</script>`'

  expect(region('prerender.mjs', from, to)).toBe(
    region('prerender-utils.ts', from, to)
  )
})
```

langz (commented):
## Final round — earlier feedback is in

Re-checked against `fd571de`, fetched fresh. Merges cleanly with `main`, and nothing on `main` has touched `vite/prod/` since the branch point. 86 tests pass, `tsc --noEmit` and Prettier are clean, all 20 checks green.

Everything raised in the earlier rounds is addressed:

- `emotionCss` goes through a function replacer now, with a test pinning it.
- `escapeHtml` lives in `html-escape.mts`, shared by both copies, and loads under plain Node's type stripping.
- The parity guard was extended over the metadata block and `buildRedirectHtml`. It bites: reverting the Emotion replacer in `prerender.mjs` alone fails *escapes and injects page metadata the same way*.
- The ampersand test was split, and `escapeHtml('<')` is the assertion that actually fails a chained implementation.
- `mdPath` is hoisted next to `desc` and `fullUrl`, and `HTML_ENTITIES` is typed.

Worth noting the escaping is not only guarding against input nobody has written. Three pages change output today, all for the better: `Fonts & Typography`, `React & TypeScript`, and the `<hr />` inside the `horizontal-rule` description. All three parse back to the same string, so nothing renders differently.

One gap in the guard below. Not a blocker.

### Carried over, still fine to leave

`escapeHtml` throws on a non-string, and `getPageMeta` produces its values with `as string` rather than a conversion, so a YAML `title: 2024` would reach `value.replace` as a number. Not reachable today — none of the 336 frontmatter files under `src/docs/` has a non-string `title` or `description`. If it is ever worth closing, the cast in `getPageMeta` is the honest place, not a coercion inside `escapeHtml`.

packages/dnb-design-system-portal/vite/__tests__/prod/prerender-utils.test.ts
```diff
@@ -643,6 +643,162 @@ describe('prerender-utils', () => {
...
+          '<style data-emotion="css">.x::after{content:"$&"}</style>'
+        const result = injectHtml(
+          template,
+          '<h1>Hi</h1>',
+          { js: [], css: [] },
+          emotionCss
+        )
+
+        expect(result).toContain(emotionCss)
+        expect(result).not.toContain('content:"</head>"')
```
langz, line 778: This line reads like it covers `</head>` inside Emotion CSS. It doesn't — it pins the `$&` substitution, and `content:"</head>"` is only what that substitution would have produced.

Feed the literal string instead and the later head injections land inside the declaration:

```html
<style data-emotion="css">.x::after{content:"    <link rel="stylesheet" crossorigin href="/a.css">
      <link rel="modulepreload" crossorigin href="/a.js">
  </head>"}</style>
```

The route CSS and modulepreloads never reach the head. The PR fixed the replacement side of `html.replace('</head>', ...)`; the search side still takes the first `</head>` in the document, and after this line that can be one the page brought with it.

Pre-existing, and nothing we author today produces it, so not a blocker. But the name promises coverage this assertion doesn't give.

langz (commented):
One more, same line as before. Not a blocker.

langz (commented):
Good to merge from my side.

Neither finding from the last round blocks: the `#root` parity gap is a hole in the guard, not in the code that ships, and the Emotion CSS `</head>` case is pre-existing and unreachable from anything we author today. Both are worth picking up, neither needs to hold this.

Would have hit Approve, but GitHub does not allow approving your own PR — this comment is the sign-off.

PR #9393, opened 2026-09-18
feat(SidebarMenu): add compact responsive navigation

tujoworker:
Adds a compact icon-rail mode between inline and Drawer layouts, plus a ready-to-use ResponsiveAside that owns scrolling, spacing, resizing, drag-to-dismiss behavior, and hamburger restoration. This lets applications preserve page space without recreating Portal-specific sidebar wiring.

Motivation: https://dnb-it.slack.com/archives/CMXABCHEY/p1789642456704379
Preview: https://feat-sidebar-menu-peek-mode.eufemia-e25.pages.dev/uilib/extensions/sidebar-menu/demos#responsive-navigation

Screenshot 2026-09-18 at 15 38 04


langz:
Reviewed the diff and ran it against the branch (not a stale checkout): the 142 `sidebar-menu` tests pass, `tsc --noEmit` is clean for the extension, ESLint and Prettier are clean, the SCSS compiles, and axe finds no violations on the `ResponsiveAside` + compact setup. Nice feature — the icon rail feels like the right middle step between inline and Drawer.

Two things I think need resolving before this comes out of draft.

### 1. The new e2e test fails deterministically in CI

`Run e2e tests` is red on `src/e2e/sidebarMenuCompact.spec.ts:160 › dismisses the full SidebarMenu by dragging and restores it` — 4/4 attempts, failing at line 200 on `scrollWidth > clientWidth`. The other two tests in the file pass, so it isn't the feature itself.

The portal e2e config runs Firefox, and the drag is dispatched synthetically:

```ts
element.dispatchEvent(
  new PointerEvent('pointerdown', { bubbles: true, button: 0, clientX: 288, pointerId: 1 })
)
```

`handlePointerDown` then calls `setPointerCapture(event.pointerId)`, which Firefox rejects for a pointer id that was never active. Minimal repro on both engines:

```
[firefox]  setPointerCapture THREW: NotFoundError: Element.setPointerCapture: Invalid pointer id
[chromium] setPointerCapture OK
```

The throw aborts the handler before `startResize()` runs, so no drag ever happens — which is why it reproduces rather than flakes, and why it likely passed locally on Chromium. `sidebarResize.spec.ts` drives the same handle with `page.mouse.move/down/move/up` and passes in the same run; reusing that shape should be enough.

### 2. `breakpoint` → `drawerAt` is a breaking rename of a shipped prop

`SidebarMenuResponsiveProvider.breakpoint` went out in `v11.14.0` (16 Sep) and is on the published properties page. This PR removes it with no alias, under a `feat:` title — with the `conventionalcommits` preset that releases as a minor, so anyone on `^11` gets a TS error, and at runtime the prop is silently dropped by the destructuring and the Drawer threshold falls back to `medium`.

Only two days of exposure, so the blast radius is probably small, but I'd rather it be explicit: either keep `breakpoint` as a deprecated alias for one release, or add a `BREAKING CHANGE:` footer so the version reflects it.

Everything else I looked at held up — first-paint CSS stays injection-safe since both thresholds are validated as `em` before interpolation, the compact mode is correctly gated off touch-only input, and the `aria-controls`/`inlineControls` wiring resolves to a real element whenever `ResponsiveAside` is used.

langz:
As this is the 1st pR of the stack, I'll start looking into this now :)

langz:
Re-reviewed at `3de8ee5` — both points from my earlier round are sorted, so I've hidden that comment 🎉 The drag test now drives `page.mouse` and `Run e2e tests` is green, and `breakpoint` is back as a deprecated alias for `drawerAt`. All checks pass on this revision.

One new thing I think should be fixed before merging.

### Hydration error on the SidebarMenu demos page

The branch preview logs `Minified React error #418` (text-content hydration mismatch) on `/uilib/extensions/sidebar-menu/demos/`. Production doesn't, and the Button demos page on the same preview doesn't either — so it's new here.

The cause is `scopeId` falling back to `useId()`. The portal's `ComponentBox` evaluates the example on the client in a different tree, so server and client ids don't line up:

```
SSR     <style> … [data-sidebar-menu-responsive-scope="sidebar-menu-_R_7iud9_"] …
client          … [data-sidebar-menu-responsive-scope="sidebar-menu-_r_1v_"]   …
```

On `main` the example used `breakpoint="medium"`, so `getFirstPaintCss()` returned `undefined` and there was no `<style>` to mismatch. Now that it sets `compactAt`, the first-paint CSS is always emitted with the scope id baked in, and React falls back to client-rendering that subtree.

`scopeId` is documented for exactly this case, so passing it in the example looks like the whole fix:

```tsx
<SidebarMenu.ResponsiveProvider
  scopeId="sidebar-menu-responsive-example"
  drawerAt="medium"
  compactAt="large"
  compactOffset="10em"
>
```

Everything else held up nicely: keyboard focus expands the rail (verified with `whatinput="keyboard"` — focused the toggle, tabbed to `Home`, `data-sidebar-menu-responsive-expanded="true"` and `clip-path: inset(0px)`), the collapsed rail keeps the full navigation in the accessibility tree, `compactAt="large"` + `compactOffset="10em"` resolves to `82em` as documented, and the properties tables for `ResponsiveInline`/`ResponsiveAside`/`ResponsiveShell` are thorough. Lovely feature 👍

langz:
Had another look at `3de8ee5`, this time poking at the rail itself in a real browser — the peek behaviour is genuinely nice to use 🙂 One thing I think we should fix before merging.

### Escape can't dismiss the hover-expanded rail (SC 1.4.13)

At `1200px` (inside the compact range), pointer over the rail, focus left on `<body>`:

```
hover         → expanded: true,  clip-path: inset(0px),  aria-expanded: "true",  activeElement: BODY
press Escape  → expanded: true,  clip-path: inset(0px),  aria-expanded: "true",  activeElement: BODY
```

Move focus into the rail first and Escape works exactly as intended (`clip-path` goes back to `inset(0px 224px 0px 0px)`). The handler is `onKeyDownCapture` on `SidebarMenuResponsiveInline`, so it only ever sees keydown originating inside the sidebar — and hover expansion never moves focus there.

That matters because the expanded panel overlays ~224px of page content (`64px` reserved vs `288px` painted), so 1.4.13 asks for a way to dismiss it *without* moving pointer hover or focus.

A `document`-level `keydown` listener while `compactHovered` is true, reusing the same `setCompactExpanded(false)` + `setCompactDismissed(true)` you already have, looks like the whole fix.

Hoverable and persistent are both already satisfied, so this is the last piece 👍

langz:
### The expanded rail's `box-shadow` never paints

Small but visible one, and an easy fix 🙂

In `dnb-sidebar-menu.scss`, the expanded state sets both on the *same* element:

```scss
&.dnb-sidebar-menu-responsive-inline--expanded
  > .dnb-sidebar-menu-responsive-inline__content {
  clip-path: inset(0);
  box-shadow: var(--shadow-default);
}
```

`inset()` clips to the border box, so the outer shadow is removed entirely. Computed style says it's there — `rgba(51, 51, 51, 0.08) 0px 8px 16px 0px` — it just never reaches the screen.

Proof on the preview: overrode the shadow to `0 0 0 14px red` and screenshotted 1px outside the panel's right edge, expanded:

```
clip-path: inset(0)  → white   (shadow clipped away)
clip-path: none      → red     (shadow paints)
```

The `box-shadow 240ms` in the transition next to it is dead for the same reason.

Since the panel expands *over* page content, that elevation is doing real work — right now the only separation is the aside's 1px `border-inline-end`. Putting the shadow on the `.dnb-sidebar-menu-responsive-inline` wrapper instead should give it back while keeping the `clip-path` animation intact.

langz:
### Missing v13 migration note for the deprecated `breakpoint`

Tiny one, and thanks for bringing `breakpoint` back as an alias 🎉

Our `AGENTS.md` asks for migration guidance in the next major-version info file whenever a public API is deprecated, and `v13-info.mdx` already collects exactly this kind of rename (`Theme`: `name` → `brand`). `SidebarMenu.ResponsiveProvider.breakpoint` isn't in there yet, so the removal in v13 would land undocumented.

A one-liner under a `SidebarMenu` heading is all it needs:

> Replace `breakpoint` with `drawerAt` on `SidebarMenu.ResponsiveProvider`.

Nothing else new from my side on this revision — the three-tier model reads really well 👍

packages/dnb-eufemia/src/extensions/sidebar-menu/SidebarMenuResponsiveAside.tsx
```diff
@@ -0,0 +1,161 @@
...
+      onWidthChange(inlineRef.current?.getBoundingClientRect().width ?? 0)
+    }
+    reportWidth()
+    if (typeof ResizeObserver === 'undefined') {
+      return undefined
+    }
+    const observer = new ResizeObserver(reportWidth)
+    observer.observe(inlineRef.current)
+    return () => observer.disconnect()
+  }, [onWidthChange])
```
langz: `onWidthChange` goes quiet for good after the first `drawerAt` crossing 🙈

The effect only depends on `onWidthChange`, but `ResponsiveInline` returns `null` below `drawerAt`, so `inlineRef.current` is swapped out without the effect ever re-running. Instrumented `ResizeObserver` over a `100em → 30em → 100em` round trip with a `useCallback`'d handler:

```
mount   observe:…-inline                       (isConnected: true)
30em    …-inline isConnected: false            (only ScrollView disconnected)
100em   observe:…-aside__scroll-view           ← no new …-inline observe
        reports: 1 at mount, none after
```

So the observer stays bound to the detached node, the consumer keeps the stale desktop width while the Drawer is in use, and it never recovers on the way back — `useCallback` is the documented-looking usage, so it won't self-heal. A callback ref for the inline element (or keying the effect on the mounted node) should cover it, and reporting `0` when it unmounts would let consumers drop the reserved space on mobile.

The current test only asserts the initial report, so a round-trip case would be a nice guard.

langz (commented):
Second deep pass at `3de8ee5` — pulled the branch, compiled the SCSS with dart-sass, ran the 43 sidebar-menu unit tests, and added a couple of throwaway probes to poke at the new lifecycles.

The rail itself is in great shape: the deprecated `breakpoint` is covered both ways, the e2e round-trips the viewport across both thresholds, and the compiled `.dnb-sidebar-menu--compact` selectors line up with the hand-written first-paint CSS. Two things below, plus the hydration one from my previous comment still open.

packages/dnb-eufemia/src/extensions/sidebar-menu/SidebarMenuResponsive.tsx
```diff
@@ -365,47 +419,376 @@ export function SidebarMenuResponsiveTrigger({
...
+        }
+      }}
+      onKeyDownCapture={(event) => {
+        onKeyDownCapture?.(event)
+        if (isCompact && event.key === 'Escape') {
+          setCompactExpanded(false)
+          setCompactDismissed(true)
+        }
+      }}
+      onPointerEnter={(event) => {
```
langz: Small one: the hover state runs even when compact mode is off.

`onPointerEnter`/`onPointerLeave` set `compactHovered` unconditionally, so a provider *without* `compactAt` still flips on a plain mouse-over:

```
before  expanded=null   --sidebar-menu-opacity: 0.55
after   expanded=true   --sidebar-menu-opacity: 1   + …-inline--expanded
        (has-compact=null, compact="false")
```

Nothing looks different, since every compact rule needs `.dnb-sidebar-menu--compact` — but each mouse enter/leave now re-renders the whole inline nav subtree for existing `ResponsiveInline` users, and the DOM advertises `expanded` when there is nothing to expand. Gating on `hasCompact`, the way `onPointerMove` already gates on `isCompact`, makes it free again. 🙂

packages/dnb-eufemia/src/extensions/sidebar-menu/SidebarMenuResponsive.tsx
```diff
@@ -231,22 +274,34 @@ export function SidebarMenuResponsiveProvider({
     ]
   )
 
-  const firstPaintCss = getFirstPaintCss(scopeId, resolvedBreakpoint)
+  const firstPaintCss = getFirstPaintCss(
```
langz: Nit: this runs in the render body and re-concatenates 6423 bytes (measured via `renderToString` with `compactAt="large"`) on every provider render. `useMemo` on `[scopeId, resolvedDrawerAt, resolvedCompactAt]` is a one-liner.

packages/dnb-eufemia/src/extensions/sidebar-menu/SidebarMenuResponsiveAside.tsx
```diff
@@ -0,0 +1,161 @@
...
+  ref,
+  style,
+  ...props
+}: SidebarMenuResponsiveAsideProps) {
+  const asideRef = useRef<HTMLElement>(null)
+  const inlineRef = useRef<HTMLDivElement>(null)
+  const combinedRef = useCombinedRef(ref, asideRef)
+  const responsive = useOptionalSidebarMenuResponsive()
+  const internalId = useId(id ?? `${responsive?.scopeId}-responsive-aside`)
+  const { collapseInline } = useSidebarMenuResponsive()
```
langz: Nit: the context is read twice. `useSidebarMenuResponsive()` already throws without a provider, so `responsive?.scopeId` can't be `undefined` — `"undefined-responsive-aside"` and `[data-sidebar-menu-responsive-provider="undefined"]` are dead paths. One `useSidebarMenuResponsiveContext()` gives you both `scopeId` and `collapseInline`.

langz (commented):
**The rail has no visual regression coverage.** `SidebarMenu.screenshot.test.ts` runs 13 cases × 2 themes, but compact mode adds none and the new example has no `data-visual-test`. The collapsed 4rem rail, the hover-expanded overlay shadow, the text-only `mask-image` fallback, the shortened divider and the toggle-island scroll shadow are all CSS-only with nothing else guarding them.

_The two nits I originally attached here (`useMemo` for the first-paint CSS, and the dead `responsive?.scopeId` fallback) turned out to duplicate https://github.com/dnbexperience/eufemia/pull/9393#issuecomment-5813979162 — hidden them to keep the thread readable._

langz:
Two more from the same pass — one I'd fix before merge, the rest are one-liners 🙂

### Two `ResponsiveAside` under one provider collide on `id`

Our `useId` returns `customId` whenever it's truthy, and `ResponsiveAside` always passes one:

```ts
const internalId = useId(id ?? `${responsive?.scopeId}-responsive-aside`)
```

The generated branch is therefore unreachable, so both asides render the same `id` — and `ResponsiveTrigger`'s default `aria-controls` (`${scopeId}-responsive-aside`) points at an ambiguous target. Since the trigger depends on that name staying deterministic, registering the aside id on the responsive context (the way `registerInlineReset` already works) is probably nicer than deriving it in two places.

### Quick nitpicks

- `compactAt` is documented as "Maximum available **content** width", but it resolves through `useMediaQuery` → `window.matchMedia`, so it's the viewport. `compactOffset` right below already says "viewport width" — aligning the wording avoids anyone expecting container queries.
- `getFirstPaintCss()` runs unmemoized in the provider's render body, and it's a 6709-character string on the demos page. A `useMemo` on `[scopeId, resolvedDrawerAt, resolvedCompactAt]` is free.
- `responsive?.scopeId` in `ResponsiveAside` can't be nullish — `useSidebarMenuResponsive()` on the next line throws without a provider. The `?.` would only ever bake `"undefined"` into `scopeSelector`.

Nothing blocking beyond the first one 👍

PR #9394, opened 2026-09-18
fix(Modal): restore focus after controlled dialogs close

tujoworker:
Controlled Dialogs and Drawers with external triggers could lose focus to the page when closing because Modal captured the active element before the actual open transition.

Restore focus to the element that opened the Modal by default, with explicit `restoreFocusTo` and `restoreFocus` controls for workflows that need another destination or manage focus themselves. The logged-out Dialog example now demonstrates the explicit target and uses the Dialog close lifecycle.

Preview: https://fix-modal-restore-focus.eufemia-e25.pages.dev/uilib/components/dialog/#logged-out-dialog


langz (approved):

langz (approved):
Approving on `d890d89`. Traced the regression on the branch and the fix lands right on it: the old code seeded `activeElementRef` from `document.activeElement` on the Modal's *first* render — almost always `<body>` — and then gated the restore behind `open === true`, which a controlled Dialog has already flipped to `false` by the time the close side effect runs. Capturing at open-request time, skipping `body`, and dropping the `open` gate fixes both halves.

Spot-checked the rest:

- Both props are destructured out of `rest`, so nothing leaks through to `ModalRoot` or the DOM.
- `restoreFocus: true` in `modalDefaultProps` is what makes `Provider Modal={{ restoreFocus: false }}` work, and `extendPropsWithContext` still lets an explicit `restoreFocus={false}` win.
- The `isConnected` filter earns its place — `ModalRoot` unmounts in the commit before `handleSideEffects` runs, so a stale in-modal target gets skipped rather than focused.
- `Dialog` and `Drawer` both build explicit `modalProps` allow-lists and both forward the new props. Their properties pages import `ModalProperties`, so the docs come along for free.
- No overlap with `ModalContent`'s `setFocus` — different scope, different phase.

Two non-blocking follow-ups: `restoreFocusTo()` is resolved outside the `try/catch`, so a throwing consumer callback would take `removeActiveState()` with it and leave `data-dnb-modal-active` stuck on `<html>`; and `Drawer` got the same forwarding as `Dialog` but no test.

Nice one — a sane default with an explicit escape hatch is the right shape here.

PR #9395, opened 2026-09-21
docs(Portal): document the company logo library

tujoworker:
Make Eufemia’s shared company-logo library discoverable in the Portal, including representative contents and links to the catalogue, package source and manifest.


langz (approved):

langz (approved):
Approving on `465ee53`.

Checked the branch preview rather than just the diff: `/foundations/` and `/foundations/logos/` both render, the section menu builds itself correctly (`Foundations` -> `Company logos`, active state included), and the build's internal link check is happy with the two new inbound links from `/brand` and `/icons`. `push-algolia.mjs` scans all of `src/docs`, so the page gets indexed for search too. All 18 checks green.

Two notes, neither blocking:

**The new top-level section has no entry point on the front page.** `MainMenu.tsx` filters on a hardcoded slug list, so `foundations` never gets a card - confirmed absent on the preview. That leaves the page reachable only through the two inline links and search, which is a little thin for something whose whole motivation is discoverability. Reasonable if `Foundations` is meant to grow first, but worth a follow-up.

**Every outbound link is DNB-internal.** `eufemia-logos.pages.dnb.ghe.com` and `dnb.ghe.com/eufemia/logos` are the first `ghe.com` links anywhere in `src/docs`. On a public eufemia.dnb.no the page is three bullet points plus three links nobody outside the network can open, with nothing explaining why - a short "internal only" note beside them would save some head-scratching.

Minor: `/quickguide-designer/logos` covers the DNB logo itself, so we now have two logo pages that don't know about each other. A cross-link would help.

Good to have this written down at last.

PR #9396, opened 2026-09-21
chore(Portal): update the meta description through its content attribute

langz:
`usePortalHead` assigned to `textContent` on the `<meta id="head-description">` element. `<meta>` is void, so that assignment never reaches the `content` attribute — the only part of the element anything reads:

```js
el.textContent = 'Updated'
// outerHTML: <meta id="head-description" name="description" content="Original">
```

The description therefore stayed at whatever the entry page was prerendered with, no matter where you navigated afterwards. The title one line above has the same shape and is correct, since `textContent` is the right channel for `<title>` — likely why this went unnoticed.

A page that declares no description of its own now falls back to the site metadata description, the same value the prerender writes into the served HTML, instead of inheriting the description of the page you came from. Only around 250 of the ~1050 MDX files declare one (others inherit it from their category), so that is the common path rather than an edge case.

Found while reviewing #9391, which escapes the same metadata on the prerender side.


langz:
Reviewed this locally. The `textContent` → `setAttribute('content', …)` change is the right fix, and the new test does fail on `main` for the expected reason.

One thing I'd like to sort out before merging: the update is behind `if (description)`, so navigating to a page that declares no description now leaves the previous page's description in place. Only ~250 of ~1050 MDX files set `description` in frontmatter (some others inherit it from their category), so this is the common path rather than an edge case.

Verified with a `rerender`:

```js
// prerendered: "Default site description"
rerender({ title: 'Button', description: 'A clickable button' }) // → "A clickable button" ✔
rerender({ title: 'Some other page', description: undefined })   // → still "A clickable button"
```

On `main` that last step kept the prerendered value, so for description-less pages this is a small step back. A fallback closes it — either capture the prerendered `content` on the first run, or let `PortalLayout` fall back to the site metadata description (the same string `prerender-utils` uses as `defaultDescription`). The new `keeps the prerendered description when a page has none` test currently locks the gap in, so it would need to move along with it.

Non-blocking: `og:description` is injected at prerender too and isn't touched here, so it stays stale after client-side navigation. Pre-existing, but worth a follow-up if we care about the live DOM values.

langz:
Addressed in `21002c3`: the hook now takes the default from `site.siteMetadata.description` — the same string the prerender writes as `defaultDescription` — so a page without its own description no longer inherits the previous one. The head elements are null-checked too, so the description update no longer depends on `#head-title` being present.

The `og:description` note stands as a separate follow-up.

langz (commented):
Final review of both commits — all good from my side.

`textContent` on a void `<meta>` is inert, so `content` is the right channel; the test fails on `main` for exactly that reason. The fallback resolves to the same string the prerender writes as `defaultDescription` — all five occurrences of it in the repo are byte-identical, so there is no new copy and no drift. Dropping the empty `try/catch` for null checks is needed now that the description update always runs, since a missing `#head-title` would otherwise have skipped it.

Ran the portal test suite, type checks and lint locally; the only failures are pre-existing timeouts in the babel/plugin/algolia tests that fail on an untouched `main` as well.

No remaining concerns. `og:description` is still only set at prerender time and stays stale on client-side navigation, but that is pre-existing and out of scope here.

Can't hit approve on my own PR, so this needs a second pair of eyes to formally approve.

langz (commented):
Final pass on `21002c3`. Merges cleanly with current `main` and the portal suite passes on the merge; both earlier notes hold up — the suite fails on `main` with `expected 'Prerendered description' to be 'A clickable button'` and passes here.

One new thing. Now that the client actually writes `content`, it overwrites the prerendered value — and the two sides don't resolve the description the same way. `PortalLayout` inherits the category description through the `fmData` merge, while `getPageMeta()` in `prerender.mjs` only falls back to the parent when the page has no title of its own, and only checks the direct parent slug.

On the preview, `/contribute/contact/` serves the site default and hydrates to `Project overview, development guides, conventions, etc.`. Running both rules over `src/docs`, 77 pages differ — 53 because the page has its own title, 24 because only a grandparent carries the description. `og:description` is prerender-only, so on those pages the two description tags now disagree.

The client value is the better one, so the fix belongs in `getPageMeta()`: inherit `description` independently of `title`, and walk up to the nearest existing ancestor. Non-blocking — this PR just makes it visible.

Two smaller ones:

- The default string now has a client-side reader. `portal-query.tsx`, `prerender-utils.ts`, `prerender.mjs` and `index.html` each hold their own hardcoded copy — byte-identical today, but if one moves, hydration quietly rewrites the description on every page that declares none.
- The title keeps its `if (title)` guard with no equivalent fallback, so a route with no resolvable title keeps the previous `document.title` where the prerender writes `Eufemia`. Deliberate?

langz:
Signing off from my side. The three points above are follow-ups, not merge blockers — the fix itself is correct and covered.

GitHub still won't let me approve my own PR, so it needs someone else to hit the button.

PR #9400, opened 2026-09-21
chore(analytics): retain a durable daily page-view rollup and expire raw records after 13 months

joakbjerk:
Anonymous portal page-view records accumulate in S3 with no expiry. They can only be safely expired once their history is preserved in a durable aggregation — but page-views never got a daily rollup like MCP and component usage have.

Preserving that history in a durable rollup is what makes it safe to then expire the raw records after 13 months (matching the MCP/component retention and the year-over-year need).

The rollup is write-only for now: the dashboard still shows the recent raw rows, and surfacing the retained long-range history (and its dimensions) is a follow-up. The store leads; the UI catches up.


langz (approved):
One thing worth fixing before this lands:

**The page-views already in `portal-views/` never reach the rollup.** `PORTAL_VIEWS_ROLLUP_DAYS = 7` only ever recomputes `dt >= today-6`, so every day between #9082 (page-view tracking went live 2026-08-26) and deploy minus 7 days gets no `portal-views-daily/` object — and this PR is what schedules exactly those rows for deletion at 395 days. MCP did not have this gap: its rollup and its raw prefix shipped together in #9320, so nothing predated the window.

A one-time wide window, or a single manual `aggregatePortalViewsRaw` run from the tracking start date, closes it before the lifecycle rule is applied.

The rest lines up: the trailing slash on `portal-views/` keeps `portal-views-daily/` out of the rule, the bucket-wide noncurrent cleanup already covers the hourly rollup overwrites, and the best-effort catch plus alarm follow the MCP/component pattern.

joakbjerk:
Thanks @langz — good catch on the pre-window history. Addressed it as a one-time backfill (your option 2) rather than baking it into the scheduled path:

- The scheduled generator stays tail-only (recomputes the recent ~7 days).
- The historical pass is a deliberate one-time invoke with an explicit `sinceDt` (e.g. `{"sinceDt":"2024-01-01"}`, the Glue projection start) — idempotent per day, documented in the analytics README, and tracked as a post-deploy step.

Went this route over a self-healing first-run backfill so there's no permanent hot-path code for a one-time migration, and no risk of a partial run silently orphaning an arbitrary subset of days. There's ~13 months of runway before any raw row expires, so it only needs to run once before then.

PR #9328, opened 2026-09-10
feat(Icons): refactor paths and add `add_reaction` and solve `icons`

tujoworker:
Adds the `add_reaction` and `solve` icons from Figma.

Motivation: https://dnb-it.slack.com/archives/C055SV1K0N6/p1789048221792529?thread_ts=1789030504.410919&cid=C055SV1K0N6

PR #8515, opened 2026-06-09
feat: update icons with accessibility, ai, clearing_house, football, holiday_home, moneybag_in_hand, moon, passkey, plot, sun, stop_square

tujoworker:
Updates the Eufemia icon library using the documented icon workflow.

Summary:
- Adds the new generated icon assets, components, exports, metadata, locks, and visual snapshots from the CI icon sync.
- Updates the Icons Library workflow to run `build:prebuild` during the icon branch flow so generated icon changes can be committed by the final `icons:commit` step.

PR #9402, opened 2026-09-21
feat(Icons): add sidebar

tujoworker:
Synchronizes the Eufemia icon library with Figma for review before release.


packages/dnb-eufemia/src/icons/dnb/icons-meta.json
```diff
@@ -4253,5 +4253,19 @@
     "name": "solve",
     "variant": "secondary",
     "category": "objects"
+  },
+  "sidebar_medium": {
+    "tags": ["AI sparkles", "KI"],
```
langz: Strange tags, hehe.

langz (approved):

langz:
Had a look at the generated output — structurally it's all there and CI is green. Two things I'd like sorted before this leaves draft.

### The tags come from the `ai` icon

```jsonc
"sidebar": { "tags": ["AI sparkles", "KI"], ... }
```

Byte-identical to the only other entries carrying those tags:

```
ai         ['AI sparkles', 'KI']
ai_medium  ['AI sparkles', 'KI']
```

Looks like the Figma component was duplicated from `ai` and the description never updated. It's user-visible rather than just metadata — `ListAllIcons.tsx:117` renders `tags.join(', ')`, so the icons page will show "AI sparkles, KI" under a sidebar glyph, and nobody searching "sidebar" or "panel" will find it. Compare `table` → `['Table', 'grid', 'module']`.

### The 16px variant looks undersized

Measured bounding boxes across the whole set:

| icon | bbox | set median |
| --- | --- | --- |
| `sidebar` | 12.25 | 14.00 (n=291) |
| `sidebar_medium` | 21.0 | 22.15 (n=294) |

`table`, `calendar`, `card` and `browser` are all 14 at 16px; below 12.25 you only find chevrons, `close` and `information`. The pair is also out of proportion with itself — 12.25/21 = 0.583 against the expected 16/24 = 0.667. Next to `table` in a toolbar the small one will read as shrunken, so worth a second pair of eyes from design.

### Not a blocker: the halved `.tgz` files

262 kB → 133 kB on an additive PR looks alarming, but nothing was lost. The archives on `main` contained every entry twice (1186 members / 594 unique); this regeneration writes each once, and the unique-set diff is exactly the two new files. Incidental fix of a pre-existing bug in `createTarWithoutCategories()`, which passes `listOfProcessedFiles` straight to `tar.create()`.

Two smaller notes while I'm here: `candle_stick`'s Figma id flip-flops on every sync (`6068:*` ↔ `6036:*`, #8515 → #9328 → this one), which suggests two components share that name — it's also the only icon still duplicated in the archives. And `EUFEMIA_CHANGELOG.mdx` still needs the new name under the release date, though I assume that's pending while this is a draft.

langz (approved):
We can merge now if we would like, as we just released our planned bugfix release v11.14.1

PR #9404, opened 2026-09-22
chore(Theme): add manual Figma synchronization of design tokens

langz:
Design tokens were the only Figma asset still exported by hand: someone had to open the Figma file, right-click two variable collections, export the modes and drop five JSON files into the repository. That made token updates rare, easy to get subtly wrong, and hard to review as a discrete change.

This mirrors the icon synchronization added in #9401, so design tokens arrive as a draft pull request that can be reviewed against the regenerated CSS and the visual snapshots before release.

The workflow needs two things before the first run: a `FIGMA_TOKENS_FILE` repository secret, and the `file_variables:read` scope on the existing `FIGMA_TOKEN`. The Figma variables API is Enterprise-only.


tujoworker (approved):
Reviewed the full implementation and the repeat-run safeguards. Looks good.

.github/workflows/tokens-sync.yml
tujoworker, line 1: maybe it would make more sense to name this file tokens-sync instead of "tokens-lib"?

langz: Thanks, I'll take a look into this.

langz: Agreed – renamed it to `tokens-sync.yml`. The `-lib` suffix was copied from `icons-lib.yml`, which really is a library build workflow, while this one only runs the sync on dispatch.

The Actions link in `tokens.mdx` is updated too. I kept the workflow display name as "Design Tokens" so the docs wording still matches what you see in the Actions list.

PR #9405, opened 2026-09-22
chore(visual-regression): make webfont loading deterministic in screenshot tests

langz:
## Is it flaky? Yes — confirmed

Comparing recent `visual-regression` runs on `main` (same committed baselines, no relevant source change):

| run | date | result | failing tests |
| --- | --- | --- | --- |
| [35588388572](https://github.com/dnbexperience/eufemia/actions/runs/35588388572) | 09-21 | pass | — |
| [35592906540](https://github.com/dnbexperience/eufemia/actions/runs/35592906540) | 09-21 | pass | — |
| [35599048755](https://github.com/dnbexperience/eufemia/actions/runs/35599048755) | 09-21 | **fail** | 8 × `Table*` (ui + sbanken) |
| [35606090567](https://github.com/dnbexperience/eufemia/actions/runs/35606090567) | 09-21 | pass | — |
| [35627091180](https://github.com/dnbexperience/eufemia/actions/runs/35627091180) | 09-21 | pass | — |
| [35706234815](https://github.com/dnbexperience/eufemia/actions/runs/35706234815) | 09-22 | **fail** | `Selection`, `Iterate.Array` |

The two failing runs have **completely disjoint** failing test sets, and `main` passes on either side of them. So these are not regressions. (The `portal-home-mdx` / `portal-sidebar-menu` failures in the same period are different — those are genuine `SidebarMenu` changes.)

Two further properties of the flake:

- **Failures cluster per test file / worker** — 8 `Table` tests in one run, forms tests in another. That points at per-browser-process state rather than a per-test race.
- **All 6 attempts (1 + 5 retries) produce byte-identical output** — the same pixel counts repeat exactly. The existing `retry: 5` therefore gives no protection at all against this class.
- The signature is always **width-only**: `265x104 -> 257x104`, `285x728 -> 281x728`, `232x352 -> 240x352`. Heights never move. Something changes text measurement while the layout structure stays intact.

## What this PR fixes

While investigating I found a concrete, reproducible hazard that produces exactly that signature and that the engine **cannot currently detect**.

Eufemia declares its webfonts with `font-display: fallback`, which gives the browser roughly a 3 second window to swap the real font in. Miss it and the page paints the fallback face for the rest of its lifetime. Verified in a browser with a delayed font response:

| scenario | webfont actually painted |
| --- | --- |
| fast font, `font-display: fallback` | yes |
| 4s font, `font-display: fallback` | **no** — while `fonts.status: loaded` and `check(): true` |
| 4s font, re-declared as `font-display: block` | yes |
| 8s font, re-declared as `font-display: block` | yes |

The engine waits on `document.fonts.ready`, which resolves happily in the failing case — so a capture made in the wrong font looks completely healthy to the tooling.

On top of that, `waitForVisualStability` memoised the font wait per page load, so any face requested by demo content rendered *after* the first wait was never waited for at all.

So: re-declare each active `@font-face` with `font-display: block` before every capture (the new face has no swap deadline), and always await `document.fonts.ready`.

**I have not proven this is what caused the two failures above** — the diff images are too low-resolution for a conclusive font identification, and I could not reproduce the CI environment locally. This removes a real and currently invisible failure mode; whether it removes *the* one needs a few runs on `main` to tell.

Two details worth calling out:

- `cssText` serializes `url()` as authored, so font URLs are resolved against their own stylesheet before being re-declared. Without this a stylesheet-relative path would resolve against the current route and 404.
- Disabled stylesheets are skipped, so the four pre-injected theme stylesheets do not leak another theme's font faces into the page.


langz (commented):
Reviewed and signed off — the diagnosis holds and the implementation is sound. Nothing blocking. (Posted as a comment because GitHub does not allow the author to submit a formal approval.)

I reproduced the hazard independently in a controlled browser run with a delayed font response, using the real `DNB-Regular.woff2`. A 40px probe measures 383.61px in the webfont and 349.22px in the fallback:

| scenario | painted |
| --- | --- |
| fast font, `fallback` | 383.61 (webfont) |
| 4s font, `fallback` | **349.22 (fallback)** — with `fonts.status: loaded`, `check(): true` |
| 4s font + this PR | 383.61 |
| 8s font + this PR | 383.61 |

So the invisible-to-the-engine part is exactly as described. A 404 font under `block` falls back immediately, so the block period can't strand a capture either.

### One finding worth acting on

Your own evidence had a loose end: every retry forces a `navigate` (`getPageResetStrategyFromMutation`, `currentRetry > 0`), so a merely-slow font should have been warm by attempt 2 — yet all 6 attempts were byte-identical.

`vite preview` explains it. It serves through `sirv(distDir, { etag: true, dev: true })` — **ETag only, no `Cache-Control`**. Measured against that header profile, the browser re-requests the font on *every* navigation, so retries re-queue on the same saturated server instead of hitting a cache.

That also means this PR currently pays for its correctness. A/B, 2s font delay, 5 sequential navigations on one page:

| variant | font HTTP | wall |
| --- | --- | --- |
| no fix | 5 | 10.3s |
| this PR | 8 | 16.2s |
| no fix + `Cache-Control: immutable` | 1 | 2.2s |
| this PR + `immutable` | 1 | 2.2s |

The portal's `vite.config.ts` has no `preview` block and every Vite asset is content-hashed, so `preview.headers` with `public, max-age=31536000, immutable` would remove the cause rather than only surviving it — and make this change free. Worth scoping to `IS_VISUAL_TEST` so `yarn serve` isn't affected. Happy for that to be a follow-up.

### Non-blocking notes

- **The override depends on being last in document order** and never re-appends itself. Safe today — prerender puts all four theme links in `<head>` at build time and in-page switching only toggles `disabled` — but I confirmed it lands mid-`<head>` once any stylesheet is appended after it. `document.head.appendChild(existing)` in the `existing` branch makes it order-proof for free.
- **Scan cost:** the active theme sheet has ~3450 top-level rules, and `waitForVisualStability` runs about twice per screenshot across ~2000 screenshots. The per-capture call is necessary for theme switching, but a cheap signature guard (sheet count + disabled flags) would skip the rescan in the common no-change case.
- `rule.type !== 5` uses the deprecated `CSSRule.type`; `instanceof CSSFontFaceRule` is more future-proof.
- An in-place `rule.style.setProperty('font-display', 'block')` achieves the same thing with **one** font request instead of two and no URL absolutization — verified working and idempotent, but only in Chromium, so I'd not switch without a Firefox check.
- The `disabled` skip is more load-bearing than the description suggests: `ui`, `carnegie` and `eiendom` all declare `DNB`/`DNBSkeleton`/`DNBMonoSkeleton`, and `sbanken` adds `Roboto`/`MaisonNeueHeadings`.

### Verified

All 11 new tests pass, `@font-face` rules are top-level in the built theme CSS (no `@media`/`@layer` nesting), font URLs are already absolute there, `font-display` cannot change metrics once loaded so no baselines churn, and the capture path already does a double-rAF before `takeScreenshot` so the final font wait needs no trailing frame.

langz:
## Cost of the per-capture scan, measured

The override re-runs before every capture, so the obvious review question is what the stylesheet walk costs. Measured in a browser against the real compiled Eufemia CSS (`dnb-ui-core`, `ui-theme-basis`, `ui-theme-components`, `dnb-ui-components` as four separate sheets):

```
stylesheets: 4
total CSS rules: 6164
@font-face rules re-declared: 12
override <style> tags (must be 1): 1

forceFontDisplayBlock: 0.36 ms per call
~2 calls per screenshot => 0.72 ms added per capture
```

Under a second across the whole suite, and idempotency holds at that scale — repeated calls reuse the single injected `<style>` rather than accumulating.

## Where this PR stands

Worth being plain about it: the `font-display: fallback` failure period is a **proven** hazard that `document.fonts.ready` and `document.fonts.check()` cannot detect, and the memoised font wait was genuinely unsound. But I have **not** shown that either caused the flaky failures that prompted the investigation. The full screenshot suite is green with this change, so it shifts no baselines.

If you would rather not carry the extra machinery on a theory, the defensible minimum is the second half alone — always awaiting `document.fonts.ready` instead of memoising it per page load. That part is pure hardening with no new behaviour. Happy to reduce it to that.

PR #9406, opened 2026-09-22
chore(analytics): add ingest route for local MCP usage telemetry

joakbjerk:
Adds the ingest side (HTTP route + validator + store) for anonymous usage telemetry from the local (stdio) Eufemia MCP server, mirroring the existing server-side Lambda MCP tracking plus an `eufemiaVersion` field. The local server runs on public/open-source machines and can't hold a secret or IAM role, so it uses the browser model — an edge-locked `POST /collect-local-mcp-usage` route (Akamai injects `X-Edge-Auth`) instead of writing S3 directly.

Part of EDS-977. Local-server beacon and dashboard breakdown follow in separate PRs.


langz (approved):
Ran both producers side by side against the same inputs to see what actually lands in the shared table. The route, the server-side `transport` stamp and the validator coverage all look right. Two things I'd like fixed before merge.

### Component and path values diverge from the web producer writing the same table

`storeMcpUsage()` writes `mcp-usage/dt=…` — the same prefix and Glue table as `tools/mcp-lambda` — and `aggregateMcpUsageRaw()` groups on `dt, tool, component, path`. The two producers canonicalise differently:

| input | web row (`mcp-lambda`) | local row (this PR) |
| --- | --- | --- |
| `Button` | `button` | `Button` |
| `DatePicker` | `datepicker` | `DatePicker` |
| `date-picker` | `date-picker` | `""` |
| `/uilib/./button.md` | `/uilib/button.md` | `/uilib/./button.md` |
| `/uilib/button.md?x=1` | `/uilib/button.md` | `""` |

So `perComponent` ends up with `Button` and `button` as two entries for the same component, and the hyphenated form loses the dimension on the local side only. The package already solved this in `src/records/component-usage.ts` — `normalizeComponentName()`, "so different import spellings … collapse to one row when Athena groups by component instead of fragmenting". `buildMcpUsageRecord()` needs the same `trim().toLowerCase()`, `COMPONENT_SEGMENT` needs to accept the hyphenated segment (`/^[A-Za-z][A-Za-z0-9-]*$/`), and `path` needs the `canonicalDocsPath()` treatment.

The module header says the two copies "must be kept in sync" — they already aren't, and nothing pins it. `mcp-lambda` pins its `normalizeDocsPath` parity by test; worth doing the same here.

### `transport` is NULL on every web row

`McpUsageRecord` in `tools/mcp-lambda/src/records/mcp-usage.ts` has no `transport` key, so the new column is empty for every row the web Lambda writes, past and future:

```json
{"tool":"docs_entry","component":"","path":"","env":"prod","timestamp":"…","createdat":"…"}
```

`store.ts` says the table is "discriminated by the `transport` field", but `WHERE transport = 'web'` matches nothing. Either stamp `transport: 'web'` in the mcp-lambda record here, or the follow-up has to read `COALESCE(transport, 'web')`. Worth deciding now, since the column ships in this PR.

CI is green. Approving so this isn't blocked on a re-review, but I'd like both sorted before it goes in.

PR #9408, opened 2026-09-22
chore(analytics): notify an SNS topic from the dashboard snapshot alarms

joakbjerk:
The dashboard snapshot generator has had CloudWatch alarms (errors, not running, empty snapshot, MCP/component usage build failures) since earlier work, but none of them had a notification target — an alarm firing was only visible if someone happened to look at the CloudWatch console.

This adds an SNS topic and wires it as `alarm_actions` on all five snapshot alarms. A subscription is optional (`snapshot_alert_email`, empty by default) so the topic can ship before an owner is chosen, or before someone decides they'd rather subscribe a Slack webhook instead — either way, only a new `aws_sns_topic_subscription` is needed, the topic and alarms don't change.

Needs a matching IAM grant in the OIDC federation repo before it can deploy (companion PR incoming).


langz (approved):
Read through the Terraform and the workflow — the wiring itself is right, and the test covers all five alarms plus the conditional subscription. Two things I'd want sorted before this lands.

### Merge order: this breaks every analytics deploy until the grant is live

`ghe-deploy-workflow.yml` runs `terraform apply -auto-approve` unattended under `set -euo pipefail`, and as you note in the description the deploy role has no SNS permissions yet. `aws_sns_topic.snapshot_alerts` is the first SNS resource in this stack, so an AccessDenied on `sns:CreateTopic` fails the apply outright — and since all five alarms reference `aws_sns_topic.snapshot_alerts.arn`, nothing downstream runs either (no `Seed snapshot`, no dashboard publish). That blocks unrelated analytics work like #9400 and #9406 too, not just this change. So: grant first, merge second.

Worth listing the full lifecycle set in the companion PR, not just create — `sns:CreateTopic`, `GetTopicAttributes`, `SetTopicAttributes`, `DeleteTopic`, `TagResource`, `UntagResource`, `ListTagsForResource` (the last three because of `tags = local.tags`, easy to miss), plus `Subscribe`, `Unsubscribe`, `GetSubscriptionAttributes`, `ListSubscriptionsByTopic` for the subscription.

And the README's "One-time bootstrap (admin, out-of-band)" section is where every other out-of-band grant is written down (both Lambda roles, the deploy role's CloudFront/S3). The SNS grant only lives in this PR description — please add it there so the next person doesn't have to dig it out of a merged PR.

### An unsubscribed or unconfirmed topic is worse than no alarm actions

`email` is only partially supported by the provider: the subscription is created in `pending confirmation` and delivers nothing until a human clicks the link, and while pending Terraform can't unsubscribe it either (`destroy` drops it from state but leaves it in AWS). With `snapshot_alert_email` defaulting to `""`, the most likely post-merge state is a topic with zero subscribers.

The config would then *look* solved while behaving exactly as before — which is a step back from the honest `# No alarm actions yet (state is visible in CloudWatch)` comments this replaces. I'd rather land it with `SNAPSHOT_ALERT_EMAIL` set and the confirmation clicked as part of the rollout, so the alarms have a real subscriber the day the topic exists.

### Smaller note: the Slack webhook suggestion doesn't hold

Both the README and the `main.tf` comment offer "an HTTPS Slack webhook" as a drop-in alternative subscriber. A Slack incoming webhook can't confirm an SNS subscription — it rejects the `SubscriptionConfirmation` payload as an invalid body, so it never moves out of pending, and it wouldn't render the alarm JSON as a message even if it did. Needs a Lambda or AWS Chatbot in between. Either drop the example or say what it actually requires.

joakbjerk:
Thanks for the review — pushed 08edc674116 addressing both points:

- Slack suggestion dropped from `main.tf` and the README: a raw incoming webhook can't complete the SNS subscription confirmation handshake (or format the alarm JSON as a message), so it's not a drop-in target. Noted AWS Chatbot or a formatting Lambda instead.
- README now lists the SNS grant alongside the existing CloudFront/S3 line documenting what's provisioned in the OIDC federation repo.

On merge order: the companion IAM grant (dnb-tooling/github-oidc-federation#2746) is merged and its deploy run succeeded, so the permissions are live, not just merged, ahead of this landing.

`SNAPSHOT_ALERT_EMAIL` is also set on the deploy repo, so the next deploy creates the topic with a real subscriber attached (just needs the one confirmation-email click).

PR #9410, opened 2026-09-22
chore(visual-regression): clear browser storage between screenshot tests

langz:
## The bug

The screenshot engine calls `clearBrowserStorage` before every fresh navigation and every hard reset, to stop one test's state leaking into the next. It has never worked.

`clearBrowserStorages` is passed to `page.evaluate`, which ships `fn.toString()` to the browser. The function called `clearStorageSafely` — a *module-scope* binding that does not exist in the page. Every call threw a `ReferenceError`, the surrounding `try/catch` swallowed it, and `page.evaluate` resolved normally. Confirmed in a browser against the shipped module:

```
before: {"local":2,"session":1}
after:  {"local":2,"session":1}
evaluate threw: no
```

## What it was corrupting

`SidebarMenuRoot` persists its expanded-accordion state. The demo declares:

```jsx
<SidebarMenu.Root
  defaultOpenItems={['products']}
  openItemsStorageKey="sidebar-menu-declarative-example"
>
```

So from a clean state Products is **expanded**, and open-state is **persisted** — a stored value overrides the default on later loads.

`SidebarMenu.screenshot.test.ts` is `describe.each(['ui', 'sbanken'])` sharing one page. `ui` runs first from a clean context, so its baselines are correct. Test 12 of 13, *"closes the declarative Products accordion with one click"*, then collapses Products and persists it — and `sbanken` inherited that for every one of its declarative tests.

The two committed baselines for `matches the declarative menu`, a test that performs **no interaction at all**, had diverged:

| baseline | Products | size |
| --- | --- | --- |
| `...for-ui-matches-the-declarative-menu-1.png` | **expanded** | 416x520 |
| `...for-sbanken-matches-the-declarative-menu-1.png` | **collapsed** | 416x368 |

A theme cannot change whether an accordion is expanded. And the old sbanken baseline for *"closes the declarative Products accordion"* showed Products **expanded** — the accordion was already collapsed from the leak, so the click **opened** it. That snapshot asserted the opposite of its own name.

Corroborating the mechanism: the `sidebar-menu-data` demo sets neither `defaultOpenItems` nor `openItemsStorageKey`, and none of its tests were affected.

## Changes

1. Make `clearBrowserStorages` self-contained. `clearStorageSafely` had no other callers and is gone with it.
2. Regenerate the 9 `SidebarMenu for sbanken` baselines (separate commit, easy to review or drop on its own). All nine now measure 416x520, matching their `for ui` twins.

The new test re-evaluates the function through `new Function(fn.toString())`, which is what `page.evaluate` effectively does. It fails against the previous implementation (`expected 1 to be +0`) and passes now — the old tests all passed because they called the helper directly, in-process, where module scope still resolved.

**On the baselines:** Playwright's Firefox will not launch in my environment, so rather than `test:screenshots:ci:update` I took the PNGs the engine captured in CI run [35720647837](https://github.com/dnbexperience/eufemia/actions/runs/35720647837) from its `visual-test-artifact` and wrote them to the paths in `report.json`. That is byte-identical to what `--update` writes. Regenerate and force-push over that commit if you would rather have them produced the normal way.

## Scope

This restores an isolation guarantee the engine already assumes. It is **not** a flakiness fix — the leaked state is deterministic within a file, so it produced a consistently wrong baseline rather than a bouncing one. Because `src/core/vitest-screenshots/` is a global-impact path, CI ran the **full** screenshot suite and only these 9 snapshots differed.


langz:
## The failing visual run is the fix working — but it needs baseline regeneration

> **Correction to an earlier version of this comment:** I originally wrote that this also explains the recurring `SidebarMenu` height flakiness on other branches. That was wrong. The runs I cited ([35605763185](https://github.com/dnbexperience/eufemia/actions/runs/35605763185), [35604220396](https://github.com/dnbexperience/eufemia/actions/runs/35604220396)) were on `chore/portal-home-mdx` and `docs/portal-sidebar-menu`, both of which modify `sidebar-menu` heavily — those failures were genuine, not flaky. `SidebarMenu` has not failed on `main` in the runs I checked.

`Run visual-regression tests` fails with 9 × `SidebarMenu for sbanken` snapshots (`416x368 -> 416x520`, `416x501 -> 416x520`). That is not a regression from this change — it is the leak becoming visible.

`SidebarMenuRoot` persists its expanded-accordion state to session/local storage. With the reset silently broken, that state survived between tests.

The proof is the pair of **committed baselines** for `matches the declarative menu`, a test that performs **no interaction at all**:

| baseline | Products accordion | size |
| --- | --- | --- |
| `SidebarMenu-for-ui-matches-the-declarative-menu-1.png` | **expanded** (Accounts / Cards / Controls) | 416x520 |
| `SidebarMenu-for-sbanken-matches-the-declarative-menu-1.png` | **collapsed** | 416x368 |

A theme cannot change whether an accordion is expanded. The file is `describe.each(['ui', 'sbanken'])` sharing one page: `ui` runs first on a fresh context and is correct, then the sibling test *"closes the declarative Products accordion with one click"* writes the collapsed state to storage, and `sbanken` inherits it. **The committed sbanken baselines are wrong.** With storage actually cleared, `sbanken` now renders the same as `ui`.

## What this PR is and is not

It restores an isolation guarantee the engine already assumes, and corrects baselines that silently encode another test's leftover state. I have **not** shown that it fixes run-to-run flakiness — the leaked state is deterministic within a file, so it produces a consistently wrong baseline rather than a bouncing one.

**It should not be merged until the 9 `SidebarMenu for sbanken` baselines are regenerated.** I have deliberately not hand-committed the CI-produced PNGs; they should come from `test:screenshots:ci:update`.

(The `Iterate.Array > have to match view container` failure in the same run is the known unrelated flake — `232x352 -> 240x352`, the same one seen on `main`.)

langz:
## Baselines regenerated

The 9 `SidebarMenu for sbanken` baselines are updated in dcbe689.

**How they were produced — worth knowing, because it is not the documented route.** Playwright's Firefox will not launch in my environment, so I could not run `test:screenshots:ci:update`. Instead I took the PNGs the engine itself captured in the failing run ([35720647837](https://github.com/dnbexperience/eufemia/actions/runs/35720647837)) from its `visual-test-artifact`, and wrote them to the snapshot paths derived from `report.json`.

That is byte-identical to what `--update` writes — `diffAndPersist` persists the same `actualBytes` to both the snapshot and the `.actual.png` in the report — and it has the advantage of coming from the CI environment rather than my machine. **If you would rather have them generated the normal way, regenerate and force-push over this commit; the content should be identical.**

Checks I ran before committing:

- All nine now measure **416x520**, matching their committed `for ui` twins exactly. Previously eight were 416x368 and one 416x501.
- Spot-checked renderings against the `ui` baselines. `closes the declarative Products accordion with one click` shows Products **collapsed** in both; `matches a hovered item` shows Products expanded with the hover highlight on `Overview` in both. The only differences are the expected theme styling (typeface, badge colour).
- The `Iterate.Array` snapshot was **deliberately excluded** — that failure is the known unrelated flake, not something this PR should rewrite.
- `git status` shows exactly 9 changed files, all under `sidebar-menu/__tests__/__image_snapshots__/`.

langz:
## Proof that the old baselines were the faulty ones

Fair challenge that the regenerated screenshots changed a lot, so here is the chain of evidence rather than just "ui and sbanken now agree" — which only shows consistency, not correctness.

**1. The demo declares Products open.** `SidebarMenuDeclarative` in the portal `Examples.tsx`:

```jsx
<SidebarMenu.Root
  defaultActiveSection="personal"
  defaultOpenItems={['products']}
  openItemsStorageKey="sidebar-menu-declarative-example"
>
```

From a clean state the Products accordion is **expanded**, and open-state is **persisted** under that key — a stored value overrides `defaultOpenItems` on later loads.

**2. Only that menu persists anything.** The `sidebar-menu-data` demo has neither `defaultOpenItems` nor `openItemsStorageKey`. And indeed only the *declarative* tests failed — `matches the data menu`, `matches wrapped accordion labels` and `closes the data Products accordion with one click` all passed untouched. The blast radius matches the mechanism exactly.

**3. Test order explains why only sbanken was affected.** `closes the declarative Products accordion with one click` is test **12 of 13**. The `ui` block runs first from a clean context, so its baselines are right. That test then collapses Products and persists it — and because the storage reset never ran, the `sbanken` block inherited the collapsed state for every one of its declarative tests.

**4. The clincher.** Here is the *old committed* sbanken baseline for the test named **"closes the declarative Products accordion with one click"**:

> Products **expanded** — chevron up, Accounts / Cards / Controls visible (416x501)

The accordion was already collapsed from the leak, so the single click **opened** it. The baseline asserted the exact opposite of the test's name. The new one shows Products collapsed, which is what the test says it does.

## Answer

It was wrong **before** this PR. Eight sbanken baselines showed a state contradicting the demo's declared default for tests that perform no interaction at all, and the ninth captured the inverse of its own assertion. The change is large because the correction is large — nine snapshots had silently encoded another test's leftover state.

tujoworker (approved):

PR #9412, opened 2026-09-22
chore(mcp): send anonymous, opt-out usage telemetry from the local server

joakbjerk:
Local MCP server telemetry to match what the web MCP Lambda already tracks, so we can see which tools, components and docs areas are actually used and prioritize accordingly. Anonymous and opt-out (`EUFEMIA_MCP_TELEMETRY=0`); sends to the ingest route from #9406.


langz (approved):
Read through it — the shape is right and the fire-and-forget path is properly defensive. Four things I'd like sorted before this goes in.

### The component regex drops the only form that resolves for multi-word components

```ts
const COMPONENT_SEGMENT = /^[A-Z][A-Za-z0-9]*$/
```

For a plain name, `conventionalDocPath` builds exactly one candidate — `` `/uilib/components/${normalizeName(name)}.md` `` (`mcp-docs-server.ts:224`) — and `normalizeName` lowercases. The docs are kebab-case (`info-card.md`, `global-status.md`, `input-masked.md`), so the input that actually resolves is `info-card`; `InfoCard` maps to `infocard.md`, which doesn't exist. Dot notation is case-insensitive too — there's an existing test asserting `componentFind({ name: 'field.address' })` resolves (`mcp-docs-server.test.ts:741`).

None of those pass the regex, so every successful multi-word lookup degrades to a tool count and what survives is single-word PascalCase: `Button`, `Input`, `Slider`. That's the headline metric of this PR quietly losing most of its rows.

The ingest validator in #9406 already handles it, with a comment saying the hyphenated doc-file form must validate:

```ts
const COMPONENT_SEGMENT = /^[A-Za-z][A-Za-z0-9-]*$/
function normalizeComponent(value: string): string {
  return value.trim().toLowerCase()
}
```

The two are meant to be kept in sync — worth matching the ingest side here.

### `path` isn't a closed vocabulary

`docs_read`'s `path` and `docs_list`'s `prefix` are free-form `z.string()` supplied by the calling model, and `isValidPath` only checks length, `..` and a charset. `/some-client-project-name` passes and goes out verbatim. `track` also doesn't await `run()`, so a path that doesn't exist in the docs at all is still reported.

That sits badly against what we state — "Only a closed, public vocabulary leaves the machine" and "no free text are ever sent". `getMarkdownFilesCached` is right there: either gate the path on a real doc, or send only the leading area segment (`/uilib/components/`), which is all "which documentation areas are most used" actually needs.

### "No ... IP address ... are ever sent"

From `tools.mdx`. Every POST discloses the client IP to the edge — unavoidable in HTTP. #9406 doesn't store it, and that's the real guarantee, so this should say *not collected* / *not stored*. Worth getting exact, since it's a public promise on the portal about traffic from people's own machines.

Related: for an opt-out default the only signal is a stderr line, which most MCP clients never surface. Since this ships inside a public npm package, I'd want privacy to have looked at the default before release rather than after. (The README says the notice prints "on first run" — `noticePrinted` is module-level, so it's once per process, i.e. every server start.)

### The endpoint doesn't exist yet

#9406 is still open and carries the route plus the Glue columns:

```hcl
route_key = "POST /collect-local-mcp-usage"
```

So this needs #9406 merged, `terraform apply` run, and Akamai injecting `X-Edge-Auth` for the new path — otherwise every tool call beacons into a 403/404. Silent, but let's not release it that way.

Approving so it isn't blocked on me, but I'd like the first two fixed before merge.

packages/dnb-design-system-portal/src/docs/uilib/usage/first-steps/tools.mdx
```diff
@@ -108,10 +108,34 @@ raicode mcp add --transport stdio eufemia -- node node_modules/@dnb/eufemia/mcp/
 ### How to use
 
 - The MCP server helps AI apply Eufemia patterns more accurately in code, but results can still be imperfect. So always review the output carefully!
-- The MCP server provides documentation context only; it does not execute code or access the network.
+- The MCP server provides documentation context only; it does not execute code. It sends anonymous, opt-out usage statistics (see [Telemetry](#telemetry) below) but never reads your code, files or search queries.
```
tujoworker: I don't think we need to mention that here. Its not related to the item topic and the "How to use" section.

joakbjerk:
Pushed fixes for the review feedback:

- **Component matching** (blocking, per @langz): `component` now validates with `/^[A-Za-z][A-Za-z0-9-]*$/` + lowercase normalization, matching what the docs server actually resolves (hyphenated and case-insensitive lookups like `info-card`, `field.address`). (913b9d2)
- **`path` as a closed vocabulary** (blocking, per @langz): a `path`/`prefix` is now gated against `computeKnownAreas()` — the set of doc areas computed once from the server's own docs — so anything outside the real docs is dropped entirely instead of just narrowed. (52a7fb9)
- **Wording accuracy** (per @langz): corrected "no IP ... ever sent" to reflect that IP is disclosed at the HTTP layer like any request but isn't stored, and "on first run" → "once per server start" to match the actual guard. (913b9d2)
- **Doc placement** (per @tujoworker): dropped the telemetry mention from "How to use" and moved the "documentation context only, no code execution" scope note to a standalone line under the top-level heading, since it applies to both the hosted and local server. (cf48af3)

tujoworker (approved):

PR #9413, opened 2026-09-22
chore(Portal): recover ComponentBox previews after hot reload

tujoworker:
ComponentBox previews could retain stale module identities during Vite hot reload, leaving the Portal error boundary stuck after a transient render failure.

This keeps live-preview scopes current during development and retries the page after the HMR update completes.


langz (approved):
Read through it — the `ErrorBoundary` reset is clean and `vite:afterUpdate` is the right hook. One thing I'd like sorted before this goes in.

### The memo key is the actual bug, and switching dev off it hides it

```ts
const hash = children as string
```

`globalThis.ComponentBoxMemo` is keyed on the code string alone — `scope`, the injected `__buildScope`, `hideCode`, `data-visual-test`, everything else sits outside the key. That is what hands back a stale identity, and `shouldMemoize = !import.meta.hot` only stops us from seeing it locally: the prerender in `vite/prod/prerender.mjs` and the shipped client still run the memoized path.

It reproduces on eufemia.dnb.no today. `LayoutComponents` exists twice with identical code and different props:

- `uilib/extensions/forms/Examples.tsx:270` — `scope={{ Value }}`
- `uilib/layout/Examples.tsx:37` — `scope={{ Field, Form }} hideCode`

Load `/uilib/layout/` directly and the Profile example is collapsed behind the `Code` toggle. Navigate client-side from `/uilib/extensions/forms/intro-examples/` to `/uilib/layout/` and the same example renders the forms variant instead — editor open, no toggle. Same mismatch this PR is chasing, just triggered by routing rather than a hot update.

Folding the props that affect the output into the key fixes both, and keeps dev and prod on one path instead of leaving the memoized one exercised nowhere we look. Worth noting `import.meta.hot` is an object under `vitest run` too, so `ComponentBox.test.tsx` only ever hits the unmemoized branch — the memoized one has no coverage at all.

Approving so it isn't blocked on me.

PR #9415, opened 2026-09-22
fix(ColorSchemeScript): allow a CSP nonce on the color scheme scripts

### LINKED ISSUE #9409: bug(ColorSchemeScript): inline scripts cannot receive a CSP nonce
### Summary

`ColorSchemeHeadScript`, `ColorSchemeBodyFirstScript` and `ColorSchemeBodyLastScript` render inline `<script>` tags with no way to attach a CSP nonce. Under a strict `script-src 'nonce-…'` policy the browser blocks them, so the colour-scheme flash they exist to prevent comes back. The way out is adding `'unsafe-inline'` to `script-src`, which defeats the point of the policy.

Every other component in the shipped source that renders an inline `<script>` or `<style>` already supports a nonce, so this looks like an oversight rather than a deliberate choice.

### Affected

`packages/dnb-eufemia/src/shared/ColorSchemeScript.tsx` — the `<script>` elements on lines 67, 81 and 95.

### Reproduction

**1. There is no prop to pass.** `tsc` rejects it:

```
TS2322: Property 'nonce' does not exist on type
  'IntrinsicAttributes & { scopeHash?: string | undefined; }'   // ColorSchemeHeadScript
TS2322: Property 'nonce' does not exist on type
  'IntrinsicAttributes'                                          // ColorSchemeBodyFirstScript
```

**2. Forced through with a cast, it is dropped** — while `TextScaleHeadScript` keeps it:

```tsx
renderToStaticMarkup(<TextScaleHeadScript nonce="abc" />)
// <script nonce="abc">(function applyTextScale() { …

renderToStaticMarkup(<ColorSchemeHeadScript {...({ nonce: 'abc' } as never)} />)
// <script>(function(){try{var t=JSON.parse(localStorage.getItem('eufemia-theme')…
```

**3. Chromium blocks it.** A page built from the real emitted markup, served with `Content-Security-Policy: script-src 'nonce-testnonce'`:

```
CONTROL  nonce'd inline script executed  : true
SUBJECT  ColorSchemeHeadScript executed  : false
  -> globalThis.__eufemiaColorScheme     : null

CSP violations reported by the browser (1):
  * Executing inline script violates the following Content Security Policy
    directive 'script-src 'nonce-testnonce''. … The action has been blocked.
```

The control running proves the policy permits nonce'd inline scripts; only Eufemia's is blocked.

### Inconsistent with the rest of the library

| Component | Inline tag | Nonce support |
| --- | --- | --- |
| `TextScaleHeadScript` | `<script>` | explicit `nonce` prop |
| `SidebarMenuPreHydrationScript` | `<script>` | spreads `ScriptHTMLAttributes` |
| `SidebarMenuResponsive` | `<style>` | documented `styleNonce` prop |
| `ColorSchemeHeadScript` / `BodyFirst` / `BodyLast` | `<script>` | **none** |

That table is the complete set of inline `<script>` / `<style>` renderers in `packages/dnb-eufemia/src`.

### Suggested fix

Accept an optional `nonce` and pass it through, mirroring `TextScaleHeadScript`:

```tsx
export function ColorSchemeHeadScript({
  scopeHash,
  nonce,
}: {
  scopeHash?: string
  nonce?: string
} = {}) {
  return (
    <script
      nonce={nonce}
      dangerouslySetInnerHTML={{ __html: getHeadScript(scopeHash) }}
    />
  )
}
```

Same for the two body scripts. Spreading `ScriptHTMLAttributes` the way `SidebarMenuPreHydrationScript` does would work too and is more flexible.

Two things worth deciding while in there:

- `getHeadScript()` also runs `documentElement.classList.add(scopeHash)`, which is lost along with the rest of the script. The [Theme docs](https://eufemia.dnb.no/uilib/usage/customisation/theming/theme/#preventing-dark-mode-flash-fouc) describe these scripts only in terms of colour scheme, so it is not clear whether anything depends on that class being set here.
- There is currently no CSP guidance anywhere in the documentation; a short note on nonces would be a natural companion to this fix.

### Workaround

`getHeadScript()`, `getBodyScript()` and `getContentScript()` are exported, so consumers can render the tag themselves:

```tsx
<script nonce={nonce} dangerouslySetInnerHTML={{ __html: getHeadScript() }} />
```

That works, but it pushes `dangerouslySetInnerHTML` into consumer code to get behaviour the components should provide.

langz:
Fixes #9409

### Motivation

`ColorSchemeHeadScript`, `ColorSchemeBodyFirstScript` and `ColorSchemeBodyLastScript` rendered a bare inline `<script>` with no way to attach a nonce. A strict `script-src 'nonce-…'` policy blocks them, so the colour-scheme flash they exist to prevent comes back, and the only way out was allowing `unsafe-inline` — which defeats the policy.

Every other inline `<script>` / `<style>` in the shipped source already supports a nonce (`TextScaleHeadScript`, `SidebarMenuPreHydrationScript`, `SidebarMenuResponsive`), so these three were the outlier.

### Verified

Loaded a page built from the real emitted markup in Chromium under `Content-Security-Policy: script-src 'nonce-testnonce'`, before and after:

| | Before | After |
| --- | --- | --- |
| `ColorSchemeHeadScript` executed | `false` | `true` |
| Resolved colour scheme | `null` | `dark` |
| `<html>` scope hash applied | `""` | `eufemia-scope--default` |
| `<body>` colour-scheme class | — | `eufemia-theme__color-scheme--dark` |
| Server-rendered `Theme` element swapped | — | `--dark` |
| CSP violations | 1 | **0** |

The full first-paint chain now works under a strict policy.

### Notes

- Follows the `Omit<ScriptHTMLAttributes<HTMLScriptElement>, 'children' | 'dangerouslySetInnerHTML'>` pattern already used by `SidebarMenuPreHydrationScript`, so `id` and `data-*` are forwarded too.
- `{...props}` is spread before `dangerouslySetInnerHTML`, and that key is excluded from the type, so the script body cannot be overridden by a caller.
- Backwards compatible: the props are optional and `scopeHash` keeps working, so existing usage is unchanged. `scopeHash` is destructured out and never reaches the DOM (covered by a test).
- The `documentElement.classList.add(scopeHash)` question raised in the issue is left as-is; this PR only restores the scripts' ability to run.


langz:
### Status

Ready for review. Three commits, three files (`ColorSchemeScript.tsx`, its test, and the Theme docs page).

### One self-inflicted failure, found and fixed

The first CI run failed on **Run type checks**. My override-protection test used `{...({ ... } as never)}`, which `tsc` rejects with `TS2698: Spread types may only be created from object types`.

Two things let it through locally, both worth noting:

- **Vitest does not typecheck.** esbuild strips types without checking them, so a green test run says nothing about `tsc`.
- I had run the type check locally but piped it through `grep ColorScheme`, which hid the error — it was in my own test file.

Fixed in `0f6de7e` by casting a named const with `as unknown as ColorSchemeHeadScriptProps`. I have since run the exact commands the `typecheck` job runs (`test:types` for both `@dnb/eufemia` and `dnb-design-system-portal`) — both clean.

### Verification

| Check | Result |
| --- | --- |
| `test:types` (`@dnb/eufemia`) | clean |
| `test:types` (`dnb-design-system-portal`) | clean |
| `src/shared/__tests__` | 539 passed (25 files) |
| `eslint --quiet` (matches CI) | clean |
| `prettier` | no changes |

Browser behaviour under `Content-Security-Policy: script-src 'nonce-testnonce'`, before vs after:

| | Before | After |
| --- | --- | --- |
| `ColorSchemeHeadScript` executed | `false` | `true` |
| Resolved colour scheme | `null` | `dark` |
| `<html>` scope hash applied | `""` | `eufemia-scope--default` |
| `<body>` colour-scheme class | — | `eufemia-theme__color-scheme--dark` |
| Server-rendered `Theme` swapped | — | `--dark` |
| CSP violations | 1 | **0** |

I also mutation-tested the override guard: inverting the `{...props}` / `dangerouslySetInnerHTML` order makes that test fail, so it is not a no-op assertion.

### Scope check

I checked whether anything else in the shipped library would still break under a strict policy. The only runtime `<style>` injection is in `SidebarMenuPreHydrationScript`, which already sets `style.nonce` from `document.currentScript.nonce`. So these three scripts were the remaining blocker, and nothing further is needed for a consumer to drop `unsafe-inline`.

### Deliberately out of scope

The `documentElement.classList.add(scopeHash)` question from #9409. It is undocumented behaviour and unrelated to making the scripts runnable, so it deserves its own decision rather than riding along here.

CI is re-running on `0f6de7e` at the time of writing.

langz:
### CI outcome

The type-check failure is resolved — **Run type checks** now passes, along with everything else that has reported:

`Run tests` · `Run build tests` · `Run type checks` · `Run lint` · `Run portal lint` · `Run e2e tests` · `Build and smoke-test consumers` · `bundlewatch` · `CodeQL` · `Analyze (javascript-typescript)` · `Run analysis (javascript)` · `security/snyk` · `Run dependency checks` · `Publish` · `Continuous Releases` · `Build and deploy (branch preview)` · `Validate PR title`

Only `Run visual-regression tests` is still running. No failures anywhere.

### Also verified on the deployed preview

The new docs section renders correctly on the branch preview, nested under *Preventing dark mode flash (FOUC)* with a working anchor.

I also checked the emitted declarations, since that is what consumers actually get:

```ts
export type ColorSchemeScriptProps = Omit<ScriptHTMLAttributes<HTMLScriptElement>, "children" | "dangerouslySetInnerHTML">;
export type ColorSchemeHeadScriptProps = ColorSchemeScriptProps & { scopeHash?: string };
export declare function ColorSchemeHeadScript({ scopeHash, ...props }?: ColorSchemeHeadScriptProps): JSX.Element;
export declare function ColorSchemeBodyFirstScript(props?: ColorSchemeScriptProps): JSX.Element;
export declare function ColorSchemeBodyLastScript(props?: ColorSchemeScriptProps): JSX.Element;
```

All props optional, so existing usage is unaffected.

Ready for review — it needs an approval to unblock, which I cannot give on this PR.

langz:
Two notes from a deeper pass over the surrounding code.

### Why `Run visual-regression tests` takes ~20 min here

Not a problem with the change. I reproduced the selection locally with this PR's changed files:

```
Source changes were detected, but no impacted screenshot tests were mapped.
Running all tests for safety.
```

`ColorSchemeScript.tsx` renders a `<script>` with no visual output, so no screenshot test maps to it, and the selector falls back to running all 108 for safety. Expected behaviour, just slow. Every other job is green.

### The `scopeHash` question from #9409 is now answered

I had written that it was "not clear whether anything depends on that class being set here". That was too cautious — it is load-bearing. The portal compiles its CSS through `postcssIsolatePlugin({ scopeHash: 'eufemia-scope--portal' })`, which prefixes every Eufemia selector:

```
.eufemia-scope--portal .dnb-accordion
```

and the head script is the only thing that puts that class on `<html>`. So for an app that relies on the head script for document-level scoping while loading `--isolated.min.css`, a CSP block means unstyled content, not just a colour flash. Apps wrapping in `<IsolatedStyleScope>` get the class from the React wrapper and only lose the colour-scheme resolution.

The behaviour itself is correct and this PR does not touch it. Details in [a comment on #9409](https://github.com/dnbexperience/eufemia/issues/9409#issuecomment-5777088922).

### Related, deliberately kept separate

While tracing the above I found that `vite/client/plugins/eufemia-theme.ts` hand-copies `getHeadScript` and `getBodyScript` from this module, guarded only by a "must stay in sync" comment. I verified both copies are currently in sync, so there is no live drift — it is a latent hazard, and `TextScaleScriptUtils.ts` already shows the pattern that removes it. Filed as #9418 rather than folded in here, since it is a build-time refactor of the portal's Vite plugin and nothing forces it to ship with a CSP fix.

langz:
### All checks green

`Run visual-regression tests` finished — all 19 checks now pass:

`Run tests` · `Run type checks` · `Run build tests` · `Run lint` · `Run portal lint` · `Run e2e tests` · `Run visual-regression tests` · `Build and smoke-test consumers` · `bundlewatch` · `CodeQL` · `Analyze (javascript-typescript)` · `Run analysis (javascript)` · `security/snyk` · `Run dependency checks` · `Install dependencies` · `Publish` · `Continuous Releases` · `Build and deploy (branch preview)` · `Validate PR title`

`mergeable: MERGEABLE`, no conflicts. The only thing holding it is `REVIEW_REQUIRED` — an approval, which I cannot give on this PR.

tujoworker (approved):

PR #9420, opened 2026-09-23
chore(ColorSchemeScript): share the inline scripts with the portal Vite plugin

### LINKED ISSUE #9418: chore(ColorSchemeScript): de-duplicate the FOUC scripts copied into the portal Vite plugin
### Summary

The colour-scheme FOUC scripts exist twice. `packages/dnb-eufemia/src/shared/ColorSchemeScript.tsx` owns `getHeadScript()` and `getBodyScript()`, and `packages/dnb-design-system-portal/vite/client/plugins/eufemia-theme.ts` hand-copies both, guarded only by a comment:

```ts
/**
 * FOUC-prevention scripts matching @dnb/eufemia/shared/ColorSchemeScript.
 * These must stay in sync with getHeadScript / getBodyScript in that module.
 *
 * We inline them here because the Vite config/plugin context cannot import
 * .tsx files from eufemia at config-load time (before Vite's JSX transform
 * is registered).
 */
```

Nothing enforces that. A change to the library scripts leaves the portal silently running the old ones.

### Current state

Not broken today. I compared both copies (normalising the scope-hash, storage-key and class-prefix interpolation):

```
getHeadScript   IN SYNC
getBodyScript   IN SYNC
```

So this is a latent hazard rather than a live defect.

Worth noting the duplication is only partial: the portal imports `getContentScript` from the library, and duplicates only the two the Vite plugin needs at config-load time.

### Why it exists, and why it no longer has to

The stated reason is that the plugin cannot import `.tsx` at config-load time. That is true — but it is already solved for the sibling feature. `TextScaleScript.tsx` keeps its logic in a plain `TextScaleScriptUtils.ts` and re-exports it:

```ts
import { applyTextScale, getTextScaleScript } from './TextScaleScriptUtils'
export { applyTextScale, getTextScaleScript } from './TextScaleScriptUtils'
```

and the very same Vite plugin imports it directly, with no copy:

```ts
import { getTextScaleScript } from '@dnb/eufemia/src/shared/TextScaleScriptUtils'
```

### Suggested fix

Apply that pattern to the colour-scheme scripts:

1. Move `getHeadScript`, `getBodyScript` and `getContentScript` into a new `ColorSchemeScriptUtils.ts` (they are pure string builders, no JSX).
2. Re-export them from `ColorSchemeScript.tsx` so the public entry point is unchanged.
3. Import them in `eufemia-theme.ts` and delete the copies, keeping the portal's own `eufemia-scope--portal` value as the `scopeHash` argument.

The generators already take `scopeHash` as a parameter, so the portal keeps its distinct scope without needing a separate implementation.

### Scope

Deliberately not folded into #9415 — that PR is a CSP fix, this is a build-time refactor touching the portal's Vite plugin, and there is no live drift forcing them together.

langz:
Fixes #9418

The portal's Vite plugin hand-copied `getHeadScript` and `getBodyScript` from `ColorSchemeScript.tsx`, with only a "must stay in sync" comment standing between them and silent drift. It did so because the plugin cannot import a `.tsx` at config-load time.

`TextScaleScriptUtils.ts` already solves that for the sibling feature, so this applies the same pattern: the builders live in an import-free `ColorSchemeScriptUtils.ts`, `ColorSchemeScript.tsx` re-exports them, and the portal imports them instead of redefining them. The public API and the emitted scripts are unchanged, and a test now fails if the plugin goes back to a local copy.


langz:
### Verification

Moved out of the description to keep that focused on motivation.

**The deployed portal is unchanged.** The head and body scripts in this PR's Cloudflare preview are byte-identical to production eufemia.dnb.no for the same page, and `eufemia-scope--portal` is preserved. That compares a portal built from the shared module against one built from the old copies.

**The published package is unchanged.** Compared the package CI published for this PR (`pkg.pr.new/@dnb/eufemia@9420`) against npm `11.14.0`:

| Check | Result |
| --- | --- |
| Exported names from `shared/ColorSchemeScript` | identical |
| `getHeadScript('x')`, `getHeadScript('eufemia-scope--portal')`, `getBodyScript()`, `getContentScript()` | byte-identical |
| `ColorSchemeScriptUtils` shipped in `shared/`, `cjs/`, `es/` | yes, each with `.d.ts` |
| Compiled `ColorSchemeScriptUtils` | zero `require()` calls, so still import-free after Babel |
| Consumer type-check against the published `.d.ts` | resolves, `scopeHash` still type-checked |

**The config-load constraint holds, in both modes.** This was the reason the copies existed, so I tested it with Vite's own `loadConfigFromFile` rather than assuming — for `serve` (what `yarn dev` hits) as well as `build`. Both load the config, find the plugin, and inject the correct head and body scripts.

**The drift guard bites.** Changing the portal's scope hash makes the new test fail.

**Suites.** `src/shared/__tests__` 539 passed, portal `plugins` + `prod` + `color-scheme` 392 passed (including `prerender-utils.test.ts`, which imports `getContentScript` through the re-export). Type checks clean in both workspaces. Prettier reports no changes.

**bundlewatch.** Reports `+2.11MB, -0B`, but so do #9413, #9415, #9416 and #9417 — a constant baseline artifact, not a size change from this PR.

### Correction

I originally wrote that ESLint was clean on all four files. It is clean on the two library files, but the portal's ESLint config ignores its whole `vite/` directory, so the two portal files are not linted at all — `--quiet` was hiding the "file ignored" notice. Nothing to fix, but the earlier claim was wrong.

### Other copies

I searched the repo for the scripts' signatures. The colour-scheme logic now exists only in `ColorSchemeScriptUtils.ts`; `Theme.tsx` merely reads the global it sets. The one other inline-script duplicate — the brand-theme `headThemeScript` in both `prerender-utils.ts` and `prerender.mjs` — is intentional, since `prerender.mjs` runs under plain Node, and it is already covered by the `parity with the prerender.mjs copy` tests. Left alone.

### Visual regression

The selector maps no screenshot test to these files and falls back to running all of them for safety, which is why that job is slow here. I reproduced the same decision locally, so it is expected rather than a sign of visual impact.

langz:
### Added a guard for the new invariant

Reviewing this again, I noticed the fix introduced the same weakness it removes: `ColorSchemeScriptUtils.ts` must stay import-free, and a comment was the only thing saying so.

I checked whether that constraint is real rather than assuming. Adding `import { getStyleScopeHash } from ...plugin-scope-hash.js` to the module makes the portal config fail to load:

```
ERR_MODULE_NOT_FOUND: Cannot find module .../src/shared/build-info/BuildInfo
  imported from .../src/plugins/postcss-isolated-style-scope/plugin-scope-hash.js
```

So it is load-bearing, and it also confirms why `scopeHash` is a required argument here rather than defaulting to `getStyleScopeHash()` — moving the function together with its default would have broken the portal build.

Added a test asserting the module has no `import` or `require(`, mutation-tested both ways: it passes as-is and fails once an import is added. The failure mode it prevents is an opaque module-resolution error during a Vite config load, which is not obviously traceable to this file.

`src/shared/__tests__` is now 540 passing.

tujoworker (approved):

langz:
### Ready for review

All 19 checks pass on the latest commit `839ce6255d3`, including a complete visual-regression run: the selector fell back to running every screenshot test (none map to these files), and all 109 test files passed.

Mergeable with no conflicts. `main` has moved on by #9404 only, which touches none of these files, so I have not rebased. Blocked solely on a required approval.

PR #9421, opened 2026-09-23
fix(Table): keep the outline visible across a rowgroup header row

### LINKED ISSUE #9419: Table outline is invisible across a scope="rowgroup" header row (ui theme token collision)
## 🐛 Bug Report

In the **ui** theme, the `Table` `outline` frame is invisible where it crosses a full-width `scope="rowgroup"` header row. The border is drawn — it is just the same colour as the row behind it, so the frame looks broken open on both sides across that band.

The component CSS is correct. The two design tokens involved simply resolve to the same value in the ui theme:

| what | token | ui theme |
| --- | --- | --- |
| `--table-outline-color` | `--token-color-stroke-neutral-subtle` | `--dnb-greyscale-100` → `#ebebeb` |
| `th[scope="rowgroup"][colspan]` background | `--token-color-background-neutral-base` | `--dnb-greyscale-100` → `#ebebeb` |

Confirmed at runtime — both compute to `rgb(235, 235, 235)`.

`--table-border-color` uses the same stroke token, so horizontal cell borders disappear across that band too.

## To Reproduce

```jsx
<Table outline border="horizontal">
  <thead>
    <Tr>
      <Th scope="col">Marked</Th>
      <Th scope="col">Åpningstid</Th>
    </Tr>
  </thead>
  <tbody>
    <Tr>
      <Th scope="rowgroup" colSpan={2}>
        Norge
      </Th>
    </Tr>
    <Tr>
      <Td>Oslo Børs</Td>
      <Td>09:00-16:30</Td>
    </Tr>
  </tbody>
</Table>
```

The frame appears open on the left and right across the "Norge" row. Setting `--table-outline-color` to any contrasting value shows the border is present all along.

## Expected behavior

The table outline stays visible along its whole perimeter, including where it passes a rowgroup header row.

## Why this is a token issue rather than a Table fix

- **Sbanken renders correctly** with the exact same component CSS, because there the two tokens differ: `--token-color-stroke-neutral-subtle` is `--sbanken-greyscale-300` while `--token-color-background-neutral-base` is `--sbanken-greyscale-100`. Only the ui theme collides.
- `src/style/themes/ui/tokens.scss` is marked `/* This file is auto generated by makePropertiesFile.ts */`, so the values come from the design token source and should not be hand-edited.
- Overriding either token inside the Table's own SCSS would be a local workaround that diverges from the token system, and it would restyle every outlined table (or every rowgroup header) rather than fix the cause.

## Options for UX

1. Give `--token-color-stroke-neutral-subtle` a value distinct from `--token-color-background-neutral-base` in the ui theme. Next step up is `--token-color-stroke-neutral-bold` = `--dnb-greyscale-200` (`#ccc`). Note this stroke token is referenced by ~41 files, so this is a design-system-wide decision.
2. Change the background token used for rowgroup headers.
3. Accept it, and document that `outline` and `scope="rowgroup"` headers should not be combined in the ui theme.

Found while working on #9407.

langz:
Fixes #9419

In the ui theme a rowgroup header row (`th[scope="rowgroup"][colspan]`) is painted with `--token-color-background-neutral-base`, which resolves to the very same `#ebebeb` as the `--token-color-stroke-neutral-subtle` used by the table outline. The outline is still drawn across that row, but with nothing to distinguish it from the band — so the frame looks broken open on both sides. `Table.Container` is affected the same way.

Measured on the rendered table:

| | outline colour | rowgroup band | |
| --- | --- | --- | --- |
| before | `#ebebeb` | `#ebebeb` | identical |
| after | `#ccc` | `#ebebeb` | distinct |

### Before / after

Before:
eufemia-9421-before

After:
eufemia-9421-after

### See it live

The repo's own `Multiple table body sections` demo is an outlined table with a rowgroup header, so it shows the bug in the published docs:

- current: https://eufemia.dnb.no/uilib/components/table/demos/#multiple-table-body-sections
- this branch: https://fix-table-rowgroup-outline-c.eufemia-e25.pages.dev/uilib/components/table/demos/#multiple-table-body-sections

### The change

Tables containing such a row now draw the outline with `--token-color-stroke-neutral-bold`. It is scoped with `:has()`, so:

- Tables without a rowgroup header keep the current outline colour — no change to existing designs.
- `:where()` zeroes the `:has()` specificity, so the `html[data-visual-test]` override still wins and the screenshot baselines are untouched. Confirmed by a green visual-regression run.
- Sbanken is a no-op, because its `stroke-neutral-subtle` and `stroke-neutral-bold` are already the same value. It never had the bug, and it does not change.

### Worth a look from UX

This keeps the change as small as possible, with the side effect that a table with a rowgroup header gets a slightly stronger frame than one without. The alternative is to raise `--table-outline-color` for **all** outlined tables, which is more consistent but changes the look of every outlined table in the ui theme. That is a design decision rather than a bug fix, so it is deliberately not done here — it is a two-line change if you prefer it, and it would not churn screenshot baselines either, since visual tests force the outline to red.


langz:
Closing this — I measured the actual effect and it does not justify the change.

**The bug is real, but the fix is imperceptible.** Pixel-diffing the two screenshots in this PR, the left frame at a rowgroup band row (x = 46..51) reads:

| | x46 | x47 | **x48** | **x49** | x50 | x51 |
| --- | --- | --- | --- | --- | --- | --- |
| before | `#ffffff` | `#ffffff` | `#ebebeb` | `#ebebeb` | `#ebebeb` | `#ebebeb` |
| after | `#ffffff` | `#ffffff` | `#cccccc` | `#cccccc` | `#ebebeb` | `#ebebeb` |

| pairing | contrast |
| --- | --- |
| `#ebebeb` outline on white (normal rows, today) | 1.19:1 |
| `#ebebeb` on `#ebebeb` band (the bug) | 1.00:1 |
| `#cccccc` on `#ebebeb` band (this fix) | 1.35:1 |

The outline was already barely visible at 1.19:1. Going to 1.35:1 across a 1px edge is not a perceptible improvement, so the change does not achieve what it sets out to do.

**It restyles far more than the bug.** The diff is 7453 pixels, `#ebebeb -> #cccccc` x6864, spread over 747 rows on each vertical edge plus the top and bottom — the whole frame, not the 224 scanlines where the outline was actually swallowed.

**And it regresses Carnegie**, which is not covered in the description:

| theme | bg-neutral-base | stroke-subtle (today) | stroke-bold (after) | had the bug? | effect |
| --- | --- | --- | --- | --- | --- |
| ui light | `#ebebeb` | `#ebebeb` | `#ccc` | yes | fixed |
| ui dark | `#333` | `#48484a` | `#48484a` | no | no-op |
| sbanken light | `#f2f2f7` | `#d9d9e4` | `#d9d9e4` | no | no-op |
| sbanken dark | `#4a4a5b` | `#3a3a4a` | `#3a3a4a` | no | no-op |
| carnegie | `#ebebeb` | `#f6eae4` | **`#000`** | no | **beige frame turns black** |

Carnegie's `--token-color-stroke-neutral-bold` resolves to `--carnegie-greyscale-1000` (`#000`), and that theme never had the collision in the first place.

**It is also not a Table problem.** `--token-color-stroke-neutral-subtle` and `--token-color-background-neutral-base` are paired in `dnb-drawer-list.scss`, `dnb-menu.scss` and `dnb-sidebar-menu.scss` too, so patching Table leaves the same trap everywhere else. In the ui theme both tokens are `--dnb-greyscale-100`; making a subtle stroke readable against a neutral base fill is a token decision, and `style/themes/ui/tokens.scss` is generated from Figma.

#9419 stays open for UX to decide between raising `stroke-neutral-subtle` in the ui theme or changing the background token used by rowgroup headers. Keeping the branch in case we revisit.

langz:
Correction to one claim in my comment above.

I wrote that the two tokens are "paired in `dnb-drawer-list.scss`, `dnb-menu.scss` and `dnb-sidebar-menu.scss` too". That was based on those files *referencing* both tokens, which is not the same as placing them adjacently. Having now checked each one:

- **`dnb-drawer-list.scss` and `dnb-menu.scss` — no collision.** Their dividers are standalone elements (or a `::before` offset into the gap between options), so they sit on the container background, never inside an item's fill. They render as intended.
- **`dnb-sidebar-menu.scss` — two real collisions**, both border-and-background on the *same element*: the selected nav item ([L210-222](https://github.com/dnbexperience/eufemia/blob/main/packages/dnb-eufemia/src/extensions/sidebar-menu/style/dnb-sidebar-menu.scss#L210-L222)) and the Dropdown trigger ([L49-65](https://github.com/dnbexperience/eufemia/blob/main/packages/dnb-eufemia/src/extensions/sidebar-menu/style/dnb-sidebar-menu.scss#L49-L65)).

So the spread is narrower than I implied, but the underlying point holds and the sidebar-menu case is more consequential than the Table one: it is a persistent selected state whose 1px border is invisible in the ui theme.

The audit also surfaced a second collision in the same ramp — `stroke-neutral-bold` and `background-neutral-bold` are both `--dnb-greyscale-200` (`#ccc`) — which means this PR's escape hatch was one step away from colliding again.

#9419 has been rescoped around the token pair, with the sidebar-menu evidence.

PR #9423, opened 2026-09-23
fix(Table): reset the sticky header row when sticky is turned off

### LINKED ISSUE #9422: Table: sticky header state is not reset when sticky is turned off
## Description

`useStickyHeader` applies the sticky state to the header row imperatively — the `sticky`, `css-position` and `is-sticky` classes, plus the `--table-offset` and `--table-top` custom properties. None of it is ever removed, so the state survives after sticky mode is switched off.

The visible consequence comes from `table-sticky.scss`:

```scss
tr.sticky {
  position: sticky;
  z-index: 4;
  transform: translate3d(0, var(--table-offset, 0), 0);
}
```

The leftover `--table-offset` keeps translating the header row by however far the user had scrolled. Measured on a table scrolled 320px, the header row ends up displaced 120px and overlaps the first body rows instead of sitting above them.

## How to reproduce

```jsx
const [sticky, setSticky] = React.useState(true)

return (
  <>
    <Button onClick={() => setSticky(!sticky)}>Toggle sticky</Button>
    <Table sticky={sticky}>...</Table>
  </>
)
```

1. Scroll down until the header sticks.
2. Toggle `sticky` off.

Expected: the header row returns to the top of the table.
Actual: the row keeps `sticky is-sticky` and the stale `--table-offset`, so it stays translated down the page, floating over the data rows, and keeps its drop shadow.

Switching between `sticky` and `sticky="css-position"` has the same root cause: the `css-position` class and `--table-top` are never cleared.

## Affected versions

Present since the sticky header was introduced. Unmounting the table is unaffected, because the row is removed along with it.

langz:
Fixes #9422

`useStickyHeader` sets the sticky state on the header row imperatively, but never took it off again. Turning `sticky` off left `sticky is-sticky` and the stale `--table-offset` on the row, and since `tr.sticky` is `transform: translate3d(0, var(--table-offset, 0), 0)`, the header stayed translated by the last scroll position — floating on top of the body rows rather than returning to the top of the table. Measured on a table scrolled 320px, the header ended up 120px down, overlapping the first rows.

Switching between `sticky` and `sticky="css-position"` had the same cause, leaving the `css-position` class and `--table-top` behind.

The reset is placed in the effect body rather than in the cleanup. That is the part worth reviewing: by the time passive cleanup runs for a deleted subtree, React has already nulled the ref, so a cleanup-based reset reads `null` and throws. Unmounting needs no reset anyway, since the row is discarded with the table — the only case that needs handling is the prop going falsy while the component stays mounted, which is exactly when the effect body re-runs.

Switching between the two sticky modes is now symmetric: the `css-position` class is reconciled with `classList.toggle`, entering css-position clears `--table-offset`, and leaving it clears `--table-top`. Without the first of those, a transform left over from a previous scroll stacked on top of the `top` offset (thanks @tujoworker for spotting that direction).

### One pre-existing test was corrected

`should set --table-top with given stickyOffset when sticky="css-position"` re-rendered with `sticky` instead of `sticky="css-position"`, so its "stickyOffset should support pixels as well" step never re-entered the branch that writes `--table-top`. It was asserting the leaked value from the previous render, and passed no matter what the pixel handling did. It now re-renders with `sticky="css-position" stickyOffset={64}` and genuinely covers the 64px → `4rem` conversion; breaking that conversion makes it fail.


packages/dnb-eufemia/src/components/table/TableStickyHeader.tsx
```diff
@@ -92,6 +90,8 @@ export const useStickyHeader = ({
               '--table-top',
               `${offsetTopPx / 16}rem`
             )
+          } else {
```
tujoworker, line 95: Non-blocking: When switching the other direction (`sticky={true}` → `sticky="css-position"`) after scrolling, `--table-offset` is not cleared. Since `tr.sticky` always applies the transform, CSS-position mode can retain the previous translation. Could we clear the transform-mode state when entering this mode and cover that direction in a regression test? This does not need to block merging the current fix.

tujoworker (approved):
Approving. The inline note is non-blocking and does not need to hold up this fix.

packages/dnb-eufemia/src/components/table/TableStickyHeader.tsx
langz: Good catch, you were right. Reproduced it as a failing test first: scroll in transform mode, switch to `sticky="css-position"`, and `--table-offset` was still `224px`.

My fix was asymmetric — I cleared `--table-top` when leaving css-position mode, but never cleared `--table-offset` when entering it. `setSizes()` now handles both directions, so the three transitions are symmetric: entering css-position clears the offset, leaving it clears the top, and turning sticky off clears everything.

Added `should remove the transform state when switching to the css-position mode` to cover that direction.

PR #9424, opened 2026-09-23
fix(Forms): distinguish label suffixes from label text

tujoworker:
Forms currently renders label suffixes with the same visual emphasis as the main label. Render optional and custom suffixes as secondary label information while keeping them inside the accessible label.

Screenshot 2026-09-23 at 09 46 13

Motivation: https://dnb-it.slack.com/archives/C081G91GCT0/p1790081558411279


langz:
Nice improvement! Looks like [`PhoneNumber-have-to-match-long-label-1.png`](https://github.com/dnbexperience/eufemia/blob/e02b4d58f7ab7923aa45a2cebe6e09c4f97a5ced/packages/dnb-eufemia/src/extensions/forms/Field/PhoneNumber/__tests__/__image_snapshots__/PhoneNumber-have-to-match-long-label-1.png) needs to be regenerated too, though.

The [`LongLabel`](https://github.com/dnbexperience/eufemia/blob/e02b4d58f7ab7923aa45a2cebe6e09c4f97a5ced/packages/dnb-design-system-portal/src/docs/uilib/extensions/forms/feature-fields/PhoneNumber/Examples.tsx#L159-L168) example uses `required={false}`, so its "(valgfritt)" turns grey as well, but the committed baseline still has it in the label color. Since screenshots have zero mismatch tolerance, that test will fail until it's updated. As far as I can see, it's the only other screenshot with an optional label :)

langz (approved):
Looking great, the optional labels are much easier to scan now 👏

Approving, just the PhoneNumber screenshot from above left :)

PR #9426, opened 2026-09-23
chore: add dedicated static asset hosting

tujoworker:
Eufemia assets need a stable public hostname without routing static files through the Lambda-backed server. This adds a private S3 and CloudFront origin plus the OIDC deployment pipeline for `assets.eufemia.dnb.no`.


tools/assets/infra/variables.tf
```diff
@@ -0,0 +1,25 @@
+variable "aws_region" {
+  type    = string
+  default = "eu-north-1"
+}
+
+variable "environment" {
+  type    = string
+  default = "dev"
+}
```
langz: `environment` defaults to `dev` and nothing overrides it — [`ghe-deploy-workflow.yml`](https://github.com/dnbexperience/eufemia/blob/e3d261fd0b5e1ac397d3fb0baa644908e37c7171/tools/assets/ghe-deploy-workflow.yml#L41-L43) passes only `TF_VAR_cost_allocation` and `TF_VAR_edge_auth_secret`, and there's no tfvars file. So `assets.eufemia.dnb.no` would be served from `eufemia-dev-assets-<account>`, with the OAC, response headers policy and edge function named to match, plus an `Environment = dev` tag.

Bucket names are immutable, so correcting this after the first apply means a new bucket, policy and OAC, and a re-upload. Passing `TF_VAR_environment: prod` from the deploy workflow now would avoid that.

`tools/analytics` has the same default, but its domain default is `analytics.dev.eufemia.tech-03.net`, so `dev` fits there. It also [validates the value](https://github.com/dnbexperience/eufemia/blob/e3d261fd0b5e1ac397d3fb0baa644908e37c7171/tools/analytics/infra/variables.tf#L11-L14) — worth copying that block over?

langz (commented):
Two small things worth looking at before merge.

`eufemia-assets-hosting` isn't registered in `verify.yml`, so its lint and tests never run on a pull request. [`assets.yml`](https://github.com/dnbexperience/eufemia/blob/e3d261fd0b5e1ac397d3fb0baa644908e37c7171/.github/workflows/assets.yml#L37-L41) is the only place that references the workspace, and it [triggers on push to `release`](https://github.com/dnbexperience/eufemia/blob/e3d261fd0b5e1ac397d3fb0baa644908e37c7171/.github/workflows/assets.yml#L3-L6).

Both lists in `verify.yml` ask to be kept complete, and [the test one](https://github.com/dnbexperience/eufemia/blob/e3d261fd0b5e1ac397d3fb0baa644908e37c7171/.github/workflows/verify.yml#L187-L190) describes this case directly:

> No other workflow runs them on a pull request, so a workspace missing from this list has its tests first execute at release or deploy time.

Since `push` needs `test`, a failing test would first surface while the release deploy is already under way. A line in [lint](https://github.com/dnbexperience/eufemia/blob/e3d261fd0b5e1ac397d3fb0baa644908e37c7171/.github/workflows/verify.yml#L74-L83) and one in [tests](https://github.com/dnbexperience/eufemia/blob/e3d261fd0b5e1ac397d3fb0baa644908e37c7171/.github/workflows/verify.yml#L197-L207) should cover it.

The second one is inline on `variables.tf`.

langz (approved):
Really nice piece of infrastructure. Private bucket with a `SourceArn`-scoped policy, OIDC the whole way through, and it slots neatly into the pattern `tools/analytics` established — easy to follow.

Approving. The two notes above are small follow-ups, not blockers.

joakbjerk:
Reviewed — clean, well-scoped addition; all CI green. Two non-blocking suggestions:

**1. Name the origin-verification header in the README.** The wiring steps say the Akamai property must send "the same value" but never name the header the CloudFront function actually checks. State it explicitly — and note it differs from the header used by the Lambda-backed origin, so the `assets.eufemia.dnb.no` property should mirror the **dashboard** (CloudFront-origin) config rather than the `server.eufemia.dnb.no` one. Without this, whoever wires Akamai can easily pick the wrong header and every request 403s.

**2. Production host is named/tagged `dev`.** `environment` defaults to `"dev"` and the deploy workflow never sets `TF_VAR_environment`, so the resources serving prod `assets.eufemia.dnb.no` are named `eufemia-dev-assets-*` and tagged `Environment = dev`. Set `TF_VAR_environment` in the deploy (or change the default) to reflect reality.

Nit: consider a `terraform validate` step in the test job — the string-match tests won't catch malformed HCL before the auto-approved apply.

Follow-ups (not this PR, per the README): the Akamai `assets.eufemia.dnb.no` property + DNS. When they land, confirm the injected value matches `EDGE_AUTH_SECRET` and wire the `origin_hostname` output.

joakbjerk (approved):

PR #9430, opened 2026-09-23
chore: increase extensions CSS budget

tujoworker:
The extensions bundle is growing as reusable extensions add production-ready responsive behavior. Increase the shared and themed extensions CSS budgets together so static component styling can remain in SCSS rather than being moved into component code solely to satisfy the previous limit.


langz (approved):
11 it is 🤝

PR #9432, opened 2026-09-23
chore(test): let Playwright's Firefox launch on macOS 27

langz:
On macOS 27, `~/Library/Application Support/Firefox` is protected when Firefox is installed. Playwright's Firefox uses the same folder, so it hangs on launch, and screenshot and e2e runs fail with a timeout, e.g. `browserType.launch: Timeout 180000ms exceeded`. This affects every contributor on macOS 27 who has Firefox installed.

The test configs now give Playwright's Firefox its own app-data folder, so no manual workaround is needed. This can be removed once we are on a Playwright version with the upstream fix (microsoft/playwright#42768).


langz:
This is just a quick draft, as I experienced issues running "yarn test:screenshots:update" locally now on macos27:

```
➜  eufemia git:(chore/table-sticky-outline-visual-test) yarn test:screenshots:update Table    

 RUN  v4.1.11 /Users/anderslangseth/dev/eufemia-2/eufemia/packages/dnb-eufemia

12:34:44 pm [vite] (client) Re-optimizing dependencies because vite config has changed

 Test Files   (1)
      Tests  no tests
     Errors  1 error
   Start at  12:34:44
   Duration  30.22s (transform 0ms, setup 0ms, import 0ms, tests 0ms, environment 0ms)


```

Not sure if any of you have upgraded to macos27, and experience the same issue?
@joakbjerk @tujoworker @snorrekim

langz:
Some findings from debugging this locally (macOS 27.0, 26A428), to help with the review:

**Root cause**
- The macOS system log shows the "App Data" privacy protection blocking Playwright's Firefox from `~/Library/Application Support/Firefox` (`service=kTCCServiceSystemPolicyAppDataDetailed`, `accessing=firefox`, `authValue=0`). It is denied silently, without a prompt, both when running from VS Code and from Terminal.app.
- Control test: starting Firefox through `XUL_APP_FILE` with an unmodified copy of its `application.ini` still hangs. Adding only `Profile=PlaywrightFirefox` makes it start in about 0.6 s. So the shared app-data folder is the cause, and a separate folder is what fixes it.

**Who is affected**
macOS 27 with Firefox installed. According to microsoft/playwright#42768, a terminal or editor with Full Disk Access is not affected, so not everyone on macOS 27 will see this. Quick check: if `ls ~/Library/Application\ Support/Firefox` prints "Operation not permitted" in your terminal, you are affected.

**With this change**
Locally, all existing Table and Tag screenshots match and the Firefox e2e specs pass. The fix only applies on macOS and leaves an existing `XUL_APP_FILE` alone. It can be removed once we upgrade to a Playwright version with the upstream fix (planned for 1.64, currently only an alpha).

joakbjerk:
## Review — chore(test): let Playwright's Firefox launch on macOS 27

**Verdict:** Sound, well-scoped dev-only fix, but it does not merge as-is — CodeQL fails with 2 high-severity alerts (required check). One cheap edit clears them. Scope is clean: test tooling only, no production/consumer code touched.

### Verified
- **Mechanism:** `isolateFirefoxAppData` guards on `process.platform !== 'darwin'` and an already-set `XUL_APP_FILE`, resolves `Contents/Resources` from the executable, injects `Profile=PlaywrightFirefox` into `[App]`, writes the derived ini to `Resources/browser/application.ini`, and points `XUL_APP_FILE` at it. Path math (`../../Resources` from `Contents/MacOS/firefox`) is correct. Not-installed is handled by the `existsSync(appIni)` guard.
- **No-op on CI:** CI is Linux → early return, so the shared pipeline is unaffected. Consumers never receive this (test config + `repo-utils`).
- **Test coverage:** good — 5 scenarios: non-darwin no-op, not-installed no-op, fresh write (asserts injected content), refresh-outdated, preserve-existing `XUL_APP_FILE`. CJS/ESM parity kept (`.js`/`.mjs` identical), `.d.ts` + README updated.
- **Blast radius:** test-only; darwin-guarded; writes identical bytes idempotently into the Playwright-managed browser install.

### Blockers
- **[Blocking] CodeQL: 2 high — file-system race (TOCTOU)** at `node-utils.js:90` and `node-utils.mjs:90`. The `!fs.existsSync(browserAppIni) || fs.readFileSync(...) !== content` check-then-`writeFileSync` is the flagged pattern. Practical risk is negligible (dev machine, own browser dir, identical bytes), but it is a required check, so it blocks merge. Cheapest fix that also clears the alert is to drop the `existsSync`/`readFileSync` pair for a single guarded read:

```js
let existing
try {
  existing = fs.readFileSync(browserAppIni, 'utf-8')
} catch {
  existing = null
}
if (existing !== content) {
  fs.writeFileSync(browserAppIni, content)
}
```

Alternatively dismiss the alert with a dev-only test-util justification — but editing is lower-friction than a dismissal on a required gate.

### Observations
- [Nitpick] The `[App]` injection is a no-op if the ini lacks an `[App]` header (regex miss → unchanged content → `XUL_APP_FILE` points at a non-isolating copy). Real Firefox always has `[App]`, so theoretical; not worth guarding.
- [Nitpick] The `if (!fs.existsSync(appIni)) return` above is the same check-then-use shape; CodeQL only flagged the write path. If refactoring for the blocker, consider collapsing this one too.

### Caveats / worth confirming
- Overwriting `Resources/browser/application.ini` mutates the Playwright-managed install. Assuming it mirrors the upstream workaround (microsoft/playwright#42768) it is fine — `playwright install --force` resets it and the code re-applies idempotently. Worth confirming `browser/application.ini` is the correct target (vs. only setting `XUL_APP_FILE`).
- The test has no `try/finally` around the platform override + tmpdir cleanup, so a mid-test assert failure leaks the stubbed `process.platform` and temp dir. Matches the file's existing plain-`assert` style, so non-blocking.

tujoworker (approved):
I'm was on 27 for a week or so, have not seen any problems actually. But nice if it solves an issue you found occurring.

langz:
Thanks for the reviews, I'll consider further what to do about this. Hopefully it will soon be fixed in a new release of playwright 🤞

PR #9436, opened 2026-09-23
fix(NumberFormat): announce bank account numbers one character at a time

langz:
A reporter using a screen reader told us that account numbers in DNB are announced as words rather than digits.

`NumberFormat.BankAccountNumber` grouped the screen reader text in pairs, so `2000 12 34567` was exposed as `20 00 12 34 56 7`. Speech synthesis reads such pairs as compound numbers ("tjue", "null", "tolv", and so on), which cannot be transcribed back into an account number. The grouping also did not match what is shown on screen, the Norwegian BBAN and the Swedish Bankgiro and Plusgiro variants ended on an orphan digit, and IBAN reused the visual blocks of four.

Account numbers are now exposed one character at a time, the same way `NumberFormat.OrganizationNumber` and the personal number part of `NumberFormat.NationalIdentityNumber` already work. The visible text is unchanged.

Amounts are deliberately left alone. A balance should be announced as "ten thousand kroner", because that is how the value is read out loud, and spelling it out digit by digit would make it harder to understand.

The `speak-as: numbers` declaration is dropped as part of this. `numbers` is a value of the `@counter-style` descriptor rather than of the CSS Speech property, and no browser implements that property, so the declaration only gave the impression that digit reading was handled.


langz (commented):
Reviewed `613a30b` against the original accessibility report.

The formatter change looks correct: account numbers retain their visible formatting while the accessible string contains individual characters, including IBAN letters. All 507 tests in the NumberFormat and NumberUtils suites pass. I did not find a code defect in this PR.

Two scope caveats before calling the report resolved:

- This PR alone does not fix `Value.BankAccountNumber`, the example sent to the reporter. #9437 is needed to stop the Forms component from discarding the accessible text.
- The report also mentions balances and teaching deafblind users. Neither PR changes amount formatting. The combined preview exposes amounts as digits, such as `150,00 kroner` and `25000,00 kroner`, not written-out number words. Hearing an amount spoken as words is normal, but that does not explain or resolve a report of words replacing digits on a braille display.

I would describe this as an account-identifier accessibility fix, not yet a confirmed fix for the complete report. Please ask the reporter to compare the [combined-preview account example](https://fix-forms-value-identifiers.eufemia-e25.pages.dev/uilib/extensions/forms/Value/BankAccountNumber/demos/#value) and [currency example](https://fix-forms-value-identifiers.eufemia-e25.pages.dev/uilib/extensions/forms/Value/Currency/demos/#value) with the affected DNB app/page, using VoiceOver and a braille display if applicable. The added spaces also affect the text available to braille, so that experience needs checking, not just pronunciation.

Non-blocking scope/verification notes for this PR. I verified rendered text and Chromium/Playwright accessibility snapshots, not actual VoiceOver speech or physical braille output. A separate `transformIn` regression is reported on #9437.

tujoworker (approved):
Works fine when testing with VO. Docs looking good too.

PR #9437, opened 2026-09-23
fix(Forms): announce identifier values to screen readers

langz:
Stacked on #9436.

`Value.BankAccountNumber`, `Value.NationalIdentityNumber`, `Value.OrganizationNumber` and `Value.PhoneNumber` rendered the visual string only and discarded the accessible text, so an account number reached the screen reader as `2000 12 34567` and was announced as compound numbers. This is the surface most applications use to display these values, so the accessible formatting in `NumberFormat` never reached users.

The four values now render through their `NumberFormat` variant, which is what `Value.Number` and `Value.Currency` already do for amounts. The visible text is unchanged.

`NumberFormat.BankAccountNumber` could only format Norwegian BBAN, while `Value.BankAccountNumber` offers five account types. It now takes a `bankAccountType` prop, so all five are covered.

Note that the `dnb-ui-lib` bundle limits go from 250 kB to 260 kB. The UMD bundle measured 249.94 kB against a 250 kB limit before this change, so it had roughly 60 bytes of headroom and any addition would have tripped it.


packages/dnb-eufemia/src/extensions/forms/Value/BankAccountNumber/BankAccountNumber.tsx
```diff
@@ -28,10 +26,12 @@ function BankAccountNumber(props: ValueBankAccountNumberProps) {
         return undefined
       }
 
-      return formatBankAccountNumberByType(
-        String(external),
-        bankAccountType
-      ).number
+      return (
+        <NumberFormatBankAccountNumber
```
langz: **[P2] Preserve the string input to `transformIn`**

Returning JSX from `toInput` breaks existing consumer transformations: [useValueProps](https://github.com/dnbexperience/eufemia/blob/ee87b202238723ee5fce50e788ad3f477d3a92c4/packages/dnb-eufemia/src/extensions/forms/hooks/useValueProps.ts#L91) evaluates `transformIn(toInput(externalValue))`, so the callback now receives a React element instead of the formatted string.

Minimal reproduction:

```tsx
<Value.BankAccountNumber
  value="20001234567"
  transformIn={String}
/>
```

This displays `2000 12 34567` on the base revision, but `[object Object]` here. Callbacks using string methods can also throw. The same input-type regression occurs in `Value.NationalIdentityNumber`, `Value.OrganizationNumber` and `Value.PhoneNumber`; I verified all four with before/after tests.

Please preserve the existing string transformation contract and create the accessible rendering afterward, with regression coverage for consumer-provided `transformIn` on all four components. The new screen-reader assertions should remain.

langz (commented):
Reviewed `ee87b20` on top of #9436.

One regression to fix before merge, detailed inline: the four identifier value components now pass a React element instead of a string to the documented `transformIn` callback. Five focused review probes pass at `613a30b` and fail here; even `transformIn={String}` now renders `[object Object]`.

The default accessibility path is improved as intended. In the [deployed account demo](https://fix-forms-value-identifiers.eufemia-e25.pages.dev/uilib/extensions/forms/Value/BankAccountNumber/demos/#value), the visible value remains `2000 12 34567`, while the Chromium/Playwright accessibility snapshot contains `2 0 0 0 1 2 3 4 5 6 7`. All five account types expose individual characters, including IBAN letters. All 581 existing tests across the seven relevant formatter/value test files pass, but they do not cover the transformation regression.

Together these PRs address account-number grouping and the Forms accessible-text gap. They do not yet establish that the complete customer report is fixed: balances are unchanged and still expose numeric text, and actual VoiceOver/braille output has not been tested here. The reporter should confirm the result in the affected DNB app/page and with the [currency example](https://fix-forms-value-identifiers.eufemia-e25.pages.dev/uilib/extensions/forms/Value/Currency/demos/#value).

packages/dnb-eufemia/src/extensions/forms/Value/NationalIdentityNumber/NationalIdentityNumber.tsx
```diff
@@ -18,7 +16,9 @@ function NationalIdentityNumber(props: ValueNationalIdentityNumberProps) {
     if (isValueEmpty(value)) {
       return undefined
     }
-    return formatNationalIdentityNumber(cleanNumber(value)).toString()
+    return (
+      <NumberFormatNationalIdentityNumber value={cleanNumber(value)} />
```
tujoworker: For all the places we now use the a component instead of the formatter; do we need the cleanNumber helper? Or does it run twice now?

langz (commented):
### Review notes

Both review threads are addressed, and I re-verified the result in a browser rather than only in jsdom.

**On the `cleanNumber` question:** yes, it ran twice. `formatNationalIdentityNumber` and `formatOrganizationNumber` already strip non-digits themselves (`String(number).replace(/[^0-9]/g, '')`), so the call was redundant. Removed from both.

**On `transformIn`:** `toInput` returns the formatted string again, so the callback keeps receiving a string. The accessible rendering is applied afterwards, via an internal `renderValue` on `Value.String`.

**What the browser run showed.** Local portal driven by Playwright, reading the real accessibility tree:

```
- strong: Bankkonto
- text: 2 0 0 0 1 2 3 4 5 6 7
```

All five account types, national identity, organization and phone numbers expose the spelled-out form while the visible text stays unchanged. Also checked with a numeric value, pre-formatted input, lowercase IBAN, a 2-digit Plusgiro and five phone shapes (`00`-prefixed, bare, `+46`, 800-number, pre-spaced).

It also caught a defect the unit tests missed: feeding the transformed string back through `NumberFormat` silently dropped what the callback returned, so `transformIn={(v) => v + ' (sperret)'}` rendered without the suffix. The accessible rendering now applies only while the value still is the formatted string, so a custom transform renders verbatim, as it did before this PR.

**Worth knowing:** a custom `transformIn` opts out of the character-by-character announcement. That is deliberate — reformatting is the only way to derive the `aria` text, and it cannot happen without discarding the consumer's string. The default path, which is what applications use, is unaffected.

**Also included here:** `formatPhoneNumber` normalises parentheses, so an already formatted dashed country code (`+1 (684) 12 34 56 7`) is detected instead of being re-parsed into `+1 68 41 23 45 67`. And the `dnb-ui-lib` limits are 255 kB rather than 260 kB, based on the CI-measured 250.01 kB (UMD) and 248.17 kB (ESM).

The earlier `Run e2e tests` failure was the flaky `focusmode.spec.ts` scroll-restore test on `/uilib/components/radio/demos/`; that commit touched only `extensions/forms/Value/**`, and the job passed on re-run. All 19 checks are green.

I consider this ready to merge, but it still needs a review from someone other than me.

langz:
## Review — approved as a draft; both open comments are resolved

### @tujoworker's question about `cleanNumber` — answered: it did run twice

`formatNationalIdentityNumberParts` cleans its own input:

```ts
// cleanup
const num = String(number).replace(/[^0-9]/g, '')
```

So `formatNationalIdentityNumber(cleanNumber(value))` was cleaning twice, and dropping the outer call is correct. It also matters beyond tidiness: `renderValue` decides whether the value is "ours" with `format(value) === value`, which only works because the formatter is idempotent — it strips non-digits before formatting, so re-formatting an already-formatted `180892 12345` returns the same string. The test asserting `data-text` is `18 08 92 1 2 3 4 5` proves that round-trip end to end.

### The earlier P2 about `transformIn` — addressed

`toInput` still returns a string and the JSX now comes from a separate `renderValue`, so `transformIn(toInput(externalValue))` keeps receiving a string. `gives the formatted string to transformIn` asserts `transformIn` is called with `'180892 12345'`, and `does not reformat a custom transformIn result` asserts a custom result is passed through untouched with no `.dnb-sr-only` added. Good coverage of exactly the regression that was raised.

**Validation.** 478 tests pass across the four `Value.*` suites and `NumberUtils`.

### Non-blocking

`renderValue` is documented as internal, but because `StringComponent` is typed with `StringComponentProps`, consumers can still pass it to `Value.String` and it becomes part of the public type surface. If you want it truly internal, exporting the component as `ValueStringProps` and casting internally would keep the prop out of consumer-facing types.

Still a draft, so leaving it for you to open when ready.

